PR #402, opened 2024-09-26
ci: Modularize CI pipeline and test Python >= 3.11

thomass-dev:
- Remove reference of `Makefile` in workflow's instructions to build front-end
- Use matrix to test back-end on all versions of python `>=3.9` (only success on `3.12` is mandatory)
- Trigger CI tasks at job level instead of workflow level
- Use one and only entry-point to define if pipeline succeed `ci-all-green`

---

Example: https://github.com/probabl-ai/skore/actions/runs/11146672648

image

### DIFF
--- a/.github/workflows/add-pr-assignee.yml
+++ b/.github/workflows/add-pr-assignee.yml
@@ -1,9 +1,11 @@
+name: Add PR assignee
+
 on:
   pull_request:
     types: [opened]
 
 jobs:
-  add-assignee:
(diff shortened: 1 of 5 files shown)


rouk1 (dismissed):
~~I strongly disapprove ^^~~

augustebaum (approved):
Alright, I think this is ready!

augustebaum (approved):

Issue #407, opened 2024-09-27
UI: Share report button does not use the custom modals


rouk1:

augustebaum:
is this closed by #414?

rouk1:
Yes it is thanks :)

Issue #377, opened 2024-09-23
Display bug: the up and down arrows for cells do not appear


sylvaincom:
When on the dashboard, the up and down arrows to move cells do not appear (but are still clickable and work fine):
<img width="313" alt="Capture d’écran 2024-09-23 à 15 18 49" src="https://github.com/user-attachments/assets/7d7e8954-1cc5-49a8-a24f-9a947d478355">

rouk1:
Those arrows will disappear in #336. 
Then some real drag'n'drop UX may arrive #408.

So I close this one.

Issue #332, opened 2024-09-11
Export: choosing the file path and file name


sylvaincom:
### Issue

When exporting a dashboard, it is not possible to choose the file path neither choose the file name.

### Expectation

I would like to be able to choose the file path and file name.

PR #414, opened 2024-09-30
feat(UI & API): Manage views in the frontend

### LINKED ISSUE #336: [UI] Manage views
Relates to #334 

When a project is created, the space where views are presented is empty, or a new project has a view initially created with the name "Default".

**User flow:**
- The user selects the path where to create a view
- A prompt for a name appears
- The empty view is displayed
- The user adds/removes items to/from the view

The user can see all their views in the list, and select another one, updating the currently displayed view.

The user can delete a view they no longer need.

rouk1:
### Backend

All project endpoints are now below the same prefix.

- `GET /api/project/items` lists all items of a project.
- `POST /api/project/views/share/{key}` returns a static shareable HTML file named `{view name}-{yyyy-MM-dd-HH-mm}.html`.
- `PUT /api/project/views/{key}` save a view in the project.
- `DELETE /api/project/views/{key}` delete a view from the project.

### Frontend

- A new `EditableList` component as been implemented to manage views. 
- A lot of renaming to get rid of the concept of "report".
- Plug everything to the new layout.
- Rework view card layout and add an action dropdown to it.
- Rework dropdowns using floating UI to make them fixed.

https://github.com/user-attachments/assets/3f94e483-ecea-4d5a-8433-50f9150fc6e6

Co-authored with @augustebaum.
Fixes #336, #407, #377, #332


frontend/src/services/api.ts
```diff
@@ -34,11 +34,11 @@ export async function fetchReport(): Promise<Report | null> {
   return null;
 }
 
-export async function putLayout(payload: Layout): Promise<Report | null> {
+export async function putView(view: string, layout: Layout): Promise<Project | null> {
   try {
-    const r = await fetch(`${BASE_URL}/report/layout`, {
+    const r = await fetch(`${BASE_URL}/project/views/${view}`, {
```
tuscland, line 39: I am not at ease with using names as identifiers in URLs. It often clashes with the `/` character found in the URL path. Could you use a query parameter instead?

For example `/project/views?name=${view}`

Same suggestion for delete and share features.

augustebaum:
Really awesome!

frontend/src/views/ProjectView.vue
```diff
@@ -1,29 +1,33 @@
...
+import { useToastsStore } from "@/stores/toasts";
 
-const reportStore = useReportStore();
+const projectStore = useProjectStore();
+const toastsStore = useToastsStore();
 const isDropIndicatorVisible = ref(false);
 const editor = ref<HTMLDivElement>();
 const isInFocusMode = ref(false);
+const views = ref<EditableListItemModel[]>([]);
+let unsavedViewsIds: string[] = [];
```
tuscland, line 22: question: Is this used to manage delayed view saving? I am not sure to know why a view would not be saved.

frontend/src/services/api.ts
augustebaum: Out of curiosity, what bothers you with this?

tuscland: If you include an identifier in the URL, you want a string over which you have control, that is, something that the program created. Here it is a user-supplied name, hence we have not a lot of control over it, unless we strictly apply validation rules.

frontend/src/components/EditableListItem.vue
```diff
@@ -1,24 +1,53 @@
...
     <div class="label-container">
-      <span class="icon" v-if="props.item.icon" :class="props.item.icon" />
-      <span class="label">
-        {{ props.item.name }}
+      <span class="icon" v-if="item.icon" :class="item.icon" />
+      <span
+        class="label"
+        :contenteditable="item.isUnnamed"
+        ref="label"
+        @blur="onItemNameEdited"
```
augustebaum: What does `blur` do? I can't find a reference to it in the Vue docs

frontend/src/components/DropdownButton.vue
augustebaum, line 1: can you explain why this necessary?

augustebaum (approved):

anasstarfaoui:
That's an amazing app so far guys, big kuddos to all of you :) 

Just one suggestion, would it be possible to fixed the position of the Views list by default? I feel like it would be preferrable to not have something who grows @rouk1 :)

frontend/src/views/ProjectView.vue
rouk1: It is used to differentiate views who are not know by the backend. It's a mapping of pending views. For example you create a view and did not gave it a name then it's local id is in this array and the project view knows that it's a creation and not a rename.

frontend/src/components/EditableListItem.vue
rouk1: It's a native event of inputs (lose of focus).
https://developer.mozilla.org/en-US/docs/Web/API/Element/blur_event

frontend/src/services/api.ts
tuscland: What about using a proper id?

rouk1:
> Just one suggestion, would it be possible to fixed the position of the Views list by default?

Could you elaborate please I don't understand your comment ? The view list is on top of the page and has a height of 20% of the height of the viewport.

thomass-dev:
Huge work 👏🏻  , the video is awesome. 
It's a bit hard for me to evaluate anything other than python, as the diff is a bit long 😄 .
Will do my best.

frontend/src/components/DropdownButton.vue
```diff
@@ -14,19 +16,63 @@ const props = withDefaults(defineProps<DropdownProps>(), { isPrimary: false, ali
...
+});
 
 function closeDropdown(e: Event) {
   if (el.value && !el.value.contains(e.target as Node)) {
     isOpen.value = false;
   }
 }
 
+// Intersection observer to close the dropdown when it is not visible
+const intersectionObserver = new IntersectionObserver(
```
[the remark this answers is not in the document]
rouk1: Dropdown were not used in scrolling area we did not have that need before...

frontend/src/services/api.ts
rouk1: I suggest that we use querystring for now and generalize the usage of id's in another issue.
WDYT ?

augustebaum (changes requested):
I found a UX issue: when first starting the UI, adding cards to the canvas doesn't work, with no explanation why. I think it's because no view is selected by default.
Some ways to solve this:
- Select the first view in the list at startup
- Allow the user to set a "default" view, and select that one at startup
- Warn the user that they need to select a view

The easiest is probably the warning. @anasstarfaoui do you see a better way of dealing with this?

EDIT: Resolved

augustebaum (commented):
I think the "intermediate view" mechanic when creating a view is a bit strange. I would expect a view either to exist, or not to exist. Some alternatives we discussed with @rouk1:
- If the user clicks away during the view naming process, either validate or cancel the view creation (like in VSCode when you create a file: if you click away and the text box is empty, the creation is aborted; if the text box is not empty, the file is created). Essentially, clicking away should be as effectful as pressing "Enter"
- Ask for the view name before creating anything (e.g. in a modal)

frontend/src/services/api.ts
rouk1: Query params are now used.

rouk1:
> I found a UX issue: when first starting the UI, adding cards to the canvas doesn't work, with no explanation why. I think it's because no view is selected by default. Some ways to solve this:
> 
>     * Select the first view in the list at startup
> 
>     * Allow the user to set a "default" view, and select that one at startup
> 
>     * Warn the user that they need to select a view
> 
> 
> The easiest is probably the warning. @anasstarfaoui do you see a better way of dealing with this?

An error toast will now pop if a user tries to double click or drag an item if no view is selected.

augustebaum (approved):

PR #420, opened 2024-10-01
fix(SklearnBaseEstimatorItem): Temporarily disable skops's security feature

thomass-dev:
Based on the skops tutorial: https://skops.readthedocs.io/en/stable/persistence.html#usage.

### DIFF
--- a/src/skore/item/sklearn_base_estimator_item.py
+++ b/src/skore/item/sklearn_base_estimator_item.py
@@ -25,8 +25,9 @@ class SklearnBaseEstimatorItem(Item):
 
     def __init__(
         self,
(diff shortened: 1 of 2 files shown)


rouk1 (dismissed):
Sounds good to me but I do not know how to build a "unsafe" estimator.

thomass-dev:
Please @tuscland can you take a look? Thanks.

tuscland (changes requested):
Overall that's looking great, however I would like to see a test succeeding in the case of an otherwise untrusted estimator.

augustebaum (approved):

tuscland (approved):

PR #421, opened 2024-10-01
refactor(launch): Replace thread with lifespan event for opening the browser

### LINKED ISSUE #301: Webapp launcher page
When the webapp is launched, we would like to open a browser so that the user can start using the application without intervention.

Serving the webapp is a blocking call, therefore the browser must be open before. But this order is inconsistent with the availability of the webapp, therefore the browser displays an error.

Instead of opening a connexion directly to the webapp, we suggest that a local HTML file is open instead, acting as a launcher. This HTML file would contain code to poll until webapp availability, then switch to the target URL.

Here is a naive implementation of such launcher page:
```html
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>Loading Application</title>
    <style>
        body {
            font-family: Arial, sans-serif;
            display: flex;
            justify-content: center;
            align-items: center;
            height: 100vh;
            margin: 0;
            background-color: #f0f0f0;
        }
        .loader {
            text-align: center;
        }
        .spinner {
            border: 4px solid #f3f3f3;
            border-top: 4px solid #3498db;
            border-radius: 50%;
            width: 40px;
            height: 40px;
            animation: spin 1s linear infinite;
            margin: 20px auto;
        }
        @keyframes spin {
            0% { transform: rotate(0deg); }
            100% { transform: rotate(360deg); }
        }
    </style>
</head>
<body>
    <div class="loader">
        <h1>Loading Application</h1>
        <div class="spinner"></div>
        <p id="message">Please wait while we connect to the server...</p>
    </div>

    <script>
        const mainAppUrl = 'http://localhost:5000/'; // Replace with your app's URL
        const checkInterval = 2000; // Check every 2 seconds

        function checkServerStatus() {
            fetch(mainAppUrl, { method: 'HEAD' })
                .then(response => {
                    if (response.ok) {
                        window.location.href = mainAppUrl;
                    } else {
                        setTimeout(checkServerStatus, checkInterval);
                    }
                })
                .catch(() => {
                    document.getElementById('message').textContent = 'Still trying to connect...';
                    setTimeout(checkServerStatus, checkInterval);
                });
        }

        checkServerStatus();
    </script>
</body>
</html>
```

Query parameters could be used to parametrize this page, passing the target URL.

augustebaum:
Closes #301

### DIFF
--- a/src/skore/cli/launch_dashboard.py
+++ b/src/skore/cli/launch_dashboard.py
@@ -1,11 +1,11 @@
 """Implement the "launch" command."""
 
-import threading
-import time
 import webbrowser
+from contextlib import asynccontextmanager
 from pathlib import Path
 
 import uvicorn
+from fastapi import FastAPI
 
 from skore.cli import logger
 from skore.project import load
@@ -16,11 +16,6 @@ class ProjectNotFound(Exception):
     """Project was not found."""
 
     project_path: Path
-
-
-def __open_browser(port: int):
-    time.sleep(0.5)
-    webbrowser.open(f"http://localhost:{port}")
 
 
 def __launch(project_name: str | Path, port: int, open_browser: bool):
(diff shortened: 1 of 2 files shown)


rouk1 (approved):
Nice : )

Issue #425, opened 2024-10-02
Add a convenient way of outputting system and dependency versions


tuscland:
Implement the equivalent of `sklearn.show_versions()`

PR #426, opened 2024-10-02
feat: Add a convenient way of outputting system and dependency versions (#425)

tuscland:


src/skore/utils/_show_versions.py
```diff
@@ -0,0 +1,84 @@
...
+    """
+    deps = [
+        "pip",
+        "setuptools",
+        "diskcache",
+        "fastapi",
+        "rich",
+        "skops",
+        "uvicorn",
+    ]
```
rouk1, line 51: You may use the built-in toml parser to read this from the pyproject.toml file.

```python
import tomllib

with open("pyproject.toml", "rb") as f:
    data = tomllib.load(f)
    print(data["project"]["dependencies"])
```

rouk1 (changes requested):

src/skore/utils/_show_versions.py
```diff
@@ -0,0 +1,82 @@
...
+        version information on relevant Python libraries
+
+    """
+    from importlib.metadata import PackageNotFoundError, version
+
+    import tomllib
+
+    deps = ["pip", "setuptools"]
+
+    with open("pyproject.toml", "rb") as f:
```
rouk1, line 49: I suppose that this path should be computed from the `__file__` variable. 
@thomass-dev will the pyproject.toml be there in a release ?

rouk1 (changes requested):

rouk1:
The `_get_deps_info` function now uses `importlib` instead of parsing the `pyproject`.
co-autorhed with @thomass-dev

thomass-dev (approved):

rouk1 (approved):

PR #440, opened 2024-10-04
chore: Move python code to its own directory

thomass-dev:
```
.
├── CONTRIBUTING.md
├── LICENSE
├── Makefile
├── README.md
├── examples
│   ├── basic_usage.ipynb
│   ├── basic_usage.py
│   ├── README.txt
│   ├── skrub_demo.ipynb
│   └── skrub_demo.py
├── skore
│   ├── pyproject.toml
│   ├── requirements-test.txt
│   ├── requirements-tools.txt
│   ├── requirements.txt
│   ├── VERSION.txt
│   ├── src
│   └── tests
└── skore-ui
    ├── index.html
    ├── package.json
    ├── package-lock.json
    ├── README.md
    ├── tsconfig.app.json
    ├── tsconfig.json
    ├── tsconfig.node.json
    ├── tsconfig.vitest.json
    ├── vite.config.lib.ts
    ├── vite.config.ts
    ├── vitest.config.ts
    ├── vitest.setup.ts
    ├── node_modules
    ├── public
    ├── src
    └── tests
```


.github/workflows/ci.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   changes:
     runs-on: ubuntu-latest
     outputs:
-      backend: ${{ steps.filter.outputs.backend }}
+      skore: ${{ steps.filter.outputs.skore }}
```
tuscland, line 9: `skore-core`? just wondering… I'm anticipating that one day, people will want to install only the "core", and not the "ui". `skore` would then become an umbrella project, composed of both libraries.

.github/workflows/release.yml
```diff
@@ -36,7 +36,7 @@ jobs:
         run: make build-skore-ui
 
       - name: Override VERSION.txt with tag
-        run: echo "${GITHUB_REF_NAME}" > VERSION.txt
+        run: echo "${GITHUB_REF_NAME}" > skore/VERSION.txt
```
rouk1, line 39: Isn't it a good time to remove this generated file from the sources N

.github/workflows/skore.yml
```diff
@@ -1,10 +1,14 @@
-name: Reusable backend workflow
+name: Reusable skore workflow
```
rouk1, line 1: I'm not found of the naming, especially has it appears in the badges. 
May I suggest to simply call the actions skore and UI ?

rouk1 (commented):
I would like to avoid symbolic links for the readme & licence files.
They are both repo wide and are not related only to the python side of things.

Side note: I like the actual layout but maybe we could have assume the "classic" monorepo layout by adding a packages folder.

.github/workflows/release.yml
thomass-dev: i will do it, but in another PR.

thomass-dev:
> I would like to avoid symbolic links for the readme & licence files. They are both repo wide and are not related only to the python side of things.
> 
> Side note: I like the actual layout but maybe we could have assume the "classic" monorepo layout by adding a packages folder.

The question of the symbolic links is right (i hesitated for a long time): if we want to include a README and a license file in the python distribution package, we need to place them inside the python directory. There can't be outside the directory (ie `../`).

I assumed we wouldn't duplicated those file, so i made symlinks.

.github/workflows/skore.yml
thomass-dev: These workflows will not longer be displayed as badges https://github.com/probabl-ai/skore/pull/444

thomass-dev: But if you want I can rename them.

.github/workflows/release.yml
rouk1: Perfect then.

rouk1:
> > I would like to avoid symbolic links for the readme & licence files. They are both repo wide and are not related only to the python side of things.
> > Side note: I like the actual layout but maybe we could have assume the "classic" monorepo layout by adding a packages folder.
> 
> The question of the symbolic links is right (i hesitated for a long time): if we want to include a README and a license file in the python distribution package, we need to place them inside the python directory. There can't be outside the directory (ie `../`).
> 
> I assumed we wouldn't duplicated those file, so i made symlinks. Should we ?

Can't they be copied at build of the release ?

.github/workflows/ci.yml
thomass-dev: Let's postpone this use case to when it's needed? It will implies building the skore-ui in a self package.

tuscland: I was just meaning to name the directories, not building things separately. Yes, we can postpone as long as we have this in mind.

rouk1 (changes requested):
Please remove symbolic links : )

rouk1 (approved):

Issue #383, opened 2024-09-23
skore.cross_validate 


MarieS-WiMLDS:
# Background

https://www.notion.so/probabl/2024-09-20-Workshop-with-Ga-l-107ef76d36b980fdbba2c6036db91a35  

# Long term goals

We want to help by: 
- allowing easy comparison 
- building report of the results once frozen

# What do we want to save?

The `cross_validate` function output a dict with 6 keys according to [scikit-learn doc](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html). Each value is an array of length n_splits.  

We want to do a skore wrapping around the sklearn function: it should behave the same way, only adding some storage and graph computing if deemed relevant.  
To reach our goals, and to provide help to the user, I need to remember the context around when the function `cross_validate` is launched.  
In the long term, I want to be able to compare the various runs of cross validate, with something that could look like this:  
![WhatsApp Image 2024-09-23 at 18 05 23](https://github.com/user-attachments/assets/e0ff50db-c56e-4bb6-9067-4f31ad750387)

It would be possible to filter on some elements, to select only one kind of estimator for instance. We would also forbid comparison on different scorings (it doesn't make sense to compare f1 with accuracy).

tuscland:
2024-09-30 -- Meeting report with @glemaitre and @ogrisel

## Proposed Enhancements:

1. **Storage of Probability and Decision Function:**
   - Store predict_proba and decision_function results to enable:
     - Calculation of additional metrics later
     - Interactive threshold adjustment without re-running cross-validation

2. **Precision/Recall Display:**
   - Show precision/recall metrics
   - Explicitly state the threshold used (typically 0.5, but could be something else)

3. **Business Metric Integration:**
   - If a user registers a business metric (via a registry?), recalculate it as well
   - Note: Scikit-learn scorers don't allow choosing scorers after the fact

4. **Dataset Statistics:**
   - Include sufficient statistics of a dataset in the cross-validation result

5. **Threshold Tuning:**
   - Using raw predictions (cross-validation results):
     - Extract code from TTCC (trivial)
     - Optimize business metric between a minimum and maximum threshold
     - Calculate average for each split of the cross-validation
   - Reference: [Scikit-learn Cost-Sensitive Learning Example](https://scikit-learn.org/dev/auto_examples/model_selection/plot_cost_sensitive_learning.html#consideration-regarding-model-refitting-and-cross-validation)

[...]

PR #443, opened 2024-10-04
feat: Implement basic `cross_validate` functionality

augustebaum:
Addresses the first part of #383 

Introduces a new item type, `CrossValidateItem`, as well as a top-level function `skore.cross_validate`, which detects the type of ML task based on the input estimator and target, adds some scorers, and runs scikit-learn's `cross_validate` with those extra arguments.
The results are stored in the `Project` given as input, always under key `cross_validation`, along with an interactive summary plot.
The output of `skore.cross_validate` is as close as possible to scikit-learn's, except for some edge cases. Also, if the user passes `scorer="my_metric"`, the output dict will contain key `test_my_metric` as well as `test_score`, whereas in the scikit-learn version the dict only has `test_score`.

TL;DR:

- Make Plotly a mandatory dependency
- Add `CrossValidateItem`
- Add `cross_validate` function
- Add `plot_cross_validation` function that plots the result of a cross-validation in Plotly
- Refactor tests (add fixture `in_memory_project` and use that instead of `project` for clarity -- it would have been great to do `from conftest import in_memory_project as project` but pytest doesn't work that way)
- Fix some issues with VegaWidget in skore-ui (the plot was originally written in altair)

---

How to test:
  - I'm in a notebook/VSCode, I have a sklearn estimator, I run `skore.cross_validate` on it just like I would with `sklearn.model_selection.cross_validate`, except I also give the argument `project=project` to let Skore know where to store things (if the argument is not passed we'll do cross-validation but nothing will be saved)
  - A plot appears in the UI and in my notebook
  - I forgot a metric; I change the cell and re-run
  - A new plot is outputted and the one on the UI updates

---

To-do:
- [x] Add Marie's tests as automated tests
- [ ] Potentially use `pytest-cases` for cross-validation tests
- [x] Deal with multi-class classification (use `["roc_auc_over_weighted", "neg_log_loss", "recall_weighted", "precision_weighted"]`)
- [x] Deal with the case where `X` and `y` are not numpy arrays
  - At least set up the infrastructure to easily support pandas, numpy and lists; if only numpy works for now, that's okay
  - Look into `struct.pack`
- [x] Remove metrics depending on e.g. if estimator has `predict_proba`
  - ~Note for example SVC has the method defined, but it raises if the model isn't initialized with `probability=True`, so a `try-except` will be more accurate than `hasattr`~
- [x] Deal with the case where the estimator params are not serializable
  - Use `repr` for the time being
- [x] Make it so when the user passes `metric="my_metric"`, the output contains `test_my_metric` as well as `test_score`
- [x] Ensure that the stored cv_results are serializable
  - [x] Serialize the numpy arrays
  - [x] Deal with the case where the user passes `return_estimator=True` (we can't pass that to `CrossValidateItem` because it's not serializable)
    - Asked Marie
  - [ ] Confirm that the `get()` output is appropriate
- [x] Fix plot not updating in the UI
- [x] Remove default metrics for clustering
- [x] Stop storing the plot directly in `CrossValidateItem`, instead store JSON
  - Rename the attribute to `plot_serialized`, add a property `plot`
- [ ] ~Investigate Marie's issue of the empty plot~


skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,175 @@
...
+if TYPE_CHECKING:
+    import sklearn.base
+
+
+def plot_cross_validation(cv_results: dict):
+    """Plot the result of a cross-validation run."""
+    import altair
+    import pandas
+
+    # df = pandas.DataFrame(cv_results).melt(var_name="metric", value_name="score")
```
tuscland, line 22: Just wondering if this is a comment meant to stay, or something you wanted to remove.

examples/basic_usage.ipynb
```diff
@@ -629,6 +629,32 @@
     "_Stay tuned for some new features!_"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "id": "e5e0c4ae",
+   "metadata": {},
+   "source": [
+    "---\n",
+    "# Cross-validation with skore"
```
tuscland, line 638: - Please use "Heading 2" `##` style.
- Move the whole one block up so that "Stay tuned for some new features" remains the last element.

examples/basic_usage.py
```diff
@@ -295,6 +295,20 @@ def my_func(x):
 # %% [markdown]
 # _Stay tuned for some new features!_
 
+# %% [markdown]
+# ---
+# # Cross-validation with skore
```
tuscland, line 300: See my notes above. (I guess I should have commented this file instead!)

examples/basic_usage.py
```diff
@@ -295,6 +295,20 @@ def my_func(x):
...
+
+# %%
+from sklearn import datasets, linear_model
+from skore.cross_validate import cross_validate
+diabetes = datasets.load_diabetes()
+X = diabetes.data[:150]
+y = diabetes.target[:150]
+lasso = linear_model.Lasso()
+
+cv_results = cross_validate(lasso, X, y, cv=3)
```
tuscland, line 310: Question: there's no need to specify the `project`?

I'm asking because the examples before do save something to the project, so I thought it'd be more consistent to specify it.

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def cross_validate(
+    *args, project: Project | None = None, **kwargs
+) -> CrossValidationItem:
+    """Evaluate estimator by cross-validation and output UI-friendly object.
+
+    This function wraps scikit-learn's [cross_validate](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html#sklearn.model_selection.cross_validate)
+    function, to provide more context
+    and facilitate the analysis.
```
tuscland, line 112: Maybe keep it on one line?

tuscland (changes requested):

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,188 @@
...
+    cross_validation_item = CrossValidationItem.factory(cv_results, estimator, X, y)
+
+    if project is not None:
+        project.put_item("cross_validation", cross_validation_item)
+
+    # If in a IPython context (e.g. Jupyter notebook), display the plot
+    try:  # noqa: SIM105
+        display(cross_validation_item.plot)
+    except NameError:
+        pass
```
thomass-dev, line 182: Really disturbing 😅 .

skore/src/skore/ui/project_routes.py
```diff
@@ -69,6 +70,11 @@ def __serialize_project(project: Project) -> SerializedProject:
         elif isinstance(item, MediaItem):
             value = base64.b64encode(item.media_bytes).decode()
             media_type = item.media_type
+        elif isinstance(item, CrossValidationItem):
+            # Convert plot to MediaItem
+            item = MediaItem.factory(item.plot)
+            value = base64.b64encode(item.media_bytes).decode()
+            media_type = item.media_type
```
thomass-dev, line 77: Take a look at my comment on `cross_validate_item.py`.

Does the user really need access to the plot in the backend? 
- If yes (what i see in `cross_validate.py` in a jupyter context), you should add a property `plot` which can dynamically reconstruct the plot from the persisted plot (json).
- If not, you can here just pass the persisted plot (json) as bytes.

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,175 @@
...
+        updated_at : str
+            The last update timestamp in ISO format.
+        """
+        super().__init__(created_at, updated_at)
+
+        self.cv_results = cv_results
+        self.estimator_info = estimator_info
+        self.X_info = X_info
+        self.y_info = y_info
+        self.plot = plot
```
thomass-dev, line 119: For my underlying comments, keep in mind that all attributes of an `Item` are persisted.
They must be environment independent and should be primitive.

Actually anything can persist, because we use `diskcache` which uses `pickle` behind it.
But in the near futur, we will use another backend not based on `pickle`.

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,188 @@
...
+    ml_task = _find_ml_task(estimator, y)
+
+    # Add scorers based on the ML task
+    if ml_task == "clustering":
+        scorers_to_add = ["homogeneity_score", "silhouette_score", "rand_score"]
+    elif ml_task == "regression":
+        scorers_to_add = ["r2", "neg_mean_squared_error"]
+    elif ml_task == "classification":
+        scorers_to_add = ["roc_auc", "neg_brier_score", "recall", "precision"]
+    elif ml_task == "unknown":
```
thomass-dev, line 157:
```suggestion
    else:
```

examples/basic_usage.py
augustebaum: I was mostly using this example to develop the plot, so I didn't need to save stuff in the project -- I'll add it.

MarieS-WiMLDS:
hi!
When I did `make install-skore` to test this PR, some files were created in the frontend folder. Git wants to follow these files. If they do not exist already in the folder, I supposed it should be in gitignore?

skore/src/skore/cross_validate.py
augustebaum: That would lead to ruff complaining that the line is too long. If this helps, in markdown new lines within paragraphs are usually discarded (although not in github-flavoured markdown apparently).

skore/src/skore/item/cross_validate_item.py
augustebaum: nice catch

examples/basic_usage.ipynb
augustebaum: done

examples/basic_usage.py
augustebaum: done

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,175 @@
...
+        -------
+        CrossValidationItem
+            A new CrossValidationItem instance.
+        """
+        if not isinstance(cv_results, dict):
+            raise TypeError(f"Type '{cv_results.__class__}' is not supported.")
+
+        estimator_info = {
+            "name": estimator.__class__.__name__,
+            "params": estimator.get_params(),
```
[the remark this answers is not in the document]
augustebaum: No; for example, according to the [sklearn docs](https://scikit-learn.org/stable/developers/develop.html#get-params-and-set-params), a parameter could even be an estimator, which itself has parameters (which you could recover with `get_params(deep=True)`

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,188 @@
...
+    return {
+        k: v
+        for k, v in cv_results.items()
+        if not any(added_scorer in k for added_scorer in added_scorers)
+    }
+
+
+def cross_validate(
+    *args, project: Project | None = None, **kwargs
+) -> CrossValidationItem:
```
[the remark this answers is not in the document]
augustebaum: My thoughts on this:
- Having it be a top-level function makes it as similar as possible to sklearn (for example, you can just change `from sklearn import cross_validate` to `from skore import cross_validate`)
- Making it a method of a Project makes it less clear what the function does (are you cross-validating the Project)?

MarieS-WiMLDS:
At first I tried this (cf screenshot). I understood after that I had to do what is commented.   
Capture d’écran du 2024-10-09 17-45-05  

Is there a technical constraint to prevent us from having the cross_validate function at the root of the lib? would it be bad practice?

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,188 @@
...
+function in order to enrich it with more information and enable more analysis.
+"""
+
+from typing import Literal
+
+from skore import logger
+from skore.item.cross_validate_item import CrossValidationItem
+from skore.project import Project
+
+
```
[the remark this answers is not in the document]
augustebaum: Yep, thanks for bringing this up. The way I usually do this is:
- if I can, write type-hints and a one-line docstring
- if the type-hints are too hard to write, or not clear enough, I use aliases to `Any` or don't add type-hints at all, and write a docstring with a `Parameters` section
 
In the case of `_expand_scorers` for example, the `scorers` parameter can be a lot of different things (see the sklearn `cross_validate` docs), so if they had some type hints I would use them, but they don't so I put no type-hint and described things in the `Parameters` section.

For the functions that are protected, I considered it lower priority to make everything consistent. I'll work on documenting everything properly.

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        y_info = None if y is None else {"hash": _hash_numpy(y)}
+
+        X_info = {
+            "nb_rows": X.shape[0],
+            "nb_cols": X.shape[1],
+            "hash": _hash_numpy(X),
+        }
```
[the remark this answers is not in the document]
augustebaum: X and y might not be numpy arrays. They could be lists of arrays, lists of lists, pandas Series, pandas DataFrames, polars DataFrames, narwhals-compatible dataframes...

This is not in the spec so I'm not sure what to do:
- Assume numpy and abort the saving to the project otherwise
- Assume more than numpy (thoughts on this @MarieS-WiMLDS?)

skore/src/skore/cross_validate.py
augustebaum: So, ruff tells me to use `with contextlib.suppress(NameError)` instead of `try: ... except: pass`, but in this case ruff then complains that `display` is not defined. `display` comes from the `IPython.display` module, so maybe I could replace this with:
```python
with contextlib.suppress(ImportError):
    from IPython.display import display

    display(cross_validation_item.plot)
```

Would that make more sense?

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,188 @@
...
+        An object containing the cross-validation results, which can be readily
+        inserted into a Project.
+
+    Examples
+    --------
+    >>> from sklearn import datasets, linear_model
+    >>> diabetes = datasets.load_diabetes()
+    >>> X = diabetes.data[:150]
+    >>> y = diabetes.target[:150]
+    >>> lasso = linear_model.Lasso()
```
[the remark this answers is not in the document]
augustebaum: done

skore/src/skore/cross_validate.py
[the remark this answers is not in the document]
augustebaum: added docs everywhere (and corrected outdated stuff too)

MarieS-WiMLDS:
It's super cool to see this live :star_struck: 

I tested: 
- creating an error in scikit-learn --> the error message is passed correctly!
- a regression problem (with diabetes dataset & lasso)
- a multi-class classification problem (with iris & random forest)
- a binary classification problem
- a classification problem

Questions and remarks:
1. Can we drop score_time? I don't find this metric very interesting, and I have the feeling that it overloads the options
2. What is the test_score?
3. The plot cross_validate in skore ui doesn't update
4. in the classif multiclass. I hadn't think of this usecase before actually. The default values for binary and multi-class shouldn't be the same. 
  a. the default I chose aren't correct, it creates an error. let's create an additional elif, and instead of `recall` and `precision`, use `recall_weighted` and `precision_weighted`. 
  b. Because the default aren't correct, there was an error in the process. Yet, some part of cross_validate in scikit learn side ran to the end, and we have some results. Is it normal not to have at least these results displayed? (cf screen shot below).  
5. Again, for classification, the default are not correct on my side. The [silhouette score](https://scikit-learn.org/1.5/modules/generated/sklearn.metrics.silhouette_score.html) exists in scikit-learn, but you have to pass it to cross_validate through a callable, and not a string. I'm surprised, so I would tend to say that we shouldn't use it as default. @sylvaincom what do you think about this? (I asked also in OS chan in slack). 

Capture d’écran du 2024-10-09 19-36-18

skore/src/skore/cross_validate.py
thomass-dev: Looks better, thanks :) .

augustebaum:
> I tested:
> - creating an error in scikit-learn --> the error message is passed correctly!
> - a regression problem (with diabetes dataset & lasso)
> - a multi-class classification problem (with iris & random forest)
> - a binary classification problem
> - a classification problem

That's really great, thanks! It would be really useful to add these use-cases in our tests directly, let's sync to do this.

> - Can we drop score_time?

It depends if we want to uphold the requirement that our `cross_validate` returns the same as scikit-learn's. IMO it makes sense to follow the principle of least surprise, so keep it.

> - What is the test_score?

If you don't pass any `scoring` parameter (or if you pass a single metric, like `scoring="r2"`), this metric is the one shown in `test_score`. If you don't pass `scoring`, in many cases scikit-learn can make a guess at what metric to use.

> - The plot cross_validate in skore ui doesn't update

Will investigate.

> - Because the default aren't correct, there was an error in the process (cf screen shot below).

This is not normal; let's sync to reproduce this.

> Let's remove all default scores for clustering

Got it!

skore/src/skore/cross_validate.py
augustebaum: done

augustebaum:
It looks like `neg_brier_score` requires the estimator to have a `predict_proba` method. I guess we can check the estimator beforehand and remove this scorer if don't have that method.

augustebaum:
@MarieS-WiMLDS The scorer `recall_weighted` does not exist; did you mean `recall_score(average="weighted")`?

augustebaum:
@MarieS-WiMLDS What should be the output of `project.get("cross_validate")`? For now, it outputs something like what scikit-learn outputs (`{"test_score": ..., ...}`), but this doesn't work when the user e.g. sets `return_estimator=True`, because skore is not meant to store anything and everything. An option could be to return the altair plot, another could be something like the scikit-learn outputs but without the non-serializable things...

skore/src/skore/ui/project_routes.py
augustebaum: done

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,175 @@
...
+
+        plot = plot_cross_validation(cv_results)
+
+        instance = cls(
+            cv_results=cv_results,
+            estimator_info=estimator_info,
+            X_info=X_info,
+            y_info=y_info,
+            plot=plot,
```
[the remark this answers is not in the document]
augustebaum: done

MarieS-WiMLDS:
For multi-class classification, the following metrics should be used to be compliant:  
- roc_auc_ovr_weighted
- recall_weighted
- precision_weighted
- neg_log_loss

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,175 @@
...
+        self.cv_results = cv_results
+        self.estimator_info = estimator_info
+        self.X_info = X_info
+        self.y_info = y_info
+        self.plot = plot
+
+    @classmethod
+    def factory(
+        cls,
+        cv_results: dict,
```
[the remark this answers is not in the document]
augustebaum: should be good now

skore/src/skore/item/cross_validate_item.py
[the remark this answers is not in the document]
augustebaum: i solved this by saving `repr(estimator.get_params())`

skore/src/skore/item/cross_validate_item.py
augustebaum: yes, this is clear

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def __init__(
+        self,
+        cv_results_serialized: dict,
+        estimator_info: dict,
+        X_info: dict,
+        y_info: dict,
+        plot_bytes: bytes,
+        created_at: str | None = None,
+        updated_at: str | None = None,
```
rouk1, line 126: You should use python typing notation.

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,232 @@
...
+    ----------
+    array : numpy array
+        The numpy array whose hash will be computed.
+
+    Returns
+    -------
+    hash : str
+        A hash corresponding to the input array.
+    """
+    return hashlib.sha256(array.tobytes()).hexdigest()
```
rouk1, line 101: Can [`struct.pack`](https://docs.python.org/3/library/struct.html#struct.pack) be used here to make this hash function generic ?

skore/src/skore/cross_validate.py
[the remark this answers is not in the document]
tuscland: I agree with @augustebaum

skore/src/skore/item/cross_validate_item.py
augustebaum: Now that we have CI for Python 3.9 this will be checked automatically; thanks for the heads-up

skore/src/skore/cross_validate.py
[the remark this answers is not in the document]
augustebaum: @thomass-dev i'll let you resolve this if that's satisfactory

skore/src/skore/item/cross_validate_item.py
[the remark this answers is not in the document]
augustebaum: done

skore/src/skore/item/cross_validate_item.py
augustebaum: struct.pack seems to only accept ints. In the end, it should be enough to simply convert X and y to numpy arrays.

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,235 @@
...
+    # Add explicit metric to result (rather than just "test_score")
+    if isinstance(scorers, str):
+        if kwargs.get("return_train_score") is not None:
+            cv_results[f"train_{scorers}"] = cv_results["train_score"]
+        cv_results[f"test_{scorers}"] = cv_results["test_score"]
+
+    cross_validation_item = CrossValidationItem.factory(cv_results, estimator, X, y)
+
+    if project is not None:
+        project.put_item("cross_validation", cross_validation_item)
```
tuscland, line 223: Question: how will someone compare two different cross validation evaluations?

Is it planned to offer a way to name or prefix the generated item keys?

cc @MarieS-WiMLDS

MarieS-WiMLDS: it's the second part of the feature to be able to compare yes indeed. At first, we will ask the user to say if it's comparable, otherwise we will consider that they are always comparable inside the same project. Each run will be named run_{n}, and will have a "info" chart that can be displayed on hoover, containing: the estimator name, number of cv, id of the data used (to be defined yet, displaying an hash is not very user-friendly).

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,233 @@
...
+        y_info = None if y is None else {"hash": _hash_numpy(y_array)}
+
+        X_array = X if isinstance(X, numpy.ndarray) else numpy.array(X)
+        X_info = {
+            "nb_rows": X_array.shape[0],
+            "nb_cols": X_array.shape[1],
+            "hash": _hash_numpy(X_array),
+        }
+
+        plot_bytes = MediaItem.factory(plot_cross_validation(cv_results)).media_bytes
```
thomass-dev, line 213:
```suggestion
        plot = plot_cross_validation(cv_results)
        plot_bytes = plot.to_json().encode("utf-8")
```

thomass-dev: You shouldn't be dependent on a third-party class item if when you reconstruct the plot, you use Altair directly.

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,233 @@
...
+            cv_results_serialized=cv_results_serialized,
+            estimator_info=estimator_info,
+            X_info=X_info,
+            y_info=y_info,
+            plot_bytes=plot_bytes,
+        )
+
+        return instance
+
+    @property
```
thomass-dev, line 225:
```suggestion
    @cached_property
```

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,233 @@
...
+        plot_bytes = MediaItem.factory(plot_cross_validation(cv_results)).media_bytes
+
+        instance = cls(
+            cv_results_serialized=cv_results_serialized,
+            estimator_info=estimator_info,
+            X_info=X_info,
+            y_info=y_info,
+            plot_bytes=plot_bytes,
+        )
+
```
thomass-dev, line 222:
```suggestion
        # add plot as cached property
        instance.plot = plot
```

skore/src/skore/item/cross_validate_item.py
```diff
@@ -0,0 +1,233 @@
...
+    @property
+    def plot(self):
+        """A plot of the cross-validation results."""
+        return altair.Chart.from_json(self.plot_bytes.decode("utf-8"))
+
+    @property
+    def cv_results(self):
+        """The cross-validation results."""
+        return self.cv_results_serialized
```
thomass-dev, line 233: Its not mandatory to do a property, because `cv_results_serialized` is exposed as public member attribute.
Furthermore, its a bit ambiguous that the property `cv_results` points to serialized one.

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,235 @@
...
+    This function wraps scikit-learn's [cross_validate](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html#sklearn.model_selection.cross_validate)
+    function, to provide more context
+    and facilitate the analysis.
+    As such, the arguments are the same as scikit-learn's cross_validate function.
+
+    Parameters
+    ----------
+    The same parameters as scikit-learn's cross_validate function, except for
+
+    project : Project or None
```
thomass-dev, line 172:
```suggestion
    project : Project, optional
```

thomass-dev (changes requested):
Pair reviewed with @rouk1 .

skore/src/skore/cross_validate.py
```diff
@@ -0,0 +1,235 @@
...
+    # Takes care both of train and test scores
+    return {
+        k: v
+        for k, v in cv_results.items()
+        if not any(added_scorer in k for added_scorer in added_scorers)
+    }
+
+
+def cross_validate(
+    *args, project: Optional[Project] = None, **kwargs
```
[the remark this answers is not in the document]
augustebaum: ruff complains about a SyntaxError when adding this; in any case, `project` is necessarily a keyword argument because we use `*args` and `**kwargs`.

MarieS-WiMLDS:
hellow!  
It looks like it's going to be published pretty soon :D!  
I spotted a few elements:  
1. the output produced by regression is not consistent with the output done by scikit-learn (cf code below)
2. I don't know if it's possible, but if: would it be possible to reset zoom when we change the metric? Otherwise since they are not the same magnitude sometimes, we completely loose the track of the new metric we want to display.
3. The cross_val plot in the UI doesn't update if the view is open. If I open another one and go back to the first, it works, but it doesn't update automatically like other variables.  

```python
from sklearn.datasets import load_diabetes
from sklearn import linear_model
from skore import cross_validate
from sklearn.model_selection import cross_validate as cross_validate_sk
diabetes = load_diabetes()
X = diabetes.data[:150]
y = diabetes.target[:150]
lasso = linear_model.Lasso()
cross_validate_sk(lasso, X, y, cv = 7, scoring = "r2")
cross_validate(lasso, X, y, cv = 7, project = project, scoring = "r2")
```

tuscland:
We need documentation about the feature, there are many interesting reasons why it has been designed like this, and we should reflect that.
@MarieS-WiMLDS just checking we are in sync as @sylvaincom is working on the larger documentation.

MarieS-WiMLDS (approved):

sylvaincom:
Hi, here are some important fix:
- [ ] For linear regression, it is redundant to say `test_r2` and `test_score`.

For the fit time:
Capture d’écran 2024-10-17 à 14 29 41

- [ ] We **must** specify the units (seconds for example) of the fit time
- [ ] and make the fit time more readable: 0.0025 seconds is not readable, use this for example:
```python
import datetime
print(str(datetime.timedelta(seconds=1_000)))
```
- [ ] We **must** distinguish scores (the higher, the better) and errors (the higher, the worse), for example this is an error but the title is still "Cross-validation scores" and the y-label is "Score": this is misleading 
Capture d’écran 2024-10-17 à 14 35 01
Maybe we can say something like "Cross-validation results" and the y-label is "Results", so that it can be appropriate for scores, errors, and times

- [ ] Why don't we include the test time and the train scores?

sylvaincom:
Agreed, `datetime.timedelta` displays nicely for large numbers of seconds but not decimal seconds
-> Maybe writing the following format: 1h23min or 23 ms is nice

tuscland:
@sylvaincom @augustebaum 

It feels like we can harvest some comments and put them in a fresh issue. Please cutoff the feature scope, so dev stops here, unless the last minute change must absolutely be included in Tuesday's release..

sylvaincom:
Follow-up issue: https://github.com/probabl-ai/skore/issues/544

rouk1:
👏

Issue #456, opened 2024-10-08
Dependabot's PR does not conform with our PR naming rules


rouk1:
### Describe the bug

PR title do not start with an uppercase letter.

### Steps/Code to Reproduce

Wait for dependabot to open a PR.

### Expected Behavior

Dependabot opens a PR the pass the `lint-pr-title` action.

### Actual Behavior

`lint-pr-title` action does not pass because PR name as no uppercase letter to start.

### Environment

```shell
Not related
```

PR #458, opened 2024-10-08
ci: Run dependabot monthly and change its commit-message prefix

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/456 .


.github/dependabot.yml
```diff
@@ -4,20 +4,20 @@ updates:
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
-      interval: "weekly"
+      interval: "monthly"
     commit-message:
-      prefix: "chore(github-actions dependencies):"
+      prefix: "chore(dependencies): GITHUB-ACTIONS"
```
rouk1, line 9: Will this fix #456 ?

augustebaum: I think it won't fit the rules either, as the format is sentence-case (it would work with e.g. "GitHub Actions"

thomass-dev: It works well :) 

```
❯ echo "chore(dependencies): GITHUB-ACTIONS" | npx commitlint --verbose
⧗   input: chore(dependencies): GITHUB-ACTIONS
✔   found 0 problems, 0 warnings
```

```
❯ echo "chore(dependencies): gITHUB-ACTIONS" | npx commitlint --verbose
⧗   input: chore(dependencies): gITHUB-ACTIONS
✖   subject must be sentence-case [subject-case]

✖   found 1 problems, 0 warnings
```

augustebaum (approved):

PR #489, opened 2024-10-14
feat(UI): Allow view items to be drag and drop

rouk1:
This PR enable drag'n'drop management of view items.

- Item of view can now be dragged from the left tree to the right pane at a specific position (including auto scroll)
- Item can be reorganized using drag'n'drop

**UI preview**

https://github.com/user-attachments/assets/5832ca08-63f1-4152-8834-956ea23583c9


tuscland:
In the video a the 9 seconds mark, I see you want to drop an item, but the drop fails. It is possible to make it so it succeeds?

rouk1:
> In the video a the 9 seconds mark, I see you want to drop an item, but the drop fails. It is possible to make it so it succeeds?

Done ✔️ 

See new video preview:


https://github.com/user-attachments/assets/f674d588-2e1e-4c89-a9bf-db110cc7058f

tuscland (approved):

rouk1:
Finally 😓 

🍾

PR #490, opened 2024-10-14
chore(python deps): Set lower bounds for python dependencies

### LINKED ISSUE #404: In pyproject.toml, define each requirement using range of versions
Actually, user can install skore without restrictions on the skore's dependencies. 
It means that user can use old (and incompatible) version of one of dependencies.
With an incompatible dependency version, some errors can occur obscurely to the user only at the runtime.

We can't be compatible with all of versions of all package.
The convention is to define a range of version between two majors release.

augustebaum:
Closes #404

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -1,4 +1,5 @@
 [project]
+
 name = "skore"
 description = "Tooling and assistance for data scientists to \"Own Your Data Science\""
 dynamic = [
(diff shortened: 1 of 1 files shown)


thomass-dev:
Please wait until the issue has been "solved".

thomass-dev:
PR automatically closed because we have rebased the main branch in order to clean the documentation commits.
Replaced by https://github.com/probabl-ai/skore/pull/847.

thomass-dev:
PR automatically closed because we have rebased the main branch to clean the documentation commits.

PR #516, opened 2024-10-17
feat(UI): Display index names if present

### LINKED ISSUE #495: feat(UI): Show the name of the indexes in dataframe widget
In a next iteration, i propose to nicely show the name of the indexes. For instance, 

image

_Originally posted by @thomass-dev in https://github.com/probabl-ai/skore/issues/493#issuecomment-2413829457_
            
---

It should be pertinent on a dataframe with multi-indexes.            

```python
In [6]: import pandas as pd
   ...: 
   ...: df = pd.DataFrame([{"age": ">70", "number": "1M", "gender": "M"}])
   ...: df.index = pd.MultiIndex.from_arrays([["france"], ["paris"], ["1"]], names=('country', 'city', 'district'))

In [7]: df
Out[7]: 
                        age number gender
country city  district                   
france  paris 1         >70     1M      M
```

rouk1:
Index names will now be visible. 
Limitation: sticky index columns are disable in this case has it is a pain to manage absolute coordinates. This could be tackled in another PR.

https://github.com/user-attachments/assets/a55cc3a9-177f-4ad2-bcfc-44ff4090c536

fixes #495


tuscland:
Can the user resize the index column?

rouk1:
No the table layout is set to [auto](https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout). I think it's the best compromise. Sizing columns needs that we switch to a fixed layout and implement our own table layout algorithm.

rouk1:
> Is it possible to prevent line wrap? (see second table in the screenshot)

It is, but we chose with @augustebaum to let the browser do line breaks when the cell contains spaces. Note that there is a white space in the test data. If users contribute large amounts of words sentence it will then print a the a text block instead of a infinite line.

TL;DR; as you wish but I recommend not to disable word warp.

augustebaum (approved):
Nice work!

PR #541, opened 2024-10-18
docs(project): Add docstrings to functions in project

MarieS-WiMLDS:
Add docstrings to the functions in the project class.

PR #545, opened 2024-10-21
fix(ci): Fix ruff being skipped in CI by prefixing destinations with `skore/` after #440

thomass-dev:
We can see in #541 that pre-commit doesn't execute ruff:

image

Now:

image


augustebaum (approved):
Nice, good catch

PR #553, opened 2024-10-21
feat: Add `/docs` used to deploy documentation to GitHub Pages

thomass-dev:
Following the [official documentation](https://docs.github.com/en/pages/getting-started-with-github-pages/configuring-a-publishing-source-for-your-github-pages-site#publishing-from-a-branch), i propose the current workflow to build the sphinx documentation:

1. (in the current iteration)
    Use a dedicated `docs/` directory to store documentation artifacts that will be automatically rendered by GitHub on GitHub Pages.
    Every time something changes in this directory on the `main` branch, the GitHub Pages will be refreshed.
    So you can manually edit the site content on a PR, and push that PR on `main` to be applied.
    Documentation can be managed independently of release.

2. (in the next iteration) 
    Populate `docs/` by triggering `sphinx` in a GitHub action when something changes in the `skore/` directory.    
    
----

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,5 +178,7 @@ tramp
 .\#*
 
 # Sphinx documentation
-doc/_build/
-doc/auto_examples/
+sphinx/build/
+sphinx/auto_examples/
+sphinx/generated/
+sphinx/sg_execution_times.rst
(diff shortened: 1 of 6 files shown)


thomass-dev:
We should think about how to provide documentation for each version of `skore` (like the drop-down selection of python documentation) in a final iteration.

PR #554, opened 2024-10-21
feat(UI): Scroll to the bottom when an item is added by click

### LINKED ISSUE #523: UI: Display newly added items at the top of the canvas instead of the bottom
**Summary**
When adding new elements to the canvas, the items are currently displayed at the bottom. To improve user experience and provide immediate visual confirmation of the action, new items should be added to the top of the canvas.

**Expected Outcome**
Newly added items should appear at the top of the canvas, giving users instant feedback that the action has been completed successfully.

**Acceptance Criteria**

- When a new item is added to the canvas, it should be positioned at the top rather than the bottom.
- Ensure the display update happens in real-time without affecting the current view or layout.
- Test to confirm that the behavior applies to all types of elements that can be added to the canvas.

rouk1:
fixes #523


augustebaum (commented):
After a manual test, I can confirm that the PR does what it says it does. On the other hand, I'm not 100% sure that this is the best behaviour for fixing #523.
Maybe someone else has a stronger opinion? Otherwise I'll merge this and we can change it if need be.

rouk1:
@MarieS-WiMLDS & @sylvaincom could you decide if this is what we want ?

MarieS-WiMLDS:
yes, it feels very natural to me!

MarieS-WiMLDS (approved):
yes, it feels very natural to me!

MarieS-WiMLDS (changes requested):
Erratum: When doing drag and drop, even if it's in the middle of the view, the screen scroll down to the bottom automatically. The item is positioned correctly in the view.

augustebaum:
> When doing drag and drop, even if it's in the middle of the view, the screen scrolls down to the bottom

Well spotted! This should now be fixed.

PR #561, opened 2024-10-22
feat(UI): use 14px almost everywhere

rouk1:
Most of the texts are now 14px.

Screenshot 2024-10-22 at 10 55 31


augustebaum (approved):
To be honest I don't see much of a difference, but at least it doesn't make the text smaller :stuck_out_tongue_closed_eyes:

PR #565, opened 2024-10-22
feat(project): Remove `on_error` argument to `Project.put`

### LINKED ISSUE #540: `Project.put` `on_error="warn"` by default which is astonishing
I see that [`Project.put`](https://github.com/probabl-ai/skore/blob/11aecbac0ea910944d38da2081a7df74f200c882/skore/src/skore/project.py#L42) has an argument `on_error`, this is so when "putting" several items at once (from a dictionary), we want to have the option to log a warning, or raise an error.

 `on_error="warn"` for single value `put`.

This is both surprising and not uniform with other Project APIs (e.g. `get` does not allow to control `on_error`).

We can make it a bit simpler, and remove the `on_error` argument. We would choose a unique behavior for both variants of the function:
- single value `put`: raise immediately.
- multiple value `put`: delay error raise to the end, collecting errors raised during individual `put`.

augustebaum:
This makes `put` raise upon error, instead of logging a warning.
When several key-value pairs are `put` at once, we `put` them in order, and stop and raise if we encounter an invalid pair.

Closes #540

### DIFF
--- a/skore/src/skore/project.py
+++ b/skore/src/skore/project.py
@@ -60,7 +60,7 @@ class Project:
         key : str | dict[str, Any]
             The key to associate with `value` in the Project,
             or dict of key-value pairs to add to the Project.
-        value : Any | None
+        value : Any | None, default: None
             The value to associate with `key` in the Project.
             If `key` is a dict, this argument is ignored.
         on_error : "warn" or "raise", optional
--- a/skore/src/skore/project.py
+++ b/skore/src/skore/project.py
@@ -48,6 +48,21 @@ class Project:
         If an item with the same key already exists, its value is replaced by the new
         one.
 
+        The dict format is the same as equivalent to running `put` for each individual
+        key-value pair. In other words,
+        ```python
+        project.put({"hello": 1, "goodbye": 2})
+        ```
+        is equivalent to
+        ```python
+        project.put("hello", 1)
+        project.put("goodbye", 2)
+        ```
+        In particular, this means that if some key-value pair is invalid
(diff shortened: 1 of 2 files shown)


skore/src/skore/project.py
```diff
@@ -52,95 +47,79 @@ def put(
...
+        then all the key-value pairs up to the first offending key-value pair will
+        be successfully inserted, *and then* an error will be raised.
 
         Parameters
         ----------
         key : str | dict[str, Any]
             The key to associate with `value` in the Project,
             or dict of key-value pairs to add to the Project.
-        value : Any | None
+        value : Any | None, default: None
```
thomass-dev, line 71:
```suggestion
        value : Any, optional
```

skore/src/skore/project.py
```diff
@@ -52,95 +47,79 @@ def put(
...
         Raises
         ------
         ProjectPutError
-            If the key-value pair(s) cannot be saved properly,
-            and `on_error` is "raise".
+            If the key-value pair(s) cannot be saved properly.
         """
         if isinstance(key, dict):
-            self.put_several(key, on_error=on_error)
+            self.put_several(key)
```
thomass-dev, line 81:
```suggestion
            for key, value in key_to_value.items():
                self.put_one(key, value, on_error=on_error)
```

thomass-dev: What do you think?

thomass-dev (changes requested):
Nice! Just minor thoughs.

thomass-dev (approved):

PR #566, opened 2024-10-22
feat(cross-validation): Show metric average and standard deviation in plot

### LINKED ISSUE #562: feat: Enhance skore.cross_validate with average and standard deviation

As a data scientist, I want to see the average of barplots displayed, and +/- standard-deviation. 

Sub issue of https://github.com/probabl-ai/skore/issues/544 

- [ ] Add the average and std on the plots, as done in the [scikit-learn MOOC](https://inria.github.io/scikit-learn-mooc/predictive_modeling_pipeline/02_numerical_pipeline_video_cross_validation.html):
Capture d’écran 2024-10-21 à 11 06 05

_No response_

### Additional context

_No response_

augustebaum:
Closes #562


sylvaincom:
Looking good! Is it possible to not display the average and std on the legend? It clutters too much, they should be visible by default
If it takes too much time, it will be for the next iteration
cc @augustebaum @MarieS-WiMLDS

augustebaum:
I thought about this; the problem is that they need to be in the legend to be clickable (so that you can remove it from view if you want). Otherwise, when all the metrics have the average + stddev shown the plot is unreadable.

skore/src/skore/item/cross_validation_item.py
```diff
@@ -54,15 +52,51 @@ def plot_cross_validation(cv_results: dict) -> plotly.graph_objects.Figure:
     fig = go.Figure()
```
sylvaincom, line 53:
```suggestion
tfytdytd
```

skore/src/skore/item/cross_validation_item.py
```diff
@@ -54,15 +52,51 @@ def plot_cross_validation(cv_results: dict) -> plotly.graph_objects.Figure:
...
 
     for col_i, col_name in enumerate(df.columns):
+        visible = True if col_i == 0 else "legendonly"
+        metric_name = dict_labels.get(col_name, col_name)
```
sylvaincom, line 56:
```suggestion
        visible = True if col_i == 0 else "legendonly" # blabla
        metric_name = dict_labels.get(col_name, col_name)
```

augustebaum:
The solution we reached is: show the average/stddev bars if and only if the metric is shown, and hide the average/stddev from the legend.

sylvaincom (approved):
LGTM, thanks!

Issue #525, opened 2024-10-17
UI: Transition toast messages from white to black and reposition to top-middle


anasstarfaoui:
**Summary**
Currently, the toast messages are displayed in white, but I think we can keep the same color for dark and white theme, which is black toasts Additionally, the toast notifications will be repositioned to the top-middle of the screen for better visibility and consistency.

**Expected Outcome**

- Toast messages should appear with a black background in dark mode, matching the current dark theme.
- Toast messages should be repositioned to the top-middle of the screen to provide immediate feedback in a more central location.

**Acceptance Criteria**

- Update the background color of toast messages to black when in dark mode.
- Reposition toast notifications to the top-middle of the screen.
- Ensure that toast notifications are still easily readable and accessible in both dark and light modes.
- Verify that all types of toast messages (e.g., success, error) maintain clear contrast with the new background color.

**Attached**
Here is a link with an example of everything we just said : https://www.figma.com/design/Qzjyxv0lPDU2kKtsH9dcp3/Skore---Workstation---High-fidelity-mock-ups---Q3?node-id=3875-13277&t=0kg64QBSvZDbnKy7-4

tuscland:
I like this, as a quick win.

@MarieS-WiMLDS @rouk1  WDYT?

MarieS-WiMLDS:
Let's go!

rouk1:
roger that

tuscland:
Sorry folks, I meant to put this comment in #523…

PR #567, opened 2024-10-22
feat(UI): Dismiss non-error toasts automatically after 20 seconds

rouk1:
New options have been added to the `addToast` function:

 - dismissible : Whether the toast can be dismissed by the user. Default is true.
 - duration : The duration in seconds for which the toast should be displayed. Default is Infinity.

Also, the toast area was moved to the top of the app.

UI preview (before the toast area was moved):

https://github.com/user-attachments/assets/3b2a0095-85f8-4de4-b038-8797df893d52


augustebaum (changes requested):
Doesn't work in dark mode:

https://github.com/user-attachments/assets/d66b8c0d-8acc-46cb-8401-040fa4ff2aae

rouk1:
> Doesn't work in dark mode: https://github.com/user-attachments/assets/86c769ac-58ef-4cfc-a314-e1c4465947a3

Toast now use same skin in dark & light mode as requested in #525. Toast area is now on center top of the user's viewport.

augustebaum:
Still doesn't work in dark mode

https://github.com/user-attachments/assets/89043a81-75f0-4079-b797-11c95e95f02e

But now:
- It also doesn't work in light mode (I can't see the feedback, like in dark mode)
- The buttons on the top bar (to collapse the side bar and to add a view) are covered by the ToastArea, so they don't work anymore
- The buttons to add items to the view also don't work -- not sure if this is from this PR.

I think the safest would be to remove the moving of the toast area from this PR, and put it in a new one. Meanwhile, I'll set this back to draft until @rouk1 comes back from vacation.

rouk1:
> Still doesn't work in dark mode
> 2024-10-23T11_18_25_screen_record.mp4
> 
> But now:
> 
>     * It also doesn't work in light mode (I can't see the feedback, like in dark mode)
> 
>     * The buttons on the top bar (to collapse the side bar and to add a view) are covered by the ToastArea, so they don't work anymore
> 
>     * The buttons to add items to the view also don't work -- not sure if this is from this PR.
> 
> 
> I think the safest would be to remove the moving of the toast area from this PR, and put it in a new one. Meanwhile, I'll set this back to draft until @rouk1 comes back from vacation.

The toast area does not intercept clicks anymore. Only the dismiss button will.
Can we check together why you do not see the countdown feedback ?

rouk1:
@augustebaum as discussed your version of firefox is outdated and does not support css @property rules.
https://caniuse.com/?search=%40property

augustebaum:
Yep thanks for going over it, all the issues seem to be solved. Eventually we might implement some kind of graceful degradation.

augustebaum (approved):

Issue #617, opened 2024-10-26
fix(PandasDataFrameItem): complex objects are persisted


koaning:
### Describe the bug

I tried running skore from Jupyter with this line:

```
! python -m skore launch "alpha-demo"
```

When I looked at the browser I saw this:

![CleanShot 2024-10-26 at 11 18 02](https://github.com/user-attachments/assets/c79d0eb3-f5e8-4b83-98d2-85d64752fee9)

The notebook contained this traceback. 

```
Running skore UI from 'alpha-demo' at URL http://localhost:22140/
ERROR:    Exception in ASGI application
Traceback (most recent call last):
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/fastapi/encoders.py", line 324, in jsonable_encoder
    data = dict(obj)
TypeError: cannot convert dictionary update sequence element #0 to a sequence

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/fastapi/encoders.py", line 329, in jsonable_encoder
    data = vars(obj)
TypeError: vars() argument must have __dict__ attribute

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/uvicorn/protocols/http/h11_impl.py", line 406, in run_asgi
    result = await app(  # type: ignore[func-returns-value]
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__
    return await self.app(scope, receive, send)
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/fastapi/applications.py", line 1054, in __call__
    await super().__call__(scope, receive, send)
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/starlette/applications.py", line 113, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 187, in __call__
    raise exc
  File "/Users/vincent/Development/probabl/.venv/lib/python3.10/site-packages/starlette/middleware/errors.py", line 165, in __call__
    await self.app(scope, receive, _send)
[...]

PR #621, opened 2024-10-28
fix(PandasDataFrameItem): Use JSON to serialize dataframe instead of pickle

thomass-dev:
Pandas Dataframes including complex objects (e.g. numpy arrays) were stored as-is.
They are now serialized using JSON to make them environment-independent.

---

Two native methods are available to serialize dataframe with multi-index, while
keepping the index names:                                                      
                                                                               
1. Using table orientation with JSON serializer:                               
   ```python                                                                   
   json = dataframe.to_json(orient="table")                                    
   dataframe = pandas.read_json(json, orient="table", dtype=False)             
   ```                                                                         
                                                                               
   This method fails when columns name is an integer.                          
                                                                               
2. Using record orientation with indexes as columns:                           
   ```python                                                                   
   dataframe = dataframe.reset_index()                                         
   json = dataframe.to_json(orient="records")                                  
   dataframe = pandas.read_json(json, orient="records", dtype=False)           
   ```                                                                         
                                                                               
   This method fails when the index has the same name as one of the columns.   
                                                                               
None of those methods being compatible, we decide to store indexes separately.

### DIFF
--- a/skore/src/skore/item/pandas_dataframe_item.py
+++ b/skore/src/skore/item/pandas_dataframe_item.py
@@ -56,8 +56,9 @@ class PandasDataFrameItem(Item):
         """
         The pandas DataFrame from the persistence.
 
-        Its content can differ from the original dataframe because it has been serialize
-        using pandas `to_json` function and not pickled, to be environment independent.
+        Its content can differ from the original dataframe because it has been
+        serialized using pandas' `to_json` function and not pickled, in order to be
+        environment-independent.
         """
         import io
 
@@ -99,7 +100,7 @@ class PandasDataFrameItem(Item):
             raise TypeError(f"Type '{dataframe.__class__}' is not supported.")
 
         # Two native methods are available to serialize dataframe with multi-index,
-        # while keepping the index names:
+        # while keeping the index names:
         #
         # 1. Using table orientation with JSON serializer:
         #    ```python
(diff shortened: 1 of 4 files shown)


skore/src/skore/item/pandas_dataframe_item.py
```diff
@@ -66,15 +88,43 @@ def factory(cls, dataframe: pandas.DataFrame) -> PandasDataFrameItem:
...
 
         if not isinstance(dataframe, pandas.DataFrame):
             raise TypeError(f"Type '{dataframe.__class__}' is not supported.")
 
-        instance = cls(dataframe_dict=dataframe.to_dict(orient="tight"))
-
-        # add dataframe as cached property
-        instance.dataframe = dataframe
+        # Two native methods are available to serialize dataframe with multi-index,
+        # while keepping the index names:
```
augustebaum, line 102:
```suggestion
        # while keeping the index names:
```

skore/src/skore/item/pandas_dataframe_item.py
```diff
@@ -66,15 +88,43 @@ def factory(cls, dataframe: pandas.DataFrame) -> PandasDataFrameItem:
...
+        #
+        # 1. Using table orientation with JSON serializer:
+        #    ```python
+        #    json = dataframe.to_json(orient="table")
+        #    dataframe = pandas.read_json(json, orient="table", dtype=False)
+        #    ```
+        #
+        #    This method fails when columns name is an integer.
```
augustebaum, line 110:
```suggestion
        #    This method fails when a column name is an integer.
```

augustebaum (changes requested):
Just some tiny changes

augustebaum (changes requested):
Would you mind adding a test where the dataframe includes a complex object, like a numpy array? Something to ensure that the https://github.com/probabl-ai/skore/issues/617 is actually solved

augustebaum (approved):

Issue #557, opened 2024-10-21
Cross_validate - improvement B.1 - Compare several cross_validate


MarieS-WiMLDS:
### Is your feature request related to a problem? Please describe.

As a DataScientist, I want to be able to compare the several cross_validation I did.  

### Describe the solution you'd like

![371590757-d7d2ff40-f163-46cf-a002-89dbf06661a3](https://github.com/user-attachments/assets/ef55b47d-49d1-4e8d-b505-51c1a7d86d4c)  
In a first iteration, to simplify things, the user will have to say when it's comparable. By default, everything lands in the same plot. As a user, I must be able to delete some runs (in case I ran too many of them).  
The "information" panel must display:  
- the training dataframe name + its uuid
- when relevant, the target name and its uuid
- the estimator name & parameters

The dataframe & the target are not saved because it would be too heavy. However, to recognize them, we can save: 
- their hash
- the following set {nb_col, nb_rows}. This is suboptimal, but good for an MVP. In an iteration, we can think about storing a skrub table report.

tuscland:
Superseded by #803

Issue #504, opened 2024-10-16
feat: Record each value of an item over time (history)


thomass-dev:
## Functional description

We have seen that for new features, such as cross-validation, we want to aggregate metrics over time.

We want to offer a way to save all updates to an item. This means saving items in an "appending" rather in "overriding" mode.
In this way, we can retain all the values of an item over time.

This feature would enable the following benefits:
- The user is confident they will not lose saved information
- The user has access to past item values
- The user can compare items
- This feature would open potential user feedback and discussion on the wider topic of tracking the evolution of results without having to code complex features

## UI/UX requirements

- visible and discoverable UI for navigating through item updates
- navigate easily through all existing updates of items (e.g. back and next buttons)
- when an item has no history, updates navigation is not visible or disabled, and we only see the creation date "Added on 2024-10-21 at 12:35:50 CET -> use browser's locale formats for date and time formatting).
- when an item has history, updates navigation is visible or active, and we see the last update date "Last updated on 2024-10-22 at 17:12:58 CET -> use browser's locale formats for date and time formatting)

## Future evolutions

The user can switch between punctual or historicized view.
This could work for:
- primitive numbers: plot the evolution of the value
- cross validation: display a historicized view of the cross validation runs (scatter plot with error bar)

---

## Technical Proposition: 5 phases

1. Change the way items are saved without breaking any API (project or UI).

2. Add a new endpoint (2.1) to retrieve all the values of an item from back-end to front-end.
Add a new selection button on widgets allowing to switch from one version to another ( reference @rouk1 ), using this last endpoint.

    We should think how to intelligibly represent all the versions of an item in this button: do we display only the date of the item, its content, etc.

    (2.1) `/item/<key>?<latest=true|false>`.
This new endpoint has a positional `<key>` parameter and returns, depending on a optional parameter `<latest>`, the latest version of the item, or all its versions.

3. Add a new developer API inside the project one, to browse programmatically all the versions of an item. 
I propose to add an new `ItemProxy` component, returned by `project.get_item`, which looks something like: 

```python
class ItemProxy:
    def latest() -> Item: ...
    def all() -> list[Item]: ...
```

4. Kill `diskcache` to efficiently browse the different versions of an item using SQL-like language.

5. Add a pruning system.

Related to #403 

## UI/UX solution

[...]

PR #624, opened 2024-10-28
feat: Record all the versions of an item in the back-end

augustebaum:
- The `ItemRepository` now works like a `dict[Key, list[Item]]` whereas before it worked like a `dict[Key, Item]`. In practice, the list of Items is ordered from oldest to newest.
- `ItemRepository` has a new method `get_item_versions` which gets the list of items corresponding to a key. The pre-existing `get_item` method itself gets the *latest* item corresponding to the key.

Co-authored-by: Thomas S <thomas@probabl.ai>

Addresses backend part of #504 without breaking end-user APIs.

First steps in comparison of items (e.g. as described by e.g. #557)


skore/src/skore/item/item_repository.py
```diff
@@ -64,33 +85,48 @@ def get_item(self, key) -> Item:
...
         item : Item
             The item to be stored.
         """
-        item_parameters = item.__parameters__
+        item = ItemRepository.__deconstruct_item(item)
 
         if key in self.storage:
-            item_parameters["created_at"] = self.storage[key]["item"]["created_at"]
+            items = self.storage[key]
+            item["item"]["created_at"] = items[0]["item"]["created_at"]
```
thomass-dev, line 125: Last-minute thought: What about retro-compatibility in this special case?
Actually, if we run this code on an existing project instance, it will break because `self.storage[key]` is not a list.
The skore file must be removed.

Should-we leave a message or ensuring retro-compatibility with `if key in self.storage and type(self.storage[key]) == list:`?

thomass-dev (changes requested):

skore/src/skore/item/item_repository.py
augustebaum: Good point; instinctively, I'm not sure we should add some code just for this purpose --- given the size of our user base, I think a message is more than enough.

On the other hand, I'm not sure how we could deal with this cleanly if we did have many users... some kind of DB migration?

skore/src/skore/item/item_repository.py
```diff
@@ -64,33 +85,48 @@ def get_item(self, key) -> Item:
...
             The retrieved item.
         """
-        value = self.storage[key]
-        item_class_name = value["item_class_name"]
-        item_class = ItemRepository.ITEM_CLASS_NAME_TO_ITEM_CLASS[item_class_name]
-        item = value["item"]
+        return ItemRepository.__construct_item(self.storage[key][-1])
 
-        return item_class(**item)
+    def get_items(self, key) -> list[Item]:
```
rouk1, line 90: I find it a bit confusing to just add a "s" at the end of `get_item`. What about being really explicit with something like `get_item_history` ?

augustebaum: Agreed, it definitely should be more explicit. However, so far `history` has been used to express all past versions, *without* the current one, so it wouldn't work here...

thomass-dev: I know that @augustebaum don't like it, but we can also use `get_item(key, *, latest=True) -> Union[Item, list[Item]]`.

skore/src/skore/item/item_repository.py
thomass-dev: Status quo: we're still in the beta phase, which means we can introduce breaking-changes without backward compatibility.

@MarieS-WiMLDS, we should advise in the corresponding release of this feature that user must drop/recreate its project.

MarieS-WiMLDS: Ok, I'll keep that in mind to say it in the release!  
Agree also that we can still do easily breaking changes for now.

skore/src/skore/item/item_repository.py
augustebaum: @thomass-dev asks: How about `get_all_items`?

augustebaum: Otherwise, let's stick with `get_items`

rouk1: What about `get_item_versions` ?

thomass-dev: Accepted by all third-parties.

rouk1 (approved):
Good job !

thomass-dev (approved):

Issue #594, opened 2024-10-23
fix(docs): In the skore cross validation example, the plotly graph is not full size in the rendering


MarieS-WiMLDS:

augustebaum:
I get the same behaviour in [`sphinx-gallery`'s documentation](https://sphinx-gallery.github.io/stable/auto_plotly_examples/plot_0_plotly.html) itself. It appears that this is a bug with either sphinx-gallery or plotly.

augustebaum:
https://github.com/sphinx-gallery/sphinx-gallery/issues/1394

augustebaum:
Related: when the Plotly figure is shown in the example, the resulting image can't be used as the thumbnail.

<details>
<summary>Detailed explanation; tldr plotly is buggy</summary>
Sphinx-gallery tells me to install orca in order to do this (see first few lines of https://sphinx-gallery.github.io/stable/auto_plotly_examples/plot_0_plotly.html), but [this issue](https://github.com/smarie/mkdocs-gallery/issues/49) in mkdocs-gallery is also relevant and is more up to date.
In fact, I tried the legacy orca method, but orca is not in PyPI so it's a no-go.

The new methd is to use plotly's kaleido, but when I tried to build the docs I got this:
https://github.com/plotly/plotly.py/issues/4832
</details>

In conclusion, using plotly images as thumbnails doesn't work yet, and the docs look like this as a result (note the big "BROKEN" image):
![Image](https://github.com/user-attachments/assets/09419e11-8103-4724-b7ab-dc60c5f89b58)

augustebaum:
I didn't expect to hit so many roadblocks. Our options (feel free to contribute more):
- De-prioritize and wait for the issues to be solved, then try again
- Instead take a screenshot of the resulting plot and use this, hoping it will be used automatically as the thumbnail
- Try out alternatives to `sphinx-gallery`

To be completely honest, I don't think the thumbnails bring much to the docs.

@MarieS-WiMLDS @sylvaincom

MarieS-WiMLDS:
I vote for a mix of 1 & 2: use a screenshot while this issue is fixed. And I agree that the thumbnails do not provide huge value, changing `sphinx-gallery` is way too much work compared to just have a thumbnail.

sylvaincom:
If you just unzoom on the page it autosizes automatically

PR #634, opened 2024-10-29
fix(docs): Add screenshot of cross-validation plot

augustebaum:
This is a quick fix to address #594.

In the end, I took a screenshot manually (by using the "Download plot as png" button of the Plotly widget) and displayed it with matplotlib.

Here are all the alternatives I considered, and what didn't work with them:
- Use SVG instead of PNG
	- For that, I'd need to use Plotly's `fig.write_image`, and the library that Plotly uses for this, Kaleido, is currently broken: https://github.com/plotly/Kaleido/issues/210
- Export plotly figure to a file during the example and read it 
	- For that I also need Plotly's `fig.write_image`
- Use Pillow for displaying, instead of matplotlib (which is not really made for that)
	- Only matplotlib is supported by sphinx-gallery, the library that renders the examples


examples/plot_03_cross_validate.py
```diff
@@ -36,3 +36,18 @@
 cv_results = cross_validate(lasso, X, y, cv=3, project=my_project_gs)
 
 my_project_gs.get_item("cross_validation").plot
+
+# %%
+# Because Plotly graphs currently do not yet properly render in our docs engine due to `a bug in Plotly <https://github.com/plotly/plotly.py/issues/4828>`_,
+# we also show a screenshot below.
+# Alternatively, if you zoom in or out the Plotly graph should display properly again.
```
MarieS-WiMLDS, line 43:
```suggestion
# Alternatively, if you zoom in or out your navigator window the Plotly graph should display properly again.
```

At first I understood that I had to zoom the plotly graph, therefore adding this small precision.

MarieS-WiMLDS (changes requested):
A tiny change, otherwise all good for me, it's much prettier, thanks Auguste!

examples/plot_03_cross_validate.py
augustebaum: done, thanks

MarieS-WiMLDS (approved):

PR #635, opened 2024-10-29
fix(UI): Accordions now emit actions even if last action is the same than the previous one

rouk1:
Multiple successive clicks to the same action of accordion where ignored by the previous implementation with models. We now use an injected event emitter to ensure that all actions bubbles up.


MarieS-WiMLDS (approved):
I'd appreciate a dev review, but on the functional side, it works!

skore-ui/src/stores/project.ts
```diff
@@ -100,7 +100,6 @@ export const useProjectStore = defineStore("project", () => {
   let _stopBackendPolling: Function | null = null;
   async function startBackendPolling() {
     _isCanceledCall = false;
-    await fetch();
```
augustebaum, line 103: Do we not want to `fetch` as soon as `startBackendPolling` runs?

skore-ui/src/components/TreeAccordion.vue
```diff
@@ -23,13 +22,9 @@ const props = defineProps<{
 const emit = defineEmits<{
   itemAction: [action: string, itemName: string];
 }>();
-const lastItemAction = ref<string | null>(null);
-const lastItemName = ref<string | null>(null);
 
-watch([lastItemAction, lastItemName], () => {
-  if (lastItemAction.value && lastItemName.value) {
-    emit("itemAction", lastItemAction.value, lastItemName.value);
-  }
```
augustebaum, line 32: Out of curiosity, do you remember why this condition was there in the first place?

augustebaum (commented):
Thanks, no changes but I have a few questions

skore-ui/src/stores/project.ts
rouk1: Yes and fetch already does that.
cf https://github.com/probabl-ai/skore/blob/1571501c06652b62bb0067569c02ac28205bf45e/skore-ui/src/services/utils.ts#L90

skore-ui/src/components/TreeAccordion.vue
rouk1: Ref's value can be null hence we check to please the compiler

augustebaum: Ah I see; I misread this and thought that we were checking whether the same item was clicked twice in a row, for some reason ^^'

PR #639, opened 2024-10-29
feat: Aggregate latest cross-validation and its history

### LINKED ISSUE #681: feat(cross-validation): Incrementally add cross-validation results to a plot
Derived from #557 

When a user calls `skore.cross_validate` several times, the results of each run should be added to a plot in the UI. The plot should have the run number on the x-axis, and the value of some user-chosen metric on the y-axis; see #557 for a mockup.

thomass-dev:
This adds a new item type, `CrossValidationAggregationItem`, which allows to both record new cross-validation data (at key `cross_validation`), and update an aggregation plot (at key `cross_validation_aggregated`) when `skore.cross_validate` is called.

Co-authored-by: Auguste Baum <auguste@probabl.ai>

Closes #681

### DIFF
--- a/skore-ui/src/components/PlotlyWidget.vue
+++ b/skore-ui/src/components/PlotlyWidget.vue
@@ -20,6 +20,13 @@ function makeLayout(): Partial<Layout> {
   };
 }
 
+async function buildPlot(containerValue: HTMLDivElement, plotData: any) {
+  const plot = await react(containerValue, plotData, makeLayout());
+  if (props.spec.frames) {
+    addFrames(plot, props.spec.frames);
+  }
+}
+
 const resizeObserver = new ResizeObserver(() => {
   if (container.value) {
     relayout(container.value, makeLayout());
@@ -29,10 +36,7 @@ const resizeObserver = new ResizeObserver(() => {
 onMounted(async () => {
   if (container.value) {
     resizeObserver.observe(container.value);
-    const plot = await react(container.value, props.spec.data, makeLayout());
-    if (props.spec.frames) {
-      addFrames(plot, props.spec.frames);
-    }
+    buildPlot(container.value, props.spec.data);
   }
 });
 
@@ -46,12 +50,9 @@ onBeforeUnmount(() => {
 watch(
   () => props.spec.data,
   async (newData, oldData) => {
-    if (container.value) {
-      if (!isDeepEqual(newData, oldData)) {
-        const plot = await react(container.value, newData, makeLayout());
-        if (props.spec.frames) {
-          addFrames(plot, props.spec.frames);
-        }
+    if (!isDeepEqual(newData, oldData)) {
+      if (container.value) {
+        buildPlot(container.value, newData);
       }
     }
   }
--- a/skore-ui/src/components/PlotlyWidget.vue
+++ b/skore-ui/src/components/PlotlyWidget.vue
@@ -1,6 +1,7 @@
 <script setup lang="ts">
-import { onBeforeUnmount, onMounted, ref } from "vue";
 import { addFrames, react, purge, relayout, type Layout } from "plotly.js-dist-min";
+import { onBeforeUnmount, onMounted, ref, watch } from "vue";
+import { isDeepEqual } from "@/services/utils";
 
 const props = defineProps<{
   spec: { data: any; layout: any; frames: any };
@@ -41,6 +42,20 @@ onBeforeUnmount(() => {
     purge(container.value);
   }
 });
+
+watch(
+  () => props.spec.data,
+  async (newData, oldData) => {
+    if (container.value) {
+      if (!isDeepEqual(newData, oldData)) {
+        const plot = await react(container.value, newData, makeLayout());
+        if (props.spec.frames) {
+          addFrames(plot, props.spec.frames);
+        }
+      }
+    }
+  }
+);
 </script>
 
 <template>
--- a/skore-ui/src/components/PlotlyWidget.vue
+++ b/skore-ui/src/components/PlotlyWidget.vue
@@ -1,6 +1,6 @@
 <script setup lang="ts">
-import { addFrames, newPlot, purge, relayout, type Layout } from "plotly.js-dist-min";
 import { onBeforeUnmount, onMounted, ref } from "vue";
+import { addFrames, react, purge, relayout, type Layout } from "plotly.js-dist-min";
 
 const props = defineProps<{
   spec: { data: any; layout: any; frames: any };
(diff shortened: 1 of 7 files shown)


skore/src/skore/ui/project_routes.py
thomass-dev, line 1: @augustebaum @rouk1 

The cross-validation object is currently exported to front-end as a single plot.
We should change this to export 2 plots, the single plot and the aggregated one.

rouk1:
I've marked this PR as draft as it sounds that the plot is not ready and its code wont go to main anyway.

skore/src/skore/ui/project_routes.py
augustebaum: For now, the aggregated plot is the responsibility of a new item type, `AggCrossValidationItem`.

skore-ui/src/components/PlotlyWidget.vue
```diff
@@ -28,7 +29,7 @@ const resizeObserver = new ResizeObserver(() => {
 onMounted(async () => {
   if (container.value) {
     resizeObserver.observe(container.value);
-    const plot = await newPlot(container.value, props.spec.data, makeLayout());
+    const plot = await react(container.value, props.spec.data, makeLayout());
     if (props.spec.frames) {
       addFrames(plot, props.spec.frames);
```
rouk1, line 34: Should this snippet be factorize in a `buildPlot` function and used here and in the watch ?

skore/src/skore/item/cross_validation_item.py
```diff
@@ -288,3 +373,68 @@ def factory(
     def plot(self):
         """A plot of the cross-validation results."""
         return plotly.io.from_json(self.plot_bytes.decode("utf-8"))
+
+
+class AggCrossValidationItem(Item):
```
rouk1, line 378: This naming is buzzing me. We do not use abbreviations that much why not something explicit ?
Maybe `CrossValidationsAggregationItem` ☕ style indeed but explicit.

skore/src/skore/cross_validate.py
```diff
@@ -246,9 +246,20 @@ def cross_validate(*args, project: Optional[Project] = None, **kwargs) -> dict:
...
+        try:
+            cv_results_history = project.get_item_versions("cross_validation")
+        except KeyError:
+            cv_results_history = []
+
+        agg_cross_validation_item = AggCrossValidationItem.factory(
+            cv_results_history + [cross_validation_item]
+        )
+
+        project.put_item("cross_validation_aggregated", agg_cross_validation_item)
```
rouk1, line 262: Could this item key be parametrized as a kwarg ? I suppose that a user may cross validate multiple models or just customize / put in a "folder" there results.

rouk1 (changes requested):
Correct me if I'm wrong:
When you call `cross_validate` once your project gets two new keys added automatically to it (cross_validation & cross_validation_aggregated) even if you only called `cross_validate` once  ?

Could it be easier to understand for a user if we add the aggregated plot to all versions of the item (except the first one maybe) then add a toggle in the UI to see the aggregated plot instead of the latest run ? This wont create extra keys in the top level of the project and will stick to the regular item behavior.

skore/src/skore/item/cross_validation_item.py
augustebaum: Done!

skore/src/skore/cross_validate.py
augustebaum: This is future work because we want to avoid changing the signature of `cross_validate` too much yet (we're discussing this kind of things a lot these days)

rouk1: Alright then !

augustebaum:
> When you call cross_validate once your project gets two new keys added automatically to it (cross_validation & cross_validation_aggregated) even if you only called cross_validate once ?

Yes

> Could it be easier to understand for a user if we add the aggregated plot to all versions of the item (except the first one maybe) then add a toggle in the UI to see the aggregated plot instead of the latest run ?

So version 0 has no aggregation, version 1 has aggregation of 0 and 1, version 1 has aggregation of 0, 1 and 2?

The toggle mechanic makes sense, but IMO it should be possible to show the two plots independently... In any case, I think it's out of scope for this PR

skore/tests/integration/test_cross_validate.py
```diff
@@ -157,3 +161,19 @@ def test_plot_cross_validation():
...
+        in_memory_project.item_repository.get_item("cross_validation_aggregated"),
+        CrossValidationAggregationItem,
+    )
+
+    cross_validate(*args, project=in_memory_project)
+    cross_validate(*args, project=in_memory_project)
+    cross_validate(*args, project=in_memory_project)
+    in_memory_project.item_repository.get_item(
+        "cross_validation_aggregated"
+    ).plot.show()
```
thomass-dev, line 179: Its a bit strange, at the top you test the item type, at the bottom you test that the plot can be shown.
What are you trying here?

skore/src/skore/item/cross_validation_item.py
```diff
@@ -196,7 +281,7 @@ def __init__(
         y_info : dict
             A summary of the target, input of scikit-learn's cross_validation function.
         plot_bytes : bytes
-            A plot of the cross-validation results, in the form of bytes.
+            A plot of the current cross-validation results, in the form of bytes.
```
thomass-dev, line 284: This change can be revert.

skore/src/skore/item/cross_validation_item.py
```diff
@@ -279,7 +364,7 @@ def factory(
             plot_bytes=plot_bytes,
         )
 
-        # Cache plot
+        # Cache plots
```
thomass-dev, line 367: This change can be revert.

thomass-dev (commented):
I can't request changes because i'm in the authors, but i have few minor comments.

I don't really like to create a separate item to store cross-validation things outside of the cross-validation, but i understand that it requires less changes on the front-end side. But, it works! Thanks for your work.

rouk1:
> The toggle mechanic makes sense, but IMO it should be possible to show the two plots independently... In any case, I think it's out of scope for this PR

Ok let's keep that in mind for a future work.

augustebaum (changes requested):
requesting changes just so that the PR doesn't get merged accidentally before @MarieS-WiMLDS or @sylvaincom could see it

MarieS-WiMLDS (approved):
Good for me for a first iteration!  
We will need to help identify the runs later.  

I like that when I delete the item cross_validation and I re-run another cross_val, the item cross_validation_aggregated is updated with only one run.  
I wonder if users will ask to have the mean displayed too. For now, I don't really imagine how we could do that nicely.

PR #641, opened 2024-10-30
feat(cross-validation plot): Show datetimes using timedelta format

augustebaum:
The metrics that end in `_time` now show more readable time when hovered
on.

Screen record:

https://github.com/user-attachments/assets/6ffcd799-c37e-4f30-8113-4357e2ea552c


sylvaincom (changes requested):
Looking good, in the hover could the feature name be something like `fit_time (timedelta)` instead of `fit_time (seconds)`?

sylvaincom:
LGTM, thanks!

sylvaincom (approved):
LGTM, thanks!

PR #645, opened 2024-10-30
feat(UI): CSS variable set is now consistent with mockups

rouk1:


skore-ui/src/assets/styles/_variables.css
```diff
@@ -1,102 +1,95 @@
...
-  --border-radius: 4px;
-
-  /* misc */
-  --transition-duration: 0.3s;
-  --transition-easing: ease;
-
-  /* brand colors */
-  --color-blue: #2329ff;
-  --color-orange: #f08b30;
+  --spacing-4: 0.28rem;
```
augustebaum, line 17: using rem is nice! do we even need to set `root-font-size`?

skore-ui/src/views/ComponentsView.vue
```diff
@@ -15,7 +15,7 @@ import DraggableList from "@/components/DraggableList.vue";
 import DropdownButton from "@/components/DropdownButton.vue";
 import DropdownButtonItem from "@/components/DropdownButtonItem.vue";
 import DynamicContentRasterizer from "@/components/DynamicContentRasterizer.vue";
-import { type EditableListItemModel } from "@/components/EditableList.vue";
+import EditableList, { type EditableListItemModel } from "@/components/EditableList.vue";
```
augustebaum, line 18: is this change here for a reason? otherwise, isn't the linter supposed to catch it?

skore-ui/src/assets/styles/_variables.css
rouk1: Yes we need to set the root size to 14px as designed. 
You can see that its done in `_typograpy.css`.

skore-ui/src/views/ComponentsView.vue
rouk1: For some reason the component page was broken because the import was missing. Sadly the linter does not catch it.

thomass-dev (approved):

Issue #644, opened 2024-10-30
fix(docs): Improve the getting started example


sylvaincom:
Do not show the broken plotly

PR #636, opened 2024-10-29
fix(docs): Improve the documentation for the new release

sylvaincom:
Fixing parts of https://github.com/probabl-ai/skore/issues/627 
- removed getting started and user guide tabs as they are in examples
- added a convincing getting started example following @koaning advice (thanks again)
- merged the two examples on the skore UI into one
- In `CONTRIBUTING.md`, made a link to the documentation
- Updated the GIF in the README and in the documentation
- In the cross-validation example, no longer use Auguste's screenshot but in the notebook I export a png that is then loaded
    - added `kaleido` to `pyproject.toml`

PR #646, opened 2024-10-30
docs: Fix various things

sylvaincom:
Minor changes following https://github.com/probabl-ai/skore/pull/636
Should fix https://github.com/probabl-ai/skore/issues/644


augustebaum (approved):

examples/plot_02_overview_skore_ui.py
```diff
@@ -104,7 +104,7 @@
 my_project_ui.get("my_string")
```
rouk1, line 104: Maybe we could outline that strings are interpreted as markdown by default and user can use markdown syntax out of the box ?

rouk1 (changes requested):

examples/plot_02_overview_skore_ui.py
sylvaincom: It is not said after?

Capture d’écran 2024-10-30 à 17 35 01

rouk1: Ok my bad.

rouk1 (approved):

PR #653, opened 2024-10-31
feat(cross-validation plot): Show time columns last and "test_score" first

### LINKED ISSUE #652: feat: `skore.cross_validate` should display the test score first
## Problem

Currently, as pointed by @koaning, `skore.cross_validate` displays the fit time first:

sphx_glr_plot_03_cross_validate_001

while the test score is more important for the user.

## Suggestion

In the legend, display the test score first. Display all scores first then all times after.

augustebaum:
Closes #652

### DIFF
--- a/skore/src/skore/item/cross_validation_item.py
+++ b/skore/src/skore/item/cross_validation_item.py
@@ -46,6 +46,14 @@ def plot_cross_validation(cv_results: dict) -> plotly.graph_objects.Figure:
 
     df = pandas.DataFrame(_cv_results)
 
+    # Move time columns to last and "test_score" to first
+    if "fit_time" in df.columns:
+        df.insert(len(df.columns) - 1, "fit_time", df.pop("fit_time"))
+    if "score_time" in df.columns:
+        df.insert(len(df.columns) - 1, "score_time", df.pop("score_time"))
+    if "test_score" in df.columns:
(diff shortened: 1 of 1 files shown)


skore/src/skore/item/cross_validation_item.py
```diff
@@ -46,6 +46,14 @@ def plot_cross_validation(cv_results: dict) -> plotly.graph_objects.Figure:
 
     df = pandas.DataFrame(_cv_results)
 
+    # Move time columns to last and "test_score" to first
```
MarieS-WiMLDS, line 49: doing these manipulations on the dataset seem a bit weird to me, isn't there a simpler way by using the [category_orders](https://plotly.com/python/legend/) function of plotly?

augustebaum: I didn't know about `category_orders`, thanks! Although, I can't find a way to use it in this case because it applies to categorical columns.
If you can see an easy way to do it, do let me know.

On the other hand, the dataframe only exists in the context of the function, so there should be no safety concerns.

sylvaincom (approved):
The feature behaves as expected, did not review the code

skore/src/skore/item/cross_validation_item.py
MarieS-WiMLDS: ok, then it's fine for me :) !

PR #657, opened 2024-10-31
fix(PandasSeriesItem): Use JSON to serialize series instead of pickle

thomass-dev:
Same problem as https://github.com/probabl-ai/skore/pull/621, but on `pandas.Series`.


augustebaum (dismissed):
Looks good to me

skore/src/skore/item/pandas_series_item.py
```diff
@@ -72,9 +100,23 @@ def factory(cls, series: pandas.Series) -> PandasSeriesItem:
...
+        #
+        # Using table orientation with JSON serializer:
+        #    ```python
+        #    json = series.to_json(orient="table")
+        #    series = pandas.read_json(json, typ="series", orient="table", dtype=False)
+        #    ```
+        #
+        #    This method fails when an index name is an integer.
+        #
+        # None of those methods being compatible, we decide to store indexes separately.
```
rouk1, line 114:
```suggestion
        # None of those methods being compatible, we store indexes separately.
```

augustebaum (approved):

PR #660, opened 2024-10-31
fix(Item): Use specific exception in `item.factory` to not swallow unexpected exceptions in `object_to_item`

thomass-dev:
Some `TypeError` exceptions are swallowed during item construction, but it should not:

image

Furthermore, during serialization, JSON can raise `TypeError`, that must not be catched.

### DIFF
--- a/skore/src/skore/item/__init__.py
+++ b/skore/src/skore/item/__init__.py
@@ -6,7 +6,7 @@ from contextlib import suppress
 from typing import Any
 
 from skore.item.cross_validation_item import CrossValidationItem
-from skore.item.item import Item
+from skore.item.item import Item, ItemTypeError
 from skore.item.item_repository import ItemRepository
 from skore.item.media_item import MediaItem
 from skore.item.numpy_array_item import NumpyArrayItem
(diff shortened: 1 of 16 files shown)


skore/src/skore/item/item.py
```diff
@@ -9,6 +9,10 @@
 from typing import Any, Optional
 
 
+class ItemTypeError(Exception):
+    """Item type exception."""
```
augustebaum, line 13:
```suggestion
    """Item type exception.
    
    Is raised when an attempt is made to convert an object to an Item, but the object's type is not supported.
    """
```

skore/src/skore/item/__init__.py
```diff
@@ -24,23 +24,23 @@ def object_to_item(object: Any) -> Item:
         PandasSeriesItem,
         NumpyArrayItem,
         SklearnBaseEstimatorItem,
-        CrossValidationItem,
```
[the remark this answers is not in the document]
augustebaum: Good catch

augustebaum (approved):
A positive change, thanks. Just added a suggestion for more precise docs.

skore/src/skore/item/item.py
thomass-dev: Applied.

rouk1 (approved):

Issue #628, opened 2024-10-29
fix(export from backend to frontend): Be resilient to item serialization failures


thomass-dev:
Actually, if one item can't be exported from backend to frontend (for instance https://github.com/probabl-ai/skore/issues/617), all the export phase will failled and the frontend will be empty.

We should skip the failed item, show a warning and go ahead with next items.

koaning:
Bonus points if the warning is explicit about which component caused the error and maybe also points to a document that explains the details on the docs.

rouk1:
We may also display "failing items" in the item tree with some special indicator (maybe leverage tooltips and warning icon) for easier tracking of "broken items". @anasstarfaoui wdyt ?

anasstarfaoui:
I think it could be a nice idea, but maybe we could do that even before, in the UI history @rouk1 ?

rouk1:
> On a technical issue, i can set the content of the item to `<not json-serializable> ` or whatever you want @rouk1 .

I would rather parse an consistent data structure. So we may add an `error` field to the `ProjectItem` interface. It's value could indeed be a message like "item is not json serializable" or maybe something more focused on how the user may fix the issue.

> I think it could be a nice idea, but maybe we could do that even before, in the UI history @rouk1 ?

Your right but we have to handle both cases. An item cant be displayed as a view card and one version of the item cant be viewed as a view card. 
This leads to another question. Should failing items be automatically removed from the history ?
@MarieS-WiMLDS & @sylvaincom wdyt ?

MarieS-WiMLDS:
> Should failing items be automatically removed from the history ?  
 
I don't think so. As a user, I would find it weird and I think I would try to load it again.  
I like better your idea to display these items in the left panel, but with an indicator.

anasstarfaoui:
Okay let's do that then, I'm just not a big fan of mixing broken and valid things inside the sidebar, I feel like it create visual clutter and distract user from using correctly the bar, especially without filters

But let's test it anyway, I will try to design something for this

anasstarfaoui:
Here is a link to what we just said, let me know if it's what you had in mind :)

https://www.figma.com/design/Qzjyxv0lPDU2kKtsH9dcp3/Skore---Workstation---High-fidelity-mock-ups---Q3?node-id=4327-13000&t=OYHGBd70M8Fpw7Mf-4

rouk1:
> Here is a link to what we just said, let me know if it's what you had in mind :)
> 
[...]

PR #661, opened 2024-10-31
fix(export from backend to frontend): Be resilient to item serialization failures

thomass-dev:
Part of https://github.com/probabl-ai/skore/issues/628.

Co-authored-by: Matthieu Jouis <matthieu@probabl.ai>

### DIFF
--- a/skore/src/skore/ui/project_routes.py
+++ b/skore/src/skore/ui/project_routes.py
@@ -1,10 +1,12 @@
 """The definition of API routes to list project items and get them."""
 
 import base64
-from dataclasses import dataclass
+import json
+from dataclasses import asdict, dataclass
 from typing import Any
 
 from fastapi import APIRouter, HTTPException, Request, status
+from fastapi.responses import JSONResponse
 
 from skore.item.cross_validation_item import CrossValidationItem
 from skore.item.media_item import MediaItem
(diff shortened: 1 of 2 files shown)


augustebaum:
Can you describe what this PR does, and potentially how to test it?

skore/src/skore/ui/project_routes.py
```diff
@@ -1,97 +1,135 @@
...
+            value = json.dumps(value, default=default)
+        except JSONError:
+            value = json.dumps(None)
+            error = True
+            traceback = format_exc()
+        else:
+            error = False
+            traceback = None
+
+        media_type = json.dumps(media_type)
```
rouk1, line 75: As this is a string shall we avoid useless serialization ?

skore/src/skore/ui/project_routes.py
```diff
@@ -1,97 +1,135 @@
...
+        error = json.dumps(error)
+        traceback = json.dumps(traceback)
+        created_at = json.dumps(item.created_at)
+        updated_at = json.dumps(item.updated_at)
```
rouk1, line 79: As those are strings shall we avoid useless serialization ?

skore/src/skore/ui/project_routes.py
```diff
@@ -1,97 +1,135 @@
...
+            value = json.dumps(None)
+            error = True
+            traceback = format_exc()
+        else:
+            error = False
+            traceback = None
+
+        media_type = json.dumps(media_type)
+        error = json.dumps(error)
+        traceback = json.dumps(traceback)
```
rouk1, line 77: Do you use json encoding to escape double quotes from the traceback ?
As the traceback wont be easily displayable in the frontend, could we have something more concise to present to the user ?

rouk1 (changes requested):
Nice work in making json encoding "composable" !

On one hand I find the implementation quite elegant. On the other hand I'm not sure that the maintainability of this is good. For example if we want to switch from arrays to dict and vice-versa we'll have to re-implement this.

The first approach we had was dirty but straight forward and relied on battle tested encoding routine (from fast API and standard lib). Do you think that we need this performance improvement more than flexibility on our exchange format ? (This is an open question.)

thomass-dev:
PR automatically closed because we have rebased the main branch to clean the documentation commits.

PR #669, opened 2024-11-05
feat(UI): Display item updates as a dropdown

rouk1:
- project store as been updated to manage updates
- updates are stored in session storage to limit memeory footprint
- card quick action have been introduced to have one click delete and browse history
- on click on an item update an event is emitted, the view listens to it and calls the store to update the list of `PresentableItem` reactivelly

UI preview:

https://github.com/user-attachments/assets/e156e38c-ec3f-4cb9-a461-7dbc9d2df61f


sylvaincom:
In the video, the "Updated about 17 hours ago" for a dozen histories of the item does not help the user track them? Maybe write the full log somewhere?

augustebaum:
Awesome stuff!!

augustebaum:
I found a small graphical bug:

https://github.com/user-attachments/assets/189aecb8-dabd-4f2c-87ed-eab8a872a753

To reproduce, open the menu, then quickly move the mouse away and then back

augustebaum:
Nitpick:
I guess the rule is that the dropdown disappears when I leave the card container?
so i can go far to the left and it stays open, but if i go slightly upwards it disappears immediately.
i would have expected something like a radial rule i suppose.

augustebaum:
The little warning symbol after "Created x minutes ago" is so small as to be indistinguishable IMO. I would put the "You are viewing an old update" on a separate line.

augustebaum (commented):
I found a small graphical bug:

https://github.com/user-attachments/assets/189aecb8-dabd-4f2c-87ed-eab8a872a753

To reproduce, open the menu, then quickly move the mouse away and then back

skore-ui/src/components/ProjectViewCard.vue
```diff
@@ -20,16 +38,35 @@ const emit = defineEmits<{
         <div class="title">{{ props.title }}</div>
         <div class="subtitle" v-if="props.subtitle">
           {{ props.subtitle }}
+          <Transition name="fade">
+            <span v-if="!isLatestUpdate" class="warning">
+              <i class="icon-warning"></i>
+              You are viewing an old update
```
augustebaum, line 44: About "You are viewing an old update": I know we said "version" should be kept for a later feature, but "an old update" sounds very weird. An update is new, by definition (it's "up to date", "à jour")

skore-ui/src/components/DropdownButtonItem.vue
augustebaum, line 1: Nitpick:
I guess the rule is that the dropdown disappears when I leave the card container?
so i can go far to the left and it stays open, but if i go slightly upwards it disappears immediately.
i would have expected something like a radial rule i suppose

skore-ui/src/components/ProjectViewCard.vue
```diff
@@ -20,16 +38,35 @@ const emit = defineEmits<{
         <div class="title">{{ props.title }}</div>
         <div class="subtitle" v-if="props.subtitle">
           {{ props.subtitle }}
+          <Transition name="fade">
+            <span v-if="!isLatestUpdate" class="warning">
+              <i class="icon-warning"></i>
```
augustebaum, line 43: The little warning symbol after "Created x minutes ago" is so small as to be indistinguishable IMO. I would put the "You are viewing an old update" on a separate line.

augustebaum (changes requested):
I found a small graphical bug:

https://github.com/user-attachments/assets/189aecb8-dabd-4f2c-87ed-eab8a872a753

To reproduce, open the menu, then quickly move the mouse away and then back

skore-ui/src/components/ProjectViewCard.vue
rouk1: Could you suggest some better wording ? Maybe @MarieS-WiMLDS & @tuscland have ideas ?

skore-ui/src/components/ProjectViewCard.vue
rouk1: Adding a second line is in my opinion not a good idea as it will lead to inconsistency in card height hence generating a scroll of everything below. Would a warning symbol without the circle do the job ?

skore-ui/src/components/DropdownButtonItem.vue
rouk1: The dropdown menu disappears in three cases:
- user clicks the button to close it
- user clicks outside
- dropdown gets scrolled outside of the viewport 

Could you explain what you mean by "radial rule" ?

rouk1:
> I found a small graphical bug:
> 2024-11-05T12_11_49_screen_record.mp4
> 
> To reproduce, open the menu, then quickly move the mouse away and then back

This is (to me ^^) not a bug. You did not close explicitly closed the drop down but moved your mouse outside of the actions which appears only on hovering. Hence actions dim out but there state persists.

What behavior would you expect ?

rouk1:
> In the video, the "Updated about 17 hours ago" for a dozen histories of the item does not help the user track them? Maybe write the full log somewhere?

Indeed it does'nt help but it's only due to my test dataset in which all updates have been generated in the same second. In a real context the dates should be very different. 

Would you like to add a number of something ? It was in the initial design but @tuscland said that giving version number to item update maybe confusing.

skore-ui/src/components/ProjectViewCard.vue
tuscland: Some suggestions:

- "You are viewing historical content" (my preferred)
- "You are viewing historical changes"
- "You are viewing past content"
- "You are viewing archived content"
- "You are viewing previous content"

I'm interested to know your feelings @augustebaum @MarieS-WiMLDS

skore-ui/src/components/ProjectViewCard.vue
augustebaum: Gotcha. Yeah, I think it would work better!

skore-ui/src/components/DropdownButtonItem.vue
augustebaum: In my manual tests it also disappears if the mouse moves away far enough (no need to click).

Yeah sorry that was unclear. In my user tests I saw that if I move e.g. 100 px to the left, the dropdown is still there, but if i go 100 px upwards, it disappears. I would expect that all directions are treated the same (i.e. the dropdown disappears if my mouse exits a circle that has radius 100px, for example).

skore-ui/src/components/ProjectViewCard.vue
augustebaum: In my opinion all of these wordings are strange, like we are deliberately avoiding the word "versions". Can you explain again why "versions" is not an option here?

augustebaum:
> This is (to me ^^) not a bug. You did not close explicitly closed the drop down but moved your mouse outside of the actions which appears only on hovering. Hence actions dim out but there state persists.
>
> What behavior would you expect ?

What's strange is that if move my mouse away, the dropdown disappears (so in my mind I think "okay, the dropdown is gone now") and then it's open again even though I haven't clicked  anything. I would expect to either keep it open, or keep it closed.

Another way of saying this is: I think moving the mouse far enough away should "explicitely" close the dropdown.

sylvaincom:
> > In the video, the "Updated about 17 hours ago" for a dozen histories of the item does not help the user track them? Maybe write the full log somewhere?
> 
> Indeed it does'nt help but it's only due to my test dataset in which all updates have been generated in the same second. In a real context the dates should be very different.
> 
> Would you like to add a number of something ? It was in the initial design but @tuscland said that giving version number to item update maybe confusing.

This "log naming" thing could be transformed into a separate more global issue? We will face the same with historization of cross-validation, we want something to identify the runs, and maybe give some comments about the run name:
- for run #2, I want to add a comment to say that I normalize the data
- I would like to be able to "squash" runs 2 to 5 because they are all wrong because they are on outdated versions on the data

WDYT of creating a new separated issue?

skore-ui/src/components/ProjectViewCard.vue
tuscland: It was mostly a defensive move that does not make a lot of sense after all. The feature I have in mind could use another name, so let's ditch this artificial constraint.

What about "You are viewing an older version"?

tuscland:
> Would you like to add a number of something ? It was in the initial design but @tuscland said that giving version number to item update maybe confusing.

Note that naming a version was not in the spec.

Think about managing versions for all items.
You could version your project as a whole, not individual items.

skore-ui/src/components/ProjectViewCard.vue
MarieS-WiMLDS: I'd prefer "you are viewing an old version" rather than "older", because it might spur the question "older than what?"

MarieS-WiMLDS (changes requested):
hello! 
I categorized my review in three parts:  

Things I loved :heart:  
- I love that the item bin is now easily accessible!
- I like the easy switch to latest version, it's helpful

Questions: 
- about version naming. I see that it's something discussed above. maybe we could call them iteration and just give them increasing numbers?
- didn't the plus button of items used to be orange?

cf this screen shot for the plus: 
image


Change requested:  

- [x] The warning logo, when I'm viewing an old item is way too small. I didn't understand what it was at first.
- [ ] Same for the icons of the right of items, they are a bit too small imo. 
- [ ] The "tick" in toaster doesn't seem aligned

cf this screenshot for the tick:  
image

skore-ui/src/components/ProjectViewCard.vue
tuscland: I prefer "you are viewing an old version" :)

tuscland:
I would not number item version because when working with many items you will not be able to easily compare those versions. Some items will "iterate" quickly while others not making it difficult to compare things.

skore-ui/src/components/ProjectViewCard.vue
augustebaum: "you are viewing an old version" sounds perfect to me

MarieS-WiMLDS:
About naming the iterations/versions/whatever instead of having "17 min ago" is important to me because I want to be able to find back which item i've seen the day before. If I find interesting to see iteration number 5, I want to be able to show it again to the rest of my team of whatever the following day, it must be easy to find if back.

skore-ui/src/components/ProjectViewCard.vue
rouk1: Screenshot 2024-11-06 at 15 44 39

The wording has been changed and the icon is a new one without the circle around it.

skore-ui/src/components/ProjectViewCard.vue
rouk1: Icon has been changed.

rouk1:
> About naming the iterations/versions/whatever instead of having "17 min ago" is important to me because I want to be able to find back which item i've seen the day before. If I find interesting to see iteration number 5, I want to be able to show it again to the rest of my team of whatever the following day, it must be easy to find if back.

I've added a simple version number counting from 1.
wdyt ?
Screenshot 2024-11-06 at 16 08 37

rouk1:
> * didn't the plus button of items used to be orange?

Nope, the idea is that views are orange and items blue.

rouk1:
> * Same for the icons of the right of items, they are a bit too small imo.

Accordion item actions are now 1.14rem.

skore-ui/src/components/DropdownButtonItem.vue
rouk1: Dropdown transition is now instant if the dropdown is invisible. I think this solves the glitch. Could you confirm ?

augustebaum: I don't see a change, I can reproduce the screenshot I made earlier.

Again, this is more or less a nitpick, I don't think it's a dealbreaker for this PR.

augustebaum (dismissed):

MarieS-WiMLDS:
I would still need all the icons (warning + branches + bin) to be a bit bigger please. 

> I've added a simple version number counting from 1. 

Very good for me! It's not pushed yet, is it?

skore-ui/src/components/DropdownButtonItem.vue
rouk1: I cant reproduce it anymore on my end. So I'll resolve this one.

rouk1:
> I would still need all the icons (warning + branches + bin) to be a bit bigger please.
> 
> > I've added a simple version number counting from 1.
> 
> Very good for me! It's not pushed yet, is it? (at least it doesn't display in my navigator)

Should be good now

Screenshot 2024-11-06 at 19 42 05

MarieS-WiMLDS (approved):
Thanks Matthieu, this will enable our next release tomorrow :) !!

PR #673, opened 2024-11-05
feat(UI): Disable drag and drop and + button for already added item

rouk1:
Bonus: now if you click an already added item the view will scroll to it and a small blink animation will be played.

UI preview:

https://github.com/user-attachments/assets/4e1eba53-2ebc-452f-b433-ce3ece6ba404


thomass-dev (approved):

thomass-dev:
Just a remark outside of the scope of this PR: its quite disturbing that i can't dnd items with my cursor on the middle on a line, but the selection cursor activates (and so the `+` icon) no matter where i am on the line.

Issue #663, opened 2024-10-31
fix(docs): Improve the example in the docstring of `skore.cross_validate`


sylvaincom:
The code snippet is not auto-contained

augustebaum:
Looks like this can be closed, given the discussions in #678? If you do close it make sure to summarize why

sylvaincom:
Yes, we can close it, we debated about it in #678 and this was not a strong opinion for me

PR #678, opened 2024-11-05
fix(docs): Improve the example in the docstring of `skore.cross_validate`

sylvaincom:
Closes https://github.com/probabl-ai/skore/issues/663

### DIFF
--- a/skore/src/skore/cross_validate.py
+++ b/skore/src/skore/cross_validate.py
@@ -218,17 +218,31 @@ def cross_validate(*args, project: Optional[Project] = None, **kwargs) -> dict:
 
     Examples
     --------
-    >>> def prepare_cv():
-    ...     from sklearn import datasets, linear_model
-    ...     diabetes = datasets.load_diabetes()
-    ...     X = diabetes.data[:150]
-    ...     y = diabetes.target[:150]
-    ...     lasso = linear_model.Lasso()
-    ...     return lasso, X, y
+    Without saving anything in a skore project:
 
-    >>> project = skore.load("project.skore")  # doctest: +SKIP
-    >>> lasso, X, y = prepare_cv()  # doctest: +SKIP
-    >>> cross_validate(lasso, X, y, cv=3, project=project)  # doctest: +SKIP
+    >>> from sklearn import datasets, svm
+    >>> import skore
+    >>> X, y = datasets.load_iris(return_X_y=True)
+    >>> clf = svm.SVC(kernel='linear', random_state=0)
+    >>> cv_results = skore.cross_validate(clf, X, y, cv=5)
+    >>> cv_results
+    {'fit_time': array(...), 'score_time': array(...), 'test_score': array(...)}
+
+    Saving the results in a skore project called ``my_project``:
+
+    >>> import subprocess
+    >>> from sklearn import datasets, svm
+    >>> import skore
+    >>> # create the skore project (deleting it beforehand)
+    >>> subprocess.run("rm -rf my_project.skore".split())
+    >>> subprocess.run("python3 -m skore create my_project".split())
+    >>> # load the project
+    >>> my_project = skore.load("my_project.skore")
+    >>> # load the data and perform the cross-validation
+    >>> X, y = datasets.load_iris(return_X_y=True)
+    >>> clf = svm.SVC(kernel='linear', random_state=0)
+    >>> cv_results = skore.cross_validate(clf, X, y, cv=5, project=my_project)
+    >>> cv_results
     {'fit_time': array(...), 'score_time': array(...), 'test_score': array(...)}
     """
     import sklearn.model_selection
--- a/skore/src/skore/cross_validate.py
+++ b/skore/src/skore/cross_validate.py
@@ -218,7 +218,7 @@ def cross_validate(*args, project: Optional[Project] = None, **kwargs) -> dict:
 
     Examples
     --------
-    Without saving anything in a skore project:
+    Classification task (without saving in a skore project):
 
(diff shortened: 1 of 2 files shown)


thomass-dev (changes requested):
Please leave the doctest disabled if we don't want the CI to explode.

skore/src/skore/cross_validate.py
```diff
@@ -218,17 +218,43 @@ def cross_validate(*args, project: Optional[Project] = None, **kwargs) -> dict:
...
-    >>> project = skore.load("project.skore")  # doctest: +SKIP
-    >>> lasso, X, y = prepare_cv()  # doctest: +SKIP
-    >>> cross_validate(lasso, X, y, cv=3, project=project)  # doctest: +SKIP
+    Classification task (without saving in a skore project):
+
+    >>> from sklearn import datasets, svm
+    >>> import skore
+    >>> X, y = datasets.load_iris(return_X_y=True)
+    >>> clf = svm.SVC(kernel='linear', random_state=0)
+    >>> cv_results = skore.cross_validate(clf, X, y, cv=5)
```
thomass-dev, line 227: ts more explicit to define the project when calling `cross_validate`: `cv_results = skore.cross_validate(clf, X, y, cv=5, project=project)`.

skore/src/skore/cross_validate.py
```diff
@@ -218,17 +218,43 @@ def cross_validate(*args, project: Optional[Project] = None, **kwargs) -> dict:
...
+    >>> cv_results
+    {'fit_time': array(...), 'score_time': array(...), 'test_score': array(...)}
+
+    Classification task saving the results in a skore project called ``my_project``:
+
+    >>> import subprocess
+    >>> from sklearn import datasets, svm
+    >>> import skore
+    >>> # create the skore project (deleting it beforehand)
+    >>> subprocess.run("rm -rf my_project.skore".split())
```
thomass-dev, line 237: I find your doctest quite complex, especially because you add some lines to setup the environment.
For instance `subprocess.run("rm -rf my_project.skore".split())`. I don't think its the place to.

thomass-dev (changes requested):
Please leave the doctest disabled if we don't want the CI to explode.

skore/src/skore/cross_validate.py
sylvaincom: I made several examples, the first one does not save anything, the second one does

sylvaincom: The examples seem short enough? It's 150 samples?

thomass-dev: Sorry, i read too fast and didn't see the 

> (without saving in a skore project)

skore/src/skore/cross_validate.py
thomass-dev: That's where we differ:  for me, doctest should only used to show usage example of the function, in particular to check that docstrings are up-to-date. IMO it shouldnt be used as tutorial.

@augustebaum , whats your feeling ?

skore/src/skore/cross_validate.py
thomass-dev: Please can you measure the execution time of these doctests? Furthermore, they failed.

skore/src/skore/cross_validate.py
augustebaum: I agree with Thomas that we should not put a full, working end-to-end tests in a docstring. The docstring is to give a short overview of how to use the function.
We already have the examples for real-life, copyable code.

sylvaincom: @MarieS-WiMLDS WDYT?

MarieS-WiMLDS: I've always been surprised by these subprocess run, I thought that were mandatory for the documentation to run. If they are not, I'd rather keep this code example as simple as possible.

augustebaum: They are more or less mandatory, but the issue here is whether the examples in *docstrings* need to be runnable at all costs. @thomass-dev and I say no (we already have runnable examples, so that should be enough); @sylvaincom says yes.

thomass-dev: The point is, we don't think that the `doctest` is the right place for a **complete** example (time-consumption in CI and maintenance cost). Here, IMO we should only focused the `doctest` on function usage example.

https://docs.python.org/3.13/library/doctest.html

> The doctest module searches for pieces of text that look like interactive Python sessions, and then executes those sessions to verify that they work exactly as shown. There are several common ways to use doctest:
> 
> - To check that a module’s docstrings are up-to-date by verifying that all interactive examples still work as documented.
> 
> - To perform regression testing by verifying that interactive examples from a test file or a test object work as expected.
> 
> - To write tutorial documentation for a package, liberally illustrated with input-output examples. Depending on whether the examples or the expository text are emphasized, this has the flavor of “literate testing” or “executable documentation”.

MarieS-WiMLDS: clearer now; I agree with Auguste & Thomas. As a user, in the docs, I want to have a very simple extract of code dedicated to the specific function I'm investigating. I'll be fed up if I'm always explained how to create a project at first (all the more that the explanation is not clear because it's not my real process to create a project).  

I agree that it would be better if we could have an automatic check of these functions though, so we don't not forget to update them as the code evolves. Maybe by having a main project for all the docs that we can use all the time?

thomass-dev:
@sylvaincom we can close the PR or the comment?

sylvaincom:
Close this PR :)

Issue #683, opened 2024-11-06
Implement basic `skore.train_test_split` with one warning


augustebaum:
Implements #492 

The first warning should be either the simplest to implement, or the one with highest priority, from the list in #492.

The warning message will be delivered using `warnings.warn`.

augustebaum:
The warning with highest priority, W4 "Use TimeSeriesSplit" is not super well defined so we'll start with the next most important, W1 "High class imbalance"

PR #690, opened 2024-11-07
feat: Implement basic `train_test_split`

augustebaum:
This PR implements the `train_test_split` function which wraps around sklearn.

The warnings are only emitted using `warnings.warn` for now; nothing is saved to a Project, as that part of the design still needs work.

In order to validate the technical design one warning is implemented, `HighClassImbalanceWarning`. This inherits from the built-in `Warning` class so that it can be used directly in `warnings.warn`.

Because we expect warnings to take a wide range of arguments, we pack all the relevant information into a `kwargs` dict inside `train_test_split`, and let the `check` static method accept `**kwargs` as well as the necessary positional arguments.

Closes https://github.com/probabl-ai/skore/issues/683

---

Todo:
- [ ] Add an example
- [ ] Test behaviour in a Jupyter notebook
	- [ ] Make warnings display nicely in Jupyter notebook: currently they show as unwrapped lines of text (so you need to scroll a lot to read the messages)
- [x] Deal with the case of `train_test_split(X, test_size=...)`; previously we took `arrays[-1]` as `y`, but now we only do this if there are at least 2 arrays
- [x] Refactor `_find_ml_task`, as a function of the same name already exists in `cross_validate.py`?


sylvaincom:
Thanks, I got some issues:

```python
from skore.sklearn.train_test_split import train_test_split
```
- [ ] should be `from skore import train_test_split `

```python
X = [[1]] * 4
y = [0, 1, 1, 1]
train_test_split(X, y)
```
- [ ] the warning message does not return to line in VS Code:

https://github.com/user-attachments/assets/80845d61-8dfc-40d4-8307-84b4701e4c80

skore/src/skore/sklearn/cross_validate.py
```diff
@@ -14,16 +14,16 @@
 )
 from skore.project import Project
 
-
-def _find_ml_task(
-    estimator, y
-) -> Literal[
+MLTask = Literal[
```
rouk1, line 17: As this type is used outside of the module, should we start a `types.py` file in project's root ?

skore/src/skore/sklearn/train_test_split.py
```diff
@@ -0,0 +1,252 @@
...
+        ml_task=ml_task,
+    )
+
+    for warning_class in [HighClassImbalanceWarning]:
+        check = warning_class.check(**kwargs)
+
+        if check is False:
+            warnings.warn(
+                message=warning_class.MSG, category=warning_class, stacklevel=1
+            )
```
rouk1, line 250: Probably in another task we could display those stings as formatted HTML if running in a notebook.

Maybe something like (not tested): 

```python
 with contextlib.suppress(ImportError):
        from IPython.core.interactiveshell import InteractiveShell
        from IPython.display import display

        if InteractiveShell.initialized():
            display("""
<div style="
  border: solid 1px var(--vscode-editorWarning-border, red);
  color: var(--vscode-editorWarning-foreground, white);
  background-color: var(--vscode-editorWarning-background, red);
">
  It seems that you have a classification problem with a high class 
  imbalance. In this 
  case, using train_test_split may not be a good idea because of high  
  variability in the scores obtained on the test set. 
  To tackle this challenge we suggest to use skore's 
  cross_validate function.
</div>""")
```

skore/src/skore/sklearn/cross_validate.py
thomass-dev: Instead of project root, why not in the `sklearn/__init__.py`?

skore/src/skore/sklearn/train_test_split.py
```diff
@@ -0,0 +1,252 @@
...
+    if type_of_target == "multiclass":
+        return "multiclass-classification"
+
+    if "continuous" in type_of_target:
+        return "regression"
+
+    if type_of_target == "unknown":
+        return "unknown"
+
+    return "unknown"
```
thomass-dev, line 53: Do you think you can factorize with `cross_validation._find_ml_task` ? 

```python
# sklearn/__init__.py

def guess_ml_task(y, estimator=None) -> MLTask:
```

skore/src/skore/sklearn/train_test_split.py
```diff
@@ -0,0 +1,252 @@
...
+    @staticmethod
+    def check(*args, **kwargs) -> bool:
+        """Perform the check.
+
+        Returns
+        -------
+        bool
+            True if the check passed, False otherwise.
+        """
+        ...
```
thomass-dev, line 70: To prepare the future iterations and to increase readability, how about splitting the codes something like:

```
sklearn/
|- cross_validate.py
|- train_test_split/
   | - __init__.py
   | - train_test_split.py
   | - warning/
      | - __init__.py
      | - interface.py
      | - first_warning_class.py
      | - second_warning_class.py
```

skore/src/skore/sklearn/train_test_split.py
```diff
@@ -0,0 +1,252 @@
...
+            return True
+
+        class_counts = pandas.Series(y).value_counts()
+
+        return (max(class_counts) / min(class_counts)) < 3
+
+
+def train_test_split(
+    *arrays: ArrayLike,
+    X: Optional[ArrayLike] = None,
```
thomass-dev, line 125:
```suggestion
    *arrays: ArrayLike,
    *,
    X: Optional[ArrayLike] = None,
```

skore/src/skore/sklearn/train_test_split.py
```diff
@@ -0,0 +1,252 @@
...
+    random_state: Optional[Union[int, RandomState]] = None,
+    shuffle: bool = True,
+    stratify: Optional[ArrayLike] = None,
+    project: Optional[Project] = None,
+):
+    """Perform train-test-split of data.
+
+    This is a wrapper over scikit-learn's `train_test_split https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html`_
+    helper function, enriching it with various warnings that can be saved in a Project.
+
```
thomass-dev, line 138:
```suggestion
    This is a wrapper over scikit-learn's `train_test_split https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.train_test_split.html`_
    helper function, enriching it with various warnings that can be saved in a Project.

    The signature is almost the same, but strictly compatible with base parameters.

```

skore/src/skore/sklearn/train_test_split.py
thomass-dev: +1 to control the way we show the warning in a notebook.

But, if i understand, you are proposing to print HTML content in a notebook, and raise a warning other-wise?
Isn't expose to side-effect if user can control its warnings (there is no more real warning in a notebook in your proposition, right?) ?

IMHO, a good trade-off would be to see how to customize the behavior of `warnings.warn` to our warnings in a notebook (i don't know if it's feasible).

rouk1: Globally yes, using `display` is also a way to enforce the visibility of our warnings as user may disable them.

rouk1: Globally yes, using display is also a way to enforce the visibility of our warnings as user may disable them.

skore/src/skore/sklearn/cross_validate.py
```diff
@@ -35,7 +35,8 @@ def _find_ml_task(
 
     Returns
     -------
-    Literal["classification", "regression", "clustering", "unknown"]
+    Literal["binary-classification", "multiclass-classification",
+    "regression", "clustering", "unknown"]
```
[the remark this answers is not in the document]
augustebaum: I thought about this but since this `MLTask` thing is more of a shortcut type I was reluctant to show it in user-facing docs... but here `_find_ml_task` is not public so maybe that's okay. Will apply

skore/src/skore/sklearn/train_test_split.py
rouk1: We discuss that and said that this may not be considered as a repetition of code until we need it in a third place ^^

skore/src/skore/sklearn/train_test_split.py
augustebaum: This results in `SyntaxError: Keyword-only parameter separator not allowed after '*' parameter`. Now we know...

skore/src/skore/sklearn/cross_validate.py
augustebaum: This results in a circular import. I can create `sklearn/types.py`?

skore/src/skore/sklearn/train_test_split.py
augustebaum: Putting it in `sklearn/__init__.py` would result in a circular import; where else could I put it? `sklearn/utils.py`?

skore/src/skore/sklearn/train_test_split.py
augustebaum: done

skore/src/skore/sklearn/cross_validate.py
augustebaum: done

skore/src/skore/sklearn/train_test_split.py
augustebaum: done

thomass-dev (approved):

PR #693, opened 2024-11-08
chore: Use dedicated `sklearn` directory for wrappers of sklearn functions

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/696


rouk1 (approved):

MarieS-WiMLDS (changes requested):
I understand the tech need and agree with you about keeping a close repository.  
for now, on the user side, I find it cumbersome without any reason for the users to ask them to add "sklearn" in their import. Is there a way to have a clean repo without adding this load to the user?

thomass-dev:
> I understand the tech need and agree with you about keeping a close repository. for now, on the user side, I find it cumbersome without any reason for the users to ask them to add "sklearn" in their import. Is there a way to have a clean repo without adding this load to the user?

If we know that, in a near future, other third-parties will be wrapped, we should now remove the alias. 
We won't be able to remove aliases easily because it implies API breaking-changes.

As i said in the issue, imagine two libraries with a same function that we want to adresss:

> Currently, if pytorch and scikit-learn have a cross_validation functions, we must change the signature of the functions:
> 
> ```python
> from skore import cross_validation_sklearn
> from skore import cross_validation_pytorch
> ```

MarieS-WiMLDS:
it's unlikely that we will wrap two different functions from two libraries which have the same function names (actually, even as an example, I can't think of anything). So we can keep the alias for now :)!

thomass-dev:
Ready to review.

thomass-dev:
Ready for review, as it doesn't have impact on user API anymore.

MarieS-WiMLDS (approved):
good for me, thanks :) !

PR #704, opened 2024-11-08
feat(UI): Rework view navigation

rouk1:
User can now navigate between their views using a dropdown visible in the page header.

- View list in dropdown leverage the previously built component. 
- The list item now have the option to be the selected one.
- Old sub view has been deleted.

UI preview:

https://github.com/user-attachments/assets/bf7949c5-1915-403d-83e4-f8d6e6ea9312

**In my opinion we should put the toasts back to the bottom of the viewport as tehy may conflict with this important element of the UI.**


augustebaum (dismissed):
Found a bug with the display of the selection: the tick mark doesn't show for the view that is actually selected

https://github.com/user-attachments/assets/ed6b6565-e91c-43cc-bcf6-04cba2bcbd85

augustebaum:
Toasts at the bottom are indeed likely better.

rouk1:
> Found a bug with the display of the selection: the tick mark doesn't show for the view that is actually selected
> 2024-11-14T15_40_29_screen_record.mp4

Nice catch thanks !
It should be fixed now.

thomass-dev (approved):

PR #705, opened 2024-11-08
ci: Skip checks after commits of `github-actions[bot]`

thomass-dev:
Currently, the documentation bot push a new commit after building the documentation using a special token to sign commits.
It implies to re-run the CI for nothing with the latest changes https://github.com/probabl-ai/skore/pull/703.

---

The solution is to skip checks after commits of `github-actions[bot]`.

### DIFF
--- a/.github/actions/sphinx/deploy/action.yml
+++ b/.github/actions/sphinx/deploy/action.yml
@@ -12,7 +12,8 @@ runs:
       run: |
         git config user.name 'github-actions[bot]'
         git config user.email 'github-actions[bot]@users.noreply.github.com'
+        git config commit.cleanup 'verbatim'
 
         git add docs/latest
-        git commit -m 'docs: Build documentation triggered by ${{ github.sha }}'
+        git commit -m $'docs: Build documentation triggered by ${{ github.sha }}\n\n\nskip-checks:true'
         git push


.github/actions/sphinx/deploy/action.yml
```diff
@@ -12,7 +12,8 @@ runs:
       run: |
         git config user.name 'github-actions[bot]'
         git config user.email 'github-actions[bot]@users.noreply.github.com'
+        git config commit.cleanup 'verbatim'
 
         git add docs/latest
-        git commit -m 'docs: Build documentation triggered by ${{ github.sha }}'
+        git commit -m $'docs: Build documentation triggered by ${{ github.sha }}\n\n\nskip-checks:true'
```
augustebaum, line 18: What is the `$` for?

thomass-dev: It's a [bash feature](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html) used to add newline characters (and more).

augustebaum: TIL, thanks!

rouk1 (approved):
Thanks for the 🌍

PR #713, opened 2024-11-12
docs: Use tempfolder instead of default project name

adrinjalali:
We should NEVER remove anything, like `rm -rf` from user's machine.


glemaitre:
I'm wondering if we could push this a bit further (in a next PR) where we could by default create the project in a temporary folder if no name is given.

adrinjalali:
> I'm wondering if we could push this a bit further (in a next PR) where we could by default create the project in a temporary folder if no name is given.

That might result in accidental loss of data by the users though (like when they restart their linux machine)

rouk1 (approved):

PR #721, opened 2024-11-12
fix(UI): Add an entry point to CLI to invoke skore

### LINKED ISSUE #717: Entry point for CLI
We need to declare the right "entry points" for our CLI, so that there's not need to write `python3 -m skore ...`

ref: https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#advanced-plugins

And then the commands become:

```sh
skore launch "my_project"
```

instead of 

```sh
python -m skore launch "my_project"
```

cc @glemaitre

glemaitre:
closes #717

This is a minimum change to add `skore` as the entry point as discussed in #717.


sphinx/getting_started.rst
```diff
@@ -31,7 +31,7 @@ Finally, from your shell (in the same directory), start the UI locally:
 
 .. code-block:: bash
 
-    python -m skore launch "my_project"
+    kkore launch "my_project"
```
glemaitre, line 34:
```suggestion
    skore launch "my_project"
```

skore/src/skore/cli/cli.py
```diff
@@ -9,7 +9,7 @@
 from skore.cli.quickstart_command import __quickstart
 
 
-def cli(args: list[str]):
```
augustebaum, line 12: Is there an advantage to this global variable approach? Despite how `parser.parse_args` behaves, I'd rather functions be as pure as possible

skore/tests/unit/cli/test_cli.py
```diff
@@ -19,14 +19,22 @@ def fake_launch(project_name, port, open_browser):
         launch_open_browser = open_browser
 
     monkeypatch.setattr("skore.cli.cli.__launch", fake_launch)
+    monkeypatch.setattr(
+        "sys.argv",
+        ["skore", "launch", "project.skore", "--port", "0", "--no-open-browser"],
+    )
 
-    cli(["launch", "project.skore", "--port", 0, "--no-open-browser"])
```
augustebaum, line 23: See comment about `cli`

thomass-dev (changes requested):
Please revert all changes on the `docs/` directory, a bot is in charge on main branch to build the documentation fro `sphinx/`.

skore/src/skore/cli/cli.py
glemaitre: Actually this is something back from when I moved in `click`. `click` will use decorator for that. I think that I can revert it in this case.

glemaitre:
> Please revert all changes on the docs/ directory, a bot is in charge to build the documentation from main:sphinx/.

I'll do. This is a weird pattern though. Why is the generation of the documentation committed in another repository?

thomass-dev:
> > Please revert all changes on the docs/ directory, a bot is in charge to build the documentation from main:sphinx/.
> 
> I'll do. This is a weird pattern though. Why is the generation of the documentation committed in this repository and not another branch or repository dedicated to the documentation?

Because we think it's really interesting to have in a same place the code and the built documentation.
User can just clone the repository and have a local access to the documentation.

glemaitre:
> User can just clone the repository and have a local access to the documentation.

If the documentation leave in the `gh-pages` then it would still be true. You only need to checkout the branch then.

thomass-dev:
@glemaitre the new Python API to create a project have been [merged](https://github.com/probabl-ai/skore/pull/744/files#).
Can you update your PR or you want someone from the dev-team to do it? If so, can we commit directly on your fork?

glemaitre:
I resolve the merge conflict. It should be good now.

skore/src/skore/cli/cli.py
```diff
@@ -66,21 +68,28 @@ def cli(args: list[str]):
...
+                overwrite=parsed_args.overwrite,
+            )
+        elif parsed_args.subcommand == "quickstart":
+            __quickstart()
+        else:
+            parser.print_help()
+        return 0
+    except Exception:
+        raise
+        return 1
```
thomass-dev, line 95: Raise and return are mutually exclusive, the return will never be reached. Can you rollback this changes?

skore/src/skore/cli/cli.py
```diff
@@ -66,21 +68,28 @@ def cli(args: list[str]):
         "quickstart", help='Create a "project.skore" file and start the UI'
     )
 
+    if args is None:
+        args = sys.argv[1:]
```
thomass-dev, line 72: I prefer the previous version, where the caller was in charge of passing parameters.

glemaitre: When creating the entry point, the executable that you get is the following:

```shell
#!/Users/glemaitre/Documents/packages/src/skore/.pixi/envs/dev/bin/python3.12                                                                                                                
# -*- coding: utf-8 -*-                                                                                                                                                                      
import re
import sys
from skore.__main__ import cli
if __name__ == "__main__":
    sys.argv[0] = re.sub(r"(-script\.pyw|\.exe)?$", "", sys.argv[0])
    sys.exit(cli())
```

So it means that no `args` are parsed explicitly. Are you aware of a way to pass the args explicitly in the entry point?

skore/src/skore/cli/cli.py
glemaitre: Yep I clean up those. Sorry about that.

skore/src/skore/cli/cli.py
thomass-dev: It's just a minor change but i would do something like (and take advantage of `rich` even in entry-point):

```python
# skore/src/skore/__main__.py

import sys
from skore.cli.cli import cli

def main():
    import rich.traceback
    
    # Display error tracebacks with Rich
    rich.traceback.install(show_locals=True)

    sys.exit(cli(sys.argv[1:]))
    
if __name__ == "__main__":
    main()
```

```toml
# skore/pyproject.toml

[project.scripts]
skore = "skore.__main__:main"
```

```python
# skore/src/skore/cli/cli.py

def cli(args: list[str]):
[...]
```

thomass-dev (approved):

thomass-dev:
Thanks @glemaitre for your contribution!

PR #722, opened 2024-11-12
refactor: Create a specific submodule to store custom exceptions

glemaitre:
Store all custom exceptions in a specific submodule for the CLI.


glemaitre:
If you have over exceptions in the library, I agree that the root folder makes more sense.

skore/src/skore/exceptions.py
```diff
@@ -0,0 +1,24 @@
+"""Custom exceptions for the CLI."""
```
thomass-dev, line 1:
```suggestion
"""Custom exceptions."""
```

thomass-dev (changes requested):
Thanks, its ready to merge for me except the suggested change. @augustebaum can you take a look?

thomass-dev (approved):

skore/src/skore/cli/launch_dashboard.py
```diff
@@ -13,12 +13,6 @@
 from skore.ui.app import create_app
 
 
-class ProjectNotFound(Exception):
-    """Project was not found."""
-
-    project_path: Path
```
thomass-dev, line 19: Note that this is not used anywhere.

PR #724, opened 2024-11-13
feat(UI): Allow to overwrite project directory

### LINKED ISSUE #719: Force the creation of project
Currently, one needs to remove manually the `.skore` folder when creating a new project.
It would be handy to have a `--force` option or `force=True/False` option to just overwrite the directory if desired.

Of course, it should not be the default but sometimes, you just want to go fast.

glemaitre:
closes #719 

This PR adds a parameter `overwrite` to `__create` such that it is possible to overwrite the project directory without the need to require calling `rm -rf ...`.

It translates into passing the flag `--overwrite` and calling the CLI: `skore create --overwrite`.


rouk1 (approved):
Nice improvement thanks.

thomass-dev:
@glemaitre can you please activate commits signature for future PRs? Thanks.

PR #729, opened 2024-11-13
fix(ci): Make `lint-pr-title` run on PR `synchronize`

augustebaum:
This fixes a bug where the workflow was skipped when a new commit was pushed to a PR with an offending title.


thomass-dev (approved):
Good point, thanks.

PR #739, opened 2024-11-14
docs: Improve example using the temporary folder and cleaning

glemaitre:
Proposes a couple of enhancement in the documentation.

- I reuse the pattern proposed by @adrinjalali. We discussed this morning IRL, and using the high-level API of the class seems better that the low-level functions.
- I move the import close to their first use because it is more intuitive when reading a notebook.
- Wrapping the line to have 88 characters in the description (so does it mean that linters are not run on those files?)
- I added links to the documentation instead of plain double back ticks when it makes sense.

### DIFF
--- a/examples/plot_01_getting_started.py
+++ b/examples/plot_01_getting_started.py
@@ -18,19 +18,23 @@ Creating a skore project, loading it, and launching the UI
 """
 
 # %%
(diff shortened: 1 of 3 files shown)


thomass-dev:
You true, currently the linter doesn't run on examples. I will change that https://github.com/probabl-ai/skore/issues/740!

Issue #740, opened 2024-11-14
ci: Run linter on `examples/`


thomass-dev:
Currently, the linter is only enabled on `^skore/(src|tests)/`.  The notebooks in `examples/` are not linted and can include mistakes.

PR #741, opened 2024-11-14
ci: Fix linter to run on `examples/`

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/740


examples/plot_01_getting_started.py
```diff
@@ -17,6 +17,8 @@
 ==========================================================
 """
 
+# ruff: noqa: E402
```
glemaitre, line 20: I would suggest to add the rule in the `pyproject.toml` to not have it in the example.

As an example: https://github.com/scikit-learn/scikit-learn/blob/main/pyproject.toml#L171-L174

glemaitre: It is less invasive and centralized.

thomass-dev: Its unfortunately not at the same level (we have chosen a mono-repo/multi-package architecture).
The configuration in `skore/` will not impact `ruff` on the `examples/` directory.
The examples are at the root level, and the python packages in dedicated sub-folders.

I wouldn't like to put examples in the `skore/` directory.

glemaitre: Arff, in this case, you can have a `ruff.toml` in the root.

thomass-dev: I made an in-between: i put the configuration in `examples/ruff.toml` to allow each package to define its own configuration without side-effect.

glemaitre (approved):
Indeed. It looks like a good trade-off.

PR #745, opened 2024-11-14
docs: Deploy documentation artifacts to S3

thomass-dev:
- Replace the use of `gh-pages` by S3 bucket: the commits history remains light and clean,
- Create a new version of the documentation after each release of `skore`,
- Add a drop-down list allowing user to select the last 10 versions (+ `dev`) of the documentation of `skore`,
- Running locally `make html` build documentation with a `dev` tag.

---

After each __commit__ on the `main` branch, the documentation is build and deployed on the `S3:dev/` directory.  
After each __release__, the documentation is build and deployed to the `S3:version/` directory, the version being
a subpart of the tag `MAJOR.MINOR.BUGFIX`: `MAJOR.MINOR`.                                                        
                                                                                                                 
For instance, with the following timeline:                                                                       
                                                                                                                 
```
       dev/ dev/ dev/ dev/ dev/                                                                                  
       0.1/                0.2/                                                                                  
main -- x -- x -- x -- x -- x -- >                                                                               
        |                   |                                                                                    
tag    0.1                 0.2       
```                                                                            
                                                                                                                 
The S3 bucket looks like:                                                                                        
```
    .                                                                                                            
    ├── 0.1/                                                                                                     
    ├── 0.2/                                                                                                     
    ├── dev/                                                                                                     
    ├── index.html                                                                                               
    └── versions.json
```


thomass-dev:
@jerome-probabl is the URL finally defined?

thomass-dev:
After closing this PR, i will:
- disable `gh-pages`,
- remove the deployment key,
- rebase all the documentation commits from the history.

.github/workflows/sphinx.yml
```diff
@@ -1,5 +1,52 @@
...
+#
+# ### How to change an old version of the documentation?
+# You can create tags wherever you want, i.e. on separate branches.
+# For example, you can create a branch from an old tag (`0.1.5`), modify the documentation and create a new tag (`0.1.6`).
+# The corresponding documentation will be automatically updated (`0.1`).
+#
+# **Side-notes**
+# ==============
+#
+# Only the 10 latest versions are listed in sphinx version-switcher.
```
augustebaum, line 47: If the bucket isn't auto-pruned, why would we stop showing the docs for older versions?

thomass-dev: Its just a security question: re-building the documentation for a specific version is tedious. I prefer we remove these versions by hand.

On the other hand, the user can access the version even if it is not in the version selector. All they have to do is specify the version on the URL

sphinx/conf.py
```diff
@@ -6,11 +6,13 @@
...
 
+import os
+
 project = "skore"
 copyright = "2024, Probabl"
 author = "Probabl"
-# version = "0"
-# release = "0"
+version = os.environ["SPHINX_VERSION"]
+release = os.environ["SPHINX_VERSION"]
```
jerome-probabl, line 15: If version is always equal to release variable why not keep only one of them?

.github/workflows/sphinx.yml
augustebaum: okay, i like that the older versions are still there even if the dropdown doesn't show them. Thanks for the explanation

.github/workflows/sphinx.yml
```diff
@@ -1,5 +1,52 @@
...
+# But, the bucket remains unchanged. You need to manually prune the bucket to remove older versions that are no longer listed.
```
rouk1, line 48: No need to prune this allows us to have legacy doc hosted somewhere without explicitly presenting it to the user. (It will still be parsed by search engines)

rouk1 (commented):
Shall we add a `stable` folder with the latest version of the documentation (a copy).
This could allow us to add a canonical link in our heads sections.

```html
<link rel="canonical" href="https://skore.probabl.ai/stable/" />
```

I think this is better SEO wise especially as shpinx switcher use permanent redirection (301).

For reference Skrub does that. 
Screenshot 2024-11-26 at 10 10 35

sphinx/conf.py
thomass-dev: Changed: release is now set with the tag.

.github/workflows/sphinx.yml
thomass-dev: Updated.

thomass-dev:
@rouk1 @jerome-probabl i propose to merge this version in state and refactor in a futur iteration with your comments.

rouk1 (approved):
ma-gni-fique 😻

PR #748, opened 2024-11-15
docs(fix): Force plotly figure resizing after HTML/CSS layout loading

glemaitre:
This should fix the issue observed on the documentation page.
It forces to resize the plotly figures once everything is loaded (and mainly the right sidebar).

Right now the bug is that the image is loaded and adapted before that the right sidebar is toggled (I think).

It feels like a hack but the rendering is better and our users should not see so much the difference since the figure are not shown as the first element of the page.

@rouk1 you might have a better understanding than me regarding all those JS stuff. So if this hack makes you think of the "right" way of fixing it, this could be cool.

### DIFF
--- a/examples/plot_01_getting_started.py
+++ b/examples/plot_01_getting_started.py
@@ -142,25 +142,7 @@ from skore import cross_validate
 cv_results = cross_validate(Ridge(), X, y, cv=5, project=my_project)
 
 fig_plotly = my_project.get_item("cross_validation").plot
-
-# %%
-# .. note::
-#   Because Plotly graphs currently do not properly render in our Sphinx
-#   auto-examples docs engine due to
-#   `a bug in Plotly <https://github.com/plotly/plotly.py/issues/4828>`_,
-#   we display its static image below.
-
-# %%
-import matplotlib.pyplot as plt
-import matplotlib.image as mpimg
-
-fig_plotly.write_image("plot_01_cross_validation.png", scale=4)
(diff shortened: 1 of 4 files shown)


glemaitre:
Just providing a quick demo (this is where having the preview is quite handy ;)):

https://github.com/user-attachments/assets/7ab73bcc-c53e-4e30-9fa1-bd01f1c2bc31

sylvaincom:
Thanks a lot, I don't go into the details of how you did it, but it works 🙏

thomass-dev (approved):

rouk1:
Hello @glemaitre, 
What you did is perfectly fine. One enhancement could be to use a resize observer.
I'm not very familiar with sphinx but it looks like the article as the classname `bd-article-container`. Hence you may write something like:


```javascript
function resizePlotlyGraphs() {
  // Find all plotly graph divs
  const plotlyDivs = document.getElementsByClassName("plotly-graph-div");

  // Iterate through each div and resize
  for (const div of plotlyDivs) {
      Plotly.Plots.resize(div);
  }
}

const resizeObserver = new ResizeObserver(() => {
  resizePlotlyGraphs();
});

document.addEventListener("DOMContentLoaded", () => {
  // Watch article size
  const article = document.querySelector(".bd-article-container");
  resizeObserver.observe(article);

  // Initial resize call after plots are created
  resizePlotlyGraphs();
});
```

Issue #731, opened 2024-11-13
cross_validate's signature, location, and purpose


adrinjalali:
Right now the signature is 

```py
cross_validate(*args, project, **kwargs)
```

This doesn't play well with auto-complete and IDEs and inspection. As a user it would be much nicer if the signature matches the scikit-learn's signature.

Another thing which would be nice is the import location. The nicest thing as a user would be to have to change only the top level import, as:

```py
from sklearn.model_selection import cross_validate
```

to

```py
from skore.model_selection import cross_validate
```

However, I wonder if it's a good pattern where we'd then need to basically patch a bunch of sklearn methods to add things to them or change their default values.

To me, it would be much more natural to get the output of sklearn's cross_validate, the estimator, and the data, and then generate what we need to show to the user, instead of calculating all of those on the fly in a custom implementation of `cross_validate`.

So in reality, I would remove this implementation, and have something like:

```py
from sklearn.model_selection import cross_validate
cv_results = cross_validate(..., return_estimator=True, return_indices=True)
project.put(X=x, y=y)
project.put("cv_result", cv_results=cv_results)
# We could generate from stored objects (names here), or actual values.
project.generate_scores_from(X="X", y="y", cv_results="cv_results")
```

cc @koaning @glemaitre

adrinjalali:
The docs also don't explain what really this function does: https://probabl-ai.github.io/skore/latest/generated/skore.cross_validate.html

adrinjalali:
What I was suggesting is more or less implemented in https://github.com/probabl-ai/skore/pull/749 now (thanks to @glemaitre )

glemaitre:
So by playing in #749, I got a couple of thoughts.

> save the predictions and/or decision function so that unplanned metrics can be later re-computed

Currently, as far I see, this is not the case. `skore.cross_validate` somehow use some caching because `sklearn.model_selection.cross_validate` will use `_MultimetricScorer` with multi-metric. However, the predictions are not persisted in a `skore` object.

So here, I would say that the wrapping is not necessary because we can manage the cache in a skore object that consume the `cv_results` from scikit-learn. The only necessary thing is that the user needs to be sure to call `sklearn.model_selection.cross_validate` with `return_estimators` and `return_indices` such that we can compute the any new metric on the fly (and persist whatever is required).

> check how it was called to generate warnings

I would say it depends on the warning.

[...]

PR #749, opened 2024-11-17
feat: Cross-validation reporter with accessors

glemaitre:
I give some thoughts to https://github.com/probabl-ai/skore/issues/731.

I wanted to see if I could come with an API that I would be happy with.

So here, this is a class implementing aggressive caching to avoid recomputation in line with the idea of the current `cross_validate`. Actually, it can go further because, we could make it work with custom scorer/metric where we would not require to recompute the model predictions if already in the cache.

In terms of API, I went on the pandas side with accessors. I find it quite need to extend upon the current scikit-learn display or metrics. In terms of feature, none of the current scikit-learn displays offer visualization for cross-validation. I implemented a dirty trick to show that we could go quite fast from today without the need to wait too much on the scikit-learn side. Once ready in scikit-learn, we could of course leverage the proposed design there.

Also, I used the `available_if` trick to expose methods that are only relevant with the estimators and target provided. If trying to access something not compatible, it will raise and explained why this is not supported.

So at this stage, there is no test and far to be the cleaner code ever but it allows to have discussion in terms of design and additional features.

### DIFF
--- a/examples/plot_04_alternative_cross_validate.py
+++ b/examples/plot_04_alternative_cross_validate.py
@@ -1,3 +1,11 @@
+"""
+==========================================================================
+Using a reporting class with accessors to display cross-validation results
+==========================================================================
+
+Just an example.
+"""
+
 # %%
 import tempfile
 from pathlib import Path
@@ -28,6 +36,8 @@ cv_results = cross_validate(
     classifier, X, y, return_estimator=True, return_indices=True
 )
 reporter = CrossValidationReporter(cv_results, X, y)
+
+# %%
 reporter.plot.roc(backend="plotly")
 
 # %%
@@ -51,6 +61,12 @@ reporter._cache  # stuff are still cached
 
 # %%
 reporter._hash
+
+# %%
+reporter.metrics.accuracy()
+
+# %%
+reporter.metrics.precision(positive_class=1)
 
 # %%
 # Cleanup the project
--- a/examples/plot_04_alternative_cross_validate.py
+++ b/examples/plot_04_alternative_cross_validate.py
@@ -31,6 +31,9 @@ reporter = CrossValidationReporter(cv_results, X, y)
 reporter.plot.roc(backend="plotly")
 
 # %%
+reporter.plot.roc(backend="matplotlib")
+
+# %%
 # I probably would like to do the following:
 # project.put("reporter", reporter)
 
@@ -44,6 +47,16 @@ reporter = joblib.load(temp_dir_path / "reporter.joblib")
 reporter.plot.roc(backend="plotly")
 
 # %%
-reporter._cache
+reporter._cache  # stuff are still cached
+
+# %%
+reporter._hash
+
+# %%
+# Cleanup the project
+# -------------------
+#
+# Remove the temporary directory:
+temp_dir.cleanup()
 
 # %%
--- a/examples/plot_04_alternative_cross_validate.py
+++ b/examples/plot_04_alternative_cross_validate.py
@@ -1,0 +1,49 @@
+# %%
+import tempfile
+from pathlib import Path
+
+temp_dir = tempfile.TemporaryDirectory(prefix="skore_example_")
+temp_dir_path = Path(temp_dir.name)
+
+# %%
+import subprocess
+
+subprocess.run(f"python3 -m skore create project --working-dir {temp_dir.name}".split())
+
+# %%
+import skore
+
+project = skore.load(temp_dir_path / "project.skore")
+
+# %%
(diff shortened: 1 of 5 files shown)


adrinjalali:
yeah this looks more like it.

glemaitre:
So some of those might be on the "cute" side but they are stuff that make me happy. I used `rich` for the help function:

image

This would restrain the output of what is available for the current cross-validation/estimator/data. Depending of the model, you will not get the same methods.

Instead of reusing, the `neg_` convention that is really difficult to explain to new comers, we could instead have the normal name of metric and use visual clues to mention if we deal with a loss or a score:

image

adrinjalali:
You probably don't wanna use green and red though 😁

glemaitre:
> You probably don't wanna use green and red though

Indeed :). Orange and blue are you friend

MarieS-WiMLDS:
I like that it splits back and front. It means that if the user wants to play with CrossValidationReporter, they can do it immediately, they don't have to get an object with un unclear type, since they are not the ones who put it.

glemaitre:
> and they'll have to pass clf again as well, to enable the use-case of tracking the model performance)

I'm not sure what this refers to here. Do you mean that `clf` would be fit and the full `X` and `y` (or single split)? This is seems unrelated to cross-validation in this case. If this is about cross-validation, I think that the `cv_results` will contain all the necessary estimators then.

glemaitre:
> Also, with the second option, put is delayed until the last step, so until the user puts, they're in danger of losing their work; this is especially dangerous in a notebook.

Actually, I might be OK having both option:

```python
reporter = CrossValidationReporter(cv_results, X, y)
project.put("my_cv_results", reporter)
```

and

```python
reporter = CrossValidationReporter(cv_results, X, y, store=project)
```

Since the `CrossValidationReporter` is not anymore close to a scikit-learn function/method. I would still be ambivalent and let the user decide the way he wants it. I could imagine that the first way of doing, we could store everything in RAM until the `put` call while the second could allow to persist in the database/disk (no idea how skore is performing on this right now) directly.

MarieS-WiMLDS:
## UX changes

Suggestion to avoid asking the user to do "return_xxx", asking them to give X and y again, and to allow for adding scorings as the current feature does.

With this piece of code, suggested by Guillaume in a discussion this afternoon: 

```python
class CrossValidationReporter:
  def __init__(self, cv_results, X, y):
    """You should probably use `from_estimator` or `from_cv_results`."""
    self.cv_results = cv_results
    self.X = X
    self.y = y
```

We would have this change.   

| before | after |
| --- | --- |
| <pre> from sklearn.model_selection import cross_validate <br> cv_results = cross_validate(classifier, X, y) </pre> | <pre> from skore import CrossValidationReporter <br> reporter = CrossValidationReporter(classifier, X, y) <br> cv_results = reporter.cv_results_ |

And the second option offers all the functions listed in the top comment.

This would answer the need: 
- change quite simple for the user, although not transparent
- possible for them to have exactly the same thing as before

## Suggestion about the "put" and parameter "project".

I'm in favor of having both. Setting `project = project` wouldn't put actually, but it would make it appear in the activity feed. This way, we avoid overloading the items list. So far, our testers are a bit surprised to see something appear in the items sections while they didn't put it, while they just learned that they can make items appear by doing using the put function. 

## Current features

Some of the current features are not included in this rework and will need to be added back: 
- add scorings (value: avoid having to rerun everything when you notice that you didn't chose correctly the scoring)
- aggregation: I think it cannot be included in the object CrossValidationReporter, it doesn't make sense. I think it would be more something like a new object, CrossValidationComparator.

## Questions remaining

imo  
- [ ] how does the object reporter appears in a view? What is displayed?
- [ ] what does the CrossValidationComparator? What are its inputs?

sylvaincom:
Quote from @MarieS-WiMLDS:
> Two backends
> For now, following a decision taken this morning with the design team (cf this [notion page](https://www.notion.so/probabl/2024-11-22-skore-workshop-with-Ga-l-and-Guillaume-146ef76d36b9808d86cec6da33699d24) for probablers), we will keep only one backend, plotly, to keep it simple and easy to develop.

In his code, @glemaitre does `from sklearn.metrics import RocCurveDisplay` so the matplotlib plots are already coded inside of scikit-learn, we only need to add the plotly version ; so if we are to go with plotly, we might as well keep matplotlib (quick-win because in scikit-learn) and thus have both?

thomass-dev:
@glemaitre your PR was automatically closed because we rebase the main branch in order to clean the documentation commits.

In addition, your issue is superseded by the issue #807. It will be the base of our implementation.

PR #791, opened 2024-11-21
shuffle true warning

augustebaum:
- **refactor(train_test_split): Remove dead tests**
- **feat(train_test_split): Add "shuffle is True" warning**


rouk1 (approved):
👍

PR #792, opened 2024-11-21
feat(project): Add first-class support for polars dataframes

### LINKED ISSUE #758: feat: Add polars support

Following https://github.com/probabl-ai/skore/issues/712, we should use polars more than pandas for data frames

The skore project should support Polars dataframes as items

augustebaum:
- Copy-pasted all the pre-existing code concerning pandas into its polars equivalent
- Adapted as needed
	- For instance, polars dataframes don't have an Index, so I deleted all mentions of it in the polars version
	- Unlike pandas, polars raises a `ComputeError` when serializing a DataFrame that contains a complex object, e.g. a numpy array.
- Added the Polars Items to the UI routes
	- Used `to_pandas` which requires `pyarrow`

It was [suggested](https://github.com/probabl-ai/skore/pull/792#issuecomment-2498980110) that we might use narwhals for the implementation. For this PR I went without it, for these reasons:
- I wanted to avoid adding a dependency (unfortunately, I still ended up adding pyarrow for the specific case of transferring data from a Project to skore-ui).
- I have never used narwhals and wanted to avoid a situation where I spend time prototyping only to realize that it doesn't perfectly fit the use-case (for example, I see that narwhals does not expose a `to_json` method).
- We can always reconsider this if we implement support for another DataFrame library.

Closes #758

TODO:
- [x] Write ADR about narwhals
- [x] Deal with complex objects in polars dataframes

### DIFF
--- a/skore/tests/unit/item/test_pandas_dataframe_item.py
+++ b/skore/tests/unit/item/test_pandas_dataframe_item.py
@@ -53,7 +53,8 @@ class TestPandasDataFrameItem:
         dataframe = DataFrame([{"key": np.array([1])}], Index([0], name="myIndex"))
         item = PandasDataFrameItem.factory(dataframe)
 
-        assert isinstance(item.dataframe["key"].iloc[0], list)
+        # NOTE: isinstance would not work because numpy.array is an instance of list
+        assert type(item.dataframe["key"].iloc[0]) is list
 
     @pytest.mark.order(1)
     def test_dataframe_with_integer_columns_name_and_multiindex(self, mock_nowstr):
--- a/skore/tests/unit/item/test_pandas_dataframe_item.py
+++ b/skore/tests/unit/item/test_pandas_dataframe_item.py
@@ -53,7 +53,7 @@ class TestPandasDataFrameItem:
         dataframe = DataFrame([{"key": np.array([1])}], Index([0], name="myIndex"))
         item = PandasDataFrameItem.factory(dataframe)
 
-        assert type(item.dataframe["key"].iloc[0]) is list
+        assert isinstance(item.dataframe["key"].iloc[0], list)
 
     @pytest.mark.order(1)
     def test_dataframe_with_integer_columns_name_and_multiindex(self, mock_nowstr):
--- a/skore/tests/unit/item/test_pandas_series_item.py
+++ b/skore/tests/unit/item/test_pandas_series_item.py
@@ -53,7 +53,8 @@ class TestPandasSeriesItem:
         series = Series([np.array([1])], Index([0], name="myIndex"))
         item = PandasSeriesItem.factory(series)
 
-        assert isinstance(item.series.iloc[0], list)
+        # NOTE: isinstance would not work because numpy.array is an instance of list
+        assert type(item.series.iloc[0]) is list
 
     @pytest.mark.order(1)
     def test_series_with_integer_indexes_name_and_multiindex(self, mock_nowstr):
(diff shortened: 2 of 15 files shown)


skore/tests/unit/item/test_pandas_dataframe_item.py
```diff
@@ -53,7 +53,7 @@ def test_dataframe_with_complex_object(self, mock_nowstr):
         dataframe = DataFrame([{"key": np.array([1])}], Index([0], name="myIndex"))
         item = PandasDataFrameItem.factory(dataframe)
 
-        assert type(item.dataframe["key"].iloc[0]) is list
+        assert isinstance(item.dataframe["key"].iloc[0], list)
```
thomass-dev, line 56: Please revert this changes: `np.array` is an instance of list.

skore/tests/unit/item/test_pandas_series_item.py
```diff
@@ -53,7 +53,7 @@ def test_series_with_complex_object(self, mock_nowstr):
         series = Series([np.array([1])], Index([0], name="myIndex"))
         item = PandasSeriesItem.factory(series)
 
-        assert type(item.series.iloc[0]) is list
+        assert isinstance(item.series.iloc[0], list)
```
thomass-dev, line 56: Please revert this changes: `np.array` is an instance of list.

skore/tests/unit/item/test_pandas_dataframe_item.py
augustebaum: uh, oh, sorry... ^^'

skore/tests/unit/item/test_polars_series_item.py
```diff
@@ -0,0 +1,68 @@
...
+
+        item1 = PolarsSeriesItem.factory(series)
+        item2 = PolarsSeriesItem(
+            series_json=series_json,
+            created_at=mock_nowstr,
+            updated_at=mock_nowstr,
+        )
+
+        assert_series_equal(item1.series, series)
+        assert_series_equal(item2.series, series)
```
thomass-dev, line 68: This test is not required, its a duplication of `def test_series(self, mock_nowstr):` as long as you don't have index in your series.

skore/tests/unit/item/test_polars_series_item.py
```diff
@@ -0,0 +1,68 @@
...
+
+    @pytest.mark.order(1)
+    def test_series_with_complex_object(self, mock_nowstr):
+        series = Series([np.array([1, 2])])
+        item = PolarsSeriesItem.factory(series)
+
+        assert isinstance(item.series[0], Series)
```
thomass-dev, line 50: `assert_series_equal` doesn't work here?

thomass-dev (changes requested):
Strong and easy job (except for your questions), nice!

skore/tests/unit/item/test_polars_series_item.py
augustebaum: The types inside the Series aren't the same because the numpy.array is converted to a list:
```
>>> item.series
shape: (1,)
Series: '' [list[i64]]
[
        [1, 2]
]
>>> series
shape: (1,)
Series: '' [array[i64, 2]]
[
        [1, 2]
]
```

I'll use `assert (item.series == series).all()` to compare the individual elements

thomass-dev: Yes i understand the types are not the same, but i supposed as pandas do, you can ignore dtypes.

PR #808, opened 2024-11-25
feat(UI): Accordion's item are now alpahabetically sorted

rouk1:
In a case insensitive fashion.

Note that the "self" items stays in top position if it exists.


augustebaum (approved):
Works for me!

PR #810, opened 2024-11-25
feat: Add support for skrub's TableReport

rouk1:
This PR adds a `SkrubTableReportItem`.

Usage: 

```python
import pandas as pd
from skore.project import create, load
from skrub import TableReport

create(project_name="project.skore")
s = load("project.skore")
df = pd.DataFrame(dict(a=[1, 2], b=["one", "two"], c=[11.1, 11.1]))
report = TableReport(df)
s.put("table report", report)
```

UI preview:

Screenshot 2024-11-25 at 17 55 43

Limitation sounds like there is no dark mode for now.


tuscland:
I like the fact that we have this feature in a basic form, allowing the user to track table reports. This also promotes skrub in our documentation.  

Can we hack the font so it looks better?

jeromedockes:
thanks for adding this! indeed that's very nice for skrub :)

- @rouk1 you are right there is no dark mode for now, we plan to add one but we've considered it low-priority -- if it's important for skore we can try to do it soon
- @tuscland if you're using the latest skrub release (0.3.1) the table report should use the same font as its parent element. if that's not the case please LMK or if you can open an issue on the skrub repo. Or, as I've noticed it is put in an iframe, you can use `TableReport.html()` rather than `TableReport.html_snippet()` to get a full page instead of just a fragment, and then the report would show in some default sans-serif font
- as a side note we're planning a skrub release tomorrow that contains some improvements to the reports

jeromedockes:
also I've noticed that adding a report makes the ui very slow and I suppose that is a problem in the tablereport itself rather than on the skore side

rouk1:
Hi @jeromedockes thanks for your comments.
We use iframe to ensure that our cards get resized when there content is updated. We control the head and body of the iframe's srcdoc. I need to force a font here.

Could you share the snippet you used to populate that report ?

rouk1:
Font family is now forced to our Geist with fallback on the default sans.
Screenshot 2024-11-26 at 11 25 06

jeromedockes:
> Could you share the snippet you used to populate that report ?

```python
from skore.project import create, load
from skrub import TableReport, datasets

create(project_name="project.skore")
s = load("project.skore")
df = datasets.fetch_employee_salaries().X
report = TableReport(df)
s.put("table report", report)
```

rouk1:
There is indeed a small freeze when the report is rendering. Especially when switching to the distribution tab.

skore/src/skore/item/skrub_table_report_item.py
augustebaum, line 1: Why do we need a new Item type? Isn't the table report just HTML that could be put via `MediaItem`?

augustebaum (commented):
Why can't we use `MediaItem`, since the TableReport is HTML?

skore/src/skore/item/skrub_table_report_item.py
thomass-dev:
```
media = MediaItem.factory(table_report.html_snippet(), media_type="text/html")
project.put_item("skrub_table_report", media)
```

rouk1: Ok but this requires users to write:

```python
from skore.item import MediaItem
from skore.project import create, load
from skrub import TableReport, datasets

df = datasets.fetch_employee_salaries().X
report = TableReport(df)
my_project.put_item(
    "table report",
    MediaItem.factory(
        report.html_sippet(), media_type="text/html"
    ),
)
```

Instead of

```python
from skore.project import create, load
from skrub import TableReport, datasets

df = datasets.fetch_employee_salaries().X
report = TableReport(df)
my_project.put("table report", report)
```

This looks cumbersome to me as it seems to me that we want to promote skrub and treat it as a first class citizen. 
Indeed this is sugar....
@MarieS-WiMLDS & @sylvaincom could we have a user point of view on this ?

thomass-dev: I'm not convinced at all of the user gain. It's in the hands of the design.
I'm afraid that open the way to sugar all media types...

If this solution remains, what do you think to inherit from `MediaItem` instead of creating a new class of item from scratch? Something like:

```python
class SkrubTableReportItem(MediaItem):
    @classmethod
    def factory(cls, table_report: TableReport) -> SkrubTableReportItem:
        if getattr(table_report, "html_snippet", None) is None:
            raise ItemTypeError(f"Type '{table_report.__class__}' is not supported.")

        html = table_report.html_snippet()
        return super().factory_str(html, media_type="text/html")
```

jeromedockes: > I'm afraid that open the way to sugar all media types...

maybe it would be nice to have a shorthand for storing such a representation of an object.

You could check if the object has a `_repr_mimebundle_`, and if not go through `_repr_html_`, `_repr_svg_`, ... ([full list](https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display)), and finally fall back to storing `repr(the_object)` as text.

This way a user could do something like `my_project.put_repr(my_object)` instead of importing and creating the MediaItem, calling the appropriate method and specifying the correct media type.

MarieS-WiMLDS: From a user point of view, this version is a no go:  
```
my_project.put_item(
    "table report",
    MediaItem.factory(
        report.html_sippet(), media_type="text/html"
    ),
)
```

As skore is not a cache anymore, could we do the MediaItem.factory(etc) stuff under the hood? in the put_item?

rouk1: @thomass-dev, the implementation now uses inheritance.

thomass-dev: @MarieS-WiMLDS `MediaItem` is only focused on display, not on cache.

Its the only way for user to define how to display a media.
For instance, if you want to display a "blabla", you can use HTML, Markdown, simple str etc.

tuscland:
I was wondering, don't we want to store the raw data used to present the report?
skrub has a JSON representation of the table report.
Maybe as a future improvement?

skore/src/skore/item/skrub_table_report_item.py
thomass-dev: @jeromedockes there are some interesting things in your proposal but we can't know in advance how the user want to show its object: imagine the functions `_repr_html_` and `_repr_svg_` are available, which one should we call? Even if no of these functions are available, i don't think the default representation is what the user want.

For instance, if the user wishes to put/display an str/bytes (content an PNG image), we can't detect such cases.
IMHO its the user responsibility to define how to show its media.

jeromedockes: I agree, but to some extent the same is true when a jupyter notebook has to display the output of a cell. Many data-science packages support the rich display mechanism of jupyter notebooks and define the appropriate method for what they think is the best rich representation of their object. So you could take advantage of that, as a way for the user to delegate the responsibility of choice you mention to the package that created the object whose representation they want to store. It wouldn't be the only (or the main) way to store some visualisation, but a shorthand to say "just store whatever I would see in a notebook". anyway that's just my 2 cents :)

sylvaincom: It'd be nice to have skrub TableReport as first citizen, but you have better visibility on code and I agree that we might have too many items, so HTML seems fine

It would be nice to have the nice display with just this:
```python
from skore.project import create, load
from skrub import TableReport, datasets

df = datasets.fetch_employee_salaries().X
report = TableReport(df)
my_project.put("table report", report)
```

jeromedockes: > I guess some kind of heuristic like "take _repr_html_ if it's there, otherwise _repr_svg_

yes exactly I would just follow the same order of preference as jupyter and then what users see in skore would match what they see in a notebook.

also I think quite a few packages will define only 1 (usually html): the one they want their users to see. that is the case for example for polars dataframes (pandas also defines `_repr_latex_`), scikit-learn estimators, skrub tablereport

skore/src/skore/item/skrub_table_report_item.py
```diff
@@ -0,0 +1,45 @@
...
+class SkrubTableReportItem(MediaItem):
+    """
+    A class to represent a skrub TableReport.
+
+    This class encapsulates a skrub TableReport
+    along with its creation and update timestamps.
+    """
+
+    @classmethod
+    def factory(cls, table_report: TableReport) -> SkrubTableReportItem:  # type: ignore[override]
```
thomass-dev, line 27:
```suggestion
    def factory(cls, table_report: TableReport) -> MediaItem:  # type: ignore[override]
```

skore/src/skore/ui/project_routes.py
```diff
@@ -19,6 +19,7 @@
 from skore.item.polars_series_item import PolarsSeriesItem
```
thomass-dev, line 19: These changes can be revert.

thomass-dev, line 19: These changes can be revert as long as your factory returns a `MediaItem`.

thomass-dev (changes requested):

skore/src/skore/item/skrub_table_report_item.py
```diff
@@ -0,0 +1,45 @@
...
+        Returns
+        -------
+        SkrubTableReportItem
+            A new SkrubTableReportItem instance.
+        """
+        if getattr(table_report, "html_snippet", None) is None:
+            raise ItemTypeError(f"Type '{table_report.__class__}' is not supported.")
+
+        html = table_report.html_snippet()
+        return super().factory_str(html, media_type="text/html")  # type: ignore
```
thomass-dev, line 45:
```suggestion
class SkrubTableReportItem:
    """
    A class to represent a skrub TableReport.

    This class encapsulates a skrub TableReport
    along with its creation and update timestamps.
    """

    @staticmethod
    def factory(table_report: TableReport) -> MediaItem:  # type: ignore[override]
        """
        Create a new SkrubTableReportItem with the current timestamp.

        Parameters
        ----------
        table_report : TableReport
            The report to store.

        Returns
        -------
        SkrubTableReportItem
            A new SkrubTableReportItem instance.
        """
        if getattr(table_report, "html_snippet", None) is None:
            raise ItemTypeError(f"Type '{table_report.__class__}' is not supported.")

        html = table_report.html_snippet()
        return super().factory_str(html, media_type="text/html")  # type: ignore
```

skore/src/skore/item/skrub_table_report_item.py
```diff
@@ -0,0 +1,45 @@
...
+from typing import TYPE_CHECKING
+
+from skore.item.item import ItemTypeError
+from skore.item.media_item import MediaItem
+
+if TYPE_CHECKING:
+    from skrub import TableReport
+
+
+class SkrubTableReportItem(MediaItem):
```
thomass-dev, line 18:
```suggestion
class SkrubTableReportItem:
```

skore/src/skore/item/skrub_table_report_item.py
```diff
@@ -0,0 +1,45 @@
...
+    """
+    A class to represent a skrub TableReport.
+
+    This class encapsulates a skrub TableReport
+    along with its creation and update timestamps.
+    """
+
+    @classmethod
+    def factory(cls, table_report: TableReport) -> MediaItem:  # type: ignore[override]
```
thomass-dev, line 27:
```suggestion
    @staticmethod
    def factory(table_report: TableReport) -> MediaItem:  # type: ignore[override]
```

skore/src/skore/item/skrub_table_report_item.py
thomass-dev, line 45:
```suggestion
        return MediaItem.factory_str(html, media_type="text/html")  # type: ignore
```

thomass-dev (commented):
What do you think about changing the class to a "namespace"?

thomass-dev (approved):

rouk1:
I was a bit to fast to accept using inheritance. HTML as never been treated as a media, it was a primitive type with `text/html` media type. We do have a bug now #823.

@thomass-dev  & @augustebaum I'll let you decide how to manage html and let me know how to fix this.

PR #817, opened 2024-11-26
fix: Missing values in pandas data frame are now replaced

rouk1:
All missing values from pandas dataframe will now be displayed with a fallback string "missing value" in the UI.
Screenshot 2024-11-26 at 16 37 48


skore/src/skore/ui/project_routes.py
```diff
@@ -55,7 +55,7 @@ def __serialize_project(project: Project) -> SerializedProject:
                 value = item.array.tolist()
                 media_type = "text/markdown"
             elif isinstance(item, PandasDataFrameItem):
-                value = item.dataframe.to_dict(orient="tight")
+                value = item.dataframe.fillna("missing value").to_dict(orient="tight")
```
thomass-dev, line 58: This should be also applied at least to numpy, polars dataframe, polars/pandas series.

thomass-dev (changes requested):
Have you tried to use ORJSON as encoder to handle NaN?

skore/src/skore/ui/project_routes.py
thomass-dev, line 58: Do you think you can easily add a test?

sylvaincom:
Why not display `NaN`?

rouk1:
> stead of `missing value`?

In the given example there are missing values in the gender column which is not numbers. I thought missing values was more generic.  As you like...

rouk1:
> Have you tried to use ORJSON as encoder to handle NaN?

Nope = ]

sylvaincom:
> > Why not display NaN instead of missing value?
> 
> 
> 
> In the given example there are missing values in the gender column which is not numbers. I thought missing values was more generic.  As you like...

Let's go with `NaN`? Cc @MarieS-WiMLDS 

See https://pandas.pydata.org/docs/user_guide/missing_data.html

skore/src/skore/ui/project_routes.py
rouk1: Done see test_ui.py

rouk1: As polars API is less permissive I chose to convert to pandas (which happens anyway) then to fill the holes.

MarieS-WiMLDS:
I'm missing the purpose here, why do we need to do that? The missing values are not displayed correctly currently?  

My main feat: if someone does put and then get, although we are not a cache anymore, it's legit that they can expect their dataset to be identical. If we do fillna... we're breaking their dataset.

rouk1:
> I'm missing the purpose here, why do we need to do that? The missing values are not displayed correctly currently?
> 
> My main feat: if someone does put and then get, although we are not a cache anymore, it's legit that they can expect their dataset to be identical. If we do fillna... we're breaking their dataset.

`fillna` is used only for frontend display. The data persists as given by the user. We need to do that because JSON encoder crashe when serializing missing values and NaN (NaN is not part of the JSON specification).

MarieS-WiMLDS:
ok thanks, it's clear for me!  

As a user, I would expect "NaN" to be displayed, like Sylvain.

skore/src/skore/ui/project_routes.py
```diff
@@ -55,13 +63,13 @@ def __serialize_project(project: Project) -> SerializedProject:
...
             elif isinstance(item, PandasSeriesItem):
-                value = item.series.to_list()
+                value = item.series.fillna("NaN").to_list()
                 media_type = "text/markdown"
             elif isinstance(item, PolarsDataFrameItem):
-                value = item.dataframe.to_pandas().to_dict(orient="tight")
+                value = pandas_dataframe_to_serializable(item.dataframe.to_pandas())
                 media_type = "application/vnd.dataframe+json"
             elif isinstance(item, PolarsSeriesItem):
                 value = item.series.to_list()
```
thomass-dev, line 75:
```suggestion
                value = item.series..fill_nan("NaN").to_list()
```

thomass-dev (approved):

PR #825, opened 2024-11-27
fix(docs): Improve the tracking example

### LINKED ISSUE #821: The cross-validate example with hyperparameter tuning looks like an anti-pattern
I was looking at this [example](https://probabl-ai.github.io/skore/latest/auto_examples/plot_04_tracking_items.html#tracking-the-results-of-several-runs-of-skore-cross-validate) and the following block:

```python
from sklearn import datasets
from sklearn.linear_model import Lasso
import skore

X = diabetes.data[:150]
y = diabetes.target[:150]
lasso = Lasso()

for alpha in [0.5, 1, 2]:
    cv_results = skore.cross_validate(
        Lasso(alpha=alpha), X, y, cv=5, project=my_project
    )
```

The fact that we do an hyperparemeter search here looks really weird to me. I would request people to use a `RandomizedSearchCV` or a `GridSearchCV` so I'm not sure that the example is relevant anymore.

One issue that you end-up with the current example is that the generic `cross_validate` does not intend to track hyperparameters: it means as a user, I'll need to store those in the order of computation as well.

Another point is about data splitting: since no random state is set, each parameter are evaluated on really different sets. The `SearchCV` will report results on consistent splits even when the random state is not set (if I'm not mistaken).

A more natural example would be that I have a model where the hyperparameters are set and I get a fresh batch of data and I want to look if I got any drift in the statistical performance. It looks to me a more appealing and real use case than the current example.
Edit: For this point, it means that somehow some feature engineering in the preprocessing stage get invalidating and that the craft model will start to not work anymore.

NB: I would put the `import` before `from`. I think this is something `isort` would do.
NB2: `lass = Lasso()` is unused.
NB3: you can load directly X and y with `X, y = datasets.load_diabetes(return_X_y=True)`.

sylvaincom:
Fix #821

### DIFF
--- a/examples/plot_04_tracking_items.py
+++ b/examples/plot_04_tracking_items.py
@@ -182,7 +182,8 @@ skore.cross_validate(Lasso(alpha=2), X, y, cv=5, project=my_project)
 
 # %%
 # Thanks to the storage of the results by :func:`skore.cross_validate` (done when
-# specifying the ``project`` argument), we can compare the metrics of each run of thecross-validation (on all splits):
+# specifying the ``project`` argument), we can compare the metrics of each
+# cross-validation run (on all splits):
 
 # %%
 fig_plotly = my_project.get_item("cross_validation_aggregated").plot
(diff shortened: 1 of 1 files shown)


rouk1 (approved):
Nice.

PR #848, opened 2024-12-02
767 activity feed UI

rouk1:


augustebaum (changes requested):
Could you animate the appearance of new activity?

augustebaum (changes requested):
I see that there is a button for navigating to the components page in the dev version, but not in the built version. I would rather limit the number of differences between the two as much as possible because otherwise `npm run dev` will become useless.

augustebaum (commented):
Plotly figures are not rendered in the activity feed


EDIT: Same for altair


https://github.com/user-attachments/assets/e5464dc3-5037-4dd4-b5f7-902081637c80

augustebaum:
Might be my dark mode thing but sklearn Pipeline HTML has a white background:

2024-12-05T10_44_39_screenshot

augustebaum:
Question (not to deal with right now): does it makes sense to delete items from the activity feed if they're deleted from skore? IMO the sane default is to add a "deleted" event.

rouk1:
> I see that there is a button for navigating to the components page in the dev version, but not in the built version. I would rather limit the number of differences between the two as much as possible because otherwise `npm run dev` will become useless. (Of course, the components page itself can stay)

I cant see what would makes `npm run dev` useless. In practice it's main purpose is to give hot module reloading. It is a common practice to have demo pages for components to streamline the development. Having those component rendered in the same context (the router view) is a quick win.

Also there is a ton of difference happening when running the production build (code is minified, css is aggregated, ...).

Could you clarify why this bothers you ? I personally like to have all the app routes following the same pattern even if they are only visible to integrators.

Note that the component page code will be tree shacked in the production build.

augustebaum:
> Could you clarify why this bothers you ?

My only fear is that one day someone misses a bug in review because they're reviewing with `npm run dev` (as I do) instead of the production build.
Admittedly, this could also be solved by only reviewing the production build -- but if we go with this, we need to stop showing screenshots of the dev build.

And yeah, fair point about the dev and prod being different in other ways -- it's just that this extra button in the main view is very obvious.
Again, the components page itself is not a problem at all, just the extra button.

rouk1:
> Question (not to deal with right now): does it makes sense to delete items from the activity feed if they're deleted from skore? IMO the sane default is to add a "deleted" event.

I suppose that a delete event should be added to cover the full spectrum of feature. For now this is not supported.

rouk1:
All types of items should now render correctly.

Deserialization of dto are now outside of the project store (which is only related to the view builder page).

anasstarfaoui:
> > Could you animate the appearance of new activity? Not high priority of course.
> 
> That is kind of non trivial. How should the arrow behave during the animation ? @anasstarfaoui any motion design insight on this ?

Super nice motion touch I like it. My one advice: Let's make it quicker, but it's very good

thomass-dev (approved):
Thanks to have separate DTO specifications and deserialization from `project.ts`.

augustebaum (approved):

Issue #767, opened 2024-11-18
Activity feed


MarieS-WiMLDS:
### Is your feature request related to a problem? Please describe.

As a user, I want to see quickly items that are created, including warnings. 

### Describe the solution you'd like

This solution is very MVP. My goal is to create a spec that requires as little tech work, rework and object creation as possible at first.  

In a view called "Activity feed", I want to see all the items created & the warnings.  
The most recent should be on top.  
Whenever an item is updated, the older version is deleted from the feed so it appears on top only.  
If possible, the warnings shouldn't be items that appear in the items list on the left. 

#### UI mockups

Application must now have two (virtual) pages. One for the project view, one for the activity feed.
There URL must be:
- /#/project
- /#/activity

A new vertical left bar (looking like a workspace switcher) will act as the main app navigation.

![Flow _ Activity feed_light](https://github.com/user-attachments/assets/21079caa-37da-4a8a-97ae-0e56374f5c9a)
*light mode*
![Flow _ Activity feed_dark](https://github.com/user-attachments/assets/6d9360a5-81cc-4017-af80-ba9a7f000231)
*dark mode*

[Figma source](https://www.figma.com/design/Qzjyxv0lPDU2kKtsH9dcp3/Skore---Workstation---High-fidelity-mock-ups---Q3?node-id=4625-10896&t=TNudaJfZ46wHBIRa-1).

#### Project's activity as an HTTP API.

> Disclaimer: even if this a perfect use case for web sockets, we will stick to a REST endpoint.

##### Needs

We need to serialize a list of item version for the frontend to display them from most recent to oldest one.

> [!NOTE]  
> This means that: if a user asks for the last 5 minutes of activity, and one item has changed 2 times during this time frame (same identifier, two different values), they will receive 2 _versions_ of this item.

##### Proposed API

`GET` `/api/project/activity?after={creation_datetime}` (fetch the project activity)

###### Parameters

| name    | type     | data type                         | description                                                                |
|---------|----------|-----------------------------------|----------------------------------------------------------------------------|
| `after` | optional | datetime as UTC string (iso 8601) | A search filter (every item that have been added after the given datetime). Adding this parameter will allow frontend to only request new data after the initial call. |

###### Response

| http code | content-type                     | response              | description                                    |
|-----------|----------------------------------|-----------------------|------------------------------------------------|
| `200`     | `application/json;charset=UTF-8` | `[{...}, {...}, ...]` | A list of item ordered by descending `create_at` datetime. |

[...]

PR #857, opened 2024-12-03
feat: Add a new backend API for activity feed

thomass-dev:
Author: Thomas S <thomas@probabl.ai>
Author: Matt J <matthieu@probabl.ai>

---

Part of #767 .

### DIFF
--- a/skore/src/skore/ui/project_routes.py
+++ b/skore/src/skore/ui/project_routes.py
@@ -158,10 +158,10 @@ async def get_activity(
     request: Request,
     after: datetime = datetime(1, 1, 1, 0, 0, 0, 0, timezone.utc),
 ):
-    """Send all recent as a JSON array.
+    """Send all recent activity as a JSON array.
 
     The activity is composed of all the items and their versions created after the
-    datetime `after`.
+    datetime `after`, sorted from newest to oldest.
     """
     project = request.app.state.project
     return sorted(
--- a/skore/src/skore/ui/project_routes.py
+++ b/skore/src/skore/ui/project_routes.py
@@ -3,24 +3,28 @@
 from __future__ import annotations
 
 import base64
-from collections import defaultdict
+import operator
 from dataclasses import dataclass
+from datetime import datetime, timezone
 from typing import TYPE_CHECKING, Any
 
 from fastapi import APIRouter, HTTPException, Request, status
 
+from skore.item import (
+    Item,
+    MediaItem,
+    NumpyArrayItem,
+    PandasDataFrameItem,
+    PandasSeriesItem,
+    PolarsDataFrameItem,
+    PolarsSeriesItem,
+    PrimitiveItem,
+    SklearnBaseEstimatorItem,
+)
 from skore.item.cross_validation_item import (
     CrossValidationAggregationItem,
     CrossValidationItem,
 )
-from skore.item.media_item import MediaItem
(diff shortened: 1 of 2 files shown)


skore/src/skore/ui/project_routes.py
```diff
@@ -138,4 +146,27 @@ async def delete_view(request: Request, key: str):
...
-    return __serialize_project(project)
+    return __project_as_serializable(project)
+
+
+@router.get("/activity")
+async def get_activity(
+    request: Request,
+    after: datetime = datetime(1, 1, 1, 0, 0, 0, 0, timezone.utc),
+):
+    """Send all recent as a JSON array.
```
augustebaum, line 157:
```suggestion
    """Send all recent activity as a JSON array.
```

skore/src/skore/ui/project_routes.py
```diff
@@ -138,4 +146,27 @@ async def delete_view(request: Request, key: str):
...
+    """
+    project = request.app.state.project
+    return sorted(
+        (
+            __item_as_serializable(key, version)
+            for key in project.list_item_keys()
+            for version in project.get_item_versions(key)
+            if datetime.fromisoformat(version.updated_at) > after
+        ),
+        key=operator.attrgetter("updated_at"),
```
augustebaum, line 170: I'm confused about the meaning of `updated_at` because of the version history. Can you remind me the difference between `updated_at` and `created_at`?

skore/tests/integration/ui/test_ui.py
```diff
@@ -98,3 +101,30 @@ def test_serialize_polars_series_with_missing_values(client, in_memory_project):
...
+        ("1", 1),
+        ("0", 0),
+    ]
+
+    now = datetime.datetime.now(tz=datetime.timezone.utc)
+
+    in_memory_project.put("4", 5)
+    in_memory_project.put("5", 5)
+
+    response = client.get("/api/project/activity", params={"after": now})
```
augustebaum, line 125: Just to be sure, if I didn't add the `after` param, the test would be
```python
    assert [(item["name"], item["value"]) for item in response.json()] == [
        ("5", 5),
        ("4", 4),
        ("4", 4),
        ("3", 3),
        ("2", 2),
        ("1", 1),
        ("0", 0),
    ]
```

skore/src/skore/ui/project_routes.py
thomass-dev: `created_at` -> date of the first version on an item
`updated_at` -> date of the current version

skore/tests/integration/ui/test_ui.py
thomass-dev: but yes, you true :)

augustebaum: oops yes i misread, thanks

skore/src/skore/ui/project_routes.py
augustebaum: Thanks. As usual, I think eventually we should only have `created_at`, because Items themselves can't be modified; it's rather that the key that points to it may point to something else.

augustebaum (dismissed):

skore/src/skore/ui/project_routes.py
rouk1: You're write this debt that we need to reimburse soon.

rouk1 (approved):

Issue #686, opened 2024-11-06
Cross-validation: Be more explicit about what the lines mean


MarieS-WiMLDS:
### Is your feature request related to a problem? Please describe.

![image](https://github.com/user-attachments/assets/ce651005-7006-4fd2-b8bc-d54c9dce5d2f)

I think that it's not 100% clear what the dash and dot lines represent. For the dash line for instance, it could be both the mean or the median. 

### Describe the solution you'd like

have a fix legend, not colored, about what these two lines represent, agnostic from the value?

### Describe alternatives you've considered, if relevant

display on the graph itself. But redundant if several values are displayed, plus might look too small.

augustebaum:
Initially they were labelled in the legend, but we decided with @sylvaincom to remove them because they took too much space. You can hover over the lines to get more details

augustebaum:
> have a fixed legend, not colored, about what these two lines represent, agnostic from the value

this seems fine to me!

sylvaincom:
Hover isn't enough?

MarieS-WiMLDS:
You have to hover on the borders of the graph. Until Auguste showed me, I believed that there was no hovering option.

rouk1:
@MarieS-WiMLDS could you specified the label you want to add to the plot ?
Maybe we could add a subtitle inviting the user to hover the plot to get more information ?

MarieS-WiMLDS:
I think it should be more natural, and therefore no need to explicitely tell the user what to do.  

I have two improvements in mind: 
- Give the possibility to see the lines texts when hovering in other places that the extrem borders (maybe by adding points inbetween the bars thanks to [bar_x](https://github.com/probabl-ai/skore/blob/main/skore/src/skore/item/cross_validation_item.py#L69)?).
- a legend to say that the dash line represents the mean, and the dots line represents the mean +/- standard value.

augustebaum:
@MarieS-WiMLDS if you still think the legend would be helpful, feel free to re-open/open another issue

PR #862, opened 2024-12-03
fix(cross-validation plot): Fix hover on horizontal lines

augustebaum:
Previously the tooltip was only shown when hovering on the extreme right or left of the plot.

Addresses #686

https://github.com/user-attachments/assets/c75ffeff-dc85-4526-bc6b-93d8afa119fe


MarieS-WiMLDS (approved):
:heart_eyes:

PR #867, opened 2024-12-05
fix: Remove undeclared dependency `pandas` from warning class

thomass-dev:
```
import skore
my_project = skore.create("my_project")

File [...]/skore/sklearn/train_test_split/warning/high_class_imbalance_warning.py:10
      6 from __future__ import annotations
      8 from typing import TYPE_CHECKING, Any, Optional, Union
---> 10 import pandas
     12 from skore.sklearn.train_test_split.warning.train_test_split_warning import (
     13     TrainTestSplitWarning,
     14 )
     16 if TYPE_CHECKING:

ModuleNotFoundError: No module named 'pandas'
```


augustebaum (approved):
Thanks!

PR #871, opened 2024-12-05
feat: Add widget for CrossValidationItem

rouk1:
This PR introduces a Cross validation report widget.

# UI preview

https://github.com/user-attachments/assets/6346cf67-938b-4648-9b2e-5929578f3d5b

# Usage

```python
from sklearn.model_selection import KFold

import skore
from skore.sklearn.cross_validation.cross_validation_reporter import (
    CrossValidationReporter,
)


def prepare_cv():
    from sklearn import datasets, linear_model

    X = diabetes.data[:150]
    y = diabetes.target[:150]
    lasso = linear_model.Lasso(alpha=2.5)
    return lasso, X, y

p = skore.load("project")
lasso, X, y = prepare_cv()
reporter = CrossValidationReporter(lasso, X, y, cv=KFold())
p.put("cv-reporter", reporter)


```


skore-ui/src/components/CrossValidationReport.vue
```diff
@@ -0,0 +1,73 @@
...
+    {
+      name: "a",
+      columns: ["Accuracy", "Precision", "Recall", "F1 Score"],
+      data: [
+        [0.8, 0.4, 0.5, 0.6],
+        [0.8, 0.4, 0.5, 0.6],
+        [0.8, 0.4, 0.5, 0.6],
+        [0.8, 0.4, 0.5, 0.6],
+      ],
+      favorability: "higher-is-better",
```
augustebaum, line 43:
```suggestion
      metric_type: "score",
```

skore-ui/src/components/CrossValidationReport.vue
```diff
@@ -0,0 +1,73 @@
...
+    {
+      name: "b",
+      columns: ["Accuracy", "Precision", "Recall", "F1 Score"],
+      data: [
+        [0.8, 0.4, 0.5, 0.6],
+        [0.8, 0.4, 0.5, 0.6],
+        [0.8, 0.4, 0.5, 0.6],
+        [0.8, 0.4, 0.5, 0.6],
+      ],
+      favorability: "lower-is-better",
```
augustebaum, line 54:
```suggestion
      metric_type: "loss",
```

skore-ui/src/components/CrossValidationReport.vue
augustebaum, line 43: Every metric can have a different "favorability", did you group the "higher-is-better" metrics together on purpose?

rouk1: Seems that this is not part of the "MVP", so I removed it. 
To be discussed in the "backend" PR imho.

augustebaum:
@rouk1 we should add a video of the widget with a real example

skore/src/skore/ui/project_routes.py
```diff
@@ -84,7 +194,10 @@ def __item_as_serializable(name: str, item: Item) -> SerializableItem:
         else:
             value = base64.b64encode(item.media_bytes).decode()
             media_type = f"{item.media_type};base64"
-    elif isinstance(item, (CrossValidationItem, CrossValidationAggregationItem)):
+    elif isinstance(item, CrossValidationItem):
+        value = __cross_validation_item_as_serializable(item)
+        media_type = "application/vnd.skore.cross_validation+json"
```
thomass-dev, line 199: Please add a dedicated test.

skore/tests/unit/item/test_cross_validation_item.py
```diff
@@ -47,7 +48,11 @@ def test_factory(self, monkeypatch, mock_nowstr):
         item = CrossValidationItem.factory(reporter)
```
thomass-dev, line 49: Please add a check for `cv_info`.

skore/src/skore/item/cross_validation_item.py
```diff
@@ -106,14 +114,15 @@ def factory(cls, reporter: CrossValidationReporter):
         if not isinstance(reporter, CrossValidationReporter):
             raise ItemTypeError(
                 f"Type '{reporter.__class__}' is not supported, "
-                "only 'CrossValidationReporter' is."
+                f"only '{CrossValidationReporter.__name__}' is."
```
thomass-dev, line 117: Why this change? `'CrossValidationReporter' == CrossValidationReporter.__name__`.

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, *args, **kwargs):
         self.estimator = args[0] if len(args) >= 1 else kwargs.get("estimator")
         self.X = args[1] if len(args) >= 2 else kwargs.get("X")
```
thomass-dev, line 125: `X` seems to be mandatory?

```suggestion
        self.X = args[1] if len(args) >= 2 else kwargs["X"]
```

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, *args, **kwargs):
         self.estimator = args[0] if len(args) >= 1 else kwargs.get("estimator")
```
thomass-dev, line 124: `estimator seems to be mandatory?

```suggestion
        self.estimator = args[0] if len(args) >= 1 else kwargs["estimator"]
```

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
```diff
@@ -124,6 +124,7 @@ def __init__(self, *args, **kwargs):
...
         self.X = args[1] if len(args) >= 2 else kwargs.get("X")
         self.y = args[2] if len(args) == 3 else kwargs.get("y")
+        self.cv = kwargs.get("cv")
```
thomass-dev, line 127: If this parameter can't be positional, please specify it in the docstring.

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
thomass-dev, line 1: Since you already know what parameters are allowed for the `Reporter`, i found it clearer to define the exact prototype of the `__init__` instead of `__init__(self, *args, **kwargs)`:

```
def __init__(
    self,
    estimator, 
    X, 
    y=None, 
    *, 
    groups=None, 
    scoring=None, 
    cv=None,
    [...]
)
```

thomass-dev (changes requested):
Huge work, review in progress.

rouk1:
> @rouk1 we should add a video of the widget with a real example

I've edited the first comment with a new screen recording.

skore/src/skore/ui/project_routes.py
```diff
@@ -51,6 +56,111 @@ class SerializableProject:
...
+    # This is done by instantiating the class with no arguments and
+    # computing the diff between the default and ours
+    try:
+        estimator_module = importlib.import_module(item.estimator_info["module"])
+        EstimatorClass = getattr(estimator_module, item.estimator_info["name"])
+        default_estimator_params = {
+            k: repr(v) for k, v in EstimatorClass().get_params().items()
+        }
+    except Exception:
+        default_estimator_params = {}
```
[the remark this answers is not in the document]
thomass-dev: Note with hindsight: to avoid environment dependency, we should computes these fields in the item. We can't assure that the module is still available during export backend to frontend.

skore/src/skore/item/cross_validation_item.py
augustebaum: If the class ever changes names, the hard-coded string could be forgotten

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
augustebaum: The thing is, I would rather `sklearn.cross_validate` be the one to issue the error if X was not passed: with your change, we'd have to add an error message that X wasn't passed.

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
augustebaum: See comments about X

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
augustebaum: It's a kwarg here because it's a kwarg in `sklearn.cross_validate`. I don't really know how to document this; in `sklearn.cross_validate` they show the actual argument list and so they have a `*`, but for us the arguments are `*args, **kwargs`... Would you have an idea of how to do this?

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
augustebaum: Yeah fair enough -- I chose to do it with `*args, **kwargs` because that way we don't have to stick to the sklearn signature where we don't need to. If we copy-paste the signature we'll have the burden of staying in sync with sklearn.
But on the other hand, sklearn is super stable...

skore/src/skore/ui/project_routes.py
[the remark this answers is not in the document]
augustebaum: I trust `get_params()` more than the signature, since it is a public method.

About computing the fields in the item: yes, that sounds much better, we should do it!

thomass-dev:
The PR needs to be merge for the incoming release. I will continue my review asynchronously in a second iteration.

skore/src/skore/ui/project_routes.py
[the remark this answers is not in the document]
thomass-dev: I don't understand your point on `get_params`. 
The signature of a function is fixed. If you want to check for non-default parameters, the signature is sufficient.
`get_params` can have side-effect on parameters, and you risk highlighting parameters that are not passed by user, but introduced by `get_params`. User can be confused.

I would prefer to store the real passed parameters in `estimator_info["params"]` and retrieve here the default parameters using signature in `default_estimator_params`.

skore/src/skore/ui/project_routes.py
```diff
@@ -51,6 +56,111 @@ class SerializableProject:
...
+        "data": list(zip(*cv_results.values())),
+    }
+
+    # Get scalar results (summary statistics of the cv results)
+    def metric_title(metric):
+        m = metric.replace("_", " ")
+        title = f"Mean {m}"
+        if title.endswith(" time"):
+            title = title + " (seconds)"
+        return title
```
thomass-dev, line 76: Can you leave this function outside of the function? So it can be defined only one-time by python in the scope of the module.

skore/src/skore/ui/project_routes.py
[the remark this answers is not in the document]
thomass-dev: And moreover,  `get_params` is only available for `scikit-learn`, but below (lines 119-127), you assume that you can received a model from a non `scikit-learn` module.

skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
augustebaum: Covered by #953

skore/src/skore/ui/project_routes.py
augustebaum: See #964

skore/tests/unit/item/test_cross_validation_item.py
augustebaum: See #965

skore/src/skore/ui/project_routes.py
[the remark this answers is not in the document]
augustebaum: However, we do assume the estimator is from sklearn [in CrossValidationItem's factory](https://github.com/probabl-ai/skore/blob/a59f0378f4d3b741db1c3b39d8c695528d64ca8e/skore/src/skore/item/cross_validation_item.py#L142) which is wrong!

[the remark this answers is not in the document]
augustebaum: On second thought, since this factory is only used to serialize information around a wrapper to sklearn's `cross_validate`, I think that we can reasonably restrict this to sklearn-compatible estimators. Still, this is a potential error that should be dealt with.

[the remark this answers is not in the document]
augustebaum: > If you want to check for non-default parameters, the signature is sufficient

We need `get_params` in general to get the parameters (e.g. to deal with sklearn `Pipeline`s, which don't have a fixed signature), so I don't think experimenting with the signature is worth it at the moment.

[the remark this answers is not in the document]
augustebaum: See #970

skore/src/skore/ui/project_routes.py
augustebaum: See #971

Issue #886, opened 2024-12-09
Forgetting the value in put function renders a very weird error


MarieS-WiMLDS:
### Describe the bug

I tried to put a dataframe in skore, forgetting to give it a name. skore output an error, so far so good, but I find it not straightforward to understand. Why do I have a type error related to cross validation item for instance?  

![Capture d’écran du 2024-12-09 14-55-44](https://github.com/user-attachments/assets/3fa841e7-09d5-48cf-ac2d-112d775157b4)

### Steps/Code to Reproduce

```python
import pandas as pd
import skore

project = skore.create("my_project_testing_a_bug", overwrite = True)
project.put(pd.DataFrame(data= {"letters": ["a", "b", "c"], "numbers": [1,2,3]}))
```

### Expected Behavior

ProjectPutError: the key provided is not a string.

### Actual Behavior

a long and weird error

### Environment

```shell
current main as of dec 9th
```

thomass-dev:
@MarieS-WiMLDS , now you will get something like:

```
TypeError                                 Traceback (most recent call last)
Cell In[1], line 6
      4 value = pd.DataFrame(data= {"letters": ["a", "b", "c"], "numbers": [1,2,3]})
      5 project = skore.create("my_project_testing_a_bug", overwrite = True)
----> 6 project.put(value)

File ~/Documents/workspace/probabl/skore/skore/src/skore/project/project.py:125, in Project.put(self, key, value)
    123     key_to_item = key
    124 else:
--> 125     raise TypeError(
    126         f"Bad parameters. "
    127         f"Find ({type(key)},{type(value) if value is not MISSING else ''})."
    128     )
    130 for key, value in key_to_item.items():
    131     if not isinstance(key, str):

TypeError: Bad parameters. Find (<class 'pandas.core.frame.DataFrame'>,).

```

MarieS-WiMLDS:
much nicer!  
I still don't understand the `Find (<class 'pandas.core.frame.DataFrame'>,).` part though, do you mean something like: `Found key-value pair: (<class 'pandas.core.frame.DataFrame'>,).`?

thomass-dev:
Is `TypeError: When value is not specified, key must be a string or a dict (found '<class 'pandas.core.frame.DataFrame'>')` better?

MarieS-WiMLDS:
yes, it's perfect, thanks!

PR #901, opened 2024-12-10
feat: Make sure CrossValidationItem creation comes from CrossValidationReporter

thomass-dev:
Part of https://github.com/probabl-ai/skore/issues/886.

As a user will never instanciate a `CrossValidationItem` himself, make sure `CrossValidationItem` creation comes from `CrossValidationReporter`, so the `object_to_item` fallback is not triggered by a simple dictionary (and remains meaningful).

- In `CrossValidationItem`, test class with `isinstance` by fixing circular imports
- In `CrossValidationItem`, remove the possibility to use factory other than with a `CrossValidationReporter`.

### DIFF
--- a/skore/src/skore/item/cross_validation_item.py
+++ b/skore/src/skore/item/cross_validation_item.py
@@ -5,21 +5,16 @@ This class represents the output of a cross-validation workflow.
 
 from __future__ import annotations
 
-import contextlib
 import hashlib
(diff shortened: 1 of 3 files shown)


augustebaum (approved):
Alright, thanks for the change!
Glad to see you're not getting the circular import problems that I had ^^'

PR #904, opened 2024-12-10
fix: `Project.put` raises meaningful exceptions

thomass-dev:


skore/src/skore/project/project.py
```diff
@@ -117,47 +116,27 @@ def put(self, key: Union[str, dict[str, Any]], value: Optional[Any] = None):
         value : Any, optional
             The value to associate with ``key`` in the Project.
             If ``key`` is a dict, this argument is ignored.
-
-        Raises
```
augustebaum, line 121: Can you replace the Raises docstring by the new possible exception types?

augustebaum (changes requested):

skore/src/skore/project/project.py
thomass-dev: of course, fixed

augustebaum (approved):
Okay, good to merge

PR #909, opened 2024-12-10
fix(API): Be resilient to item serialization failures

thomass-dev:
Part of https://github.com/probabl-ai/skore/issues/628.


thomass-dev:
Unrelevant as the skore-ui has been removed.

PR #916, opened 2024-12-11
ci: Test with the 3 latest major versions of `scikit-learn`

thomass-dev:
Refactor `skore.yml` workflow:
- launch tests with the 3 latest major versions of `scikit-learn` on `os: "ubuntu-latest"` and `python: "3.12"`,
- speed-up tests by installing dependencies in `venv`and caching it.

And make `_add_scorers` compatible with `sklearn ==1.4`.

### DIFF
--- a/.github/workflows/skore.yml
+++ b/.github/workflows/skore.yml
@@ -30,7 +30,8 @@ jobs:
           python-version: ${{ matrix.python }}
           cache: "pip"
 
-      - uses: actions/cache/restore@v4
+      - name: Restore python-venv
+        uses: actions/cache/restore@v4
         id: cache-python-venv
         with:
           path: 'skore/venv'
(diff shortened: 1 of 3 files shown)


thomass-dev:
> Not sure `skore` is critical to the point of making our CI run 4 times slower

Oldest versions of `scikit-learn` are only tested on `ubuntu-latest python 3.12`.

glemaitre:
So as a workaround for the older scikit-learn function, I think that you can:

- add a parameter `estimator` to `_add_scorers`
- pass `self.estimator` to the call to `_add_scorers` in the `CrossValidationReporter`.

Basically, in the case that `scoring is None`, we would default to `self.estimator.score`. However, in practice it will never happen because we are creating a list of scorer to `scoring` is never `None`. So it just a workaround.

I think it is worth a small comment to know why we pass the estimator around and actually don't expect to use it.

glemaitre:
I open https://github.com/probabl-ai/skore/pull/1062 that should solve the issue.

thomass-dev:
Thanks @glemaitre for your fix, all good. It's ready to merge @rouk1 .

rouk1 (dismissed):
Nice work, the planet says thx for caching venvs 💚

PR #922, opened 2024-12-12
feat: Remove `CrossValidationAggregationItem`

augustebaum:


MarieS-WiMLDS (approved):
I guess there was no way to keep the code but hide it from the doc? anyway, the code is still accessible thanks to git.

sylvaincom:
The current rendering of the example is this, not good

Capture d’écran 2024-12-12 à 12 31 03

augustebaum:
> The current rendering of the tracking example is this, not good

@sylvaincom can you elaborate?

augustebaum:
> I guess there was no way to keep the code but hide it from the doc? anyway, the code is still accessible thanks to git.

Yes I wanted to take the opportunity to start fresh since we don't know how aggregation should work.

sylvaincom:
> > The current rendering of the tracking example is this, not good
> 
> @sylvaincom can you elaborate?

In the screenshot, the example does not make any sense, as we don't show the plot comparing the several runs

augustebaum:
> > > The current rendering of the tracking example is this, not good
> > 
> > 
> > @sylvaincom can you elaborate?
> 
> In the screenshot, the example does not make any sense, as we don't show the plot comparing the several runs

Okay, maybe we could talk about `project.get_item_versions()` then? Or maybe just remove the example altogether?

thomass-dev (approved):

PR #933, opened 2024-12-12
feat(UI): Activity feed is now the default route

rouk1:
Screenshot 2024-12-12 at 15 54 10

### DIFF
--- a/skore-ui/src/router.ts
+++ b/skore-ui/src/router.ts
@@ -12,18 +12,18 @@ export enum ROUTE_NAMES {
 const routes: RouteRecordRaw[] = [
   {
     path: "/",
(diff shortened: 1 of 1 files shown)


skore-ui/src/router.ts
```diff
@@ -12,18 +12,18 @@ export enum ROUTE_NAMES {
 const routes: RouteRecordRaw[] = [
   {
     path: "/",
-    name: ROUTE_NAMES.VIEW_BUILDER,
-    component: ProjectView,
+    name: ROUTE_NAMES.ACTIVITY_FEED,
```
tuscland, line 15: Can we name this route, and have a redirect from / to this one at launch?

tuscland (approved):

PR #937, opened 2024-12-12
feat(UI): Underline sklearn doc links

rouk1:
Screenshot 2024-12-12 at 16 59 58


tuscland:
This solution does not indicate the intent, that is, clicking the estimator name leads out of skore.

augustebaum:
When hovering on a link the browser displays its target, which is already sufficient IMO as a user

PR #951, opened 2024-12-13
doc: Simplify docstring example of CrossValidationReporter

glemaitre:
Simplify the example available in the docstring.

### DIFF
--- a/skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
+++ b/skore/src/skore/sklearn/cross_validation/cross_validation_reporter.py
@@ -99,17 +99,12 @@ class CrossValidationReporter:
 
     Examples
     --------
-    >>> def prepare_cv():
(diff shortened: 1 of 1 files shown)


glemaitre:
Not really sure what is the error provided by the CI. It does seem related to any change from the PR itself.

MarieS-WiMLDS:
I agree with this simplication.  
I don't know what happens with the CI though, asking either Thomas or Sylvain.

sylvaincom:
Thanks!
@augustebaum dealt with the examples of the docstring

augustebaum (approved):
Before `CrossValidationReporter` was introduced, this doctest was skipped because we would have needed to create a Project, which would have made the test needlessly complex.

But now we don't anymore, so this is great; thanks!

glemaitre:
> from a programming view I feel like Project is well scoped
It makes sense.

PR #952, opened 2024-12-13
docs: Improve style of logo and moto lines

glemaitre:
Improve the style of logo and the moto lines of the landing page:

image

vs.

image


MarieS-WiMLDS (commented):
Thanks for making the docs & the readme aligned! It's much cleaner.
From a technical point of view, it's functional, I tested.    
In the readme, the subtitle ("Elevate ML Dev [...]") is in bold. I prefer when it's not, as in your suggestion, because it's too much text in bold in my opinion. However I prefer the text in white or black depending on the color mode of the user.  
What's your opinon @sylvaincom & @marie-auxille, since you worked on the readme?

glemaitre:
I don't mind adapting the style. Here, the chosen style is the same than on the skrub website. It was my starting thoughts.

image

sylvaincom (approved):
Indeed, looks much better, thanks!

PR #960, opened 2024-12-15
feat(cli): Improve colors in CLI

glemaitre:
Improve the coloring to have a nicer CLI interaction.
I was just playing around. This PR is not ready for review.

But right now, it will look at something like that:

image


glemaitre:
After playing with `rich` in #962, I think that we should start adding some coloring based on the `skore` color (cyan/orange) to show the infor.

glemaitre:
This is the current coloring in line with the other CLI:

thomass-dev (approved):

PR #962, opened 2024-12-15
feat: Use friendly verbose and colorish

### LINKED ISSUE #959: Use printing instead of logging in non-verbose mode in CLI
Right now, when using the CLI, we get verbose message that might be OKish in a terminal but really not user friendly in a notebook:

image

From what I encountered in other library, I think that we should print in the console the information that we expect our user to see and have a "verbose" option that print detailed information from the logging. On such example is `pixi`.

### Example with `pixi`

This the output of `pixi update`. It is a set of progress bar with a final message that says if something has been updated or not. So the info is really limited but useful feedback for the user.

image

`pixi update --verbose` provides more insights with the output of the logging in this case.

image

glemaitre:
closes #959 

This PR proposes to:

- use a default `logger` that does not show any information by default
- introduce a `--verbose` option for the CLI or `verbose` parameter for the functions
- a decorator that register a rich handler depending of the `verbose` parameter
- introduce a `rich` console with cyan/orange colors (as the default `skore` theme) for showing the pertinent information.

### DIFF
--- a/skore/src/skore/__init__.py
+++ b/skore/src/skore/__init__.py
@@ -2,7 +2,8 @@
 
 import logging
 
-import rich.logging
+from rich.console import Console
+from rich.theme import Theme
 
 from skore.project import Project, create, load
 from skore.sklearn import CrossValidationReporter, train_test_split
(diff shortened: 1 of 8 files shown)


glemaitre:
It would currently look like.

#### Terminal

image

#### Notebook

image

#### VS code




#### Documentation website

image

glemaitre:
So I switch from a decorator that was implicitly consuming `verbose` to a context manager that explicitly consume the `verbose` parameter. The former has a bug and was too much magical at my test. The context manager is much nicer in this regard.

glemaitre (commented):
OK, so it should be ready for review.

@MarieS-WiMLDS saw that we could refactor the argument in the CLI and use `parent_parser`. I don't add it in this PR to have the minimum diff, but it could be worth to do so for maintainability.

Another limitation of `argparse` here: I cannot pass `verbose` at the top parser and the subparser to allow `skore --verbose launch ...` and `skore launch --verbose`. This is something that `click` can nicely deal with and it is a pattern allowed in term of UX that I saw in other package (e.g. `pixi`).

Finally, @MarieS-WiMLDS spotted something weird in the `quickstart` where we can provide a `working_dir` for the `create` part but we don't do any `cd` for this directory in the `launch`. So it should fail. I'll try to craft a minimal non-regression test and it can be handled in another PR.

glemaitre:
@thomass-dev This last run will be fun for you :). In addition of the issue with the coverage (I maybe need to merge main), I segfault in Windows ;)

MarieS-WiMLDS (approved):

sylvaincom:
@thomass-dev I leave to you to merge this? 🙏

thomass-dev (approved):

PR #972, opened 2024-12-16
ci: Test FORK based on #981

thomass-dev:


thomass-dev:
https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

image

PR #974, opened 2024-12-17
fix: Correct serialization of CrossValidationItem for UI

augustebaum:
Fixes a bug introduced in https://github.com/probabl-ai/skore/pull/964/ where the plot was outputted to a non-serializable object

### DIFF
--- a/skore/src/skore/ui/project_routes.py
+++ b/skore/src/skore/ui/project_routes.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import base64
 import copy
 import importlib
+import json
 import operator
 import re
 import statistics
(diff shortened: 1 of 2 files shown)


skore/src/skore/ui/project_routes.py
```diff
@@ -136,7 +137,7 @@ def __cross_validation_item_as_serializable(item: CrossValidationItem) -> dict:
         "plots": [
             {
                 "name": "cross-validation results",
-                "value": item.plot,
+                "value": plotly.io.to_json(item.plot),
```
rouk1, line 140: This will return a json string. Hence needs a double josn decoding. I do not think that it's what we want.

augustebaum: gr

rouk1 (approved):

PR #979, opened 2024-12-17
Theme toggle

rouk1:
This PR introduces a theme toggle that overwrites user preferences.
Depends on #926 that introduces class based theme management.

# UI preview

https://github.com/user-attachments/assets/5a220d2b-d201-4787-98a2-db0b5c9917f7

@anasstarfaoui let me know if the integration suits you.


anasstarfaoui:
Hey! Looks okay overall, a bit stacked but it's fine for now. 

In the future, we will aim to have a settings part in skore hub, so it will be inside it!

PR #993, opened 2024-12-20
chore: Add a test target to the Makefile

rouk1:
This PR adds 3 target to the Makefile.

- `make test-frontend` will run vitest tests.
- `make test-backend` will run pytest tests.
- `make test` will run all tests.


Makefile
```diff
@@ -22,3 +22,12 @@ serve-skore-ui:
 		--host 0.0.0.0 \
 		--port 22140 \
 		--timeout-graceful-shutdown 0
+
+test-frontend:
+	cd skore-ui && npm install
+	cd skore-ui && npm run test:unit
+
+test-backend:
+	cd skore && pytest tests
```
thomass-dev, line 31:
```suggestion
	cd skore && python -m pytest --cache-clear --doctest-modules src/ tests/
```

rouk1: This command outputs warning which could be puzzling...

> Users/rouk1/dev/skore/venv/lib/python3.12/site-packages/coverage/inorout.py:508: CoverageWarning:
> Module src/ was never imported. (module-not-imported)
> /Users/rouk1/dev/skore/venv/lib/python3.12/site-packages/coverage/inorout.py:508: CoverageWarning:
> Module tests/ was never imported. (module-not-imported)
> /Users/rouk1/dev/skore/venv/lib/python3.12/site-packages/coverage/control.py:892: CoverageWarning:
> No data was collected. (no-data-collected)
> 
> WARNING: Failed to generate report: No data to report.

Makefile
```diff
@@ -22,3 +22,12 @@ serve-skore-ui:
 		--host 0.0.0.0 \
 		--port 22140 \
 		--timeout-graceful-shutdown 0
+
```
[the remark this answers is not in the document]
rouk1: ✅

Makefile
```diff
@@ -22,3 +22,15 @@ serve-skore-ui:
...
+lint:
+	pre-commit run --all-files
+
+test-frontend:
+	cd skore-ui && npm install
+	cd skore-ui && npm run test:unit
+
+test-backend:
+	cd skore && pytest tests
```
augustebaum, line 34: Could it also run the docs? like `make html`

augustebaum (changes requested):

Makefile
augustebaum: Solution: don't compute coverage (add `--no-cov`)

Makefile
augustebaum: `git clean -dx && make html` to empty the cache?

thomass-dev: `cd skore/sphinx && git clean -dx && make html`

rouk1: To me building the doc is not a form of testing. Shall we create another target (not included in the test target) for this ?

thomass-dev: We must be careful with `git clean` because this Makefile can be used during development.
It should not remove WIP, even on documentation.

Makefile
rouk1: Or cd to the correct folder and get coverage + color output :)

Makefile
glemaitre: > To me building the doc is not a form of testing. Shall we create another target (not included in the test target) for this ?

Usually, there is a top level `make docs` or equivalent in a lot of package as target

glemaitre: > git clean -dx && make html to empty the cache?

Having the cache is actually a feature :). It allows to not have to recompute some stuff. It would make sense to have a `make clean-docs` that remove the `auto-examples` and `generated` folders (I agree with @thomass-dev) regarding the `git` command).

rouk1: That was suppose to be a quick PR to help @MarieS-WiMLDS running tests.
Could we split documentation helper target in another PR ?

Makefile
thomass-dev: Please can you update to include cache cleaning/doctest as proposed (with you cd)?

Makefile
glemaitre: > Could we split documentation helper target in another PR ?

Absolutely, they are orthogonal issues. For future reference, when it comes to my comments, I might comment because it seems relevant to a particular discussion but they might not required to be address in the PR. I should have probably be explicit that I had not expectations to be in this PR.

thomass-dev (approved):

PR #994, opened 2024-12-20
chore: Allow to run tests in parallel using pytest-xdist

glemaitre:
`pytest-xdist` allows to run the test in parallel.
As the test suite will grow, it might be handy for little cost in terms of maintenance.

### DIFF
--- a/.github/workflows/skore.yml
+++ b/.github/workflows/skore.yml
@@ -39,7 +39,7 @@ jobs:
           wheel=(dist/*.whl); python -m pip install "${wheel}[test]"
 
           # Test
-          python -m pytest --no-cov src/ tests/
+          python -m pytest --no-cov src/ tests/ -n auto
 
   coverage-skore:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/skore.yml
```diff
@@ -59,7 +59,7 @@ jobs:
           python -m pip install -e .[test]
 
           # run coverage
-          python -m pytest --junitxml=coverage.xml --cov=skore src/ tests/ | tee pytest-coverage.txt
+          python -m pytest --junitxml=coverage.xml --cov=skore src/ tests/ | tee pytest-coverage.txt -n auto
```
thomass-dev, line 62:
```suggestion
          python -m pytest -n auto --junitxml=coverage.xml --cov=skore src/ tests/ | tee pytest-coverage.txt
```

thomass-dev:
I used to to use it, nothing against, thanks :).

thomass-dev (approved):

Issue #834, opened 2024-11-29
feat(back): Estimator Report


MarieS-WiMLDS:
### Is your feature request related to a problem? Please describe.

The steps of a ML project is roughly the following (without trying to be fully exhaustive yet):  
- Data exploration
  - EDA
  - cleaning
- Feature Engineering
  - ideation
  - evaluation
  - selection
  - imputer when required
- Model
  - hyperparameters tuning
  - evaluation & selection
  - analysis
    - [feature importance](https://scikit-learn.org/1.5/auto_examples/ensemble/plot_forest_importances.html)
    - metrics
      - accuracy, RMSE, precision... 
      - confusion matrix
    - plots
      - [ROC](https://scikit-learn.org/1.5/modules/generated/sklearn.metrics.RocCurveDisplay.html#sklearn.metrics.RocCurveDisplay.from_estimator) & [PR](https://scikit-learn.org/1.5/modules/generated/sklearn.metrics.PrecisionRecallDisplay.html#sklearn.metrics.PrecisionRecallDisplay.from_estimator) curves for classification
      - QQ-plots or residuals vs fitted values plot for regression
    - calibration
    - robustness
    - [learning curves](https://scikit-learn.org/dev/modules/generated/sklearn.model_selection.learning_curve.html)
- Business Eval
  - threshold choice
  - ROI measure (quite related to above)
- Documentation

We tackled the model evaluation & selection with `CrossValidationReporter`. I would very much like to have something that helps the user to have everything they want to know to evaluate their model for supervised learning.  

### Describe the solution you'd like

My `SupervisedModelAnalyzer` (name suggested) would need to answer the 6 points listed above.  
We have several options when it comes to computing the subparts this object: 
1. everything is given by the user at the beginning. heavy, but ensure consistency of the object
2. the object is like an empty shell, and whenever the user is interested in a subpart, they provide the required information. Light at the beginning, but can turn actually heavy in terms of UX + no insurance that there is consistency.
3. Some parameters are shared, let's ask them at the beginning.  

I suggest going for option 3.  
All the items require:
- fitted estimator
- X_validation 
- y_validation

Addition for each: 
- feature importance
  - option to go with permutation or by impurity in the case of trees (possible to have a default, here i'd rather have permutation)
     - in the case of permutation: scoring to evaluate (possible to have a default)
- learning curves
  - training set & target
  - train sizes (possible to have a default, same as sklearn)
[...]

PR #997, opened 2024-12-20
feat: Design of `EstimatorReport`

glemaitre:
closes https://github.com/probabl-ai/skore/issues/834

Investigate an API for a `EstimatorReport`.

#### TODO

- [x] Metrics
  - [x] handle string metrics has specified in the accessor
  - [x] handle callable metrics
  - [x] handle scikit-learn scorers
  - [x] use efficiently the cache as much as possible
  - [x] add testing for all of those features
  - [x] allow to pass new validation set to functions instead of using the internal validation set
  - [x] add a proper help and rich `__repr__`
- [x] Plots
  - [x] add the roc curve display
  - [x] add the precision recall curve display
  - [x] add prediction error display for regressor
  - [x] make proper testing for those displays
  - [x] add a proper `__repr__` for those displays
- [x] Documentation 
  - [x] (done for the checked part) add an example to showcase all the different features
  - [x] find a way to show the accessors documentation in the page of `EstimatorReport`. It could be a bit tricky because they are only defined once the instance created.
    - We need to have a look at the `series.rst` page from pandas to see how they document this sort of pattern.
  - [x] check the autocompletion: when typing `report.metrics.->tab` it should provide the autocompetion. **edit**: having a stub file is actually working. I prefer this than type hints directly in the file.
- Open questions
  - [x] we use hashing to retrieve external set.
  - use the caching for the external validation set? To make it work we need to compute the hash of potentially big arrays. This might more costly than making the model predict.

#### Notes

This PR build upon:
- https://github.com/probabl-ai/skore/pull/962 to reuse the `skore.console`
- https://github.com/probabl-ai/skore/pull/998 to be able to detect clusterer in a consistent manner.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -461,7 +461,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
     @available_if(
         _check_supported_ml_task(supported_ml_tasks=["binary-classification"])
     )
-    def brier_score(self, *, data_source="test", X=None, y=None, pos_label=None):
+    def brier_score(self, *, data_source="test", X=None, y=None):
         """Compute the Brier score.
 
         Parameters
@@ -481,38 +481,15 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             New target on which to compute the metric. By default, we use the target
             provided when creating the reporter.
 
-        pos_label : int, default=None
-            The positive class.
-
         Returns
         -------
         pd.DataFrame
             The Brier score.
         """
-        if self._parent._ml_task == "binary-classification" and pos_label is None:
-            # for consistency with the precision and recall, we are going to compute
-            # the brier score for the all possible labels and return a dataframe with
-            # the brier score for each label
-            labels = self._parent._estimator.classes_
-            brier_scores = [
-                self._compute_metric_scores(
-                    metrics.brier_score_loss,
-                    X=X,
-                    y_true=y,
-                    data_source=data_source,
-                    response_method="predict_proba",
-                    metric_name=(
-                        f"Brier score {self._SCORE_OR_LOSS_ICONS['brier_score']}"
-                    ),
-                    pos_label=label,
-                )
-                for label in labels
-            ]
-            results = pd.concat(brier_scores, keys=labels, axis=1)
-            # the labels will be the level 0 and we need to swap it with the metric name
-            results.columns = results.columns.swaplevel(0, 1)
-            return results
-
+        # The Brier score in scikit-learn request `pos_label` to ensure that the
+        # integral encoding of `y_true` corresponds to the probabilities of the
+        # `pos_label`. Since we get the predictions with `get_response_method`, we
+        # can pass any `pos_label`, they will lead to the same result.
         return self._compute_metric_scores(
             metrics.brier_score_loss,
             X=X,
@@ -520,7 +497,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             data_source=data_source,
             response_method="predict_proba",
             metric_name=f"Brier score {self._SCORE_OR_LOSS_ICONS['brier_score']}",
-            pos_label=pos_label,
+            pos_label=self._parent._estimator.classes_[-1],
         )
 
     @available_if(
--- a/skore/tests/unit/sklearn/test_estimator.py
+++ b/skore/tests/unit/sklearn/test_estimator.py
@@ -566,12 +566,12 @@ def test_estimator_report_report_metrics_binary(
     result = report.metrics.report_metrics(pos_label=pos_label)
     expected_metrics = ("precision", "recall", "roc_auc", "brier_score")
     # depending on `pos_label`, we report a stats for each class or not for
-    # precision, recall and brier_score
-    expected_nb_stats = 3 * nb_stats + 1
+    # precision and recall
+    expected_nb_stats = 2 * nb_stats + 2
     _check_results_report_metrics(result, expected_metrics, expected_nb_stats)
 
     # Repeat the same experiment where we the target labels are not [0, 1] but
-    # ["neg", "pos"]. We check that the brier score is computed for each label.
+    # ["neg", "pos"]. We check that we don't get any error.
     target_names = np.array(["neg", "pos"], dtype=object)
     pos_label_name = target_names[pos_label] if pos_label is not None else pos_label
     y_test = target_names[y_test]
@@ -580,8 +580,8 @@ def test_estimator_report_report_metrics_binary(
     result = report.metrics.report_metrics(pos_label=pos_label_name)
     expected_metrics = ("precision", "recall", "roc_auc", "brier_score")
     # depending on `pos_label`, we report a stats for each class or not for
-    # precision, recall and brier_score
-    expected_nb_stats = 3 * nb_stats + 1
+    # precision and recall
+    expected_nb_stats = 2 * nb_stats + 2
     _check_results_report_metrics(result, expected_metrics, expected_nb_stats)
 
     estimator, X_test, y_test = binary_classification_data_svc


examples/diagnose/plot_estimator_report.py
```diff
@@ -0,0 +1,338 @@
...
+Get insights from any scikit-learn estimator
+============================================
+
+This example shows how the :class:`skore.EstimatorReport` class can be used to
+quickly get insights from any scikit-learn estimator.
+"""
+
+# %%
+#
+# TODO: we need to describe the aim of this classification problem.
```
glemaitre, line 12: @GaelVaroquaux @jeromedockes Could you provide me a bit more details on what is the purpose of this dataset in `skrub`. It would be nice to give a short introduction to the ML problem we try to solve.

glemaitre:
I think that I'll stop here since it implements the feature describe in #834.
One should read the example to have an overview on what is going in this PR.

I'll try to build upon it to showcase what I expect when it comes to aggregation in the `CrossValidationReport`.

skore/src/skore/sklearn/_estimator.py
```diff
@@ -0,0 +1,1353 @@
...
+            response_method="predict",
+            pos_label=pos_label,
+            metric_name=f"Recall {self._SCORE_OR_LOSS_ICONS['recall']}",
+            average=average,
+            use_cache=use_cache,
+        )
+
+    @available_if(
+        _check_supported_ml_task(
+            supported_ml_tasks=["binary-classification", "multiclass-classification"]
```
glemaitre, line 1048: We should not support multiclass here.

MarieS-WiMLDS (commented):
hello! 
I've started to play around with ModelReport. 
I like it how it's created, and the menu displayed to explain what's inside. 

Here are a couple of questions I have about it: 
1. Is it correct that we can't access the plots yet, or is it that I didn't find how to do it?
2. Is it possible to be more explicite about the cache? From scratch, I don't understand why it can be necessary to clean a cache, and what it contains. 
3. Also about the cache, when calling the cleaning function, the menu appears, is it possible to remove it please? (cf screen capture)
4. About the metrics, what is the purpose of using a dataframe? I expected a float when I'm calling a unique metric.  

screen 1:  
Capture d’écran du 2025-01-02 17-38-23

screen 2:  
image

glemaitre:
> Is it correct that we can't access the plots yet, or is it that I didn't find how to do it?

Nop. But the current plot for a classifier (roc + precision recall) are only defined for a binary problem. I assume that here you got a multiclass problem I guess.?

For multiclass, we would need to extend the display and check what we want to represent between OvO / OvR and the different averages: https://scikit-learn.org/stable/auto_examples/model_selection/plot_roc.html#sphx-glr-auto-examples-model-selection-plot-roc-py

> Is it possible to be more explicite about the cache? From scratch, I don't understand why it can be necessary to clean a cache, and what it contains.

As a first draft, I wanted to have an easy way to clean the cache to show-off improvement. I don't think that we should expose what is the cache, this is the magic internal part.

Then, I agree that it is kind of debatable if we should expose a `clean_cache`. I think that I went this way because:

- it makes the example easy to write (but it is not a good argument)
- `joblib.Memory` has a `clear` (https://joblib.readthedocs.io/en/latest/memory.html) method and it is a bit the same paradigm (you don't know how it internally works)

So I think that it might be better to find a better example where we don't need to clear the cache.

> Also about the cache, when calling the cleaning function, the menu appears, is it possible to remove it please? (cf screen capture)

I need to think about it. Here, I assume that the cell is triggering a call to `__repr__`. The `clean_cache` does not return `self` or anything like that so. So we would need to have a workaround.

> About the metrics, what is the purpose of using a dataframe? I expected a float when I'm calling a unique metric.

It is true in single output regression or binary classification. But with multioutput regression or multiclass classification, then if you want a single float, you need to make a choice for the averaging and it starts to be tricky.

There is not a good average by default because they all loose some type of information and you are confronted to some application trade-off. For instance for the roc-auc: https://github.com/scikit-learn/scikit-learn/issues/28585#issuecomment-1989484965

I would argue that the good default here is to not average by default and expose the average parameter to the user (it is already exposed but we use an average by default).

In scikit-learn, it might be an historical choice because we wanted to use this metric in grid-search and you wanted to optimize a scalar number.

All of this to say that it is the reason to use a ataframe. The result is therfore consistent independently if it is a single scalar or an array of metrics.

An add-on is that it is possible to concatenate several metrics together with pandas then:

```
pd.concat([df_metric_1, df_metric_2], axis=1)
```

even though we could argue that this is the job of the `report_metrics` method.

glemaitre:
By the way, I'm thinking that we should add a parameter:

```python
data_source : {"test", "train", "X_y"}, default="test"
    The data source to use.
    
    - "test" : use the test set provided when creating the reporter.
    - "train" : use the train set provided when creating the reporter.
    - "X_y" : use the provided `X` and `y` to compute the metric.
```

To allow switching of source of information when computing the plot and metrics. And it allows to potentially cache the prediction for arbitrary `X` and `y`. I started to program this in my branch before Christmas but I did not push it yet.

MarieS-WiMLDS:
- For the plots, indeed I used multi-class classification to have something a bit more funny. I tried with with binary with what I understood from the API based on the helper displayed:  
Capture d’écran du 2025-01-03 10-40-54

- As the metric menu is called metric**s**, I would also expect plot**s**.  

- Indeed `reporter.clean_cache()` is better. I was sure I tried both, but I might have mixed it with other tests.  

- I'm not sure to understand your answer for the metrics presented as a dataframe. Even when presented like this, I never precised how to average. Maybe a solution would be to have no default, and throw an error if no averaging method is given by the user? The options could be `{‘raise’, ‘ovr’, ‘ovo’}` (as in [scikit-learn](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.roc_auc_score.html#sklearn.metrics.roc_auc_score))

MarieS-WiMLDS:
I like the idea of data source! It allows to compute both train & test values, and check the diff!

glemaitre:
> As the metric menu is called metrics, I would also expect plots.

It is interesting. I did not consider but I think it makes sense: under metrics, you can get either table or graphics. Some same of model inspection.

I need to think a bit more about it but your intuitions make sense.

glemaitre:
> I'm not sure to understand your answer for the metrics presented as a dataframe. Even when presented like this, I never precised how to average. 

Let me illustrate. First some boilerplate:

```python
from skrub import tabular_learner


X, y = make_classification(
    n_samples=1_000,
    n_classes=3,
    n_clusters_per_class=1,
    class_sep=0.01,
    random_state=42,
)
estimator = tabular_learner("classifier")
reporter = EstimatorReport(
    estimator, fit=True, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
)
reporter
```

Calling:

```python
reporter.metrics.precision()  # equivalent to `average="weighted"` because multiclass
```

will return now:

image

because by default we get the weighted average. This choice has been made in this part of the code:

https://github.com/probabl-ai/skore/blob/ef116ed9fa22dd9e639a68563fcc288cd9ca381f/skore/src/skore/sklearn/cross_validation/cross_validation_helpers.py#L42-L55

In scikit-learn, we actually raise an error because `average="binary"` and we explicitly ask the user to tell us what to do.

A better choice to me is to have as default `average=None` (for multiclass):

```python
reporter.metrics.precision(average=None)
```

that gives:

image

and thus a dataframe with the precision for each of the class.

So we could revisit the default values for the average depending on the ML problem such that we are neutral and still allow to tune it.

glemaitre:
I added the one-vs-rest strategy for the ROC curve. Here is an example:

```python
import numpy as np
from skrub import tabular_learner
from skore import train_test_split, EstimatorReport

X, y = make_classification(
    n_samples=1_000, n_classes=3, n_clusters_per_class=1, random_state=42
)
target_names = np.array(["class #0", "class #1", "class #2"], dtype=object)
y = target_names[y]

estimator = tabular_learner("classifier")
reporter = EstimatorReport(
    estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
)
reporter.metrics.plot.roc()
```

image

During this change, it made me realized that we should probably reimplement the display from scratch:

- we don't care to expose the `from_estimator` or `from_predictions` class method because we are in charge to create the object. Only the `.plot` should be available (and the attributes to alter the plot)
- because we are using the internal cache, we are always goint to use `from_predictions`. However, we also have extra data that a super handy such as the estimator (to access the `classes_` for instance) or the `._ml_task` (avoiding to recall `type_of_target`). So it makes a lot of sense to reimplement our own `_from_predictions` that is a private class method to initialize the class.
- since we have extra information like `data_source` and `estimator.classes_`, we can make a much better job for the legend. Indeed, the reporter is intended to show different ROC curve of the training/testing set or classes for a given model. The display for scikit-learn is potentially used to compare the ROC curve of several models. We probably need to revisit the visual when we are going to have a `GridSearchCV` where we actually want to compare models.

Here, I need to have more tests for the displays because I was relying on the scikit-learn code before. Since we diverge, we should strengthen this part.

skore/src/skore/sklearn/_estimator.py
```diff
@@ -0,0 +1,1552 @@
...
+
+    @property
+    def estimator(self):
+        return self._estimator
+
+    @estimator.setter
+    def estimator(self, value):
+        raise AttributeError(
+            "The estimator attribute is immutable. "
+            "Please use the `from_fitted_estimator` or `from_unfitted_estimator` "
```
augustebaum, line 213: Out of date?

glemaitre: Yep the second sentence. Let me fix it.

rouk1 (commented):
As I have low statistical skills my review is mostly questions : )

skore/src/skore/sklearn/_plot/prediction_error.py
```diff
@@ -0,0 +1,9 @@
+from sklearn.metrics import PredictionErrorDisplay
+
+from skore.sklearn._plot.utils import HelpDisplayMixin
+
+
+# FIXME: we should write our own class here to not expose the class methods from
+# scikit-learn and only relied on the estimator predictions.
+class PredictionErrorDisplay(HelpDisplayMixin, PredictionErrorDisplay):
+    pass
```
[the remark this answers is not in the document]
glemaitre: I'm currently revamping this one to have a consistent API as other display so this will go away.

skore/src/skore/sklearn/_estimator.py
```diff
@@ -0,0 +1,1612 @@
...
+class _MetricsAccessor(_BaseAccessor):
+    _SCORE_OR_LOSS_ICONS = {
+        "accuracy": "📈",
+        "precision": "📈",
+        "recall": "📈",
+        "brier_score": "📉",
+        "roc_auc": "📈",
+        "log_loss": "📉",
+        "r2": "📈",
+        "rmse": "📉",
```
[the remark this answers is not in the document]
glemaitre: I'm open to any suggestion. I see that I'm already struggling with matplotlib if the font is not available. Here, my only requirement is to have something going up for score and going down for error.

sylvaincom:
When building the doc, I previewed the example, and the emoji do not appear:
Capture d’écran 2025-01-07 à 15 09 14

skore/src/skore/sklearn/_estimator.py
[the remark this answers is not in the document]
glemaitre: We will have to indeed. I'm yet unsure of how we should split it. But we are going to have additional accessor, each of them specialized in something. So we could create a module `estimator`, and split the accessor per file.

I will not make the change right now, in case @augustebaum and @thomass-dev are looking at the PR but after the first round of review, I'll do that.

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -0,0 +1,384 @@
...
+# %%
+print(f"Time taken to compute the metrics: {end - start:.2f} seconds")
+
+# %%
+#
+# Since we obtain a pandas dataframe, we can also use the plotting interface of
+# pandas.
+import matplotlib.pyplot as plt
+
+ax = metric_report.T.plot.barh()
```
sylvaincom, line 135: When building the doc, I previewed the `plot_estimator_report` example, and the emoji do not appear for the metric names:

Capture d’écran 2025-01-07 à 15 09 14

skore/src/skore/sklearn/_estimator.py
glemaitre: Oh I forgot about my own comment. It is a limitation on scikit-learn. The implementation is only for `binary-classification`. Indeed, there is some future works in scikit-learn with the multiclass implementation: https://github.com/scikit-learn/scikit-learn/pull/22046

Here, we use the `log_loss` as a default score when we want to measure the quality of the probabilities given by a classifier in a multiclass setting (which is what the Brier score is trying to assess).

skore/src/skore/sklearn/_estimator.pyi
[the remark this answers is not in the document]
glemaitre: I personally don't like typing right now. The reason is that if you generate code with sphinx then you get "programmatic" type that are not necessarily human readable in the documentation. Usually, the numpydoc docstring is more readable in this regard.

However, within an IDE, types are useful. And I would scream if I try to write:

```
report.metric.->tab
```

and that my IDE does not provide any hints. So the middle ground (at least to me), is to provide a "stubs" file that are used by IDE that give the expected UI for auto-completion and having a clean documentation.

Also the implementation of the accessor make the type completion difficult and the stubs are helping at it if I'm not wrong.

For instance `pandas` has this type of approach.

In `scikit-learn`, microsoft and vs code poeple maintain the stubs: https://github.com/microsoft/python-type-stubs/tree/main/stubs/sklearn

We should have a system that create the stubs from the numpy documentation actually but it is also work in progress (https://github.com/scientific-python/docstub)

examples/model_evaluation/plot_estimator_report.py
glemaitre: Yep this is the limitation pointed out here: https://github.com/probabl-ai/skore/pull/997#discussion_r1905181294

We have to find a workaround. One of those should be available in `skrub` with similar issue. However, if @rouk1 has a nice proposal to explore, I'm happy to go this road as well.

skore/src/skore/sklearn/_estimator.py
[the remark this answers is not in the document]
rouk1: 👍🏻

skore/src/skore/sklearn/_estimator.pyi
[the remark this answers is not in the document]
rouk1: Ok thanks.

skore/src/skore/sklearn/_estimator.py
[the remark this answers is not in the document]
glemaitre: I would prefer a visual clue. Actually, you have some unicode ↗︎↘︎, that would be good enough as a start (and not to weird).

skore/src/skore/sklearn/_estimator.py
```diff
@@ -834,16 +834,16 @@ def _get_help_title(self):
...
+        "accuracy": "↗︎",
+        "precision": "↗︎",
+        "recall": "↗︎",
+        "brier_score": "↘︎",
+        "roc_auc": "↗︎",
+        "log_loss": "↘︎",
+        "r2": "↗︎",
+        "rmse": "↘︎",
+        "report_metrics": "",
+        "custom_metric": "",
```
sylvaincom, line 846:
```suggestion
        "accuracy": "(↗)",
        "precision": "(↗)",
        "recall": "(↗)",
        "brier_score": "(↘)",
        "roc_auc": "(↗)",
        "log_loss": "(↘)",
        "r2": "(↗)",
        "rmse": "(↘)",
        "report_metrics": "",
        "custom_metric": "",
```

skore/src/skore/sklearn/_estimator.pyi
```diff
@@ -0,0 +1,174 @@
...
+        y: Optional[ndarray] = None,
+        ax: Optional[matplotlib.axes.Axes] = None,
+        kind: Literal[
+            "actual_vs_predicted", "residual_vs_predicted"
+        ] = "residual_vs_predicted",
+        subsample: Optional[Union[int, float]] = 1_000,
+    ) -> PredictionErrorDisplay: ...
+
+class _MetricsAccessor(_BaseAccessor):
+    plot: _PlotMetricsAccessor
```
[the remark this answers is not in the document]
glemaitre: I did not think about it but why not. We already use plural for `metrics`. I thought at `.plot` because of pandas (https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.html) but it is just different visual representation from a single source of data.

Before to decide, let's think of the other potential accessor that we can have and if we expect to have verbs or we don't care.

skore/src/skore/sklearn/_estimator.py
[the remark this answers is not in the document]
glemaitre: Oh I actually misread the PR. So I'm pretty much in line with what you did in the UI. With the new coloring done with `rich` we are pretty close and consistent.

skore/src/skore/sklearn/_estimator.py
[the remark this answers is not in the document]
glemaitre: I finally just split the file now.

skore/src/skore/sklearn/_estimator/base.py
```diff
@@ -0,0 +1,143 @@
...
+
+        Returns
+        -------
+        X : array-like of shape (n_samples, n_features)
+            The requested dataset.
+
+        y : array-like of shape (n_samples,)
+            The requested dataset.
+
+        data_source_hash : int or None
```
augustebaum, line 112: Out-of-sync with arguments

glemaitre: Nop. I return either a int hash or None.

glemaitre: Ah you mean in the parameters above.

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -0,0 +1,384 @@
...
+from skore import train_test_split
+
+X_train, X_test, y_train, y_test = train_test_split(df, y, random_state=42)
+
+# %%
+# TODO: we have a perfect case to show useful feature of the `train_test_split`
+# function from `skore`.
+#
+# Now, we need to define a predictive model. Hopefully, `skrub` provides a convenient
+# function (:func:`skrub.tabular_learner`) when it comes to get strong baseline
```
sylvaincom, line 49:
```suggestion
# function (:func:`skrub.tabular_learner`) when it comes to getting strong baseline
```

skore/src/skore/sklearn/_estimator/base.py
```diff
@@ -0,0 +1,149 @@
...
+import joblib
+from rich.console import Console
+from rich.panel import Panel
+from rich.tree import Tree
+
+from skore.externals._sklearn_compat import is_clusterer
+
+
+class _HelpMixin:
+    """Mixin class providing help for the `help` method and the `__repr__` method."""
```
sylvaincom, line 13:
```suggestion
    """Mixin class providing help for the ``help`` method and the ``__repr__`` method."""
```
(for Sphinx rst)

skore/src/skore/sklearn/_estimator/base.py
```diff
@@ -0,0 +1,149 @@
...
+
+    def _get_methods_for_help(self):
+        """Get the methods to display in help."""
+        methods = inspect.getmembers(self, predicate=inspect.ismethod)
+        filtered_methods = []
+        for name, method in methods:
+            is_private_method = name.startswith("_")
+            # we cannot use `isinstance(method, classmethod)` because it is already
+            # already transformed by the decorator `@classmethod`.
```
sylvaincom, line 22:
```suggestion
            # we cannot use `isinstance(method, classmethod)` because it is already
            # transformed by the decorator `@classmethod`.
```

skore/src/skore/sklearn/_estimator/report.py
```diff
@@ -0,0 +1,371 @@
...
+        return predictions
+
+    ####################################################################################
+    # Methods related to the help tree
+    ####################################################################################
+
+    def _get_help_panel_title(self):
+        return (
+            f"[bold cyan]📓 Tools to diagnose {self.estimator_name} "
+            "estimator[/bold cyan]"
```
sylvaincom, line 324:
```suggestion
            f"[bold cyan]📓 Tools to diagnose estimator "
            "{self.estimator_name}[/bold cyan]"
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -0,0 +1,1101 @@
...
+
+###############################################################################
+# Metrics accessor
+###############################################################################
+
+
+class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
+    """Accessor for metrics-related operations.
+
+    You can access this accessor using the `metrics` attribute.
```
sylvaincom, line 31:
```suggestion
    You can access this accessor using the ``metrics`` attribute.
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -0,0 +1,1101 @@
...
+    """
+
+    _SCORE_OR_LOSS_ICONS = {
```
sylvaincom, line 34:
```suggestion
    # (↗︎) denotes a score (higher is better), (↘︎) denotes a loss or error (lower is better)
    _SCORE_OR_LOSS_ICONS = {
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: No need for the double backticks because I turn the `default_role` to `literal` that make single bacticks the same as double one.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: I don't think we need a comment. Otherwise, we just need to rename the variable instead.

skore/src/skore/utils/_accessor.py
```diff
@@ -0,0 +1,18 @@
+from pandas.core.accessor import DirNamesMixin, _register_accessor  # noqa: F401
```
thomass-dev, line 1: Can you add a comment to explain that these imports is done here as a shortcut for other files.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
sylvaincom: Ok, maybe this will be in the user guide or something like that

skore/src/skore/sklearn/_estimator/report.py
```diff
@@ -0,0 +1,371 @@
...
+        )
+
+    def _create_help_tree(self):
```
thomass-dev, line 327: Can you add to the helper the representation of the attributes of the reporter.
For instance, it can help users to know that the reporter contains the fitted estimator.

skore/src/skore/sklearn/_estimator/report.py
```diff
@@ -0,0 +1,371 @@
...
+                response_method=response_method,
+                pos_label=pos_label,
+                data_source=data_source,
+            )
+            for response_method, pos_label, data_source, X in product(
+                response_methods, pos_labels, data_sources, Xs
+            )
+        )
+        # trigger the computation
+        list(
```
rouk1, line 178: Maybe we could have a list of indeterminated progress instead of one progress bar that "jumps".

skore/src/skore/utils/_accessor.py
thomass-dev: As seen together, can you vendor the code instead of a direct import?

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -0,0 +1,1101 @@
...
+        )
+
+    def _get_help_panel_title(self):
+        return f"[bold cyan]{self._icon} Available plot methods[/bold cyan]"
+
+    def _get_help_tree_title(self):
+        return "[bold cyan]reporter.metrics.plot[/bold cyan]"
+
+
+register_metrics_accessor("plot")(_PlotMetricsAccessor)
```
rouk1, line 1101: Can this call be somwhere more discoverable ?

skore/src/skore/sklearn/_estimator/metrics_accessor.pyi
```diff
@@ -0,0 +1,168 @@
+from typing import Any, Callable, Literal, Optional, Union
+
+import matplotlib.axes
+import numpy as np
+import pandas as pd
```
thomass-dev, line 5: Please add numpy and pandas as direct dependencies.

skore/src/skore/sklearn/_estimator/metrics_accessor.pyi
```diff
@@ -0,0 +1,168 @@
+from typing import Any, Callable, Literal, Optional, Union
```
thomass-dev, line 1: To-do: check if removing the stub files breaks the auto-completion or not (ping @augustebaum).

skore/src/skore/sklearn/_estimator/metrics_accessor.py
thomass-dev: Maybe put all the declaration of accessors in a central place

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: However, we can have this part in the `help` as a kind of legend.

skore/src/skore/sklearn/__init__.py
```diff
@@ -1,9 +1,11 @@
 """Enhance `sklearn` functions."""
 
+from skore.sklearn._estimator import EstimatorReport
```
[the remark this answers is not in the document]
glemaitre: Basically, I want the user to be able to do

```python
skore.EstimatorReport
```

or

```python
skore.sklean.EstimatorReport
```

but I don't want to expose in a lower level. In scikit-learn (and other package), whenever you don't want people to import from the private module, you add an `_` even if it is a folder.

For instance, I would probably to the same for `cross_validation`.

However, it is something that we can discuss later.

skore/tests/conftest.py
```diff
@@ -7,6 +7,16 @@
...
 
 
+def pytest_configure(config):
+    # Use matplotlib agg backend during the tests including doctests
+    try:
+        import matplotlib
+
+        matplotlib.use("agg")
+    except ImportError:
+        pass
```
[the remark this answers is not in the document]
glemaitre: because up to now it was not :). In scikit-learn, since it is optional we were doing this. But if adding `pandas` and `matplotlib` as dependence then I don't need to be defensive anymore.

skore/tests/conftest.py
```diff
@@ -37,3 +47,22 @@ def in_memory_project():
...
         view_repository=view_repository,
     )
+
+
+@pytest.fixture(scope="function")
+def pyplot():
+    """Setup and teardown fixture for matplotlib.
+
+    This fixture checks if we can import matplotlib. If not, the tests will be
+    skipped. Otherwise, we close the figures before and after running the
```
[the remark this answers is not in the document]
glemaitre: I don't have a definitive answer since I did not write in scikit-learn. What I can infer is that some test might fail and might not end in the teardown maybe. So the subsequent test is here to make a clean start. However, I'm unsure.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: Done.

skore/src/skore/sklearn/_estimator/report.py
glemaitre: Happy to see what we can do to improve the current state.

skore/src/skore/sklearn/_estimator/report.py
glemaitre: I added an ending branch listing all getter and init attributes.

image

glemaitre:
OK. It should be good to go and we should be able to iterate.

sylvaincom (approved):
Many thanks for this very useful PR @glemaitre and the whole team for reviewing it! Let's iterate on sub-issues if needed

PR #998, opened 2024-12-21
fix: Use estimator whenever possible to detect the ML task

glemaitre:
I think that I saw the feedback somewhere and try to consolidate here.
Using `estimator` to find the type of ML-task is more robust that using the target `y`.

Here, I change the logic to use `estimator` as much as possible. However, if not fitted, I fallback on the target to detect which type of classification we are facing. If only `y` is provided, I use the previous approach only relying on `y`.

I added a test file.

NB: `is_clusterer` is only available from sklearn 1.6. I vendor a file `_sklearn_compat.py` that contains utility to make it easy to have developer tools working from sklearn 1.2 to 1.6. While some of them are not useful right now, I just want to vendor it completely. The package itself is tested and developed here: https://github.com/sklearn-compat/sklearn-compat.

### DIFF
--- a/skore/src/skore/sklearn/_sklearn_compat.py
+++ b/skore/src/skore/sklearn/_sklearn_compat.py
@@ -1,0 +1,870 @@
+# ruff: noqa
+"""Ease developer experience to support multiple versions of scikit-learn.
+
+This file is intended to be vendored in your project if you do not want to depend on
+`sklearn-compat` as a package. Then, you can import directly from this file.
+
+Be aware that depending on `sklearn-compat` does not add any additional dependencies:
+we are only depending on `scikit-learn`.
+
+Version: 0.1.3
+"""
+
+from __future__ import annotations
+
+import functools
+import inspect
+import platform
+import sys
+import types
+from dataclasses import dataclass, field
+from typing import Callable, Literal
+
+import sklearn
+from sklearn.utils.fixes import parse_version
+
+sklearn_version = parse_version(parse_version(sklearn.__version__).base_version)
+
+
+########################################################################################
+# The following code does not depend on the sklearn version
+########################################################################################
+
+
+# tags infrastructure
+def _dataclass_args():
+    if sys.version_info < (3, 10):
+        return {}
+    return {"slots": True}
+
+
+def get_tags(estimator):
+    """Get estimator tags in a consistent format across different sklearn versions.
+
+    This function provides compatibility between sklearn versions before and after 1.6.
+    It returns either a Tags object (sklearn >= 1.6) or a converted Tags object from
+    the dictionary format (sklearn < 1.6) containing metadata about the estimator's
+    requirements and capabilities.
+
+    Parameters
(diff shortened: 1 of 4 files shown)


skore/src/skore/sklearn/_sklearn_compat.py
```diff
@@ -0,0 +1,870 @@
+# ruff: noqa
```
thomass-dev, line 1: Shouldn't we add this file to the list of files to be ignored when computing coverage?

thomass-dev: Can you add in the `pyproject.toml` something like:

```yml
[tool.coverage.report]
exclude_also = [
    # Comments to turn coverage on and off:
    "no cover: start(?s:.)*?no cover: stop",
    # A pragma comment that excludes an entire file:
    "\\A(?s:.*# pragma: exclude file.*)\\Z",
    ]
```

And add a comment in this file `# pragma: exclude file`.

https://coverage.readthedocs.io/en/7.6.10/excluding.html#multi-line-exclusion-regexes

skore/src/skore/sklearn/find_ml_task.py
```diff
@@ -11,36 +16,49 @@ def _find_ml_task(y, estimator=None) -> MLTask:
     y : numpy.ndarray
         A target vector.
     estimator : sklearn.base.BaseEstimator, optional
-        An estimator.
+        An estimator. If fitted, we use mainly the estimator
```
thomass-dev, line 19:
```suggestion
        An estimator, used mainly if fitted.
```

skore/src/skore/sklearn/find_ml_task.py
```diff
@@ -11,36 +16,49 @@ def _find_ml_task(y, estimator=None) -> MLTask:
...
             return "multiclass-classification"
-
-    if type_of_target == "unknown":
-        return "unknown"
-
-    if "continuous" in type_of_target:
-        return "regression"
-
-    return "classification"
+        return "unsupported"
```
thomass-dev, line 64: Can you change the `unknown` type to `unsupported` in `skore/sklearn/types.py` or reuse `unknown` in this file?

thomass-dev (changes requested):
Thanks for this improvement :)

skore/src/skore/sklearn/_sklearn_compat.py
glemaitre: > Shouldn't we add this file to the list of files to be ignored when computing coverage?

Yes indeed.

Another question that I had was: do we want this file to be inside a specific module like "externals" to acknowledge that we vendor something. This is something that we are doing in some open-source project.

thomass-dev: > inside a specific module like "externals" to acknowledge that we vendor something

I agree completely.

thomass-dev (approved):

PR #1013, opened 2024-12-23
feat: Hide view functions from project

MarieS-WiMLDS:
We have functions about views in project which are documented & public, while we offer no specific feature related to a view. So far, it's something we want to deal with internally.  
Suggesting this solution, but not sure it's the best since the view functions still appear in the autocomplete. Also, there is still the "view_repository" as one of the parameters of Project while it would be best to be hidden.  

Capture d’écran du 2024-12-23 17-35-26

### DIFF
--- a/skore/src/skore/project/create.py
+++ b/skore/src/skore/project/create.py
@@ -143,7 +143,7 @@ def create(
         ) from e
 
     p = load(project_directory)
-    p.put_view("default", View(layout=[]))
+    p._put_view("default", View(layout=[]))
 
     logger.info(f"Project file '{project_directory}' was successfully created.")
     return p
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -252,11 +252,11 @@ class Project:
         """
         self.item_repository.delete_item(key)
 
-    def put_view(self, key: str, view: View):
+    def _put_view(self, key: str, view: View):
         """Add a view to the Project."""
         self.view_repository.put_view(key, view)
 
-    def get_view(self, key: str) -> View:
+    def _get_view(self, key: str) -> View:
         """Get the view corresponding to ``key`` from the Project.
 
         Parameters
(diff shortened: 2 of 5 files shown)


thomass-dev:
> We have functions about views in project which are documented & public, while we offer no specific feature related to a view. So far, it's something we want to deal with internally.

Managing a view from the developer interface was in the original design. We don't want to do this anymore? (ping @tuscland )

tuscland:
I agree that it is weird to have those repositories exposed. This is private stuff. 

However, being able to build views programmatically can be useful. I would like to discuss with Marie offline about her intent in creating this issue. Maybe it is part of a larger project that I am not aware of at the moment.

thomass-dev:
I've re-drafted the PR as long as there is an open discussion.

MarieS-WiMLDS:
My intent is to have a library that is as simple as possible for the user. When they see a function, it really provides value to them, it helps them for a given use case they have appart from skore.  
It will be useful to create repositories programmatically when the user can do something with their views. Today, it helps to visualise, but that's it, there is no sharing nor export. Therefore it's useful only for them: it's not something that will be done in high amounts, hence it seems to me that there is no need to create it automatically.

tuscland:
@thomass-dev as discussed with Marie, it is a good decision to reduce the public API space and relegate View to private scope.

glemaitre:
So if I follow well enough, this PR is superseded by https://github.com/probabl-ai/skore/pull/1052 and do not need any review and will be closed by https://github.com/probabl-ai/skore/pull/1052?

MarieS-WiMLDS:
No because #1052 is about the items, while this is about the views.

glemaitre:
OK I see, I got confused because of the related issue (https://github.com/probabl-ai/skore/issues/1045#issuecomment-2575390055) where it is also mentioned that `views` should become private.

It would really help to have a proper description in PR summary.

glemaitre:
Yes, it is just that we can add `closes #...` in the summary of the PR to close it at merge.

thomass-dev:
Closed in favor of #1052 .

PR #1048, opened 2025-01-06
fix: Fix quickstart instruction in CLI

### LINKED ISSUE #995: [cli] Quickstart working-directory option not working
### Describe the bug

When using the working directory option in quickstart, it fails. 

### Steps/Code to Reproduce

```
mkdir tmp
skore quickstart --working-dir ./tmp
```

### Expected Behavior

it should create a project in ./tmp and open it. 

### Actual Behavior

skore creates a projet but can't find it. It rejects a `ProjectLoadError`. 

### Environment

```shell
python dependencies:
        skore: 0.5.0
          pip: 24.2
   setuptools: 75.1.0
    diskcache: 5.6.3
      fastapi: 0.115.6
 plotly<6,>=5: None
      pyarrow: 18.1.0
         rich: 13.9.4
scikit-learn<1.7: None
        skops: 0.11.0
      uvicorn: 0.32.1
```

thomass-dev:
Closes #995 .

### DIFF
--- a/skore/src/skore/cli/launch_dashboard.py
+++ b/skore/src/skore/cli/launch_dashboard.py
@@ -15,7 +15,10 @@ from skore.utils._logger import logger_context
 
 
 def __launch(
-    project_name: Union[str, Path], port: int, open_browser: bool, verbose: bool = False
+    project_name: Union[str, Path],
+    port: int,
+    open_browser: bool,
+    verbose: bool = False,
 ):
     """Launch the UI to visualize a project.
 
--- a/skore/src/skore/cli/quickstart_command.py
+++ b/skore/src/skore/cli/quickstart_command.py
@@ -1,7 +1,7 @@
 """Implement the "quickstart" command."""
 
 from pathlib import Path
-from typing import Optional, Union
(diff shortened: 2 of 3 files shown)


MarieS-WiMLDS (commented):
I tried the fix and it works for me, thanks!  
However I still don't understand why the test was crashing, did you understand? I have some doubts that it's doing its job since it didn't catch the error before.

augustebaum:
> I tried the fix and it works for me, thanks! However I still don't understand why the test was crashing, did you understand? I have some doubts that it's doing its job since it didn't catch the error before.

I'm guessing it's because the test is very basic and doesn't actually run the real `launch` command.

augustebaum (approved):
Nice, thanks!

thomass-dev:
@MarieS-WiMLDS the test was wrong, this is why it was green.

image

It tested that the directory was well passed to the create function, but not to the launch function.
If we had tested `assert launch_project_name == "hello/my_project.skore"`, we would have discovered the bug.
Thanks for the feedback, well spotted.

PR #1053, opened 2025-01-07
feat(UI CrossValidationReporter): Indicate whether metrics are lower/higher-means-better

### LINKED ISSUE #903: (UI) FEAT: In skore.cross_validate plot, for each metric, specify when lower / higher means better
For each metric we display, we would like to know when lower/higher is better.  

We want to display it in several places:  
- in the CrossValidationReporter.plots.scores plot (we will have some dev constraints here because it's a legend created through plotly).
- in the dataframe "cross validation results" in the primary results page
- in the first line in the primary results page.

cf issue https://github.com/probabl-ai/skore/issues/860.

rouk1:
This PR adds icon to help understand if metric value is better when higher or better if lower.
Fixes #903 

# UI preview

https://github.com/user-attachments/assets/e6c521e9-5ab9-496c-af34-72bce9f39b0d


augustebaum (approved):
Looks good

PR #1054, opened 2025-01-07
feat(ui): Add tooltips on favorability icons

rouk1:
This PR adds tooltips on favorability icons.


augustebaum (approved):
Reviewed and tested. Nice!

PR #1060, opened 2025-01-08
feat: Use HTML to present warnings of train_test_split when in notebooks

rouk1:
This PR modifies the behavior of `train_test_split`.
From now on, in a notebook context (vscode or jupyter), messages will no longer be warnings but HTML, in order to avoid horizontal scrolling.

## Why do this?
The `train_test_split` feature is dedicated to an experimentation phase and has an educational purpose. We anticipate that most of our users will use notebooks or an interactive context. Some will disable warnings...
Therefore, I believe that warning messages are not the right communication medium during the experimentation phase. Additionally, it's not very visually appealing. While waiting to create a proper `TrainTestSplitReport`, this PR proposes an alternative.

## UI preview

https://github.com/user-attachments/assets/0b174281-a96a-4aa1-bb6a-cf475614bb46

https://github.com/user-attachments/assets/db3f5f3a-5d24-4523-b6e5-9e6125656463


glemaitre:
It might be fine temporary. My short term vision is that those warnings should be collected and added the `EstimatorReport` and access from an accessor.

sylvaincom:
Thanks, could we also display the title of the warning such as `ShuffleTrueWarning`? To kind of have a key takeaway for the user

sylvaincom:
Thanks! LGTM

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -158,10 +204,36 @@ class labels.
...
+                                re.sub("([A-Z]+)", r" \1", warning_class.__name__),
+                            ).lstrip(),
+                        )
+                        for warning, warning_class in to_display
+                    ]
+                )
+                display(HTML(markup))
+    else:
+        for warning, warning_class in to_display:
             warnings.warn(message=warning, category=warning_class, stacklevel=1)
```
glemaitre, line 237: An alternative that would work in both HTML and in prompt would be to leverage `rich`:


```python
from skore import console  # avoid circular import
    from rich.panel import Panel

    for warning_class in TRAIN_TEST_SPLIT_WARNINGS:
        warning = warning_class.check(**kwargs)

        if warning is not None:
            # Only check if warning is filtered/ignored
            if not warnings.filters or not any(
                f[0] == "ignore" and f[2] == warning_class for f in warnings.filters
            ):
                console.print(
                    Panel(
                        title=warning_class.__name__,
                        renderable=warning,
                        style="orange1",
                        border_style="cyan",
                    )
                )
```

It would provide the following outputs:

image

And here it is configure such that someone filter the warnings with ignore, then we don't show it.

It might be a bit less invasive than checking for the environment (while this code could be useful for some other stuff along the road).

sylvaincom: Using rich everywhere does provide coherence (with the EstimatorReport of #997)

sylvaincom: Can rich display `False` with code style for example?

rouk1: > Can rich display `False` with code style for example?

Yes rich is capable to display markdown. 

I hadn't considered using pure textual, it's cool and lightweight. Only down side is that we cant match hosting styles.
I'm happy to refactor that way. @MarieS-WiMLDS can you please settle this ?

MarieS-WiMLDS: What do you mean by matching hosting style? Is it light/dark theme, or is it about the font?  
I really don't mind if we don't have the default font of the user, it's more troublesome if it's about the theme, because it might affect readability.

glemaitre: It is about the theme. But the color for rich will work in a dark/light theme. However if you have a theme that use an orange background then it will not work. As a first approx, I would personally go with `rich` because it is a smaller diff and complexity.

glemaitre: An note that we have a similar limitation with the logging when creating project, etc. where we use `rich`.

MarieS-WiMLDS: Agree with Sylvain, very unlikely for a user to have a orange background!  
The "we can do it later" stuff tends to be never done, especially when it's about having clean code, so I'd prefer that we go for rich before merging, if it's compatible with your deadline of tuesday evening.

sylvaincom: You're the best ⚡️

sylvaincom:
Superseded by #1060

sylvaincom:
Superseded by https://github.com/probabl-ai/skore/pull/1060

Issue #1041, opened 2025-01-03
Annotate the version of an item


tuscland:
### Is your feature request related to a problem? Please describe.

As a data scientist, I want to annotate the specific version of an item because it helps me keep track of what is happening.

As a data scientist, I want to update or delete the note I have previously created.

As a data scientist, the notes I create from the Python API must be visible in the UI.

### Describe the solution you'd like

**Python API**

The value of the `note` kwarg is some Markdown text.

Add an note:

```python
skore.put(
    "feat_imps",
    fig,
    note="added AggJoiner with max_observation=2"
)
```

More programmatic CRUD:
```python
item_version = skore.get_item_versions("feat_imps")[0]

# If there is no note, it creates a note, otherwise update the note
item_version.set_note("set frunobulax level at 11")
note = item_version.get_note()

item_version.delete_note()
```

**UI**

Let the user interact with an item version:
- Create / Update a note using Markdown text.
- Delete the note.
- View the note of an item (last version)
- View the note of a specific item version

### Additional context

Related to #889

- [x] https://github.com/probabl-ai/skore/issues/905
- [x] https://github.com/probabl-ai/skore/pull/1084
- [x] https://github.com/probabl-ai/skore/pull/1101
- [x] https://github.com/probabl-ai/skore/pull/1082

augustebaum:
Without `put_item` it seems harder to achieve the `set_note` behaviour, as Items are just dataclasses with no link to a Project.

rouk1:
To me, we need to introduce IDs for items. They could be a simple string interpolation `{name}-{counter}`.
Then note may be a new entity with a foreign key. Even without a proper db engine we can do this in repositories.

augustebaum:
Yes, then `set_note` could be a method of `Project` instead:
```python
project.put("my_key", 3)

# Annotates the latest version by default
project.set_note("my_key", "this is 3 because 3 is good")

# Annotates version 10
project.set_note("my_key-10", "this is 3 because 3 is good")
```

thomass-dev:
@augustebaum , we shouldn't add the version in the key, prefer a dedicated parameter.

In your example, imagine you have two items in your project with keys `my_key` and `my_key-10`.
How to differentiate the version 10 of the `my_key` item to the `my_key-10` item?

I would prefer something like: 
```
 class Project:
     def set_note(self, key, note, *, version=-1):
         ...
```

MarieS-WiMLDS:
[...]

PR #1082, opened 2025-01-10
feat(UI): Add markdown editor to annotate item

rouk1:
Part of #1041 

- [x] upgrade dependencies (eslint 9 \o/)
- [x] new icons
- [x] minimal skinnable rich text editor
- [x] expose python's `set_note` as a JSON API 
- [x] plug the API in the frontend

## UI preview

https://github.com/user-attachments/assets/33cdf005-9ddd-42db-8059-c84c6a4a2433

## implementation choices

Finding a good and not too opinionated rich text editor is no easy job. After struggling with a few libs I decided that rolling a minimal editor may do the job (at least for now). 

The RichTextEditor component (probably badly named) tries to:
- help user with basic markdown syntax
- keep interactions button outside of it for best skinnability

In a future work we may move to something more powerful.


augustebaum (changes requested):
The note should be specific to an item *version*; right now the note stays the same regardless of the version

skore-ui/src/components/RichTextEditor.vue
```diff
@@ -0,0 +1,95 @@
...
+function markItalic() {
+  replaceSelectedTextWith(`*${selectedText}*`);
+}
+
+function markList() {
+  if (model.value) {
+    const lines = model.value.split("\n");
+    const textBeforeSelection = model.value.substring(0, selectionStart);
+    const selectedLineIndex = textBeforeSelection.split("\n").length - 1;
+    const list = lines.map((l, i) => (i === selectedLineIndex ? `- ${l}` : l));
```
augustebaum, line 46:
```suggestion
    const list = lines.map(l => `- ${l}`);
```

skore-ui/src/components/RichTextEditor.vue
```diff
@@ -0,0 +1,95 @@
...
+    model.value = list.join("\n");
+  }
```
augustebaum, line 48:
```suggestion
  replaceSelectedTextWith(`${selectedText.split("\n").map(l => `- ${l}`).join("\n")}`);

```

rouk1: This will potentially add a - in the middle of a line...

rouk1:
> The note should be specific to an item _version_; right now the note stays the same regardless of the version

This is fixed thanks for the report :)

skore-ui/src/components/RichTextEditor.vue
rouk1: I think the new implementation is better:

https://github.com/user-attachments/assets/234aa14c-a2d5-4955-9b5b-b682d82c9a43

augustebaum (approved):
Alright, good stuff!

sylvaincom:
Very nice, thanks for shipping this fast!

PR #1084, opened 2025-01-10
feat: Implement note-related methods

augustebaum:
New methods have been added to `Project`:

```python
# The default is to act on the latest version
project.set_note(key, message, version=-1)

project.get_note(key, message, version=-1)

project.delete_note(key, message, version=-1)
```

To this end,
- `Item` now has an additional attribute, `note`, which is a `str | None`
- `ItemRepository` has similar methods

The note related methods fail with a KeyError when the key-version combination does not exist, but:
- `get_note` returns `None` if the note is `None` (no error)
- `delete_note` never errors if the key-version combination exists (even if the note is already `None`, it will be set to `None` again)

`set_note` fails with a TypeError if `key` or `message` is not a string.

Addresses part of #1041

### DIFF
--- a/skore/src/skore/item/cross_validation_item.py
+++ b/skore/src/skore/item/cross_validation_item.py
@@ -14,7 +14,7 @@ import json
 import re
 import statistics
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, Literal, TypedDict
+from typing import TYPE_CHECKING, Any, Literal, TypedDict, Union
 
 import numpy
 import plotly.graph_objects
@@ -136,12 +136,12 @@ class CrossValidationItem(Item):
         cv_results_serialized: dict,
         estimator_info: EstimatorInfo,
         X_info: dict,
-        y_info: dict | None,
+        y_info: Union[dict, None],
         plots_bytes: dict[str, bytes],
         cv_info: dict,
-        created_at: str | None = None,
-        updated_at: str | None = None,
-        note: str | None = None,
+        created_at: Union[str, None] = None,
+        updated_at: Union[str, None] = None,
+        note: Union[str, None] = None,
     ):
         """
         Initialize a CrossValidationItem.
@@ -167,7 +167,7 @@ class CrossValidationItem(Item):
             The creation timestamp in ISO format.
         updated_at : str
             The last update timestamp in ISO format.
-        note : str | None
+        note : Union[str, None]
             An optional note.
         """
         super().__init__(created_at, updated_at, note)
--- a/skore/src/skore/item/cross_validation_item.py
+++ b/skore/src/skore/item/cross_validation_item.py
@@ -141,6 +141,7 @@ class CrossValidationItem(Item):
         cv_info: dict,
         created_at: str | None = None,
         updated_at: str | None = None,
+        note: str | None = None,
     ):
         """
         Initialize a CrossValidationItem.
@@ -166,8 +167,10 @@ class CrossValidationItem(Item):
             The creation timestamp in ISO format.
         updated_at : str
             The last update timestamp in ISO format.
+        note : str | None
+            An optional note.
         """
-        super().__init__(created_at, updated_at)
+        super().__init__(created_at, updated_at, note)
 
         self.cv_results_serialized = cv_results_serialized
         self.estimator_info = estimator_info
--- a/skore/src/skore/item/cross_validation_item.py
+++ b/skore/src/skore/item/cross_validation_item.py
@@ -14,7 +14,7 @@ import json
 import re
 import statistics
 from functools import cached_property
-from typing import TYPE_CHECKING, Any, Literal, TypedDict, Union
+from typing import TYPE_CHECKING, Any, Literal, TypedDict
 
 import numpy
 import plotly.graph_objects
@@ -136,11 +136,11 @@ class CrossValidationItem(Item):
         cv_results_serialized: dict,
         estimator_info: EstimatorInfo,
         X_info: dict,
-        y_info: Union[dict, None],
+        y_info: dict | None,
         plots_bytes: dict[str, bytes],
         cv_info: dict,
-        created_at: Union[str, None] = None,
-        updated_at: Union[str, None] = None,
+        created_at: str | None = None,
+        updated_at: str | None = None,
     ):
         """
         Initialize a CrossValidationItem.
--- a/skore/src/skore/item/item.py
+++ b/skore/src/skore/item/item.py
@@ -34,6 +34,8 @@ class Item(ABC):
         The creation timestamp of the item. If None, the current time is used.
     updated_at : str | None, optional
         The last update timestamp of the item. If None, the current time is used.
+    note : str | None
+        An optional note.
 
     Attributes
     ----------
@@ -41,17 +43,21 @@ class Item(ABC):
         The creation timestamp of the item.
     updated_at : str
         The last update timestamp of the item.
+    note : str | None
+        An optional note.
     """
 
     def __init__(
         self,
         created_at: Optional[str] = None,
         updated_at: Optional[str] = None,
+        note: Optional[str] = None,
     ):
         now = datetime.now(tz=timezone.utc).isoformat()
 
         self.created_at = created_at or now
         self.updated_at = updated_at or now
+        self.note = note
 
     @classmethod
     @abstractmethod
@@ -120,4 +126,5 @@ class Item(ABC):
         return {
             "updated_at": self.updated_at,
             "created_at": self.created_at,
+            "note": self.note,
         }
--- a/skore/src/skore/item/item.py
+++ b/skore/src/skore/item/item.py
@@ -30,11 +30,11 @@ class Item(ABC):
 
     Parameters
     ----------
-    created_at : str | None, optional
+    created_at : Union[str, None], optional
         The creation timestamp of the item. If None, the current time is used.
-    updated_at : str | None, optional
+    updated_at : Union[str, None], optional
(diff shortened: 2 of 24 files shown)


skore/src/skore/item/item.py
```diff
@@ -34,24 +34,30 @@ class Item(ABC):
         The creation timestamp of the item. If None, the current time is used.
     updated_at : str | None, optional
         The last update timestamp of the item. If None, the current time is used.
+    note : str | None
```
thomass-dev, line 37:
```suggestion
    note : str | None, optional
```

skore/src/skore/item/item.py
```diff
@@ -34,24 +34,30 @@ class Item(ABC):
...
+        An optional note.
 
     Attributes
     ----------
     created_at : str
         The creation timestamp of the item.
     updated_at : str
         The last update timestamp of the item.
+    note : str | None
```
thomass-dev, line 46:
```suggestion
    created_at : str | None, optional
        The creation timestamp of the item.
    updated_at : str | None, optional
        The last update timestamp of the item.
    note : str | None, optional
```

skore/src/skore/item/cross_validation_item.py
```diff
@@ -136,11 +136,12 @@ def __init__(
...
         X_info: dict,
-        y_info: Union[dict, None],
+        y_info: dict | None,
         plots_bytes: dict[str, bytes],
         cv_info: dict,
-        created_at: Union[str, None] = None,
-        updated_at: Union[str, None] = None,
+        created_at: str | None = None,
+        updated_at: str | None = None,
+        note: str | None = None,
```
thomass-dev, line 144: Union type using `|` is only available from `python 3.10`. Can you use `Optional` instead?

https://docs.python.org/3.12/library/typing.html#typing.Union

skore/src/skore/project/project.py
thomass-dev, line 1: I see you don't have addressed the need to add a parameter `note` to the `put` function? Any reason?

image

I know it's firstly a question of API design, especially with the put many feature, but i think it's a good feature.
What do you think @MarieS-WiMLDS ? Can we postponed this, and add "only" `<set|get|delete>_note` functions?
Or it has to come together?

skore/tests/unit/project/test_notes.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_set_note_version(in_memory_project):
+    """By default, `set_note` only attaches a message to the latest version
+    of a key."""
+    in_memory_project.put("key", "hello")
+    in_memory_project.put("key", "goodbye")
+    in_memory_project.set_note("key", "message")
+    assert in_memory_project.get_note("key", version=-1) == "message"
+    assert in_memory_project.get_note("key", version=0) is None
```
thomass-dev, line 19: Just a remark: it's not unit as your test focused on both `set_note` and `get_note`.

skore/src/skore/item/item_repository.py
```diff
@@ -153,3 +153,72 @@ def keys(self) -> list[str]:
             A list of all keys in the storage.
         """
         return list(self.storage.keys())
+
+    def set_item_note(self, key: str, message: str, *, version=-1):
+        """Attach a note to key ``key``.
```
thomass-dev, line 158:
```suggestion
        """Attach a note to the item associated with `key` in the storage.
```

skore/src/skore/project/project.py
```diff
@@ -302,3 +302,91 @@ def list_view_keys(self) -> list[str]:
...
+        --------
+        # Annotate latest version of key "key"
+        >>> project.set_note("key", "message")  # doctest: +SKIP
+
+        # Annotate first version of key "key"
+        >>> project.set_note("key", "message", version=0)  # doctest: +SKIP
+        """
+        return self.item_repository.set_item_note(
+            key=key, message=message, version=version
+        )
```
thomass-dev, line 334: Can you test that `key` and `message` are instance of `str` to ensure that we will put no other type?

thomass-dev (changes requested):
Thanks for your work. 

There is 2 important comments, about the user API and the type checking in `<set|get|delete>_notes`.
The other comments are about documentation and typing, which can be applied to all item files.

thomass-dev:
@MarieS-WiMLDS fyi, this PR can be considered as a breaking-change because it will intend to get a new field in our persistence.

When calling `get_note` and `set_note` on older skore project, an meaningless error for the user will be raises `*** KeyError: 'note'`.

thomass-dev (changes requested):
(i don't know why my last request changes is not taken into account by GH)

skore/src/skore/item/cross_validation_item.py
augustebaum: Yeah I was surprised because all other item types use `|`, and we haven't seen any problems in CI. Why do you think that is? If things don't break I would rather use `|` than `Union`.

skore/src/skore/project/project.py
augustebaum: I skipped it for now because I didn't know how to integrate it with the dict passing feature, yes

skore/tests/unit/project/test_notes.py
augustebaum: Would you like me to move it to `integration`?

skore/src/skore/item/cross_validation_item.py
thomass-dev: It's because we have disabled `mypy` in the CI, but if someone want to run `mypy` with python 3.9, it will fail.
I don't know... we just want to uniform things.

skore/src/skore/item/item_repository.py
augustebaum: How about "an item" rather than "the item", since there can be several items associated with a key? Otherwise great

skore/src/skore/project/project.py
MarieS-WiMLDS: I'm ok doing it in another PR if it allows to split the work and makes the review easier, but I think it's important to have it for the release.   

for several notes, we could have something like:  
```python
# option 1
project.put({
                    "keys": ["key1", "key2"], 
                    "values": [var1, var2],
                    "notes": ["text1", "text2"], # this is optional
})

# option 2
project.put(
       [
            {"key": "key", "value": var, "note": "text"},
            {"key": "key", "value": var, "note": "text"},
            {"key": "key", "value": var, "note": "text"},
       ]
)

# to be compared with singular put: 
project.put("key", var)

# and current multiplie: 
put({"key1": var1, "key2": var2})
```

I think I like option 2 best, because the notes are optional. Therefore, it can be troublesome to align the three lists.

(for your other comment: duly noted for the breaking change)

skore/src/skore/item/item_repository.py
thomass-dev: As u want 😄

skore/src/skore/project/project.py
thomass-dev: Your proposition change drastically the actual user API `project.put({"key1": "value1", "key2": "value2"})`.
What is the benefit of 

```python
project.put(
       [
            {"key": "key", "value": var, "note": "text"},
            {"key": "key", "value": var, "note": "text"},
            {"key": "key", "value": var, "note": "text"},
       ]
)
```

over

```python
project.put("key", var, "text"})
project.put("key", var, "text"})
project.put("key", var, "text"})
project.put("key", var, "text"})
```

As we add new parameters, i wonder about the pertinence of the "put several" feature.

skore/tests/unit/project/test_notes.py
thomass-dev: Yes please.

skore/src/skore/item/item.py
augustebaum: These are the attributes, they're not optional

skore/src/skore/project/project.py
MarieS-WiMLDS: you are completely right. I'm in favor of removing it too.

skore/src/skore/item/item_repository.py
augustebaum: Should I also update the docstrings for the `Project` methods?

skore/src/skore/project/project.py
augustebaum: Done

skore/src/skore/item/cross_validation_item.py
augustebaum: Alright, let's use Union everywhere

skore/tests/unit/project/test_notes.py
augustebaum: I'm looking into it and this would require splitting the notes tests into two far-away files, even though the tests operate at the same level of abstraction. I would rather keep them together, and do the split when the feature is tested at different levels of abstraction.

skore/src/skore/item/cross_validation_item.py
augustebaum: Done

augustebaum: Added an ADR to refer to next time the question is asked.

skore/src/skore/project/project.py
augustebaum: I'll do this in a new PR, and make the current one dependent on the new one.

augustebaum: See https://github.com/probabl-ai/skore/pull/1100. Adding to `note="..."` option to `put` will be done in a separate PR.

thomass-dev: Thanks @augustebaum .

thomass-dev (approved):

PR #1086, opened 2025-01-10
feat: Use rich Panel for showing warning in train_test_split

glemaitre:
closes #1060 

It is the alternative to #1060 using `rich`. I added a test to check that we can filter the warning since we are not using the usual `warnings` module.

In the future, we could factor out the code in a utils to be sure that we can also transform the warnings into error.


sylvaincom (approved):
It seems to work fine, many thanks!

sylvaincom (dismissed):
Please address failing tests

sylvaincom (approved):
LGTM

PR #1089, opened 2025-01-10
chore: Use emoji markup as proposed in rich

glemaitre:
Instead of using the direct emoji, let's change it to the markup one.
However, it does not fix the issue in width in ipython.


sylvaincom (approved):
Sure, thanks

PR #1090, opened 2025-01-10
docs: Add a source button for the API documentation to go to the source code on GitHub

### LINKED ISSUE #1067: docs: Add a `source` button for the API documentation to go to the source code on GitHub

API documentation

Contrary to skore v0.5.1, scikit-learn has this `source` button from the API documentation that enables to jump to the source code on GitHub (on the video my laptop is lagging a little) :

https://github.com/user-attachments/assets/7e2da473-5aed-4762-9b8e-a7041b40e992

### Suggested improvement

Implement the same Sphinx trick as in scikit-learn.

### Additional context

_No response_

sylvaincom:
Fix #1067


glemaitre:
Actually, this configuration is to get the button on the right "Source".

To get access to the source code online, I think that we use "linkcode" extension. You can check this config from imbalanced-learn: https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/doc/conf.py

It is less complex than the scikit-learn one. Search for all the match for `linkcode`. Basically, you need to:

- add a new file in a folder of the `sphinx` directory and add a path in the `conf.py`
- then add the extension at the list of the extension
- then define the path to the repository such that the source code is retrieved

sylvaincom:
Should be good now. Should I remove the stuff about https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/source-buttons.html?

glemaitre:
You can try to see the difference. If it has no effect then you can do that.

thomass-dev:
> But it does not work for now.

Please re-draft your PR is it's still the case.

sylvaincom:
>Please re-draft your PR til it's good.
My top comment was not updated, my PR is good as soon as I updated it

sylvaincom:
>You can try to see the difference. If it has no effect then you can do that.

Done, it's removed

sphinx/sphinxext/github_link.py
```diff
@@ -0,0 +1,86 @@
+# This file is copied from https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/doc/sphinxext/github_link.py
```
augustebaum, line 1:
```suggestion
"""Link docs with GitHub lines.

This file is copied from https://github.com/scikit-learn-contrib/imbalanced-learn/blob/master/doc/sphinxext/github_link.py
"""
```

PR #1097, opened 2025-01-13
docs: Fix the type of  parameter

glemaitre:
Fixing the type of `pos_label` in the different docstrings.
Here, it is the most consistent types that are accepted.


sylvaincom (approved):
LGTM, thanks

PR #1099, opened 2025-01-13
chore: Move base.py outside from _estimator module

glemaitre:
It is a refactor to ease the review in #1091.

It is only moving the file from `sklearn/_estimator/base.py` to `sklearn/_base.py` since some base class will be in common between the `EstimatorReport` and `CrossValidationReport`.

In addition, I factor out an additional base class link to the help of the `*Report` class and the `_get_cached_response_values` that will also be used in the `CrossValidationReport`.

So in short, there is no new code here.

### DIFF
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -327,10 +327,12 @@ def _get_cached_response_values(
     else:
         cache_key = (estimator_hash, prediction_method, data_source)
 
-    if data_source_hash is None and data_source == "X_y":
-        data_source_hash = joblib.hash(X)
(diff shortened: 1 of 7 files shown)


thomass-dev (dismissed):

thomass-dev:
Can you resolve the conflict @glemaitre before i merge? Thanks.

glemaitre:
Let me put the PR in draft for a couple of minutes, the time to move one test related to the base class.

glemaitre:
I'll ping when ready

glemaitre:
OK ready for review. I adapted one integration test to a unit tests. It should be fine now.

thomass-dev (approved):

Issue #1045, opened 2025-01-03
feat: As a user, i want to retrieve a `CrossValidationReporter` from a project


MarieS-WiMLDS:
### Is your feature request related to a problem? Please describe.

(I know it's not really a bug, but I wanted to highlight how weird it is for the user).

```python
# %%
from skore import CrossValidationReporter

# %% 
rep = CrossValidationReporter(clf, X, y)
project.put("rep_cv", rep)
rep # outputs screenshot 1

# %%
rep_get = project.get("rep_cv")
rep_get # outputs screenshot 2

# %% 
rep_get_item = project.get_item("rep_cv")
rep_get_item # outputs screenshot 3
```

Screen 1:  
![image](https://github.com/user-attachments/assets/bfd486d0-11cc-41ef-861a-95aa2cb2d2b9)

Screen 2:  
![image](https://github.com/user-attachments/assets/19dabfe9-5708-4f69-890c-ab557c55a750)

Screen 3:  
![image](https://github.com/user-attachments/assets/a62589bf-5d88-4735-9c94-be510e45c9b9)

This is highly inconsistant, and weird to the user. How do I know how to manipulate each of these?

### Describe the solution you'd like

The solution to me is two folds:  
- `get` and `get_item` should render a CrossValidationReporter. If I close my notebook, I want to be able to get my cv_reporter, and still use the interesting features such as `cv_rep.plots.scores`.
- having the display in the notebook is really nice! Let's keep it, but with a different function. It could be something like `skore.prettify(reporter)`

### Describe alternatives you've considered, if relevant

Remove completely the get_item function. It's in this discussion: https://github.com/probabl-ai/skore/discussions/990. Actually, I still think we should do it :).  

### Additional context

The solution proposed would also have the advantage to not loose information, while today put/get does.

thomass-dev:
It's not a bug, it's a question of design.

In the current:
- `get` returns the original data when it is possible, or a transformed data close to it,
- `get_item` returns the Item, i.e. the object that contains your (transformed) data that can be persisted.

**Currently and in state**, after you have put your object in a project, you have zero guarantee that you can programmatically retrieve your original data (for example, you can't retrieve your `CrossValidationReporter`).

I know it's an opened and actual question, we should think about it.
I removed the bug label and changed the title, because it's not a bug and involves more thinking.

glemaitre:
I think it is related with https://github.com/probabl-ai/skore/issues/949

With my user perspective, I would expect to write something like:

[...]

PR #1100, opened 2025-01-14
feat!(project): Remove "put-several" mechanic

augustebaum:
This is for several reasons:
- it is not as explicit as several simple `put` calls: the mechanic
is ambiguous with regards to atomicity (if a key-value pair is
invalid, does it make the whole operation fail?)
- the mechanic makes it complicated to add options to `put`,
e.g. the `note` option proposed in
https://github.com/probabl-ai/skore/issues/1041
or the `display_as` option proposed in
https://github.com/probabl-ai/skore/issues/1045#issuecomment-2575390055.


skore/src/skore/project/project.py
```diff
@@ -127,21 +101,10 @@ def put(self, key: Union[str, dict[str, Any]], value: Optional[Any] = MISSING):
         NotImplementedError
             If the value type is not supported.
         """
-        if value is not MISSING:
```
thomass-dev, line 130: You can remove the `MISSING` value declaration too.

skore/src/skore/project/project.py
```diff
@@ -80,44 +80,18 @@ def __init__(
...
-        be successfully inserted, *and then* an error will be raised.
-
         Parameters
         ----------
-        key : str | dict[str, Any]
-            The key to associate with ``value`` in the Project,
-            or dict of key-value pairs to add to the Project.
+        key : str
+            The key to associate with ``value`` in the Project.
         value : Any, optional
```
thomass-dev, line 93:
```suggestion
        value : Any
```

skore/src/skore/project/project.py
augustebaum: done

thomass-dev (approved):
Nice, thanks!

PR #1101, opened 2025-01-14
feat(project.put): Add `note` argument

augustebaum:
Implementation notes:

At first I used `set_note` directly, to delegate the type-checking process, but that meant touching the storage 2 times instead of 1.

### DIFF
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -1,7 +1,7 @@
 """Define a Project."""
 
 import logging
-from typing import Any, Union
+from typing import Any, Optional, Union
 
 from skore.item import (
     CrossValidationItem,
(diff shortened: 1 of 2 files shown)


skore/src/skore/project/project.py
```diff
@@ -77,7 +77,7 @@ def __init__(
         self.item_repository = item_repository
         self.view_repository = view_repository
 
-    def put(self, key: str, value: Any):
+    def put(self, key: str, value: Any, *, note: Union[str, None] = None):
```
thomass-dev, line 80:
```suggestion
    def put(self, key: str, value: Any, *, note: Optional[str] = None):
```

augustebaum: done

thomass-dev (approved):

PR #1106, opened 2025-01-14
feat: Allow for nested progress bar

glemaitre:
This PR is in anticipation of #1091 but is already in use and tested:

- we use a progress in standalone mode in `EstimatorReport.cache_predictions`
- we test the error handling and nested mode in a specific tests (and also standalone)

The nested mode will come when cross-validation is in.


skore/src/skore/__init__.py
```diff
@@ -1,14 +1,55 @@
...
 
+from rich import jupyter
 from rich.console import Console
 from rich.theme import Theme
 
 from skore.project import Project, open
 from skore.sklearn import CrossValidationReporter, EstimatorReport, train_test_split
 from skore.utils._show_versions import show_versions
 
+########################################################################################
```
glemaitre, line 14: If used in VS Code, the `ipywidget` as a white background. There is no actual way to avoid it apart of patching it.

It is pretty safe I think for the moment to do so and it improve the UI in the IDE.

glemaitre: @rouk1 you might have done a trick like this one in the previous `_environment.py` file that I recall. If you can have a look.

rouk1: That `_environment.py` was only intended to detect if the env is "HTML capable". To me has plots are on white background, I'll accept white background for progress.

rouk1 (approved):
Appart for the transparent background stuff LGTM : )

skore/src/skore/__init__.py
glemaitre: works for me

PR #1109, opened 2025-01-14
feat: Private API to expose `data_source_hash` when computing metric

glemaitre:
This PR exposes private metrics to expose the `data_source_hash` such we can compute it once and pass it around whenever possible. It is particularly interested because the hash might be expensive to compute on huge dataset.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -157,7 +157,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
                 else:
                     metric_fn = partial(self._custom_metric, metric_function=metric)
                     if scoring_kwargs is None:
-                        metrics_kwargs = {"data_source_hash": data_source_hash}
+                        metrics_kwargs = {}
                     else:
                         # check if we should pass any parameters specific to the metric
                         # callable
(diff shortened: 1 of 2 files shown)


skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -725,6 +920,13 @@ def custom_metric(
             The name of the metric. If not provided, it will be inferred from the
             metric function.
 
+        data_source : {"test", "train", "X_y"}, default="test"
+            The data source to use.
+
+            - "test" : use the test set provided when creating the reporter.
+            - "train" : use the train set provided when creating the reporter.
+            - "X_y" : use the provided `X` and `y` to compute the metric.
+
```
augustebaum, line 929: Can you add documentation for `data_source_hash`?

glemaitre: You mean on each private functions (because we don't expose it on the public one)?

glemaitre: Since they are private and only the this particular parameter change, we can find a middle ground to not repeat everything:

```
"""Private interface of {metrics} to be able to pass `data_source_hash`.

`data_source_hash` is either an `int` when we already computed the hash
and are able to pass it around or `None` and thus trigger its computation
in the underlying process.
"""
```

Would this fit what you had in mind @augustebaum?

augustebaum: Yep that works

glemaitre:
I added of additional to improve the coverage of 3 missing lines as well. But the private functions are already called by the previous code. It is that the reporter is able to called directly the private functions.

augustebaum (approved):
LGTM

PR #1113, opened 2025-01-14
chore: Differentiate __repr__ and help for report and accessors

### LINKED ISSUE #1103: Do not use the output of `EstimatorReport.help()` as a representation
As a feedback, we go that the `repr` of the `EstimatorReport` was not the best choice to be aligned with the `help` method.

We could think of a condensed `repr` that potentially indicate to use `.help()` to get more information.

glemaitre:
closes #1103 

As suggested in #1103, let's differentiate the output of `help` with the output of the `repr`. Let's have a concise output for `repr` to show where to get the information.

### DIFF
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -70,10 +70,17 @@ class _HelpMixin:
 
         console.print(self._create_help_panel())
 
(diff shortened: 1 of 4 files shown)


glemaitre:
Thanks @thomass-dev

PR #1114, opened 2025-01-14
feat: Add scoring_names to overwrite the column of the score dataframe

### LINKED ISSUE #1111: Add a parameter `scoring_names` to `EstimatorReport.report_metrics` to overwrite the default name
Currently there is no way to overwrite the names of score in `report_metrics` and it could annyoing when you pass your own score because they are inferred from the function names.

Instead we should provide a way to pass a list of strings to start that should overwrite the score of the columns in the resulting dataframe.

glemaitre:
closes #1111 

Add a `scoring_names` parameter allowing to overwrite the columns of dataframe returned by the `report_metrics`.

Build upon #1109 that expose some private functions for our internal metrics. It is nice because we can therefore expose as well the `metric_name` consistently in both `_custom_metric` and our defined metric.


augustebaum (approved):
Looks good to me, thanks

glemaitre:
Merging this one as per the approval of @augustebaum

Issue #1104, opened 2025-01-14
feat(skore): create an HTML representation for `.help` in the differen reports


glemaitre:
Right now, `EstimatorReport.help` shows a tree representation of all available functionalities available for the reporter using `rich`. However, it will become overwhelming at the time that we add more and more accessors.

Therefore, I think that we should still have `rich` backend whenever we are in a terminal but we should come with a richer representation with dropdown menu whenever we are in a notebook or IDE environment where we can leverage HTML. It will the UI/UX more enjoyable. I can also foresee to be able to click on a specific method to copy and then paste the access to a specific method of an accessor.

I'm thinking that we could then reduce the amount of information that we display with rich: never go beyond 1 level but make the help available at each level. The dynamic aspect of the HTML makes that it will be easier to handle.

anasstarfaoui:
Hello!

Some questions:

- Should descriptions be visible immediately or on hover/click?
- What's the expected screen size/environment constraints?
- Do users need to see relationships between different methods?

anasstarfaoui:
**Feedbacks** 

- On hover : Indication on parameters 
- Being able to quickly jump to documentation
- No necessity to see relations there
- Need to be easy to switch to sckit-learn documentation


2 design : 1 for terminal and one UI (HTML)

sylvaincom:
Thanks, +1

PR #1115, opened 2025-01-14
fix: Remove icon from rich representation to have a cleaner display

glemaitre:
This PR is removing the icon used in the different `help` method such that we don't have issue with spacing. https://github.com/probabl-ai/skore/issues/1104 will provide a richer output where we can leverage those types of icons instead to purely rely on textual that are grant in terminal mainly.


thomass-dev:
@glemaitre can you please resolve the conflict?

thomass-dev (approved):

PR #1126, opened 2025-01-15
fix: Support passing pos_label as scorer param or as method param

glemaitre:
This PR makes sure that we can pass `pos_label` either as a parameter to a scikit-learn scorer or as a parameter of the skore function. If both are provided, we just make sure that the values of both input are coherent. Otherwise, we raise an informative error.


sylvaincom (approved):
I see only green flags, thanks!

PR #1130, opened 2025-01-16
fix(notes api): Make ItemRepository save notes properly

augustebaum:
The storage API was not used correctly which resulted in not saving the notes properly; this affected `set_note` and `delete_note`.
The issue was not detected in unit tests because they relied on an in-memory project, not a "real" on-disk project.


rouk1 (approved):
thanks : )

thomass-dev (approved):

PR #1141, opened 2025-01-16
ci: Fix timeout by upgrading scikit-learn to latest bugfix

glemaitre:
Looking a bit more, it seems not only the `coverage` job is problematic. So since we are using `-n auto`, it means that we potentially allocate too many cores (logical + physical) and potentially some test allocate even more parallel jobs (usually I tried some jobs with `n_jobs=2`). So we could have a problem of over-subscription and not problem of deadlocks.

Let see if we have the same issue by limiting the number of worker.

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -127,6 +127,11 @@ jobs:
           # Install `skore` without its dependencies, which are present in the venv
           wheel=(dist/*.whl); python -m pip install --force-reinstall --no-deps "${wheel}"
 
+      - name: Show dependencies versions
+        working-directory: skore/
+        run: |
+          python -c "import skore; skore.show_versions()"
+
       - name: Test without coverage
         if: ${{ ! matrix.coverage }}
         timeout-minutes: 10
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -131,7 +131,7 @@ jobs:
         if: ${{ ! matrix.coverage }}
         timeout-minutes: 10
         working-directory: skore/
-        run: python -m pytest -n 3 src/ tests/ --no-cov
+        run: python -m pytest src/ tests/ --no-cov
 
       - name: Test with coverage
         if: ${{ matrix.coverage }}
@@ -139,7 +139,7 @@ jobs:
         working-directory: skore/
         run: |
           mkdir coverage
-          python -m pytest -n 3 src/ tests/ --junitxml=coverage/coverage.xml --cov-config=pyproject.toml --cov | tee coverage/coverage.txt
+          python -m pytest src/ tests/ --junitxml=coverage/coverage.xml --cov-config=pyproject.toml --cov | tee coverage/coverage.txt
 
       - name: Upload coverage reports
         if: ${{ matrix.coverage && (github.event_name == 'pull_request') }}
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -49,17 +49,17 @@ jobs:
       matrix:
         os: ["ubuntu-latest", "windows-latest"]
         python: ["3.9", "3.10", "3.11", "3.12"]
-        scikit-learn: ["1.6"]
+        scikit-learn: ["1.6.1"]
         include:
           - os: "ubuntu-latest"
             python: "3.12"
-            scikit-learn: "1.4"
+            scikit-learn: "1.4.2"
           - os: "ubuntu-latest"
             python: "3.12"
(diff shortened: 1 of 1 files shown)


thomass-dev:
Can you link a failing pipeline?

glemaitre:
@thomass-dev for instance on `main`: https://github.com/probabl-ai/skore/actions/runs/12813764685/job/35728620204

Usually we are < 1 minute for the job, but this Python 3.12 with sklearn 1.4, goes sideways.
In the last build without `pytest-xdist`, we are as fast as with it. I don't know if it would be more stable. It was still slower on the last build from this PR.

Let's launch a couple of time and see if we trigger the timeout still.

.github/workflows/backend.yml
```diff
@@ -127,11 +127,16 @@ jobs:
           # Install `skore` without its dependencies, which are present in the venv
           wheel=(dist/*.whl); python -m pip install --force-reinstall --no-deps "${wheel}"
 
+      - name: Show dependencies versions
```
glemaitre, line 130: I added this part because it is useful to check the dependency version to have some context.

glemaitre:
OK the job is back to 27 seconds. I recall that it was due some sorting algorithm in C++ that was impacting the trees.

Now the question is @thomass-dev how to do you want to alter the actions file.

My feeling is roughly:

- we should install scikit-learn with `~=` specifier (i.e. `scikit-learn~=1.4.0` to get 1.4.2). In this way we get the latest bug fix release for all minor releases tested
- however, the cache mechanism that you have is at the level of the minor number
- it would nice to fetch the number of the latest available bug fix release to invalidate the cache at a new available release without the need to change the matrix.

.github/workflows/backend.yml
```diff
@@ -102,7 +102,7 @@ jobs:
         run: |
           python -m pip install --upgrade "pip"
           python -m pip install --upgrade "build"
-          python -m pip install --upgrade "scikit-learn ==${{ matrix.scikit-learn }}"
+          python -m pip install --upgrade "scikit-learn~=${{ matrix.scikit-learn }}.0"
 
           # Install `skore` and its dependencies
           python -m pip install --upgrade ".[test]"
```
thomass-dev, line 108:
```suggestion
          python -m pip install --upgrade --upgrade-strategy=eager ".[test]"
```

.github/workflows/backend.yml
```diff
@@ -102,7 +102,7 @@ jobs:
         run: |
           python -m pip install --upgrade "pip"
           python -m pip install --upgrade "build"
-          python -m pip install --upgrade "scikit-learn ==${{ matrix.scikit-learn }}"
+          python -m pip install --upgrade "scikit-learn~=${{ matrix.scikit-learn }}.0"
```
thomass-dev, line 105:
```suggestion
          python -m pip install --upgrade "scikit-learn ==${{ matrix.scikit-learn }}.*"
```

I prefer to have less changes. Can you add a comment to explain that without this, it will not update to last bugfix?

.github/workflows/backend.yml
thomass-dev: Just to ensure that we upgrade dependencies of dependencies 😮‍💨

.github/workflows/backend.yml
glemaitre: Yep. done

.github/workflows/backend.yml
```diff
@@ -127,11 +129,16 @@ jobs:
...
+        working-directory: skore/
+        run: |
+          python -c "import skore; skore.show_versions()"
+
       - name: Test without coverage
         if: ${{ ! matrix.coverage }}
         timeout-minutes: 10
         working-directory: skore/
-        run: python -m pytest -n auto src/ tests/ --no-cov
+        run: python -m pytest src/ tests/ --no-cov
```
thomass-dev, line 141: Can you please rollback this change and reset `-n auto` in coverage too?

.github/workflows/backend.yml
```diff
@@ -127,11 +127,16 @@ jobs:
           # Install `skore` without its dependencies, which are present in the venv
           wheel=(dist/*.whl); python -m pip install --force-reinstall --no-deps "${wheel}"
 
+      - name: Show dependencies versions
+        working-directory: skore/
+        run: |
+          python -c "import skore; skore.show_versions()"
```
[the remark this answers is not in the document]
glemaitre: I would keep this first and later we can add the `pip freeze` if it is not enough.

thomass-dev (approved):

PR #1144, opened 2025-01-17
feat(UI): Blur markdown editor when user clicks outside 

rouk1:
And save a trimmed version of the text.

As requested in feedbacks.

### DIFF
--- a/skore-ui/src/stores/project.ts
+++ b/skore-ui/src/stores/project.ts
@@ -96,8 +96,11 @@ export const useProjectStore = defineStore("project", () => {
    */
   let _stopBackendPolling: (() => void) | null = null;
   async function startBackendPolling() {
-    _isCanceledCall = false;
-    _stopBackendPolling = await poll(fetch, 1500);
+    // ensure that there is only one polling running
+    if (_stopBackendPolling === null) {
+      _isCanceledCall = false;
+      _stopBackendPolling = await poll(fetch, 1500);
+    }
   }
 
   /**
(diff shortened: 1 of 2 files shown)


augustebaum (changes requested):
Doesn't seem to work on firefox. In the screenrecord I enter text and then click outside of the box, and nothing happens.

https://github.com/user-attachments/assets/3760eac6-2685-49c0-8e7b-a4b50ca71ef2

augustebaum (changes requested):
I found a strange bug on Firefox.
Clicking out of the box triggers a bunch of requests in quick succession, which takes up my RAM and persists across refreshes (even hard refresh). Only killing the browser process makes it stop.

See attached screenrecord: at first I save with `Shift+Enter` (which btw seems to leave a newline), it works; next I click outside the box and the behaviour occurs.


https://github.com/user-attachments/assets/179ab2bf-d01d-4780-ab92-df4ea03e358f

rouk1:
> I found a strange bug on Firefox. Clicking out of the box triggers a bunch of requests in quick succession, which takes up my RAM and persists across refreshes (even hard refresh). Only killing the browser process makes it stop.
> 
> See attached screenrecord: at first I save with `Shift+Enter` (which btw seems to leave a newline), it works; next I click outside the box and the behaviour occurs.
> 2025-01-17T15_52_43_screen_record.mp4

Each keystroke is sent to the backend to auto save you're note. Firefox network panel is very bad at handling tons of request. Cant do much on this...

auguste-probabl:
Bug was dealt with. Thanks!

auguste-probabl (approved):
Great work!

PR #1148, opened 2025-01-17
feat(UI): New rules to format numbers in cv report

### LINKED ISSUE #394: Numerical precision in skore-ui
Numbers in DataFrame tables have too many digits. This results in a poor presentation.

I would not recommend to rely on the browser locale to format things. It would make things difficult to compare for teams that work in international environments (some will have a comma decimal separator, and other will have a dot). The tool we are building is suited to data-savvy users.

So, we can offer a parameter to specify number formatting as project-level settings. The default format should allow for a precision of 3 digits.

Here is a suggested approach for formatting numbers:

```javascript
const formatter = new Intl.NumberFormat('en-US', {
  minimumFractionDigits: 0,
  maximumFractionDigits: 3
});
function formatNumber(number) {
  return formatter.format(number);
}
```

Which would result in 3 project-level settings:

```
ui.number-format.locale = 'en-US'
ui.number-format.minimum-fraction-digits = 0
ui.number-format.maximum-fraction-digits = 3
```

I'm open to improved or different ideas.

rouk1:
Closes #394

### DIFF
--- a/skore-ui/src/components/CrossValidationReportResults.vue
+++ b/skore-ui/src/components/CrossValidationReportResults.vue
@@ -21,13 +21,10 @@ function exponential(x: number) {
   if (typeof x !== "number") {
     return x;
   }
-  const expoForm = x.toExponential(2);
-  const expo = expoForm.split("e")[1];
-  if (Math.abs(parseInt(expo)) < 2) {
-    return x.toFixed(2);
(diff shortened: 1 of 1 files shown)


augustebaum (approved):
Next time consider adding a screenshot

Issue #1178, opened 2025-01-20
docs: In Sphinx examples, remove the temp files


sylvaincom:
### Which part of the documentation needs improvement?

Sphinx examples (user guide)

### Describe the problem found in the documentation

Currently, in the Sphinx examples in the user guide, we have the following:

<img src="https://github.com/user-attachments/assets/1edce0af-9c25-4498-81ce-fa50e7778e26" width="450">

then at the end:
```python
temp_dir.cleanup()
```

Using temp files is not intuitive for a new user.

But we can't have `project.create(overwrite=True)` in the quick start for example, as it sends the message that this is what users should do.

We want to make sure to not have too many things stored in the place where to doc is deployed (S3 or whatever)? As Sphinx actually runs the examples at each build, we would add several histories in the skore project on the S3. **In real life, a user would not want to clear the cache, but when building the Sphinx examples we want to clear the cache.**

### Suggested improvement

*Quickfix "hack" suggested by @MarieS-WiMLDS*

Instead of cleaning the temp dir at the end, we do `project.create(overwrite=True)`. This way, this code that we don't want to be hastly copy/pasted is at the end of the notebook.

![Image](https://github.com/user-attachments/assets/b5e26258-e51e-4251-94f3-01c7df3dd41c)

glemaitre:
I personally prefer to keep as-is for the moment and await for creating a `project.clear()` because it looks weird to create a project at the end of the file.

MarieS-WiMLDS:
isn't this function redundant with overwrite? They both remove everything from the project and start with an empty one.

glemaitre:
The problem for me is not overwrite but using create to remove something. I find it counter intuitive. 

Also the clear could have two ways of working: just removing the items and make it possible to still use the project or completely wiping out the project to really clean up and make the project in a stage that you need to recreate one if needed. 

I can see usage of both and in the documentation it is the second one.

sylvaincom:
Setting this issue back to "In preparation"

PR #1193, opened 2025-01-21
remove tempfiles from docs

MarieS-WiMLDS:
fixes https://github.com/probabl-ai/skore/issues/1178

### DIFF
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -272,3 +272,11 @@ class Project:
         >>> project.delete_note("key", version=0)  # doctest: +SKIP
         """
         return self.item_repository.delete_item_note(key=key, version=version)
+
+    def clear(self):
+        """Delete all the contents of the project."""
+        # delete all the items
+        for item_key in self.keys():
+            self.delete(item_key)
+        for view_key in self.view_repository:
+            self.view_repository.delete_view(view_key)
--- a/skore/tests/unit/project/test_project.py
+++ b/skore/tests/unit/project/test_project.py
@@ -266,6 +266,13 @@ def test_put_wrong_key_and_value_raise(in_memory_project):
         in_memory_project.put(0, (lambda: "unsupported object"))
 
 
+def test_clear(in_memory_project):
+    in_memory_project.put("an int", 1)
+    in_memory_project.put("a str", "some text here to have fun")
+    in_memory_project.clear()
+    assert len(in_memory_project.keys()) == 0
+
+
 test_cases = [
     (
         "a" * 250,


skore/src/skore/project/project.py
```diff
@@ -272,3 +272,11 @@ def delete_note(self, key: str, *, version=-1):
...
         """
         return self.item_repository.delete_item_note(key=key, version=version)
+
+    def clear(self):
+        """Delete all the contents of the project."""
+        # delete all the items
+        for item_key in self.keys():
+            self.delete(item_key)
+        for view_key in self.view_repository:
+            self.view_repository.delete_view(view_key)
```
rouk1, line 282: This two lines are untested : )

rouk1 (changes requested):

skore/src/skore/project/project.py
MarieS-WiMLDS: so you spotted it... How do you usually test things related to views? I can't create a view programmatically, can I?

auguste-probabl: Something like this should work:

```python
from skore.persistence.view.view import View

project.view_repository.put_view("my_view", View(layout=["key1", "key"2]))
```

auguste-probabl (changes requested):
I don't love this solution because it makes all the examples dependent on one-another. If example 1 fails halfway through, then `clear` won't be run and example 2 will start with a non-empty project, which could make it fail, etc.

examples/getting_started/plot_skore_product_tour.py
```diff
@@ -35,38 +35,13 @@
...
-temp_dir_path = Path(temp_dir.name)
-
-# %%
-# Then, we create and load the skore project from this temporary directory:
-
 # %%
 import skore
 
-my_project = skore.open(temp_dir_path / "my_project")
+my_project = skore.open("my_project", create=True)
```
auguste-probabl, line 44:
```suggestion
my_project = skore.open("my_project")
```
`create=True` is the default

skore/src/skore/project/project.py
thomass-dev: You should put a default view as it is done in the `create` script.

auguste-probabl:
But the `clear` method is fine IMO

skore/src/skore/project/project.py
thomass-dev: And to test, you have an access to the repository and it's storage.

examples/getting_started/plot_skore_product_tour.py
MarieS-WiMLDS: I like that it's explicit and therefore as a user I feel that I can copy/paste even if I don't have anything before.  
Is it bad practice? or misleading?

examples/getting_started/plot_skore_product_tour.py
```diff
@@ -288,7 +263,7 @@
 # Cleanup the project
 # -------------------
 #
-# Removing the temporary directory:
+# Let's clean the skore project to avoid conflict with other examples.
```
thomass-dev, line 266: I don't think it's relevant as you can just name your project differently.

thomass-dev: Let's rename each example project differently, and find IMO a better description for the use of `clear`?

MarieS-WiMLDS: if we don't use it for the documentation, we can remove the clear function altogether, because it's kind of doing the same thing as `skore.open("a_prj_name", overwrite=True)`.   
Also, in a user test this morning, the user did several copy/paste from various examples. It triggered a problem when the projects were not named alike, so I'd prefer to have it everywhere the same.

skore/src/skore/project/project.py
MarieS-WiMLDS: test and default view added!

examples/getting_started/plot_skore_product_tour.py
thomass-dev: > Also, in a user test this morning, the user did several copy/paste from various examples. It triggered a problem when the projects were not named alike, so I'd prefer to have it everywhere the same.

Can you add a test showing this problem?

auguste-probabl:
> > I don't love this solution because it makes all the examples dependent on one-another. If example 1 fails halfway through, then clear won't be run and example 2 will start with a non-empty project, which could make it fail, etc.
> 
> Is this likely to happen? Or it will trigger a red CI and therefore shouldn't land in main? (unless we force it)

Yes, tests fail all the time, and we would prefer that 1 example failed rather than all of them; that way it's easier to understand what went wrong and fix things

Another thing I thought of later is that this problem prevents us from running the examples in parallel

examples/getting_started/plot_skore_product_tour.py
auguste-probabl: I don't mind explicit, but the goal of `open` is to do the "right thing" without having to think much about all the arguments. Maybe this is telling us that `open` is a bad name for this function?

examples/getting_started/plot_skore_product_tour.py
MarieS-WiMLDS: it's not something that can be really caught by a test, the user copy/pasted something from [this ex](https://skore.probabl.ai/0.6/auto_examples/getting_started/plot_quick_start.html), where the project is called `quick_start`, and then from [this oher one](https://skore.probabl.ai/0.6/auto_examples/getting_started/plot_skore_product_tour.html), where the project is called `my_project`. When you don't yet clearly understand the launch command, and you're reading docs in diagonal, it's not clear that these variables quick_start and my_project are the projects names, and that it should be always the same.

MarieS-WiMLDS:
I can't find any alternative solution for now to answer to all our goals. I created an issue to formalize this: https://github.com/probabl-ai/skore/issues/1200. In the mean time, if it's ok for you, I'd like to proceed with this solution to remove tempfile for this afternoon.

auguste-probabl:
> If example 1 fails halfway through, then clear won't be run and example 2 will start with a non-empty project, which could make it fail, etc.

To mitigate this could you follow https://sphinx-gallery.github.io/stable/configuration.html#abort-build-on-first-fail? This way if an example fails, we can see quicker where the problem comes from (rather than 10 other examples failing just because the first one failed)

examples/getting_started/plot_skore_getting_started.py
```diff
@@ -34,38 +34,13 @@
...
-temp_dir_path = Path(temp_dir.name)
-
-# %%
-# Then, we create and load the skore project from this temporary directory:
-
 # %%
 import skore
 
-my_project = skore.open(temp_dir_path / "my_project")
+my_project = skore.open("my_project", create=True)
```
sylvaincom, line 43: By default, `create=True`
```suggestion
my_project = skore.open("my_project")
```

examples/getting_started/plot_tracking_items.py
```diff
@@ -14,23 +14,12 @@
...
-
-# %%
-# We create and load the skore project from this temporary directory:
+# We create and load the skore project in the current directory:
 
 # %%
 import skore
 
-my_project = skore.open(temp_dir_path / "my_project")
+my_project = skore.open("my_project", create=True)
```
sylvaincom, line 22:
```suggestion
my_project = skore.open("my_project")
```

examples/getting_started/plot_working_with_projects.py
```diff
@@ -14,23 +14,12 @@
...
-
-# %%
-# We create and load the skore project from this temporary directory:
+# We create and load the skore project from the current directory:
 
 # %%
 import skore
 
-my_project = skore.open(temp_dir_path / "my_project")
+my_project = skore.open("my_project", create=True)
```
sylvaincom, line 22:
```suggestion
my_project = skore.open("my_project")
```

examples/model_evaluation/plot_cross_validate.py
```diff
@@ -19,23 +19,12 @@
...
-
-# %%
-# We create and load the skore project from this temporary directory:
+# We create and load the skore project from the current directory:
 
 # %%
 import skore
 
-my_project = skore.open(temp_dir_path / "my_project")
+my_project = skore.open("my_project", create=True)
```
sylvaincom, line 27:
```suggestion
my_project = skore.open("my_project")
```

examples/model_evaluation/plot_train_test_split.py
```diff
@@ -12,25 +12,13 @@
...
-
 # %%
-# We create and load the skore project from this temporary directory:
+# We create and load the skore project from the current directory:
 
 # %%
 import skore
 
-my_project = skore.open(temp_dir_path / "my_project")
+my_project = skore.open("my_project", create=True)
```
sylvaincom, line 21:
```suggestion
my_project = skore.open("my_project")
```

sylvaincom (changes requested):

sylvaincom (commented):
Could we also apply this to the quick start example? 🙏

examples/getting_started/plot_skore_getting_started.py
```diff
@@ -347,7 +322,7 @@
 # Cleanup the project
 # -------------------
 #
-# Removing the temporary directory:
+# Let's clean the skore project to avoid conflict with other examples.
```
sylvaincom, line 325:
```suggestion
# Let's clear the skore project (to avoid conflict with other documentation examples).
```

examples/getting_started/plot_skore_getting_started.py
MarieS-WiMLDS: it was also a remark from Auguste, and here is my answer:  
image  

Do you still want me to remove `create=True`?

examples/getting_started/plot_quick_start.py
```diff
@@ -69,3 +69,12 @@
 # .. admonition:: What's next?
 #
 #    For a more in-depth guide, see our :ref:`example_skore_getting_started` page!
+
+# %%
+# Cleanup the project
+# -------------------
+#
+# Let's clean the skore project to avoid conflict with other examples.
```
sylvaincom, line 77:
```suggestion
# Let's clear the skore project (to avoid any conflict with other documentation examples).
```

examples/getting_started/plot_tracking_items.py
```diff
@@ -130,7 +119,7 @@
 # Cleanup the project
 # -------------------
 #
-# Removing the temporary directory:
+# Let's clean the skore project to avoid conflict with other examples.
```
sylvaincom, line 122:
```suggestion
# Let's clear the skore project (to avoid any conflict with other documentation examples).
```

examples/model_evaluation/plot_cross_validate.py
```diff
@@ -170,7 +159,7 @@
 # Cleanup the project
 # -------------------
 #
-# Removing the temporary directory:
+# Let's clean the skore project to avoid conflict with other examples.
```
sylvaincom, line 162:
```suggestion
# Let's clear the skore project (to avoid any conflict with other documentation examples).
```

examples/model_evaluation/plot_train_test_split.py
```diff
@@ -258,7 +246,7 @@
 # Cleanup the project
 # -------------------
 #
-# Removing the temporary directory:
+# Let's clean the skore project to avoid conflict with other examples.
```
sylvaincom, line 249:
```suggestion
# Let's clear the skore project (to avoid any conflict with other documentation examples).
```

sylvaincom (approved):
Many thanks!

examples/getting_started/plot_skore_product_tour.py
MarieS-WiMLDS: I think the open name is ok, but i would it more natural to have "create = False" by default.

Issue #1160, opened 2025-01-20
feat: Refine the constructor of Project to be user-friendly


thomass-dev:
After https://github.com/probabl-ai/skore/pull/1052 which refactors the user API to hide all notions of `Item`, we can go further and hide all notions of `Repository` too.

Partially addresses https://github.com/probabl-ai/skore/issues/1000.

augustebaum:
Can you explain how you intend to do this?

MarieS-WiMLDS:
to close this, another PR must be done to fix this:  
> Using clear, I noticed that it doesn't refresh the activity feed: we still have old items displayed, unless I change window or close and restart the ui. When doing clear, is it possible to force this refresh?

MarieS-WiMLDS:
Reopening because of comment above; maybe it's more for @rouk1 to tackle this last mile?

rouk1:
To me the activity feed is like a log book. I find it a bit strange to rewrite the past. Maybe adding a log in the feed saying that  the project as been cleared could help a bit ?

Anyhow is one deletes only one item should it be removed from the activity feed ?

MarieS-WiMLDS:
That's a very good question. Indeed, I would expect to have the same behavior when deleting & clearing. I would also expect the activity feed to be the same when I turn on/off skore UI.  
Therefore, yes, I would say something deleted should be deleted from the activity feed.  
I like the idea of adding a log saying the project has been cleared. Instead of this text we have when we start from scratch, if could be slightly different, to "The project has been cleared. New items will be added automatically."

![Image](https://github.com/user-attachments/assets/05318bae-5f66-4e37-b840-3218276c5acc)

rouk1:
Hence it's not an activity feed anymore, its a project observer. Deleting items from the feed is complicated to achieve with a good UX. For example if the item you've delete is outside of your viewport how do we notify the user of this. 

Is what you want a "view" of the project with all the active items ? Hence should we group again item versions ?

MarieS-WiMLDS:
Ok, it seems that it's not that trivial. Let's keep the current behavior, and we will improve it once we know more about what we want to do with the activity feed. If it's purely personal, maybe the current behavior is fine, while if it's collective, we might want to be more precise in terms of deletion.

PR #1157, opened 2025-01-19
feat: Make `open` to spin-off UI server and to be used in CLI and notebook

glemaitre:
closes #1002 

This PR proposed to resolve the issue discussed in #1002:

- `open` should provide the possibility to spin-off a UI server
- `open` can therefore be used in the notebook and with the CLI
- `kill` to kill all alive servers (normally we tested that we should not have some case or zombie processes).

So in short, we allow for the following:

```python
import skore

project = skore.open("my_project")
project.shutdown_web_ui()
```

### Implementation details

We launch the UI in a separate process. Why process: if we would attach to the async loop for the notebook or run in a separate thread from the main process, we might get into trouble when this main process starts to execute large computation making the UI unresponsive.

So we are starting a new process with the webserver inside. For the moment, we expect to have 1 skore project associated with 1 skore UI server. So we don't allow to spin a new webserver if one already exist and is associated with a running skore project. Instead, we provide the link to the server.

Also, we have a `rejoin` mechanism: if a project was closed but the UI was not, then when opening a project, we reattach the web server to it.

Finally, we found with @rouk1 that you have the following use-cases:

- CI/CD: you want `serve=False` when calling `open`
- standalone script: You might be interested in `serve=True` and in this case, you most probably want to keep the webs server alive (so deamon + atexit is not what we want)
- jupyter notebook: You might be interested in `serve=True` and you will be interested to have access to the UI the time that the notebook is alive. The current setup will kill the web server.

So we added a `keep_alive` option that detect if you are in a specific environment and choose a sensible default but it can be easily overwritten.

### Notes

Some of the covering is not really working with the multiprocessing because we explicitly check for the console message in the test and the coverage mentioned that we don't cover it.

### TODO

- [x] write tests for the `keep_alive` option
- [x] write tests for the `kill` cli

Issue #1002, opened 2024-12-22
skore.show_activity() a simple helper function to get users started


GaelVaroquaux:
### Is your feature request related to a problem? Please describe.

As mentioned in https://github.com/probabl-ai/skore/issues/1001 , we need to bridge the skore notions of store, activity feed, UI, to the way many data scientists currently work: interactive work in vscode and jupyter notebooks.

Getting immediate benefits of skore requires setting up a project, and opening a web browser. This is a central sequence of operation that we can expect users to be doing over and over.

It requires knowing (and remembering) that "skore launch project_name" must be run.

It requires to open a shell, and run the corresponding command, in the right python environment. Shell commands and Python environment mess is a place where it is easy to loose users.

### Describe the solution you'd like

I would like a small function to do all this for me. 

```python
>>> import skore
>>> p = skore.show_activity()   # Here a browser is opened with the activity feed
>>> p.put(reporter)
```

Short term, it helps getting people started (eg by inserting it in every one of our example) and stupid mistakes. In the long term, we can insert more cleverness (like detecting that we are on a compute cluster and not actually starting a web browser there :) , avoid starting multiple browsers for the same project on the same computer, potentially closing the server at the exit of the Python executable where it was started, to avoid pollution of servers).

Unlike `skore.create`, `skore.show_activity` should by default:
1. Create the project with a default name
2. Not complain if the project already exists, but not overwrite it

glemaitre:
> Unlike skore.create, skore.show_activity should by default:
>   - Create the project with a default name
>   - Not complain if the project already exists, but not overwrite it

To me it seems like exposing the current `__quickstart` private function that is already available via the CLI. But it should be fixed as spotted in #995

glemaitre:
Let's start a discussion about the scope of functionality here.

@MarieS-WiMLDS provides her views on the feature in https://github.com/probabl-ai/skore/pull/1194#pullrequestreview-2580631526

> I disagree with @glemaitre on the fact that the creation function should also open the UI. these are two different things, we should have a python func to launch UI indeed, but it would be dedicated to this and only this.

@GaelVaroquaux mentioned his wishes as:

> Unlike skore.create, skore.show_activity should by default:
> - Create the project with a default name
> - Not complain if the project already exists, but not overwrite it

[...]

PR #1194, opened 2025-01-21
feat: Refine the constructor of Project to hide repositories

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/1160

This PR intends to ease the project API by centralizing all the action in the constructor.

To note:
- Remove all actions from the CLI except "launch UI", which implicitly create the specified project.
- Rename the command `skore` to `skore-ui`.
- Release the constraints on project name.
- Remove all project management scripts and `open` function in favor of the project class constructor.
- Create implicitly the project when calling the constructor.
- Add project `clear` function to allow user to **explicitly** clear a project.

---

```python
# create project

import skore
project = skore.Project("project.skore")
```
```python
# work on existing project

import skore
project = skore.Project("project.skore", exist_ok=True)
```
```python
# work on existing project and clear explicitly

import skore
project = skore.Project("project.skore", exist_ok=True)
project.clear()
```
```python
# work on existing project, but without knowing it

import skore
project = skore.Project("project.skore")

FileExistsError                           Traceback (most recent call last)
[...]
FileExistsError: Project 'project.skore' already exists.
```


skore/src/skore/project/project.py
```diff
@@ -36,11 +43,38 @@ class Project:
 
     def __init__(
         self,
-        item_repository: ItemRepository,
-        view_repository: ViewRepository,
+        path: Optional[Union[str, Path]] = "project.skore",
```
glemaitre, line 46: Interesting thing. I actually going this path in #1157

thomass-dev: I hope to finish this PR before merging yours.

glemaitre: :P

glemaitre:
I'm almost happy with what I see:

- I like some simplification here.
- I don't like that we don't have the `open` anymore. We should be able to trigger the UI launch on the Python side. Having this in the `Project` constructor does not make any sense. So we will need a kind of factory. This `open` will have a symmetry between the CLI and Python. But it is orthogonal with the work here: It only need to create the project and attach the server info to it.

skore/src/skore/project/project.py
```diff
@@ -36,11 +33,62 @@ class Project:
...
+        """
+        Initialize a Project.
+
+        Initialize a project, by creating a new project or by loading an existing one.
+
+        Parameters
+        ----------
+        path : str or Path, optional
+            The path of the project to initialize, default "./project.skore".
+        exist_ok: bool, optional
```
MarieS-WiMLDS, line 49: I find this parameter not easy to grasp. I'd rather have something as `load` instead (as we would use to open a json file for instance).

skore/src/skore/project/project.py
```diff
@@ -36,11 +33,62 @@ class Project:
 
     def __init__(
         self,
-        item_repository: ItemRepository,
-        view_repository: ViewRepository,
+        path: Optional[Union[str, Path]] = "project.skore",
+        *,
```
MarieS-WiMLDS, line 37: what is this for? is there an additional parameter we can give to the project as a user?

skore/src/skore/project/project.py
```diff
@@ -36,11 +33,62 @@ class Project:
...
+        )
+        self.view_repository = ViewRepository(
+            DiskCacheStorage(self.__view_storage_dirpath)
+        )
+
+        # Ensure default view is available
+        if "default" not in self.view_repository:
+            self.view_repository.put_view("default", View(layout=[]))
+
+    def clear(self):
```
MarieS-WiMLDS, line 82: Using clear, I noticed that it doesn't refresh the activity feed: we still have old items displayed, unless I change window or close and restart the ui. When doing clear, is it possible to force this refresh?

skore/src/skore/cli/cli.py
```diff
@@ -39,87 +38,24 @@ def cli(args: list[str]):
...
-        __quickstart(
-            project_name=parsed_args.project_name,
-            overwrite=parsed_args.overwrite,
-            port=parsed_args.port,
-            open_browser=parsed_args.open_browser,
-            verbose=parsed_args.verbose,
-        )
-    else:
-        parser.print_help()
+def cli(args: list[str]):
```
MarieS-WiMLDS, line 51: the cli is not working properly anymore, and the helper doesn't tell me all the functions I can access:  
image

MarieS-WiMLDS (changes requested):
Thank you for these changes!! It should really make it easier for users to grasp how to create a project.  

I have two additional comments:  
- the first one is about the `delete` function, it looks like a bit a function that would delete the project now. I don't really have an idea how to solve this. A mitigation could be to have a delete_project function (function that i'd like to have anyway, but it's not a great solution to be explicit about the purpose of delete)
- I disagree with @glemaitre on the fact that the creation function should also open the UI. these are two different things, we should have a python func to launch UI indeed, but it would be dedicated to this and only this.

skore/src/skore/cli/cli.py
thomass-dev: Please look at `skore -h`

skore/src/skore/project/project.py
thomass-dev: I think it's a UI issue, which should be addressed in another PR.

skore/src/skore/project/project.py
thomass-dev: The `*` is used to force kwargs only argument.

skore/src/skore/project/project.py
thomass-dev: This parameter is the same as the one in the `mkdir` python function. I find it more explicit than your proposition `load`.

With `load=True`, there is an ambiguity when the project doesn't exist.
With `exist_ok=True`, it is clear that it's only refer to the existence of the project.

skore/src/skore/cli/cli.py
thomass-dev: @glemaitre i see with this traceback that there is a strange behaviour with `ColorArgumentParser` on unrecognized parameter. Any idea?

glemaitre:
> I disagree with @glemaitre on the fact that the creation function should also open the UI. these are two different things, we should have a python func to launch UI indeed, but it would be dedicated to this and only this.

@MarieS-WiMLDS Let's move the discussion in this https://github.com/probabl-ai/skore/issues/1002 since it will not impact this PR.

skore/src/skore/cli/cli.py
thomass-dev: @MarieS-WiMLDS i can add a general description to the command.

```
usage: skore [-h] [--version] [--port PORT] [--open-browser | --no-open-browser] [--verbose] project_name

Launch the skore-ui on a defined project. Description well described here =).

positional arguments:
  project_name          the name or path of the project to be created or opened

options:
  -h --help
                        show this help message and exit
  --version
                        show program's version number and exit
  --port PORT
                        the port at which to bind the UI server (default: 22140)
  --open-browser --no-open-browser
                        whether to automatically open a browser tab showing the web UI (default: True)
  --verbose
                        increase logging verbosity
```

What would you like to read?

thomass-dev: Done in https://github.com/probabl-ai/skore/pull/1194/commits/8cbbce819913db763c5bf5fc17d897f5159158ed.

glemaitre: Yep, the parsing is not anymore the same. I'll have a look how to fix it. It is a bit the problem that we need to do manual scrapping with `argparse`.

skore/src/skore/cli/cli.py
```diff
@@ -9,7 +9,10 @@
 
 def argumentparser():
     """Argument parser for the Skore CLI."""
-    parser = ColorArgumentParser(prog="skore")
+    parser = ColorArgumentParser(
+        prog="skore-ui",
```
glemaitre, line 13: OK so you found the bug :). I was about to tell you :)

skore/src/skore/project/project.py
MarieS-WiMLDS: it's true that it's unclear what should happen with the param `load` if the project doesn't exist. But i'm really puzzled by `exist_ok`. I understand that it will not raise an error if it exists, but what will happen? create a duplicate? overwrite? load?  
therefore, new suggestions: `load_existing=bool` or `load_if_exists=bool` (I prefer the first because there is less _). What do you think?

skore/src/skore/cli/cli.py
```diff
@@ -39,87 +41,24 @@ def cli(args: list[str]):
...
-            verbose=parsed_args.verbose,
-        )
-    else:
-        parser.print_help()
+def cli(args: list[str]):
+    """CLI for Skore."""
+    parser = argumentparser()
+    arguments = parser.parse_args(args)
+
+    launch(
```
glemaitre, line 59: Here, it would be better to have an if/else, and if it is not `launch` then we show the help as before.

skore/src/skore/cli/cli.py
MarieS-WiMLDS: good idea, I feel that's clearer now, thanks!

skore/src/skore/project/project.py
MarieS-WiMLDS: ok!

skore/src/skore/project/project.py
MarieS-WiMLDS: ok! I wrote it in the main issue so we don't forget.

skore/src/skore/cli/cli.py
glemaitre: We can delay but I would like to write some test for this class because we did not write any.

skore/src/skore/project/project.py
```diff
@@ -2,22 +2,19 @@
...
-
-logger = logging.getLogger(__name__)
-logger.addHandler(logging.NullHandler())  # Default to no output
-logger.setLevel(logging.INFO)
+logger = getLogger(__name__)
+logger.addHandler(NullHandler())  # Default to no output
+logger.setLevel(INFO)
 
 
 class Project:
```
glemaitre, line 20: Since we touch it and that now `Project` instantiation is intended to for user, we need to document it better in terms of convention:

```python
class Project:
    ...
    Parameters
    ----------
    path : str or Path, default="project.skore"
        ...

    Attributes
    ----------
    items_repository : ...
```

skore/src/skore/project/project.py
glemaitre: I agree with @MarieS-WiMLDS that we can choose a better name even if your choice is meaningful for Python people.
I like `load_if_exists`, a bit longish but hopefully we use it almost always with it's default.

skore/src/skore/cli/cli.py
thomass-dev: It works well, thanks!

skore/src/skore/project/project.py
glemaitre: So if I'm not missing something, we don't really have anymore the "overwrite" feature?

skore/src/skore/project/project.py
thomass-dev: Fixed. As we don't want to publicly expose the repositories, i shadowed the attributes.

skore/src/skore/ui/app.py
```diff
@@ -20,11 +20,6 @@ def create_app(
 ) -> FastAPI:
```
glemaitre, line 20: So you should update the type above because project is not optional anymore (from what @rouk1 said in my PR :))

skore/src/skore/project/project.py
thomass-dev: Fixed in ffcf3fd059602ad7ce497f9beba640937b7ef92a.

I have replace the parameter `exist_ok=<True|False>` by `if_exists=<"raise"|"load">`.

skore/src/skore/ui/app.py
thomass-dev: Indeed, fixed.

MarieS-WiMLDS (approved):
image

Great simplification :) !!

rouk1:
I know I'm late to the show but `if_exists` default value to `raise` is very awkward to me. Indeed how do one rerun an cell containing a `Project()` statement if you don't now that kwarg ? (note that its not discoverable by IDE probably cause we do not have stubs I suppose).

To me reloading an existing project is something more common than creating one.

glemaitre:
> Indeed how do one rerun an cell containing a Project() statement if you don't now that kwarg ? 

For sure, you error message should mention this kwargs to make you do the right choice (I did not read the updated code however).

rouk1:
Another issue that I have with this PR is that as the `project` parameter of the `create_app` factory is no longer optional the server app is no longer easily debuggable. 

i.e. when you run it from the command line you can not pass a path hence it crashes when run.

```bash
uvicorn  skore.ui.app:create_app --factory --reload --reload-dir skore/src --host 0.0.0.0 --port 22140 --timeout-graceful-shutdown 0 --log-level info
INFO:     Will watch for changes in these directories: ['/Users/rouk1/dev/skore/skore/src']
INFO:     Uvicorn running on http://0.0.0.0:22140 (Press CTRL+C to quit)
INFO:     Started reloader process [19757] using StatReload
ERROR:    Error loading ASGI app factory: create_app() missing 1 required positional argument: 'project'
```

This how you get to debug the FastAPI app in vscode.

```jsonc
{
  // launch.json
  "version": "0.2.0",
  "configurations": [
    {
      "name": "Run webapp",
      "type": "debugpy",
      "request": "launch",
      "module": "uvicorn",
      "args": [
        "skore.ui.app:create_app",
        "--factory",
        "--reload",
        "--reload-dir",
        "skore/src",
        "--host",
        "0.0.0.0",
        "--port",
        "22140",
        "--timeout-graceful-shutdown",
        "0",
        "--log-level",
        "info"
      ],
      "console": "integratedTerminal",
      "jinja": true,
      "env": {
        "DEBUG": "1",
      }
    }
  ]
}
```

thomass-dev:
Thanks @rouk1 for your review, i will revert the `create_app` changes.

PR #1212, opened 2025-01-23
docs: Minor changes of the doc examples for v0.6

sylvaincom:
Some minor changes

### DIFF
--- a/examples/use_cases/plot_employee_salaries.py
+++ b/examples/use_cases/plot_employee_salaries.py
@@ -32,7 +32,7 @@ project = skore.open("my_project", create=True)
 
 # %%
 #
-# We use a `skrub` dataset that is non-trivial.
+# We use a skrub dataset that is non-trivial.
 from skrub.datasets import fetch_employee_salaries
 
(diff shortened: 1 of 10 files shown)


examples/use_cases/plot_employee_salaries.py
```diff
@@ -170,10 +170,10 @@ def periodic_spline_transformer(period, n_splines=None, degree=3):
 
 # %%
 # We observe that the cross-validation report detected that we have a regression task
-# and provides us only a subset of the metrics and plots that make sense for our
+# and provides us with (a subset of) metrics and plots that make sense for our
```
MarieS-WiMLDS, line 173: why the parenthesis here?

sylvaincom: I'd rather completely delete this notion of subset actually

sylvaincom:
```suggestion
# and provides us with some metrics and plots that make sense for our
```

MarieS-WiMLDS (approved):

Issue #1008, opened 2024-12-23
docs: Modify the doc following the v0.5 release, part 2


sylvaincom:
Following #1007

Taking into account comments from @tuscland, @MarieS-WiMLDS and @sylvaincom:
- overall
    - [ ] try to unify the colors on those of skore: hover, links, etc
    - [x] do not use temporary files --> https://github.com/probabl-ai/skore/pull/1213
- landing page
    - [x] remove "Contents" at the bottom --> https://github.com/probabl-ai/skore/pull/1191
- user guide
    - [x] remove the skore logo as examples' preview  --> https://github.com/probabl-ai/skore/pull/1219
- API
    - [x] `CrossValidationItem` has an ugly signature
    - [x] `CrossValidationReporter`: the description of its parameters, we say "See sklearn blabla" each time, which is too redundant. --> canceled because it will be removed in the next release.

MarieS-WiMLDS:
isn't this issue too big?  
contributing could be an issue per se, and about the API, the CVItem signature might be fixed by item visibility switched off + CVReporter might be if we change the signature as asked in this issue https://github.com/probabl-ai/skore/issues/953.

MarieS-WiMLDS:
I edited the text of the issue to remove everything related to contributing guide, to transform this epic to a regular issue.

sylvaincom:
Ok thanks, have you created a separate issue for the stuff about the contributing guide?

MarieS-WiMLDS:
it's already listed in #882

sylvaincom:
Perfect

PR #1213, opened 2025-01-23
docs: Isolate examples

### LINKED ISSUE #1200: docs: improve examples harmony and isolation

the examples. 

Our constraints are: 
- no more tempfile (or at least hidden in a dropdown), to avoid giving the impression that's it represent the complexity of the library
- the same variables names are used everywhere (check the reason [here](https://github.com/probabl-ai/skore/pull/1193#discussion_r1925013596))
- the examples can run in parallel (because otherwise it takes quite some time)
- the examples are run (to avoid having to check them by hand that any api change affect them)

### Suggested improvement

What we tried so far:  
- using `cd`, but it's not allowed
- dropdown to hide the code, but in this case it's not executed

### Additional context

_No response_

MarieS-WiMLDS:
fixes #1200  
participates to https://github.com/probabl-ai/skore/issues/1008


auguste-probabl (commented):
Really cool, glad to see that `chdir` works in CI. Have you tested it locally too (with `make html`)? I recall there were problems with `chdir` then...

examples/getting_started/plot_quick_start.py
```diff
@@ -12,13 +12,20 @@
...
 
+# sphinx_gallery_start_ignore
+import os
+import tempfile
+from pathlib import Path
+
+temp_dir = tempfile.TemporaryDirectory()
+temp_dir_path = Path(temp_dir.name)
+os.chdir(temp_dir_path)
+# sphinx_gallery_end_ignore
```
thomass-dev, line 23: Is this script intend to be used only sphinx? Not by a user locally?

auguste-probabl: A user might copy-paste it from the docs directly to get started -- although I expect it to be a bit rare

MarieS-WiMLDS:
>  Really cool, glad to see that chdir works in CI. Have you tested it locally too (with make html)? I recall there were problems with chdir then...

Yes, it's working :) !

examples/getting_started/plot_quick_start.py
MarieS-WiMLDS: like Auguste says, plus it's also to remove all unnecessary content for the user in the docs, and focusing on the valuable parts. The tempfiles are only there to the examples can be run in parallel and that there is no interaction between them, so it's purely internal.

MarieS-WiMLDS:
I think I'll need help to tackle the sphinx error: it worked on friday, but now not anymore, while the merge didn't even affect this bugging example... :thinking:   
image

auguste-probabl (approved):
Nice work!

glemaitre:
I see that we removed the `clear` function. I was wondering if it would have not been a convenience function to have?

MarieS-WiMLDS:
It was not useful anymore, so I removed it to avoid dead code. All the more that it's somehow redundant with the overwrite option, so I really don't want to keep it if we don't have a clear use case for `clear`.

glemaitre:
I would have personally two distinct expectations:

`open` intend to open a project and with a couple of options: creating the project, overwriting it, launching the UI.

`clear` is more about cleaning my messy project in an interactive mode without the expectation to reopen the project.

In an interactive setting (like in notebook), it seems weird to do:

```python
skore.open("my_project", serve=True)

skore.put(...)
```

and if I want just to clear everything (for instance to rexecute a subset of `put` method), it is weird to me to write, because it is not what I intend to (even though the result is the same).

```python
skore.open("my_project", overwrite=True, serve=False)
```

In case, that we tell me that I should either restart my kernel (optionally) or execute the first `open` cell, then it is annoying because I just not have `overwrite=True` in most of the cases because I can just loose my project. Then, the `serve=True/False` will also annoyed me because I need to change this `open` line before each execution depending of what I'm doing.

So until, it is tested, I would argue it is not dead code and since we show cases in which we never make mistakes, it is right that we have almost no usage of `clear` in our examples (the same way that we don't need `clear_cache`).

thomass-dev:
Please take a look at https://github.com/probabl-ai/skore/pull/1194 which simplify API, by removing `open` in favor of the project constructor. In this PR, i reintroduced the clear function.

PR #1226, opened 2025-01-24
fix: Allow sending primitive item with `float("nan")` to frontend

thomass-dev:
Allow sending primitive item with `float("nan")` to frontend by using ORJSON as FastAPI JSON serializer.


rouk1 (approved):
orjson ! why not : )

PR #1227, opened 2025-01-24
docs: No need to create a skore project in some examples

### LINKED ISSUE #1221: docs: In some examples, there is no need to create a skore project

`examples/model_evaluation/plot_train_test_split.py`

At the beginning, we create a skore project. And the end, we clear the project. But this example does not require us to store any items.

### Suggested improvement

Check if there are other examples like this.

### Additional context

_No response_

sylvaincom:
Fix #1221 

and also minor change in the quick start guide


glemaitre (approved):
LGTM. Thanks @sylvaincom

PR #1234, opened 2025-01-27
feat(UI): Add markdown cells in views

rouk1:
...


auguste-probabl:
I think this can be closed

PR #1242, opened 2025-01-28
fix: Persist base64 string instead of bytes to ensure items are JSONable

thomass-dev:
This PR intends to persist b64 string inplace of bytes.
The latest isn't JSONable, so items can't be easily persisted without a pickle backend.


auguste-probabl (approved):
Looks great, thanks

Issue #1159, opened 2025-01-20
feat: Change the way NumPy array are persisted


thomass-dev:
Originally planned in https://github.com/probabl-ai/skore/issues/1045 but postponed to another PR because it's not a blocking point : 

> Is it planned to use `np.save` and use the NumPy format to dump NumPy objects? It will be safer when reloading it.

_Originally posted by @glemaitre in https://github.com/probabl-ai/skore/pull/1052#discussion_r1920114598_

PR #1250, opened 2025-01-29
feat(NumpyArrayItem): Use `numpy.save` instead of `array.tolist()`

auguste-probabl:
Only for storage for now; `array.tolist()` is still used for sending data to the UI.

The `allow_pickle=False` setting is used to keep the same behaviour for arrays.

Closes https://github.com/probabl-ai/skore/issues/1159


thomass-dev (changes requested):
Well done! Just a docstring issue.

thomass-dev (approved):

Issue #1251, opened 2025-01-29
enh: hide progress bar once the execution is done


glemaitre:
Right now, whenever we launch some processing that trigger the progress bars, we keep them at 100% once the execution done. I think that they take vertical space in the output of the cell and are not really useful.

It should be sufficient to pass `transient=True` to the `Progress` class:

https://github.com/probabl-ai/skore/blob/66ca391158bae3d4edca561f8f93c95451a515a2/skore/src/skore/utils/_progress_bar.py#L36-L46

glemaitre:
@MarieS-WiMLDS What do you think?

ictorv:
> [@MarieS-WiMLDS](https://github.com/MarieS-WiMLDS) What do you think?

Yes I think by adding transient=True, it would hide progress bars,

 progress = Progress( 
     SpinnerColumn(), 
     TextColumn("[bold cyan]{task.description}"), 
     BarColumn( 
         complete_style="dark_orange", 
         finished_style="dark_orange", 
         pulse_style="orange1", 
     ), 
     TextColumn("[orange1]{task.percentage:>3.0f}%"), 
     expand=False, 
     transient=True
 ) 
Can I add this feature?

glemaitre:
@ictorv Yep feel free to take this issue.

glemaitre:
And you can tag my GitHub handle when submitting the PR such that I can review it.

PR #1253, opened 2025-01-29
 Fixed Progress bar

harshitnagar22:
Fixed the Progress Bar so it will  hide progress bar once the execution is done and  fixed Doctest.  

Resolved issue #1251

### DIFF
--- a/skore/src/skore/utils/Fixed Progress bar
+++ b/skore/src/skore/utils/Fixed Progress bar
@@ -1,5 +1,5 @@
+import sys
 from functools import wraps
-
 from rich.progress import (
     BarColumn,
     Progress,
@@ -43,6 +43,7 @@ def progress_decorator(description):
                     ),
                     TextColumn("[orange1]{task.percentage:>3.0f}%"),
                     expand=False,
+                    transient="pytest" not in sys.modules,  # Disable transient during tests
                 )
                 progress.start()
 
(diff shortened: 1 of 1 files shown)


skore/src/skore/utils/Fixed Progress bar
auguste-probabl, line 1: Thanks for the PR! I don't think renaming the file is necessary, do you mind changing it back?

auguste-probabl (changes requested):

skore/src/skore/utils/Fixed Progress bar
```diff
@@ -43,6 +43,7 @@ def wrapper(*args, **kwargs):
                     ),
                     TextColumn("[orange1]{task.percentage:>3.0f}%"),
                     expand=False,
+                    transient="pytest" not in sys.modules,  # Disable transient during tests
```
thomass-dev, line 46: I don't think this is necessary. We should instead fix all falling doctest.

Related PR https://github.com/probabl-ai/skore/pull/1254

skore/src/skore/utils/Fixed Progress bar
harshitnagar22: Hi,

Thank you for the review and your suggestion! I will revert the file name change as per your recommendation. Let me know if there's anything else you'd like me to adjust.

Best regards,
Harshit Nagar

skore/src/skore/utils/Fixed Progress bar
harshitnagar22: Hi,

Thank you for your input! I understand your point. Instead of reverting the changes, I agree that fixing the failing doctests is a better approach.

Thanks for pointing this out, and I will proceed accordingly.

Best regards,
Harshit Nagar

glemaitre:
Sorry I'll be closing this PR because @ictorv already claimed it and we already started to review the associated PR. We will soonish add more items that can be tackled in the issue tracker that might be easily done.

PR #1267, opened 2025-01-30
chore(UI): Remove view builder

### LINKED ISSUE #1266: Remove the views section from the UI
As discussed this feature no longer makes perfect sense. Hence it should be removed.

- [x] remove `skore-ui/src/views/project/ProjectView.vue` and all of its dependencies
- [x] remove view related endpoint in `skore/src/skore/ui/project_routes.py`
- [x] remove `skore/src/skore/persistence/repository/view_repository.py`
- [x] remove app navigation vertical bar
- [x] add a page header that displays the logo, the path of the current project and the theme switch
- [x] move note on items to the activity feed
- [x] add the current project name to the header

rouk1:
Closes #1266.

This PR removes all view builder related code.
Activity feed item now have a version number in backend aswell as in frontend.

User can now annotate there items in the activity feed.

## UI preview:

https://github.com/user-attachments/assets/7f283b74-17a4-4f4b-a6cc-a043a2415b24


MarieS-WiMLDS (changes requested):
I tried with the window covering half of my screen, and it's great!! I really like that there is little margin, in a context where I need all the space to be useful. I also like where the annotation pencil is now.    
image  

I noticed these three points, none of them is a big deal.

## point 1

image  

when hovering the project name, it gives me the location. It's nice so i'm not lost, but i'd like to have it complete. Is it possible to display it on several lines? or on the right? 

## point 2
I don't think that it's something introduced in the PR, but if it's not too difficult, when the item contains text, I'd like to have it in a grey rectangle like the other objects, otherwise it get mixed with the notes (cf below).

## point 3
I like the idea to have a `#n` after the object name to display that there is another one! But it's very small, i had a hard time reading.

image


image

rouk1:
## Point 1

Tooltip will now never be larger than 50% of the viewport and placed at the bottom start of the related element.

Screenshot 2025-01-31 at 18 23 24

## Point 2

Nice catch !
This would apply to all "markdown" items. So I'd like to propose something else, I've added a left border to the notes to use the same pattern as citation you can find pretty much everywhere. wdyt ? In the mean time @anasstarfaoui may have some better ideas : )

Screenshot 2025-01-31 at 18 30 37

## Point 3

Ok it's now the same size as the name of the item.
Screenshot 2025-01-31 at 18 36 25

MarieS-WiMLDS:
it's good for me from a functional point of view, thanks :) !  
I'll let a dev give his opinion for the technical aspects.

auguste-probabl (approved):
At a glance this looks fine to me

PR #1268, opened 2025-01-30
WIP: modelcard first demo

koaning:
This is just me helping out @glemaitre by independantly having a quick stab at a model card. First draft, but looks alright. 

This code:

```python
render_model_card(
    report,
    name="Demo classifier", 
    description="This is really just a demo of a model card. The model is really silly.",
    intended_use="The goal of this pipeline is to show a bunch of properties that can be displayed in a model card.",
    limitations="There is no reason to ever use this system. Ever."
)
```

Can generate this: 

CleanShot 2025-01-30 at 16 48 30

There is an opportunity to also attach training logs, but we should think a bit about how if we want to go there.

### DIFF
--- a/demo.ipynb
+++ b/demo.ipynb
@@ -1,0 +1,69 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "8bbb5efa-b7dc-4999-a178-e21495634f3a",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from sklearn.datasets import make_classification \n",
+    "from sklearn.pipeline import make_pipeline\n",
+    "from sklearn.linear_model import LogisticRegression\n",
+    "from sklearn.preprocessing import StandardScaler\n",
+    "from skore import CrossValidationReport\n",
+    "\n",
+    "pipe = make_pipeline(StandardScaler(), LogisticRegression())\n",
+    "X, y = make_classification(n_samples=16000, n_features=10)\n",
+    "report = CrossValidationReport(pipe, X, y)"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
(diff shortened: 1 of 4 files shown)


adrinjalali:
This is a nice start, and the general API is certainly okay. However, for implementation, there things we need to consider.

Two points to start with:

- Model cards are artifacts which will be modified outside our "platform". Therefore it's a common usecase where we'd need to "load" the edited model card for further modifications.
- Markdown is the format which people find easiest to edit, and therefore most model cards out there are in markdown format.

`skops.card` model cards allow you to do all of that, and the templates are somewhat jinja templates. We use `pandoc` to load stored cards into a structured form in the memory, to modify them. I'm not 100% set on the API there, but the basics are pretty nice and we've worked quite a bit (including with Meg Mitchel, the original model card co-author) to get it to that point.

I don't mind adding methods so that we can get an HTML rendered version of the card to show inside a notebook, but usually the markdown is the only source of truth, and platforms render it nicely (GH, HuggingFace, etc), and there's not much of a need to render it inside a notebook. Getting a pdf version of it is more usual that getting the html version.

I agree with @glemaitre that it probably would make sense to have a layer of abstraction in `skore` to make things easier for the user, yet add a bunch of functionalities we need to `skops.card` itself.

adrinjalali:
I don't think Vincent is gonna be working on this anymore 😉

koaning:
Yeah, this work as kind of meant as a "get the ball rolling" in a direction, not so much something that I can pick up again now.

MarieSacksick:
Let's close this PR - it gives ideas of elements to include in model cards, and we are working on it in other PRs.

adrinjalali:
@thomass-dev would be nice if you could include an "automated message" text in automated posts if they're not posted by a "bot" account in case it happens again in the future 😉

PR #1281, opened 2025-02-03
feat(UI): report widget

rouk1:


skore-ui/src/assets/fixtures/estimator-report.json
```diff
@@ -0,0 +1,113 @@
+{
+  "scalar_results": [
+    { "name": "toto", "value": 4.32, "favorability": "greater_is_better" },
```
thomass-dev, line 3: Toto is always greater than other <3

PR #1282, opened 2025-02-03
chore: Serialization of items is now only in UI module

rouk1:
As a preparation work to serialize reporters and to limit the footprint of UI to only it's module, this PR moves every serialization related code to the UI module.


thomass-dev (approved):
Nice, it's much clearer, thanks!

PR #1283, opened 2025-02-03
fix(api): Return most native type for metrics in EstimatorReport

### LINKED ISSUE #1275: api: return Python native type when calling metrics function
As discussed in https://github.com/probabl-ai/skore/issues/1259, we should be returning the most appropriate Python type (float or dict) when calling a function delegating the concatenation possibility to the `.report_metrics` methods.

glemaitre:
closes #1275 

Make metric methods from `EstimatorReport` return the most native type possible: `float`, `dict` or numpy array.

### TODO:

- [x] Adapt the `CrossValidationReport`


glemaitre:
It is ready for review @thomass-dev @auguste-probabl @sylvaincom 

I see that we should most probably drop the `plot_estimator_report.py` soonish since we split it in use case.

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -736,15 +735,18 @@ def custom_metric(
...
               Split #1  52.6...
         """
-        return self._compute_metric_scores(
-            report_metric_name="custom_metric",
-            data_source=data_source,
-            aggregate=aggregate,
-            metric_function=metric_function,
+        scorer = make_scorer(
+            metric_function,
+            greater_is_better=True,
```
auguste-probabl, line 740: Are you doing this because you expect all custom metrics to be "greater-is-better"? If so this should be documented. Note that in the docstring example the MAE is "lower-is-better".

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -325,41 +387,20 @@ def _compute_metric_scores(
...
-                    names=["Metric", "Output"],
-                )
-                score = score.reshape(1, -1)
-        else:  # unknown task - try our best
-            columns = [metric_name] if score.shape[0] == 1 else None
-        return pd.DataFrame(
-            score, columns=columns, index=[self._parent.estimator_name_]
-        )
+        ) and isinstance(score, np.ndarray):
+            return dict(zip(self._parent._estimator.classes_, score))
```
auguste-probabl, line 400: nit: Does `zip` preserve order? It would be nice if it did.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1280,10 +1251,8 @@ def custom_metric(
         >>> report.metrics.custom_metric(
         ...     metric_function=mean_absolute_error,
         ...     response_method="predict",
-        ...     metric_name="MAE (↗︎)",
         ... )
-        Metric        MAE (↗︎)
-        Ridge   44.9...
+        44.9...
```
auguste-probabl, line 1255: Could you add examples where the output is a dict/ndarray?

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -960,11 +986,10 @@ def predict(self, X):
     report = EstimatorReport(estimator, fit=False, X_test=X_test, y_test=y_test)
     result = report.metrics.custom_metric(
         metric_function=lambda y_true, y_pred: 1,
-        metric_name="Custom Metric",
         response_method="predict",
     )
-    assert result.columns.tolist() == ["Custom Metric"]
-    assert result.to_numpy()[0, 0] == 1
+    assert isinstance(result, Real)
```
auguste-probabl, line 991: Why not `float`?

auguste-probabl (commented):
Great, just a few questions/comments

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
glemaitre: Tricky one here. The reason for `greater_is_better=True` even for losses is that I don't want the `scorer` to automatically `*-1` internally. I want solely call the function provided by the user and it without any scikit-learn convention on the top.

It is worth a comment for sure.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: Since I have arrays and list, it will preserve the order. `dict` in the passed where not necessarily preserving it but it is old Python version (and the need for `OrderedDict`).

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: Yep I can add an extra example.

skore/tests/unit/sklearn/test_estimator.py
glemaitre: The test returns an integer so `Real` account for both `int` and `float`. It is a bit the intended spirit when reading the test (the dtype does not matter).

auguste-probabl (approved):

PR #1239, opened 2025-01-27
docs(future feat): Example driven development

MarieS-WiMLDS:

PR #1286, opened 2025-02-04
feat: Add `ComparisonReport` to compare instances of `EstimatorReport`

### LINKED ISSUE #1245: feat: `ComparisonReport`

As a user, once I have created several EstimatorReport or CrossValidationReport, I want to compare them. 

We want to be able to compare the metrics inside an estimator report.

To see an example of how it would be used: https://github.com/probabl-ai/skore/pull/1239

We could want to have a comparison system that can compare anything: compare ints with each others, compare plots, compare data... But it raises plenty of technical and UX questions. While a person using skore, all the more given the number of features we have now, use it mainly for its reports. Therefore, let's focus on comparing reports for now. It makes more sense with current product environment, and also we are more in control as we know the datasets, the models... 

### Additional context

The goal is to have a v0. Whenever it can be simplified, simplify. We will iterate.   
Only one issue doesn't mean only one epic.

thomass-dev:
- [x] Rename to `ComparisonReport`
- [x] Rebase on top of #1239 and adapt
- [x] Raise if `report.metrics.accuracy(data_source="train")` is called with at least one EstimatorReport that does not have training data
- [x] Test
- [x] Docstrings
    - [x] MetricsAccessor
- [x] Move index column "#0" in front of each metric
- [x] Pass report names in comparator
- [ ]  ~Update plots legend~ see #1309 
    - The actual `RocCurveDisplay` needs a full refactor to be splitted by use-case: estimator report, cross-validation report and finally comparison report. In each of these use-cases, there is two scenarios with binary classification and multi-class classification. Otherwise, it will be unmaintainable.
- [ ] ~Investigate missing metrics in `report_metrics`~ **(deferred to future PR)**
    - The logic is split between `report_metrics` and `available_if`; it should be merged (ideally everything in `available_if`?)
- [ ] ~Refactor to make `CrossValidationReport` depend on it~ **(deferred to future PR)**
- [x] ~Change EstimatorReport `repr`?~ Issue https://github.com/probabl-ai/skore/issues/1293

Closes #1245 

Co-authored-by: Auguste <auguste@probabl.ai>
Co-authored-by: Sylvain <sylvain@probabl.ai>


skore/src/skore/sklearn/comparator/comparator.py
```diff
@@ -0,0 +1,113 @@
...
+class Comparator(_BaseReport, DirNamesMixin):
+    """"""
+
+    _ACCESSOR_CONFIG = {
+        "metrics": {"name": "metrics"},
+    }
+
+    def __init__(
+        self,
+        estimator_reports: list[EstimatorReport],
```
glemaitre, line 28: Quick thought: the comparator should eat some `EstimatorReport` but also some `CrossValidationReport`. In terms of naming, we might maybe be a bit more generic.

auguste-probabl: This is out of scope for now, but we're thinking about it

glemaitre: To be clear I don't request to support cross validation report but rather think slightly ahead a not name a variable that we will need to change in the next iteration. 

Since it is specifically about code I don't see the point to discuss it in the issue.

thomass-dev:
> Could you elaborate on why it makes sense for the cross validation report to reuse the comparator. Is it about inheritance or about some internal routines?

The goal is to see if some code can be factorized between comparison and cross-validation reports.

skore/src/skore/sklearn/comparator/comparator.py
auguste-probabl: The argument will be renamed to `reports`.

sylvaincom (commented):
I believe the `ComparisonReport` does not appear in the API?

auguste-probabl:
> I believe the `ComparisonReport` does not appear in the API?

Oh sorry, we forgot! On it

examples/model_evaluation/plot_comparison_report.py
```diff
@@ -0,0 +1,234 @@
+"""
```
glemaitre, line 1: I see that we are going to potentially increase the number of examples pretty quick (it is already a huge issue in scikit-learn).

I think that it would be best to think of incorporating the usage of the comparator in existing example. I think that we could add a section at the following one: https://skore.probabl.ai/dev/auto_examples/use_cases/plot_employee_salaries.html

The idea would to train the linear & gbdt model and make the comparison using the comparator in the section.

Later when the comparator supports cross-validation report, then we can remove the sections and directly compare the results from the cross-validation.

examples/model_evaluation/plot_comparison_report.py
```diff
@@ -0,0 +1,234 @@
...
+
+# %%
+# Metrics
+# ^^^^^^^
+
+# %%
+# Let us display the result of our benchmark:
+
+# %%
+df_report_metrics = comp.metrics.report_metrics()
```
glemaitre, line 96: Looking at the rendering and at a first glance I find the `level_0` integral index weird. I need to think more about it.

examples/model_evaluation/plot_comparison_report.py
```diff
@@ -0,0 +1,234 @@
...
+import matplotlib.pyplot as plt
+
+display = comp.metrics.plot.roc()
+plt.tight_layout()
+
+# %%
+# .. note::
+#
+#   In a near future version of skore, the legends will be correct with the names of
+#   the estimators.
```
glemaitre, line 143: I'll not merge until it is fixed ;P

examples/model_evaluation/plot_comparison_report.py
```diff
@@ -0,0 +1,234 @@
...
+    y_train=y2_train,
+    X_test=X2_test,
+    y_test=y2_test,
+)
+
+# %%
+# Now, let us try to compare these estimator reports:
+
+# %%
+try:
```
glemaitre, line 199: I would say that there is not need to show that :)

examples/model_evaluation/plot_comparison_report.py
```diff
@@ -0,0 +1,234 @@
...
+# %%
+# As desired, skore raises an error.
+
+# %%
+# Comparing 2 estimator reports that do not hold their data
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Methodologically, we always want to compare estimator reports on the exact same
+# dataset. If we do not know where the datasets come from, then we expect skore to
+# throw a warning:
```
glemaitre, line 219: For me this delay should be delayed at computation of the metric. We are able to have a `hash` on the data or to know that they differ and thus warn properly.

Otherwise we will be raising too many potential true positive.

glemaitre (commented):
Some thinking once I check the example

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -105,13 +105,30 @@ def __init__(
...
+                raise ValueError("Cannot compare reports without testing data")
+
+            ml_tasks.add(report._ml_task)
+            test_dataset_hashes.add(joblib.hash((report.X_test, report.y_test)))
+
+            if len(ml_tasks) > 1:
+                raise ValueError("Not all estimators are in the same ML usecase")
+
+            if len(test_dataset_hashes) > 1:
+                raise ValueError("Not all estimators have the same testing data")
```
auguste-probabl, line 131:
```suggestion
        if len(ml_tasks) > 1:
            raise ValueError("Not all estimators are in the same ML usecase")

        if len(test_dataset_hashes) > 1:
            raise ValueError("Not all estimators have the same testing data")
```

auguste-probabl: Okay, I thought it was a typo, feel free to resolve

glemaitre:
Working on a review.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1286 @@
...
+                for report in self._parent.estimator_reports_
+            )
+            results = []
+            for result in generator:
+                results.append(result)
+                progress.update(main_task, advance=1, refresh=True)
+
+            results = pd.concat(results, axis=0, ignore_index=True)
+            results.index = pd.MultiIndex.from_tuples(
+                enumerate(self._parent.report_names_),
```
glemaitre, line 192: Let's start simple and avoid to add the enumerate since the user can override the `report_names`.

glemaitre (commented):
Here, is a partial review. I'll continue after the demo

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1286 @@
...
+
+    def report_metrics(
+        self,
+        *,
+        data_source="test",
+        scoring=None,
+        scoring_names=None,
+        pos_label=None,
+        scoring_kwargs=None,
+        aggregate=None,
```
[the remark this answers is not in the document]
auguste-probabl: Yeah we discussed this and we thought it might still be interesting, e.g. if you want to compare a model with the mean or something like this.

skore/tests/unit/sklearn/test_comparison.py
```diff
@@ -0,0 +1,327 @@
...
+
+    with pytest.raises(
+        TypeError, match="Only instances of EstimatorReport are allowed"
+    ):
+        ComparisonReport([estimator_report, None])
+
+    with pytest.raises(
+        TypeError, match="Only instances of EstimatorReport are allowed"
+    ):
+        ComparisonReport([None, estimator_report])
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/tests/unit/sklearn/test_comparison.py
```diff
@@ -0,0 +1,327 @@
...
+    estimator, _, X_test, _, y_test = usecase("binary-logistic-regression")
+    estimator_report = EstimatorReport(
+        estimator,
+        fit=False,
+        X_test=X_test,
+        y_test=y_test,
+    )
+
+    with pytest.raises(
+        TypeError, match="object of type 'EstimatorReport' has no len()"
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+    from skore import console
+
+    console.print(
+        Panel(
+            title=title,
+            renderable=message,
+            style="orange1",
+            border_style="cyan",
+        )
+    )
```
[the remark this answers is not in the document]
auguste-probabl: removed

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+        the scores are parallelized.
+        When accessing some methods of the `ComparisonReport`, the `n_jobs`
+        parameter is used to parallelize the computation.
+        ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
+        ``-1`` means using all processors.
+
+    Attributes
+    ----------
+    estimator_reports_ : list of EstimatorReport
+        The estimator reports for each split.
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+            if len(test_dataset_hashes) > 1:
+                raise ValueError("Not all estimators have the same testing data")
+
+        if report_names is None:
+            self.report_names_ = [report.estimator_name_ for report in reports]
+        elif len(report_names) == len(reports):
+            self.report_names_ = report_names
+        else:
+            raise ValueError(
+                "There should be as many report names as there are reports"
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        if len(reports) < 2:
+            raise ValueError("At least 2 instances of EstimatorReport are needed")
+
+        ml_tasks = set()
+        test_dataset_hashes = set()
+
+        for report in reports:
+            if not isinstance(report, EstimatorReport):
+                raise TypeError("Only instances of EstimatorReport are allowed")
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+
+            if (report.X_test is None) or (report.y_test is None):
+                raise ValueError("Cannot compare reports without testing data")
+
+            ml_tasks.add(report._ml_task)
+            test_dataset_hashes.add(joblib.hash((report.X_test, report.y_test)))
+
+            if len(ml_tasks) > 1:
+                raise ValueError("Not all estimators are in the same ML usecase")
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/metrics_accessor.py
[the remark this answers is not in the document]
glemaitre: I would think that it is rarely useful. I would only dispatch when having `CrossValidationReport`.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+    >>> estimator_report_2 = EstimatorReport(
+    ...     estimator_2,
+    ...     X_train=X_train,
+    ...     y_train=y_train,
+    ...     X_test=X_test,
+    ...     y_test=y_test
+    ... )
+    >>> report = ComparisonReport([estimator_report_1, estimator_report_2])
+    ...
+    """
```
glemaitre, line 79: It might be interesting to compute offer a `cache_predictions` method.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+
+import joblib
+import numpy as np
+
+from skore.externals._pandas_accessors import DirNamesMixin
+from skore.sklearn._base import _BaseReport
+from skore.sklearn._estimator.report import EstimatorReport
+
+
+class ComparisonReport(_BaseReport, DirNamesMixin):
```
glemaitre, line 16: From the demo today, I think the feedback was to allow the following.

```python
report = ComparisonReport.from_estimators(
    estimators=[est1, est2], X_train, y_train, X_test, y_test
)
report = ComparisonReport.from_reports(
    reports=[report1, report2], ...
)
```

`from_reports` is actually the current `__init__` while the `from_estimators` is something that should create the `EstimatorReport` and pass it to the `from_reports` afterwards.

skore/tests/unit/sklearn/test_comparison.py
```diff
@@ -0,0 +1,314 @@
...
+def test_estimator_report_metrics_linear_regression(
+    monkeypatch, metric_name, metric_value, regression_model
+):
+    def dummy(*args, **kwargs):
+        return pd.DataFrame([[metric_value]])
+
+    monkeypatch.setattr(
+        f"skore.sklearn._estimator.metrics_accessor._MetricsAccessor.{metric_name}",
+        dummy,
+    )
```
glemaitre, line 295: I would not monkey patch here but instead really test on dummy dataset. Maybe we don't need to check the score exactly but rather what the dataframe look like in terms of columns and indices.

glemaitre:
I would also directly incorporate the feedback from https://github.com/probabl-ai/skore/issues/1307 in terms of API.

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: Could we defer this to a future PR?

examples/model_evaluation/plot_comparison_report.py
thomass-dev: Fixed in https://github.com/probabl-ai/skore/pull/1309

examples/model_evaluation/plot_comparison_report.py
auguste-probabl: done

examples/model_evaluation/plot_comparison_report.py
auguste-probabl: This one needs to be clarified in light of the newer discussions. I can simply remove this part of the docs for now?

skore/tests/unit/sklearn/test_comparison.py
auguste-probabl: done. there doesn't seem to be a way to assert that two dfs had the same shape, so i went a step further and set the expected dfs.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: @thomass-dev brings the argument that by default people will not name their reports -- in which case the simplest is to identify them somehow.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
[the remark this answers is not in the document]
auguste-probabl: removed

skore/src/skore/sklearn/_comparison/report.py
glemaitre: We can potentially differ at first.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: I would think that people will try different models family rather than several hyperparameters.

If we can avoid as well the multi-index in the first iteration, it could also be great.

skore/src/skore/sklearn/_comparison/report.py
glemaitre: If we go with partial feature then yes.

skore/src/skore/sklearn/_comparison/report.py
glemaitre: So let's differ

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: On the contrary, I could easily imagine a user using a ComparisonReport for doing grid search over hyperparameters.

If we replace the multi-index with a single index I would rather it have different labels for each row, something like "LinearRegression #1", etc.

examples/model_evaluation/plot_comparison_report.py
auguste-probabl: Duplicate of the other multi-index discussion https://github.com/probabl-ai/skore/pull/1286#discussion_r1950601230

examples/model_evaluation/plot_comparison_report.py
sylvaincom: You don't want too many examples for computation reasons when deploying the doc? Or because users reading the doc would get lost?

I like the fact that we have one "short" example per feature so that the feature is super clear. As of now we only have 3 so that feels ok?

And, for sure, we can have some more "use case" examples, where one example encompasses several skore features to show skore in action

If we introduce the `ComparisonReport` in `plot_employee_salaries`, as you mentioned, it can not compare the `CrossValidationReport` at hand, so that kind of breaks the narrative of the example. A way around that would be to do the `CrossValidationReport`'s as is, then extract the estimator reports of the first fold of each estimator and compare them.

WDYT?

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: For hyper parameter search, a scikit learn user should use a GridSearchCV object. 

Right now, you can put the entire model in the report but we are not yet able to properly consume and show the results from some analysis and I would not request people to make the search by hand. 

At the time that we have our own or patched GridSearchCV containing a cv_results_ dict rich enough to be used as an input of a ComparisonReport, then I'll agree that we should rethink the matter. It is to early now for this use case.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+    """Report for comparison of :class:`skore.EstimatorReport`.
+
+    Parameters
+    ----------
+    reports : list of :class:`skore.EstimatorReport`s
+        Estimator reports to compare.
+
+    report_names : list of str, default=None
```
glemaitre, line 24: Thinking about this parameter I think that @MarieS-WiMLDS wanted to support a list and a dict but avoid two lists. We might want to stick to this?

MarieS-WiMLDS: I prefer the dict solution by a lot indeed; I hate double lists as far as I'm concerned, they always turn out to be not the same length. (I think it comes from all the pandas or numpy errors saying "blablabla, not the same size").  
It would look like `{"report_name_1": report, "report_name_1": report}`.

examples/model_evaluation/plot_comparison_report.py
sylvaincom: Ok, agreed, I'll look into how to make the `plot_employee_salaries` more natural with the `ComparisonReport`

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: Alright, I'll remove the enumerate

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1286 @@
...
+    ):
+        """Report a set of metrics for the estimators.
+
+        Parameters
+        ----------
+        data_source : {"test", "train"}, default="test"
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,169 @@
...
+
+        ml_tasks = set()
+        test_dataset_hashes = set()
+
+        for report in reports:
+            if not isinstance(report, EstimatorReport):
+                raise TypeError("Only instances of EstimatorReport are allowed")
+
+            if (report.X_test is None) or (report.y_test is None):
+                raise ValueError("Cannot compare reports without testing data")
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done

examples/model_evaluation/plot_comparison_report.py
auguste-probabl: closed as duplicate

skore/src/skore/sklearn/_comparison/metrics_accessor.py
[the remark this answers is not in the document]
auguste-probabl: Also, I don't see why `X_y` mode would not be of interest for CrossValidationReports

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1286 @@
...
+        "precision": "(↗︎)",
+        "recall": "(↗︎)",
+        "brier_score": "(↘︎)",
+        "roc_auc": "(↗︎)",
+        "log_loss": "(↘︎)",
+        "r2": "(↗︎)",
+        "rmse": "(↘︎)",
+        "report_metrics": "",
+        "custom_metric": "",
+    }
```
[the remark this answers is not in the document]
auguste-probabl: done. This was way more work than i would have liked -- i'm looking forward to have some code shared between CrossValidationReport and ComparisonReport

auguste-probabl:
At this point I think all major issues are either solved or we have agreed to defer them to future PRs. I would recommend that we remove the plots while #1309 is ready, and merge this

auguste-probabl:
> * the second one is about the space that appears between the cell and the metrics in a notebook. But it's not a blocker if you don't see easily where it comes from.

If I had to guess I'd say this comes from the progress bar.

glemaitre:
> If I had to guess I'd say this comes from the progress bar.

Indeed. We should check if the issue was reported to `rich`.

examples/getting_started/plot_skore_getting_started.py
```diff
@@ -125,16 +125,65 @@
...
 # .. seealso::
 #
 #   For more information about the motivation and usage of
 #   :class:`skore.CrossValidationReport`, see :ref:`example_use_case_employee_salaries`.
 
+# %%
+# Comparing estimators reports
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Skore has implemented a :func:`skore.ComparisonReport` feature that enables users to
```
glemaitre, line 142: Let's go active voice ;)

```python
:class:`skore.ComparisonReport` enables to compare several estimator reports ...
```

examples/getting_started/plot_skore_getting_started.py
```diff
@@ -125,16 +125,65 @@
...
+
+# %%
+from skore import ComparisonReport
+
+comp_report = ComparisonReport(
+    reports=[
+        log_reg_report,
+        rf_report,
+    ]
+)
```
glemaitre, line 168: Can you gain some vertical space?

```python
comp_report = ComparisonReport(reports=[log_reg_report, rf_report])
```

also I would prefer to "comparison" instead of "comp" or call the variable "comparator" without the "report" name?

examples/getting_started/plot_skore_getting_started.py
```diff
@@ -125,16 +125,65 @@
...
+
+# %%
+comp_report.help()
+
+# %%
+# Let us display the result of our benchmark:
+
+# %%
+df_comp_report_metrics = comp_report.metrics.report_metrics()
+df_comp_report_metrics
```
glemaitre, line 182: Call this variable benchmark_metrics ;)

examples/getting_started/plot_skore_getting_started.py
sylvaincom: Thanks, done

examples/getting_started/plot_skore_getting_started.py
sylvaincom: Thanks, done

examples/getting_started/plot_skore_getting_started.py
```diff
@@ -125,16 +127,60 @@
...
 # .. seealso::
 #
 #   For more information about the motivation and usage of
 #   :class:`skore.CrossValidationReport`, see :ref:`example_use_case_employee_salaries`.
 
+# %%
+# Comparing estimators reports
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# :func:`skore.ComparisonReport` enables users to compare several estimator reports
```
glemaitre, line 144:
```suggestion
# :class:`skore.ComparisonReport` enables users to compare several estimator reports
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1359 @@
...
+        return "[bold cyan]report.metrics[/bold cyan]"
+
+    def __repr__(self):
+        """Return a string representation using rich."""
+        return self._rich_repr(
+            class_name="skore.ComparisonReport.metrics",
+            help_method_name="report.metrics.help()",
+        )
+
+    @progress_decorator(description="Computing predictions for display")
```
glemaitre, line 1088: We need to remove the code below if we don't want the display in the first iteration.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1359 @@
...
+        method_name = method_name.ljust(22)
+        if name in self._SCORE_OR_LOSS_INFO and self._SCORE_OR_LOSS_INFO[name][
+            "icon"
+        ] in ("(↗︎)", "(↘︎)"):
+            if self._SCORE_OR_LOSS_INFO[name]["icon"] == "(↗︎)":
+                method_name += f"[cyan]{self._SCORE_OR_LOSS_INFO[name]['name']}[/cyan]"
+                return method_name.ljust(43)
+            else:  # (↘︎)
+                method_name += (
+                    f"[orange1]{self._SCORE_OR_LOSS_INFO[name]['name']}[/orange1]"
```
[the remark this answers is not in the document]
glemaitre: We should the same non-regression test that I have in the PR that solve this bug in CrossValidationReport.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,168 @@
...
+from skore.sklearn._estimator.report import EstimatorReport
+
+
+class ComparisonReport(_BaseReport, DirNamesMixin):
+    """Report for comparison of :class:`skore.EstimatorReport`.
+
+    Parameters
+    ----------
+    reports : list of :class:`skore.EstimatorReport`s or dict with estimator names as
+    keys and :class:`skore.EstimatorReport`s as values
```
glemaitre, line 22:
```suggestion
    reports : list of :class:`skore.EstimatorReport`s or dict with estimator names as\
            keys and :class:`skore.EstimatorReport`s as values
```

glemaitre: I would like to have a shorter type definition here. I have to check how we can handle that.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,168 @@
...
+        Estimator reports to compare.
```
glemaitre, line 23:
```suggestion
        Estimator reports to compare.
        If `reports` is a list, the name of each estimator is used.
```

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,168 @@
...
+        Number of jobs to run in parallel. Training the estimators and computing
+        the scores are parallelized.
+        When accessing some methods of the `ComparisonReport`, the `n_jobs`
+        parameter is used to parallelize the computation.
+        ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
+        ``-1`` means using all processors.
+
+    Attributes
+    ----------
+    estimator_reports_ : list of EstimatorReport
```
glemaitre, line 37: I don't think that we need to expose this attribute because the user is passing it and we don't modify it.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,168 @@
...
+        The estimator reports for each split.
+
+    report_names_ : list of str
+        The names of the compared estimator reports.
```
glemaitre, line 41: The same here. I don't think that we need it.

skore/src/skore/sklearn/_comparison/report.py
glemaitre: If actually, we create `estimator_reports_` to always be a dict whatever the input, then it makes sense to expose it.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,168 @@
...
+            if (report.X_test is None) or (report.y_test is None):
+                raise ValueError("Cannot compare reports without testing data")
+
+        test_dataset_hashes = {
+            report: joblib.hash((report.X_test, report.y_test)) for report in reports
+        }
+        if len(set(test_dataset_hashes.values())) > 1:
+            raise ValueError(
+                "Expected all estimators to have the same testing data; "
+                f"got {test_dataset_hashes}"
```
glemaitre, line 128: Not sure that the hash will be helpful here. Let's just mention that some of the input dataset is not the same between the estimator. Potentially, we could iterate over all estimators and show the `id(X)` and `id(y)` but it might be an overkill.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,168 @@
...
+                f"Expected all estimators to have the same ML usecase; "
+                f"got {ml_tasks}"
+            )
+
+        if report_names is None:
+            self.report_names_ = [report.estimator_name_ for report in reports]
+        else:
+            self.report_names_ = report_names
+
+        self.estimator_reports_ = deepcopy(reports)
```
glemaitre, line 143: I'm thinking that a copy will also copy the input data. It will start to be expensive in this respect. The `ComparisonReport` will never change any internal of the object apart of calling some of the methods that trigger some computation stored into the cache.

So I would probably not do any copy.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1359 @@
...
+        if metric_kwargs != {}:
+            # we need to enforce the order of the parameter for a specific metric
+            # to make sure that we hit the cache in a consistent way
+            ordered_metric_kwargs = sorted(metric_kwargs.keys())
+
+            for key in ordered_metric_kwargs:
+                if isinstance(metric_kwargs[key], (np.ndarray, list, dict)):
+                    cache_key += (joblib.hash(metric_kwargs[key]),)
+                else:
+                    cache_key += (metric_kwargs[key],)
```
glemaitre, line 164: This is not covered by the test. I assume that we can take the case used in the cross-validation reporter.

skore/tests/unit/sklearn/test_comparison.py
```diff
@@ -0,0 +1,599 @@
...
+    estimator, _, _, _, _ = binary_classification_model
+    estimator_report = EstimatorReport(estimator, fit=False)
+
+    with pytest.raises(ValueError, match="Cannot compare reports without testing data"):
+        ComparisonReport([estimator_report, estimator_report])
+
+
+def test_comparison_report_init_different_ml_usecases(
+    binary_classification_model, regression_model
+):
```
glemaitre, line 89: If you can add a single sentence for all the tests that do not have one, it could be cool.

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done

examples/getting_started/plot_skore_getting_started.py
auguste-probabl: done

glemaitre (commented):
OK here is a round of review.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1359 @@
...
+
+    def _format_method_name(self, name):
+        """Override format method for metrics-specific naming."""
+        method_name = f"{name}(...)"
+        method_name = method_name.ljust(22)
+        if name in self._SCORE_OR_LOSS_INFO and self._SCORE_OR_LOSS_INFO[name][
+            "icon"
+        ] in ("(↗︎)", "(↘︎)"):
+            if self._SCORE_OR_LOSS_INFO[name]["icon"] == "(↗︎)":
+                method_name += f"[cyan]{self._SCORE_OR_LOSS_INFO[name]['name']}[/cyan]"
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/sklearn/_comparison/metrics_accessor.py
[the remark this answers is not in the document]
auguste-probabl: done

skore/tests/unit/sklearn/test_comparison.py
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: Coming back to this because not copying made a test fail. Unless we deepcopy, a user might
- Create an EstimatorReport
- Create a ComparisonReport from that EstimatorReport
- Change something in the EstimatorReport
- Now the ComparisonReport is inconsistent

EstimatorReports are already somewhat "immutable" (e.g. X_train/X_test setters are disabled) but this is Python, so there is no guarantee.

I'll add a notice in the ComparisonReport docs, and remove the test.

auguste-probabl: done

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: done

sylvaincom (commented):
_deleted_

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: my coverage report says it is covered; what makes you think it's not?

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: see https://github.com/probabl-ai/skore/issues/1346

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: It was from the report in the PR but maybe it was outdated :)

skore/src/skore/sklearn/_comparison/report.py
glemaitre: Thanks. I think think that we can go this way and check if really people mess up the instance but I would have the intuition that it could be OK>

glemaitre:
I have this for the current report:

image


I will try locally to see if it is the case.

examples/model_evaluation/plot_comparison_report.py
auguste-probabl: Should be fixed. The new behaviour is
 - estimator reports don't have to contain test data (but if you try to compute metrics with `data_source="test"`, you'll get errors from EstimatorReport logic
 - if estimator reports do have test data, they need to have the same (reports without test data are not counted)

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,162 @@
...
+
+    Parameters
+    ----------
+    reports : list of :class:`~skore.EstimatorReport` instances or dict
+        Estimator reports to compare.
+
+        * If `reports` is a list, the class name of each estimator is used.
+        * If `reports` is a dict, it is expected to have estimator names as keys \
+        and :class:`~skore.EstimatorReport` instances as values. \
+        If the keys are not strings, they will be converted to strings.
```
glemaitre, line 32:
```suggestion
        * If `reports` is a dict, it is expected to have estimator names as keys
          and :class:`~skore.EstimatorReport` instances as values.
          If the keys are not strings, they will be converted to strings.
```

No need for the `\`, you only need to align the next line under the text.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,162 @@
...
+                f"Expected all estimators to have the same ML usecase; "
+                f"got {ml_tasks}"
+            )
+
+        if report_names is None:
+            self.report_names_ = [report.estimator_name_ for report in reports]
+        else:
+            self.report_names_ = report_names
+
+        self.estimator_reports_ = reports
```
glemaitre, line 137: we still have two 2 attributes and there are not documented. I would probably only store one that is `self.reports_` that is a `dict` containing the names and the report.

glemaitre: then when we need only the name we can always get it with `self.reports_.keys()`

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -0,0 +1,162 @@
...
+        report_names = (
+            list(map(str, reports.keys())) if isinstance(reports, dict) else None
+        )
+        reports = list(reports.values()) if isinstance(reports, dict) else reports
+
+        if not all(isinstance(report, EstimatorReport) for report in reports):
+            raise TypeError("Expected instances of EstimatorReport")
+
+        test_dataset_hashes = {
+            report: joblib.hash((report.X_test, report.y_test))
```
glemaitre, line 118: Maybe we don't need to store the `report` and only store the hash since we don't use it.

sphinx/reference/report/index.rst
```diff
@@ -42,6 +42,17 @@ scikit-learn estimators by cross-validation, and reporting the results.
 
    cross_validation_report
 
+Comparison Report
+-----------------------
+
+:class:`skore.ComparisonReport` provides comprehensive capabilities for comparing :class:`skore.EstimatorReport` instances, and reporting the results.
```
glemaitre, line 48: If you can also make it 88 chars max, it is nice on a small screen ;) (even if it is not python code).

glemaitre (commented):
Looks good. I don't recall if I mentioned the `dict` solution for the report in the past?

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: `estimator_reports_` is the attribute used in `CrossValidationReport`, so for consistency I'd like to keep it like this.

auguste-probabl: I'll add documentation

sphinx/reference/report/index.rst
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: done, thanks!

glemaitre:
Nice work @thomass-dev @auguste-probabl 🎉

PR #1295, opened 2025-02-07
docs: Change colors of links

MarieS-WiMLDS:
Closes https://github.com/probabl-ai/skore/issues/1008

### DIFF
--- a/sphinx/_static/css/custom.css
+++ b/sphinx/_static/css/custom.css
@@ -107,3 +107,13 @@ div.docutils.container.index-box {
     padding-top: 1.5rem !important;
     padding-bottom: 2.5rem !important;
 }
+
+/* Change the color of links to the orange of skore */
+a {
+    color: #F89A36; /* Change this to your desired color */
+}
+
+/* Change the color of visited links to the blue of skore */
+a:visited {
+    color: #3499CD; /* Change this to your desired color */
+}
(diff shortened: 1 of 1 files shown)


sphinx/_static/css/custom.css
```diff
@@ -107,3 +107,13 @@ div.docutils.container.index-box {
     padding-top: 1.5rem !important;
     padding-bottom: 2.5rem !important;
 }
+
+/* Change the color of links to the orange of skore */
+a {
+    color: #F89A36; /* Change this to your desired color */
```
glemaitre, line 113:
```suggestion
    color: #F89A36;
```

sphinx/_static/css/custom.css
```diff
@@ -107,3 +107,13 @@ div.docutils.container.index-box {
...
+}
+
+/* Change the color of visited links to the blue of skore */
+a:visited {
```
glemaitre, line 117: I would also suggest the hover to be the blue color to avoid to have a 3rd color (purple right now).

```suggestion
a:visited,
a:hover {
```

sylvaincom:
I think there are too many colors different colors, especially `Project` is underlined in orange

Capture d’écran 2025-02-10 à 13 03 43

glemaitre:
That's why you better redefined the color rather than modify the HTML objects.

MarieS-WiMLDS:
yes, the colors for light mode are not exactly those of the logo because the contrast was not nearly high enough, but it's the same idea, orange and blue, only darker.

PR #1296, opened 2025-02-07
ci: Ensure pyenv works when generating lockfiles

thomass-dev:
Currently, if the developer has some `pyenv` global settings or use `direnv`, we can't set the python version in a subshell to generate lockfiles. Instead, the developer python version is used silently.

---

We fix it by overloading `PYENV_VERSION`, which is the highest way to define the python version we want.

https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-shell

> Sets a shell-specific Python version by setting the PYENV_VERSION environment variable in your shell. This version overrides application-specific versions and the global version.

---

```diff
$ bash pip-compile.sh
Generating requirements: python==3.9 | scikit-learn==1.6 (1/6)
- Something wrong setting 'python-3.9', get '3.12'
```

```bash
$ bash pip-compile.sh
Generating requirements: python==3.9 | scikit-learn==1.6 (1/6)
Generating requirements: python==3.10 | scikit-learn==1.6 (2/6)
Generating requirements: python==3.11 | scikit-learn==1.6 (3/6)
Generating requirements: python==3.12 | scikit-learn==1.4 (4/6)
Generating requirements: python==3.12 | scikit-learn==1.5 (5/6)
Generating requirements: python==3.12 | scikit-learn==1.6 (6/6)
```

### DIFF
--- a/skore/ci/pip-compile.sh
+++ b/skore/ci/pip-compile.sh
@@ -36,6 +36,7 @@ set -eu
     # Move to `TMPDIR` to avoid absolute paths in requirements file
     cd "${TMPDIR}"
 
+    counter=1
     for combination in "${COMBINATIONS[@]}"
     do
         IFS=";" read -r -a combination <<< "${combination}"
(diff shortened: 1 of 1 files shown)


auguste-probabl (approved):
Looks good to me!

PR #1299, opened 2025-02-07
fix(api): Do not call .plot() in display factory

### LINKED ISSUE #1298: api: Do not call `.plot()` when creating the display
Following the changes in #1274, the display factory should not call internally `.plot()` and instead only return the display.

glemaitre:
closes #1298 

This PR makes sure that the display factory does not call `.plot()` directly.
It allows:

- to pass plotting parameters to be passed only to `display.plot()`
- to have a consistent api as for pandas object like dataframe:

```python
report.metrics.report_metrics().plot(kind="barh")
report.metrics.roc().plot()
```


thomass-dev (approved):
Much clearer, thanks!

PR #1300, opened 2025-02-07
feat(api): Allow to flatten index in reports

### LINKED ISSUE #1271: enh: expose a parameter to return a flat/multiindex in `report_metrics`
As discussed in https://github.com/probabl-ai/skore/issues/1260, we should provide a parameter in `report_metrics` to provide a flat or multiindex when we output the resulting dataframe.

glemaitre:
closes #1271

Add a parameter to allow to get a flat index to ease indexing.


skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
auguste-probabl, line 1: Can you adapt this for ComparisonReport?

glemaitre: done

glemaitre:
I change a bit the rule to not have any upper case, spaces and hash characters and replace with lower case, underscore, and nothing, respectively.

It would be the type of naming that I would use if I want to index a dataframe. I documented this behaviour in the different docstrings.

auguste-probabl:
Time to rebase ^^

auguste-probabl (approved):
Nice work!

PR #1301, opened 2025-02-10
docs(fix): Explicitely call .plot() in example to show the output

glemaitre:
Maybe due to the resolution of a merge conflict in #1299, it is missing some call to `.plot()`.

Here, we call explicitly and thus show the plot.

ping @thomass-dev


sylvaincom (approved):
LGTM, thanks

PR #1304, opened 2025-02-10
feat: Change repr method of `EstimatorReport` and `CrossValidationReport`

### LINKED ISSUE #1293: `EstimatorReport` repr method has unintended side-effects
The `__repr__` method of EstimatorReport prints a help message:
```python
>>> report
╭─────── skore.EstimatorReport ────────╮
│ Get guidance using the help() method │
╰──────────────────────────────────────╯
```

This can cause strange things to happen when printing things containing an `EstimatorReport`:
```python
>>> [report, report]
[╭─────── skore.EstimatorReport ────────╮
 │ Get guidance using the help() method │
 ╰──────────────────────────────────────╯,
 ╭─────── skore.EstimatorReport ────────╮
 │ Get guidance using the help() method │
 ╰──────────────────────────────────────╯]
```

In VSCode it looks even less intelligible:

Image

Perhaps a strategy similar to sklearn can be used, where the repr shows an HTML widget when the estimator is shown on its own, but it shows a string when the estimator is inside another object.

Originally reported by @MarieS-WiMLDS

auguste-probabl:
This makes it more practical to print objects containing such reports,
e.g. `[report]`.

Closes #1293

### DIFF
--- a/skore/src/skore/sklearn/_cross_validation/report.py
+++ b/skore/src/skore/sklearn/_cross_validation/report.py
@@ -323,7 +323,5 @@ class CrossValidationReport(_BaseReport, DirNamesMixin):
         )
 
     def __repr__(self):
-        """Return a string representation using rich."""
-        return self._rich_repr(
-            class_name="skore.CrossValidationReport", help_method_name="help()"
-        )
+        """Return a string representation."""
+        return f"{self.__class__.__name__}(estimator={self.estimator_}, ...)"
--- a/skore/src/skore/sklearn/_estimator/report.py
+++ b/skore/src/skore/sklearn/_estimator/report.py
@@ -326,7 +326,5 @@ class EstimatorReport(_BaseReport, DirNamesMixin):
         )
 
     def __repr__(self):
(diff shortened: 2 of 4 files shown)


thomass-dev (dismissed):
Nice!

skore/src/skore/sklearn/_cross_validation/report.py
```diff
@@ -323,7 +323,5 @@ def _get_help_legend(self):
         )
 
     def __repr__(self):
-        """Return a string representation using rich."""
-        return self._rich_repr(
-            class_name="skore.CrossValidationReport", help_method_name="help()"
-        )
+        """Return a string representation."""
+        return f"{self.__class__.__name__}(estimator={self.estimator_}, ...)"
```
thomass-dev, line 327: Nitpick: `self.__class__.__name__` is different from `skore.CrossValidationReport`.

auguste-probabl: You're right; I think `self.__class__.__name__` is correct here though.

PR #1305, opened 2025-02-10
feat(_find_ml_task): Refine ML task detection logic

### LINKED ISSUE #1005: Have distinct problem for single-output regression and multi-output regression
Right now, when detecting the ML task, we return a `"regression"` string for both single- and multi-output regression.

However, with `EstimatorReport` and some plotting it might not be straightforward to support `multi-output` regression out of the box. It could be handy to distinguish both to provide a fine grain solution in the different tools.

auguste-probabl:
ML detection now also works for multioutput targets (e.g. where `y` is a 2d array).

Added 3 new MLTask variants: `multioutput-binary-classification`, `multioutput-multiclass-classification` and `multioutput-regression`. `regression` still means "single-output regression".

The detection uses the same mechanism described in https://github.com/probabl-ai/skore/commit/be29a2641839e6ae83831ca1210b278d7f515c2a to discriminate between an array of integers that is actually for classification vs. one that is for regression.

Closes #1005

### DIFF
--- a/skore/src/skore/sklearn/find_ml_task.py
+++ b/skore/src/skore/sklearn/find_ml_task.py
@@ -107,4 +107,9 @@ def _find_ml_task(y, estimator=None) -> MLTask:
             if _is_sequential(y) and 0 in y:
                 return "multiclass-classification"
             return "regression"
+        if target_type == "multiclass-multioutput":
+            # If y contains integers, type_of_target considers
+            # the task to be multiclass classification.
+            # We refine this analysis a bit here.
+            return "multioutput-multiclass-classification"
         return "unknown"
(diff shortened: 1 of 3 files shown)


glemaitre (commented):
There is also another "bug" that I realised. In l.99-100, we return `"unknown"`.

In short, we use the estimator and if the estimator does not provide us the require information then we return "unknown". I think it would be better to fall back to the target inference, and try to get the ML task from it.

I realized that if someone provide a non-compatible estimator, we would not be inferring the task properly.

glemaitre (commented):
For the rest of the PR, all look good.

thomass-dev (approved):

PR #1308, opened 2025-02-11
ci: Fix backend workflow to cache venv using exact python version

thomass-dev:
Cache venv using exact python version to fix failing pipelines https://github.com/probabl-ai/skore/actions/runs/13260129137/job/37014601717?pr=1305.

References:

https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#check-latest-version
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#python-version

---

image
image

### DIFF
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -86,8 +86,10 @@ jobs:
 
       - name: Setup Python
         uses: actions/setup-python@v5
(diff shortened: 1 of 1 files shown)


auguste-probabl (approved):
So we're not actually locking the version, just ensuring that we don't go backwards?

PR #1309, opened 2025-02-11
feat: Add dedicated `Display` classes to ComparisonReport

thomass-dev:
Duplicate the plot classes for `ComparisonReport` use-cases:
- `RocCurveDisplay`,
- `PredictionRecallCurveDisplay`,
- `PredictionErrorDisplay`.

I've chosen to simplify a few things from the original classes, in particular by removing the user's ability to customize the plot.
The plot tests strictly compare pixel by pixel the matplotlib result with an expected image.

The next iteration should be to factorize these duplicated classes with the one in `skore.sklearn._plot`.

---

**Roc curves**
binary-classification
multiclass-classification

**PR curves**
pr-binary-classification
pr-multiclass-classification

**Prediction error curves**
Figure_1
Figure_2

### DIFF
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -1088,6 +1088,8 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
     def _get_display(
         self,
         *,
+        X,
+        y,
         data_source,
         response_method,
         display_class,
@@ -1097,11 +1099,18 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
 
         Parameters
         ----------
-        data_source : {"test", "train"}, default="test"
+        X : array-like of shape (n_samples, n_features)
+            The data.
+
+        y : array-like of shape (n_samples,)
+            The target.
+
+        data_source : {"test", "train", "X_y"}, default="test"
             The data source to use.
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
+            - "X_y" : use the provided `X` and `y` to compute the metric.
 
         response_method : str
             The response method.
@@ -1130,17 +1139,21 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             display = self._parent._cache[cache_key]
         else:
             y_true, y_pred = [], []
+
             for report in self._parent.estimator_reports_:
-                X, y, _ = report.metrics._get_X_y_and_data_source_hash(
-                    data_source=data_source
+                report_X, report_y, _ = report.metrics._get_X_y_and_data_source_hash(
+                    data_source=data_source,
+                    X=X,
+                    y=y,
                 )
-                y_true.append(y)
+
+                y_true.append(report_y)
                 y_pred.append(
                     _get_cached_response_values(
                         cache=report._cache,
                         estimator_hash=report._hash,
                         estimator=report._estimator,
-                        X=X,
+                        X=report_X,
                         response_method=response_method,
                         data_source=data_source,
                         data_source_hash=None,
@@ -1167,7 +1180,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             supported_ml_tasks=["binary-classification", "multiclass-classification"]
         )
     )
-    def roc(self, *, data_source="test", pos_label=None, ax=None):
+    def roc(self, *, data_source="test", X=None, y=None, pos_label=None, ax=None):
         """Plot the ROC curve.
 
         Parameters
@@ -1177,6 +1190,15 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
+            - "X_y" : use the provided `X` and `y` to compute the metric.
+
+        X : array-like of shape (n_samples, n_features), default=None
+            New data on which to compute the metric. By default, we use the validation
+            set provided when creating the report.
+
+        y : array-like of shape (n_samples,), default=None
+            New target on which to compute the metric. By default, we use the target
+            provided when creating the report.
 
         pos_label : int, float, bool or str, default=None
             The positive class.
@@ -1219,6 +1241,8 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         response_method = ("predict_proba", "decision_function")
         display_kwargs = {"pos_label": pos_label}
         return self._get_display(
+            X=X,
+            y=y,
             data_source=data_source,
             response_method=response_method,
             display_class=RocCurveDisplay,
@@ -1230,7 +1254,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             supported_ml_tasks=["binary-classification", "multiclass-classification"]
         )
     )
-    def precision_recall(self, *, data_source="test", pos_label=None):
+    def precision_recall(self, *, data_source="test", X=None, y=None, pos_label=None):
         """Plot the precision-recall curve.
 
         Parameters
@@ -1240,6 +1264,15 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
+            - "X_y" : use the provided `X` and `y` to compute the metric.
+
+        X : array-like of shape (n_samples, n_features), default=None
+            New data on which to compute the metric. By default, we use the validation
+            set provided when creating the report.
+
+        y : array-like of shape (n_samples,), default=None
+            New target on which to compute the metric. By default, we use the target
+            provided when creating the report.
 
         pos_label : int, float, bool or str, default=None
             The positive class.
@@ -1282,6 +1315,8 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         response_method = ("predict_proba", "decision_function")
         display_kwargs = {"pos_label": pos_label}
         return self._get_display(
+            X=X,
+            y=y,
             data_source=data_source,
             response_method=response_method,
             display_class=PrecisionRecallCurveDisplay,
@@ -1293,6 +1328,8 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         self,
         *,
         data_source="test",
+        X=None,
+        y=None,
         subsample=1_000,
         random_state=None,
     ):
@@ -1307,6 +1344,15 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
+            - "X_y" : use the provided `X` and `y` to compute the metric.
+
+        X : array-like of shape (n_samples, n_features), default=None
+            New data on which to compute the metric. By default, we use the validation
+            set provided when creating the report.
+
+        y : array-like of shape (n_samples,), default=None
+            New target on which to compute the metric. By default, we use the target
+            provided when creating the report.
 
         subsample : float, int or None, default=1_000
             Sampling the samples to be shown on the scatter plot. If `float`,
@@ -1355,6 +1401,8 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         """
         display_kwargs = {"subsample": subsample, "random_state": random_state}
         return self._get_display(
+            X=X,
+            y=y,
             data_source=data_source,
             response_method="predict",
             display_class=PredictionErrorDisplay,
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -5,7 +5,12 @@ from sklearn.metrics import make_scorer
 from sklearn.utils.metaestimators import available_if
 
 from skore.externals._pandas_accessors import DirNamesMixin
-from skore.sklearn._base import _BaseAccessor
+from skore.sklearn._base import _BaseAccessor, _get_cached_response_values
+from skore.sklearn._comparison.precision_recall_curve_display import (
+    PrecisionRecallCurveDisplay,
+)
+from skore.sklearn._comparison.prediction_error_display import PredictionErrorDisplay
+from skore.sklearn._comparison.roc_curve_display import RocCurveDisplay
 from skore.utils._accessor import _check_supported_ml_task
 from skore.utils._progress_bar import progress_decorator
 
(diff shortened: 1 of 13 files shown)


skore/src/skore/sklearn/_comparison/prediction_error_display.py
```diff
@@ -0,0 +1,147 @@
...
+        y_range_perfect_pred = [np.inf, -np.inf]
+        for y_true, y_pred in zip(self.y_true, self.y_pred):
+            if kind == "actual_vs_predicted":
+                min_value = min(y_pred.min(), y_true.min())
+                max_value = max(y_pred.max(), y_true.max())
+                x_range_perfect_pred[0] = min(x_range_perfect_pred[0], min_value)
+                x_range_perfect_pred[1] = max(x_range_perfect_pred[1], max_value)
+                y_range_perfect_pred[0] = min(y_range_perfect_pred[0], min_value)
+                y_range_perfect_pred[1] = max(y_range_perfect_pred[1], max_value)
+            else:
```
auguste-probabl, line 65:
```suggestion
            else:  # kind == "residual_vs_predicted"
```

skore/src/skore/sklearn/_comparison/prediction_error_display.py
```diff
@@ -0,0 +1,147 @@
...
+        **kwargs,
+    ):
+        if ml_task != "regression":
+            raise ValueError
+
+        random_state = check_random_state(random_state)
+        if isinstance(subsample, numbers.Integral):
+            if subsample <= 0:
+                raise ValueError(
+                    f"When an integer, subsample={subsample} should be positive."
```
auguste-probabl, line 114:
```suggestion
                    f"When an integer, subsample should be positive; got {subsample}."
```

skore/src/skore/sklearn/_comparison/prediction_error_display.py
```diff
@@ -0,0 +1,147 @@
...
+                )
+        elif isinstance(subsample, numbers.Real) and (subsample <= 0 or subsample >= 1):
+            raise ValueError(
+                f"When a floating-point, subsample={subsample} should be in the "
+                "(0, 1) range."
```
auguste-probabl, line 119:
```suggestion
                f"When a floating-point, subsample should be between 0 and 1; got {subsample}."
```

auguste-probabl (approved):
Haven't tested manually but so far so good!

skore/src/skore/sklearn/_comparison/precision_recall_curve_display.py
```diff
@@ -0,0 +1,200 @@
...
+                    class_linestyle = LINESTYLE[(class_idx % len(LINESTYLE))]
+
+                    (line_,) = ax.plot(
+                        recall,
+                        precision,
+                        color=report_color,
+                        linestyle=class_linestyle,
+                        alpha=0.6,
+                        label=(
+                            f"{report_name} #{report_idx + 1} - class {str(class_)} "
```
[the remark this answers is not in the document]
thomass-dev: Fixed, typing mistake after removing `str(class_).title()`.

auguste-probabl (dismissed):
Looks good to me

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
```diff
@@ -0,0 +1,126 @@
...
+        plt.gcf().savefig(tmp_path / "rc-multiclass-classification.png")
+
+        assert (
+            compare_images(
+                tmp_path / "rc-multiclass-classification.png",
+                CWD / "rc-multiclass-classification.png",
+                0,
+            )
+            is None
+        )
```
glemaitre, line 126: It is a trap :)

More seriously I think that we don't need to got this road. It would be best to only check the programmatic matplotlib important element that one can access.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
```diff
@@ -0,0 +1,126 @@
...
+    estimators = (FakeEstimator(classes), FakeEstimator(classes))
+
+    return RocCurveDisplay._from_predictions(
+        y_true=y_true,
+        y_pred=y_pred,
+        estimators=estimators,
+        estimator_names=["BC-E1", "BC-E2"],
+        ml_task="binary-classification",
+        data_source="test",
+    )
```
glemaitre, line 32: Instead of calling the private factory, let's called the `ComparisonReport` and generate the report from there.
It will allow us to not have to retouch the test once we come with a refactoring because we are going through the public interface that will not change.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
```diff
@@ -0,0 +1,126 @@
...
+            {
+                0: [(0.0, 1.0, 1.0), (0.0, 0.0, 0.0, 1.0)],
+                1: [(0.0, 1.0, 1.0), (0.0, 0.0, 0.0, 1.0)],
+                2: [(0.0, 1.0, 1.0), (0.0, 0.0, 0.0, 1.0)],
+            },
+        )
+        assert_equal(display.roc_auc, {0: [1.0, 0.0], 1: [1.0, 0.0], 2: [1.0, 0.0]})
+        assert_equal(display.estimator_names, ["MC-E1", "MC-E2"])
+        assert_equal(display.ml_task, "multiclass-classification")
+        assert_equal(display.pos_label, None)
```
glemaitre, line 93: When we deal with Python type, I prefer to keep the pure `assert`:

```python
assert display.pos_label is None
```

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
glemaitre: Basically, I would reuse almost the same test than in `test_roc_curve.py` and the only change is the really the legend in your case. I assume that once we refactor, we will be able to drop some of the tests because they go through the same code path.

glemaitre (commented):
I'll have a look at the display code now.

skore/src/skore/sklearn/_comparison/roc_curve_display.py
```diff
@@ -0,0 +1,304 @@
...
+        self.roc_auc = roc_auc
+        self.estimator_names = estimator_names
+        self.ml_task = ml_task
+        self.pos_label = pos_label
+        self.data_source = data_source
+
+    def plot(
+        self,
+        *,
+        plot_chance_level=True,
```
glemaitre, line 116: I would replicate the exact interface of the current roc curve display:

```python
        self,
        ax=None,
        *,
        estimator_name=None,
        roc_curve_kwargs=None,
        plot_chance_level=True,
        chance_level_kwargs=None,
        despine=True,
```

It is temporary fine to get a different display but we should make sure that a user do not see a different interface.

skore/src/skore/sklearn/_comparison/roc_curve_display.py
```diff
@@ -0,0 +1,304 @@
...
+        """Plot visualization.
+
+        Parameters
+        ----------
+        plot_chance_level : bool, default=True
+            Whether to plot the chance level.
+        despine : bool, default=True
+            Whether to remove the top and right spines from the plot.
+        """
+        _, ax = plt.subplots()
```
glemaitre, line 128: we need to store the axis in `self.ax_` such that one can further tweak it. It is the same reasoning with `self.lines_`.

skore/src/skore/sklearn/_comparison/roc_curve_display.py
```diff
@@ -0,0 +1,304 @@
...
+    (5, (10, 3)),
+    (0, (5, 10)),
+    (0, (5, 5)),
+    (0, (5, 1)),
+    (0, (3, 10, 1, 10)),
+    (0, (3, 5, 1, 5)),
+    (0, (3, 1, 1, 1)),
+    (0, (3, 5, 1, 5, 1, 5)),
+    (0, (3, 10, 1, 10, 1, 10)),
+    (0, (3, 1, 1, 1, 1, 1)),
```
glemaitre, line 33: I would say that I have no idea what does are :).
My GPT friend came with nice names that are descriptive and nice to read.

```python
linestyles = OrderedDict([
    ('solid',               (0, ())),
    ('loosely dotted',      (0, (1, 10))),
    ('dotted',              (0, (1, 5))),
    ('densely dotted',      (0, (1, 1))),
    ('loosely dashed',      (0, (5, 10))),
    ('dashed',              (0, (5, 5))),
    ('densely dashed',      (0, (5, 1))),
    ('loosely dashdotted',  (0, (3, 10, 1, 10))),
    ('dashdotted',          (0, (3, 5, 1, 5))),
    ('densely dashdotted',  (0, (3, 1, 1, 1))),
    ('loosely dashdotdotted', (0, (3, 10, 1, 10, 1, 10))),
    ('dashdotdotted',         (0, (3, 5, 1, 5, 1, 5))),
    ('densely dashdotdotted', (0, (3, 1, 1, 1, 1, 1)))
])
```

skore/src/skore/sklearn/_comparison/prediction_error_display.py
```diff
@@ -0,0 +1,221 @@
...
+        cls,
+        y_true,
+        y_pred,
+        *,
+        estimator_names,
+        ml_task,
+        data_source=None,
+        subsample=1_000,
+        random_state=None,
+        **kwargs,
```
glemaitre, line 145: We probably don't need this kwargs apparently?

glemaitre (commented):
The same comment would be true for the other plots.

skore/src/skore/sklearn/_comparison/roc_curve_display.py
thomass-dev: What's the use-case for modifying the plot outside the `plot()` function ?

skore/src/skore/sklearn/_comparison/roc_curve_display.py
thomass-dev: Fixed,  i added the description :).

thomass-dev: Tell your GPT that dict is ordered since Python 3.7 :trollface: .

skore/src/skore/sklearn/_comparison/prediction_error_display.py
thomass-dev: This is for the `estimators` parameter, passed from `metrics_accessor.py`.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
thomass-dev: Can you describe your fears? I don't understand. 

That way, we can be exhaustive and test the entire plot.
Furthermore, it avoids to set specific attributes on the display object, needed only for the test.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
```diff
@@ -0,0 +1,126 @@
...
+from numpy.testing import assert_equal
+from pytest import fixture
+from skore.sklearn._comparison.roc_curve_display import RocCurveDisplay
+
+CWD = Path(__file__).parent.resolve()
+
+
+class FakeEstimator:
+    def __init__(self, classes: list[int]):
+        self.classes_ = classes
```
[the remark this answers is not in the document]
thomass-dev: In theory, i totally agree with you. But in that case, we only need a class with the `classes_` attribute.
I would avoid the cost of the `estimator.fit` call to get this attribute. No other estimator feature are used.

Do you think `classes_` can be dropped by scikit-learn in any world ? If so, i will change.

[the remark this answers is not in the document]
thomass-dev: I think we should test compatibility with scikit-learn in one integration test.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
thomass-dev: You true, a too fast search and replace... i will fix it.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
thomass-dev: This is intentional. In this unittest i only want to test at the lowest level the feature `RocCurveDisplay` and all its functions.
Do you want me to add a more higher integration test?

skore/src/skore/sklearn/_comparison/prediction_error_display.py
thomass-dev: Fixed.

skore/src/skore/sklearn/_comparison/prediction_error_display.py
thomass-dev: Fixed.

skore/src/skore/sklearn/_comparison/roc_curve_display.py
glemaitre: For instance:

```
display.ax_.set_title("Put a nice title")
```

but you can set a bunch of stuff.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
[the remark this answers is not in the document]
glemaitre: It is part of the spec. One thing that can happen is that an estimator that pretend to be scikit-learn compatible is not compatible. But it is not what we are going to test here.

> I would avoid the cost of the estimator.fit

I agree with the fact that you want to have a very cheap operation. But you can actually achieve it with a real estimator such as `DummyClassifier` that does not `fit` anything and that already have the tested API for us. I would certainly always go for a linear model when testing with a small dataset and we will be at the level of milliseconds as well.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
glemaitre: To summarize an IRL discussion, let's go with a unit test + integration test :P

skore/src/skore/sklearn/_comparison/roc_curve_display.py
thomass-dev: Fixed.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
thomass-dev: Fixed.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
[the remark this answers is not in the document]
thomass-dev: Duplicate of https://github.com/probabl-ai/skore/pull/1309#discussion_r1959799812 .

skore/src/skore/sklearn/_comparison/precision_recall_curve_display.py
```diff
@@ -0,0 +1,310 @@
...
+        drop_intermediate : bool, default=False
+            Whether to drop some suboptimal thresholds which would not appear
+            on a plotted precision-recall curve. This is useful in order to
+            create lighter precision-recall curves.
+
+        Returns
+        -------
+        display : PrecisionRecallCurveDisplay
+        """
+        estimator_classes = map(attrgetter("classes_"), estimators)
```
glemaitre, line 250:
```suggestion
        estimator_classes = [est.classes_ for est in estimators]
```

It is a legit code but you will have the following issue:

- The Zen of Python will tell you that the readability count and the map/reduce is usually replaced by a list comprehension if memory is not an issue
- Related to the previous one, `map` will return a generator. Thus, we will not raise an error before actually consuming the iterator and it could be inside a part of the code where it will be easy to understand what `classes_` is.
- Because we have a generator, if we consume it once, then we cannot access the classes a second time. So it could be prone to error in this specific case because we don't really care about the memory footprint but we might need to access several time.

glemaitre (commented):
Looks good. A couple of comments that could be expanded to each occurrence of the pattern.

skore/src/skore/sklearn/_comparison/precision_recall_curve_display.py
thomass-dev: > The Zen of Python will tell you that the readability count and the map/reduce is usually replaced by a list comprehension if memory is not an issue

This is the equivalent of emacs/vim debate, there is pros and cons, i use one or the other depending on the situation.
Otherwise, the code assumes that we only have to consume one time the classes of estimator.
This is why i use a generator.

No worries to change, fixed for you in 2ed61d0b4a260ce157551f79e8905f4161175670.

thomass-dev: Just keep in mind that generator is ideal when you want to make your code more readable without overhead.

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
```diff
@@ -0,0 +1,183 @@
...
+        assert_equal(display.lines_[3].get_ydata(), (0.0, 0.0, 0.0, 1.0))
+        assert_equal(display.lines_[4].get_ydata(), (0.0, 0.0, 0.0, 1.0))
+        assert_equal(display.lines_[5].get_ydata(), (0.0, 0.0, 0.0, 1.0))
+        assert display.lines_[0].get_label() == "MC-E1 #1 - class 0 (AUC = 0.50)"
+        assert display.lines_[1].get_label() == "MC-E1 #1 - class 1 (AUC = 0.50)"
+        assert display.lines_[2].get_label() == "MC-E1 #1 - class 2 (AUC = 0.50)"
+        assert display.lines_[3].get_label() == "MC-E2 #2 - class 0 (AUC = 0.50)"
+        assert display.lines_[4].get_label() == "MC-E2 #2 - class 1 (AUC = 0.50)"
+        assert display.lines_[5].get_label() == "MC-E2 #2 - class 2 (AUC = 0.50)"
+        assert display.lines_[0].get_color() != display.lines_[3].get_color()
```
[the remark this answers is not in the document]
thomass-dev: As it is, even though there are a lot of lines, i find it easier to read and understand (i have tested both before your comment). At a glance, you can see how the lines must be ordered.

examples/model_evaluation/plot_comparison_report.py
```diff
@@ -0,0 +1,219 @@
+"""
```
thomass-dev, line 1: What about this example @sylvaincom @glemaitre @auguste-probabl ?

skore/tests/unit/sklearn/comparison/test_roc_curve_display.py
thomass-dev: I'm adding integration tests to ensure that `ComparisonReport.roc` and so on work.

examples/model_evaluation/plot_comparison_report.py
glemaitre: We modified an existing instead of introducing a new one. We could show the display there.

thomass-dev: So i will remove this file and improve existing examples, thanks.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -1078,3 +1083,280 @@ def __repr__(self):
...
+        data_source,
+        response_method,
+        display_class,
+        display_kwargs,
+    ):
+        """Get the display from the cache or compute it.
+
+        Parameters
+        ----------
+        data_source : {"test", "train"}, default="test"
```
glemaitre, line 1100: We should support `"X_y"` case as well since that we have `EstimatorReport` under the hood for the moment.

glemaitre: Then we will need a test as well (we should have this one when testing the `EstimatorReport`).

glemaitre (commented):
Otherwise, I'm fine with the state and we can latter refactor.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: Cool. Easy one then ;)

thomass-dev: But in fact, there is no `X` and `y` parameters in any plot functions (`roc`, `prediction_error` or `precision_recall`), either in `comparison/metrics_accessor.py`, or `cross-validation/metrics_accessor.py`.

Should i add need to add these parameters as it is done in `estimator/metrics_accessor.py`?
I think this should be postpone to another PR.

thomass-dev: I'm currently added the support of `X_y` **only** in comparison reports.

glemaitre: Nop. For the moment, the comparison report does not support `"X_y"`. We can see in another PR, if it is something that we want but discussing with @auguste-probabl yesterday, it is something that we can envision.

But here, you only need to change the ComparisonReport.

thomass-dev: Done.

glemaitre (approved):

PR #1312, opened 2025-02-12
fix(api): Add favoribility metric as a column indicator

### LINKED ISSUE #1260: fix: Remove the arrow up and down from the column name in the dataframe
When showing a score dataframe, e.g. `report.metrics.report_metrics()`, the column name contains some arrow up/down depending if we show a score or a loss. However, adding this information directly in the column is annoying to index a specific score.

We should come with another way to provide this information. Maybe, a dirty way but possible way would be to patch the HTML repr from pandas for those specific dataset to incorporate the information in the HTML representation itself.

glemaitre:
closes #1260 

Add whether a metric is greater is better or lower is better as a column indicator instead to have it in the name of the index.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -50,6 +50,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         scoring_names=None,
         pos_label=None,
         scoring_kwargs=None,
+        indicator_favorability=False,
     ):
         """Report a set of metrics for our estimator.
 
@@ -88,6 +89,10 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
 
         scoring_kwargs : dict, default=None
             The keyword arguments to pass to the scoring functions.
+
+        indicator_favorability : bool, default=False
+            Whether or not to add an indicator of the favorability of the metric as
+            an extra column in the returned DataFrame.
 
         Returns
         -------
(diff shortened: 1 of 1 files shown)


auguste-probabl (dismissed):
Looks good to me!

auguste-probabl:
Consider adding an example of the new behavior (in some doctest or in the PR description)

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
auguste-probabl, line 1: I'd rather delay this change until #1286 is merged

glemaitre:
To not forget, let's use this feature in the example use-case.

glemaitre:
Found a new bug and fix the example

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -168,14 +177,27 @@ def _compute_metric_scores(
                 keys=[f"Split #{i}" for i in range(len(results))],
             )
             results = results.swaplevel(0, 1, axis=1)
+
+            # pop the favorability column if it exist to:
+            # - no use it in the aggregate operation
+            # - later to only report a single column and not by split columns
```
auguste-probabl, line 183:
```suggestion
            # Pop the favorability column if it exists, to:
            # - not use it in the aggregate operation
            # - later to only report a single column and not by split columns
```
What is meant by the second bullet point?

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
auguste-probabl: Please add this to comparison report as well

glemaitre: done

auguste-probabl (approved):
Ready to merge when the tests pass

PR #1313, opened 2025-02-12
docs: Example of the feature plot feature importance

MarieSacksick:
Doc specification.


glemaitre:
I would just point out that here, it is only the permutation importance that is one type of feature importance. I think it is nice to have the following reference: https://christophm.github.io/interpretable-ml-book/ when it comes to model's interpretability and plots.

PR #1318, opened 2025-02-13
refactor: Remove `CrossValidationReport._MetricsAccessor`

auguste-probabl:
Since a `CrossValidationReport` can be considered a special case of a `ComparisonReport`, this PR uses inheritance to avoid code duplication. Of note are the following design choices:
- `ComparisonReport` metrics have the `aggregate` parameter in order to make them available to `CrossValidationReport`.
- `ComparisonReport` can now accept just 1 `EstimatorReport`, so that even if cross-validation is interrupted, the initialization succeeds.

Todo:
- [ ] Make sure plots are correct for both Reports


glemaitre:
I would delay this refactoring once the `ComparisonReport` is accepting `CrossValidationReport`, i.e.:

```python
ComparisonReport(reports={
    "LR": CrossValidationReport(LogisticRegression(), X=X, y=y),
    "GBDT": CrossValidationReport(HistGradientBoostingClassifier(), X=X, y=y),
})
```

I want to know if we should have a `BaseComparisonReport` from which both classes should inherit from or we are fine if the current proposal.

Here, for the moment I find it weird that some child of `ComparisonReport` are given to the `ComparisonReport` itself.

thomass-dev:
Please delay this PR after the merge of https://github.com/probabl-ai/skore/pull/1309.

auguste-probabl:
#1512 makes it apparent that the two classes `ComparisonReport` and `CrossValidationReport` are different enough. I'll close this and keep it around for reference.

PR #1322, opened 2025-02-14
feat(project): Add parameter to `Project.clear` to delete project

### LINKED ISSUE #1294: Enh(project): Allow to delete a project from a command line

today, the user has to do either a rm -rf, which is cumbersome, or do it by hand, which is not nice either, if they want to delete a project.  

add a parameter to the project clear function, `delete_project`, a boolean, False by default, which can be turned to True. This way, the project would be deleted.  

See discussion here: https://github.com/probabl-ai/skore/discussions/1258.  

### Additional context

_No response_

auguste-probabl:
`project.clear(delete_project=True)` now deletes the entire project, while `project.clear(delete_project=False)` (the default) removes every item from the project.

Closes #1294


skore/src/skore/project/project.py
```diff
@@ -102,8 +103,18 @@ def __init__(
...
+
+        Parameters
+        ----------
+        delete_project : bool
+            If set, the project will be deleted entirely.
+        """
+        if delete_project is True:
+            shutil.rmtree(self.path)
+            return
+
```
thomass-dev, line 117: I'm not a big fan because i can `clear(delete_project=True)` a project and continue to use the project object.
It will lead to errors.

The deletion shouldn't be done in a method, or you should delete the object itself.

thomass-dev: I think it's a question of design. The question is: what do we want to do with a project that have been deleted.

Ping @MarieS-WiMLDS .

MarieS-WiMLDS: I think you are right.  
What do you suggest instead?

auguste-probabl: It's a little weird, but with the on-disk storage it does work fine: If you delete the project and then later do something with it (like `put`) it will re-create the folder under the hood.

auguste-probabl: At the very least I can add a warning that the user should probably avoid using the project variable once they have run `clear(delete_project=True)`

MarieS-WiMLDS: I find it very unexpected that the folder is re-created.  
What do you think of `skore.delete(project_name = "/examples/my_project")`?

auguste-probabl:
The solution I gave doesn't work on Windows, because Windows is stricter about deleting files while the `project` variable still exists. I don't really know how to deal with this.

skore/src/skore/project/project.py
auguste-probabl: This is better IMO, but the issue remains: if you delete the directory, but the `project` variable still exists, then the folder will be re-created and Windows will still complain

glemaitre (commented):
I think that I'm fine that `myproject.clear(delete_project=True)` invalidate completely the project.

However, I also agree that we should not have weird side-effect or magical stuff happening under the hood. If the project was deleted, I would expect a proper error message telling me that I need to call `skore.create` or `Project` constructor. It should be easy to track using a sentinel, e.g. `Project._storaged_initalized` that is set in the constructor to `True` and `False` in the `clear`. All methods from project, could check that this sentinel is well set before to do an operation.

I can think that we could so a small decorator for this.

skore/src/skore/project/project.py
```diff
@@ -102,8 +103,18 @@ def __init__(
 
         self._server_info = ServerInfo.rejoin(self)
 
-    def clear(self):
-        """Clear the project."""
+    def clear(self, delete_project=False):
+        """Remove all items from the project.
```
glemaitre, line 107: we could add a big warning directive to document the behaviour


```suggestion
        """Remove all items from the project.

        .. warning::
           `delete_project=True` will invalidate the `Project` instance. One needs to call
           :class:`~skore.Project` constructor or :func:`skore.open` function to recreate a
           project.
```

skore/src/skore/project/project.py
```diff
@@ -102,8 +103,18 @@ def __init__(
...
+
+        Parameters
+        ----------
+        delete_project : bool
+            If set, the project will be deleted entirely.
+        """
+        if delete_project is True:
```
glemaitre, line 114:
```suggestion
        if delete_project:
```

skore/src/skore/project/project.py
```diff
@@ -102,8 +103,18 @@ def __init__(
...
+            shutil.rmtree(self.path)
```
glemaitre, line 115:
```suggestion
            shutil.rmtree(self.path)
            self._storage_initialized = False
```

glemaitre: If we go with a sentinel.

skore/src/skore/project/project.py
MarieS-WiMLDS: Let's go for this solution then.  
About the `project` variable, it seems that Guillaume offers a solution below.

skore/src/skore/project/project.py
auguste-probabl: done

skore/src/skore/project/project.py
auguste-probabl: done

skore/src/skore/project/project.py
auguste-probabl: done

skore/src/skore/project/project.py
auguste-probabl: I ended up going for Guillaume's sentinel suggestion (which we had discussed @thomass-dev), which at least makes Windows happy. I kept the original design, `project.clear(delete_project=True)`.

thomass-dev (approved):

skore/src/skore/project/project.py
```diff
@@ -16,6 +17,29 @@
...
+
+
+def _raise_if_deleted(method):
+    """Raise if the underlying Project has been deleted, otherwise execute `method`.
+
+    This wrapper makes it safe to "delete" a Project, even if the Project instance
+    still exists.
+    """
+
+    def run(self, *args, **kwargs):
```
thomass-dev, line 31: Please use https://docs.python.org/3.12/library/functools.html#functools.wraps

thomass-dev (changes requested):
Nice, thanks for your investigation. Just a quick fix.

thomass-dev (approved):

PR #1327, opened 2025-02-14
fix(_find_ml_task): Make `_find_ml_task` work for non-numpy arrays

### LINKED ISSUE #1324: Bug: _find_ml_task does not accept series anymore
### Describe the bug

 _find_ml_task do not accept series anymore.

I think it's because of the line [`y = y.flatten()`](https://github.com/probabl-ai/skore/blob/b5f84adb03cfb659233d3085739b33ab086710d9/skore/src/skore/sklearn/find_ml_task.py#L25)

### Steps/Code to Reproduce

```python
from sklearn import datasets

X, y = datasets.load_iris(return_X_y=True, as_frame = True)
train_test_split(X,y)
```

renders 

Image

### Expected Behavior

The piece of code above should run.

### Actual Behavior

It doesn't run.  

### Environment

```shell
System:
    python: 3.12.8 | packaged by Anaconda, Inc. | (main, Dec 11 2024, 16:31:09) [GCC 11.2.0]
executable: /home/marie/anaconda3/envs/skore_test/bin/python
   machine: Linux-6.8.0-52-generic-x86_64-with-glibc2.35

Python dependencies:
        skore: 0.0.0+unknown
          pip: 24.2
   setuptools: 75.1.0
    diskcache: 5.6.3
      fastapi: 0.115.6
       joblib: 1.4.2
   matplotlib: 3.9.3
        numpy: 2.2.0
       orjson: 3.10.15
       pandas: 2.2.3
 platformdirs: 4.3.6
       plotly: 5.24.1
       psutil: 5.9.0
      pyarrow: 18.1.0
         rich: 13.9.4
 scikit-learn: 1.6.0
        skops: 0.11.0
      uvicorn: 0.32.1

Current `comparator` branch (as of Feb 14th)
```

auguste-probabl:
Closes #1324
closes https://github.com/probabl-ai/skore/issues/1357

### DIFF
--- a/skore/src/skore/sklearn/find_ml_task.py
+++ b/skore/src/skore/sklearn/find_ml_task.py
@@ -11,10 +11,10 @@ from skore.sklearn.types import MLTask
 
 def _column_is_classification(y) -> bool:
     """Check whether ``y`` is a 1-d array of sequential integer values containing 0."""
-    y_values = np.unique(y)
-    if np.any(y_values == 0):
-        sequential = np.arange(y_values[0], y_values[-1] + 1)
-        return np.array_equal(y_values, sequential)
+    y_unique = np.unique(y)
+    if np.any(y_unique == 0):
+        sequential = np.arange(y_unique[0], y_unique[-1] + 1)
+        return np.array_equal(y_unique, sequential)
     return False
 
 
--- a/skore/src/skore/sklearn/find_ml_task.py
+++ b/skore/src/skore/sklearn/find_ml_task.py
@@ -12,10 +12,10 @@ from skore.sklearn.types import MLTask
 def _column_is_classification(y) -> bool:
     """Check whether ``y`` is a 1-d array of sequential integer values containing 0."""
     y_values = np.unique(y)
-    if 0 not in y:
-        return False
-    sequential = np.arange(y_values[0], y_values[-1] + 1)
-    return np.array_equal(y_values, sequential)
+    if np.any(y_values == 0):
+        sequential = np.arange(y_values[0], y_values[-1] + 1)
+        return np.array_equal(y_values, sequential)
+    return False
 
 
 def _is_classification(y) -> bool:
--- a/skore/src/skore/sklearn/find_ml_task.py
+++ b/skore/src/skore/sklearn/find_ml_task.py
@@ -10,7 +10,7 @@ from skore.sklearn.types import MLTask
 
 def _is_sequential(y) -> bool:
     """Check whether ``y`` is vector of sequential integer values."""
-    y_values = np.sort(np.unique(y))
+    y_values = np.unique(y)
     sequential = np.arange(y_values[0], y_values[-1] + 1)
     return np.array_equal(y_values, sequential)
 
@@ -18,12 +18,17 @@ def _is_sequential(y) -> bool:
 def _is_classification(y) -> bool:
     """Determine if `y` is a target for a classification task.
 
-    If `y` contains integers, sklearn's `type_of_target` considers
-    the task to be multiclass classification.
-    This function makes the analysis finer.
+    If `y` contains integers, sklearn's `type_of_target` considers the task
+    to be multiclass classification. This might not be the case, so we add the
+    constraints that `y` must contain sequential values and contain 0.
+
+    If `y` does not contain numbers (e.g. strings), then this function returns True.
     """
     y = np.array(y).flatten()
-    return _is_sequential(y) and 0 in y
+    try:
+        return _is_sequential(y) and 0 in y
+    except TypeError:
+        return True
 
 
 def _find_ml_task(y, estimator=None) -> MLTask:
--- a/skore/src/skore/sklearn/find_ml_task.py
+++ b/skore/src/skore/sklearn/find_ml_task.py
@@ -25,13 +25,27 @@ def _is_classification(y) -> bool:
     to be multiclass classification. This might not be the case, so we add the
     constraints that `y` must contain sequential values and contain 0.
 
-    If `y` does not contain numbers (e.g. strings), then this function returns True.
+    If `y` is a 2-d array, we check each column independently; if at least one
+    column does not fit the constraints, then we consider the task to be regression.
+
+    Note that this can cause false positives, e.g. a classification task with classes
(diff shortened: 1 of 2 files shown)


skore/src/skore/sklearn/find_ml_task.py
```diff
@@ -22,7 +22,7 @@ def _is_classification(y) -> bool:
     the task to be multiclass classification.
     This function makes the analysis finer.
     """
-    y = y.flatten()
+    y = np.array(y).flatten()
```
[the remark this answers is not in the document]
glemaitre: I think that you need additional safeguards here. The following is going to fail.

```
_find_ml_task(["a", "a", "b", "b", "c"])
```

I think that we need to refine this sentence in the documentation: "This function makes the analysis finer.".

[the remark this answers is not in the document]
auguste-probabl: Addressed; let me know what you think about the type checking problem.

skore/src/skore/sklearn/find_ml_task.py
```diff
@@ -10,20 +10,25 @@
...
-    This function makes the analysis finer.
+    If `y` contains integers, sklearn's `type_of_target` considers the task
+    to be multiclass classification. This might not be the case, so we add the
+    constraints that `y` must contain sequential values and contain 0.
+
+    If `y` does not contain numbers (e.g. strings), then this function returns True.
     """
-    y = y.flatten()
-    return _is_sequential(y) and 0 in y
+    y = np.array(y).flatten()
```
glemaitre, line 27: I'm thinking that we could use `sklearn.utils.check_array` that should help.

```suggestion
    y = check_array(y, dtype="numeric", ensure_2d=False)
```

Whatever `ValueError` should be related to the type not being "numeric"

skore/tests/unit/sklearn/test_utils.py
```diff
@@ -80,3 +81,15 @@ def test_find_ml_task_unfitted_estimator():
...
+def test_find_ml_task_pandas():
+    y = pandas.Series([0, 1, 2])
+    assert _find_ml_task(y, None) == "multiclass-classification"
+
+    y = pandas.DataFrame([0, 1, 2])
+    assert _find_ml_task(y, None) == "multiclass-classification"
+
+
+def test_find_ml_task_string():
+    assert _find_ml_task(["0", "1", "2"], None) == "multiclass-classification"
```
glemaitre, line 95: Looking at the code more, I see that we have "multiclass-multioutput" and "multilabel-indicator" (https://scikit-learn.org/stable/modules/multiclass.html is good page to get around those definition).

The use of `flatten` is problematic in this case because those two types assumes that the shape of `y` is 2-dimensional and `flatten` will make a flat array and mix allow the array.

It would be safer to not flatten and apply the `_is_sequential` on each dimension of `y` I think.

auguste-probabl: I'm not sure I follow; here are some random thoughts, but please give examples of what you mean so that we can add corresponding tests.

- We do have `multioutput-binary-classification` in our MLTask type (which corresponds to multilabel-indicator), we're not skipping this case
- > not flatten and apply the _is_sequential on each dimension of y

  This makes me think of edge cases: for example, in `[[0, 0], [0, 2]]` the 2nd column is not sequential, so `_find_ml_task` currently returns `multioutput-regression`

skore/src/skore/sklearn/find_ml_task.py
auguste-probabl: I don't mind `y` containing strings, in this case we know that the task is classification

skore/tests/unit/sklearn/test_utils.py
glemaitre: > right now your example would result in "classification"

Yes it is but for the wrong reason. We should not mix labels from different output because they can represent really different thing.

> if we change it would result in "regression"

My question would be a bit more "meta": does the `_is_sequence` and sequential enforcement necessary for classification? Basically, it looks like we are chasing a case that is really ambiguous (in single and multiple output). Could you recall what was the reason for introducing this rule?

auguste-probabl: rule has been changed for 2d arrays

skore/src/skore/sklearn/find_ml_task.py
auguste-probabl: we use check_array now

skore/src/skore/sklearn/find_ml_task.py
```diff
@@ -91,10 +113,12 @@ def _find_ml_task(y, estimator=None) -> MLTask:
         if is_clusterer(estimator):
             return "clustering"
         if is_regressor(estimator):
-            if np.ndim(y) == 1:
+            if y is None:
```
glemaitre, line 116: Uhm I did not think about this case. Is it related when the user does not pass any `y_test`?

I think that we cannot do much. I think that I'll open an issue in scikit-learn to have a better interface. In the meanwhile, we could try our best:

- if the model has `intercept_`, we could check the number of elements and it will tell us the number of output
- if the model has an `n_outputs_` attribute then we have the answer
- otherwise we default to "regression" as you do now.

glemaitre: This case should be added in the test. I don't think that we properly cover it.

auguste-probabl: I brought the coverage up to 100%, let me know if this works for you

skore/src/skore/sklearn/find_ml_task.py
```diff
@@ -2,28 +2,54 @@
...
+    Similarly, if `y` is a 2-d array where some column contains 0 and 2 but not 1,
+    the whole array would be considered "regression".
+
+    If `y` does not contain numbers (e.g. strings) then this function returns True.
     """
-    y = y.flatten()
-    return _is_sequential(y) and 0 in y
+    try:
+        y = check_array(y, dtype="numeric", ensure_2d=False)
+    except ValueError:
```
glemaitre, line 45:
```suggestion
    except ValueError:
        # The conversion in `check_array` failed meaning
        # that we are in a classification case with non-numeric
        # data type.
```

glemaitre (approved):

PR #1329, opened 2025-02-14
fix: Use arrow in help instead of metric names

glemaitre:
Just spotted a regression that I introduced during a refactor. I used metric names instead of icon in the help of the `CrossValidationReport`.

Solving the issue and introducing two non-regression tests for the reporters.


auguste-probabl (approved):
Thanks

PR #1330, opened 2025-02-14
docs: Document the different displays as part of public API

### LINKED ISSUE #1328: docs: Document the different displays
Right now, we can get displays using reporters:

```python
    *load_breast_cancer(return_X_y=True), random_state=0
)
classifier = LogisticRegression(max_iter=10_000)
report = EstimatorReport(
    classifier,
    X_train=X_train,
    y_train=y_train,
    X_test=X_test,
    y_test=y_test,
)
display = report.metrics.roc()
display
```

One issue is that we don't publicly expose those displays in the documentation. So we should expose them publicly. Hopefully, we only have the `.plot` interface with which a user should interact.

glemaitre:
closes #1328 

Document the different displays since they are part of the public API.

### DIFF
--- a/skore/src/skore/__init__.py
+++ b/skore/src/skore/__init__.py
@@ -9,6 +9,9 @@ from skore.project import Project, open
 from skore.sklearn import (
     CrossValidationReport,
     EstimatorReport,
+    PrecisionRecallCurveDisplay,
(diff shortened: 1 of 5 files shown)


auguste-probabl (approved):
Alright, thanks!

PR #1331, opened 2025-02-15
chore: Use pixi for backend tests

glemaitre:
Giving a try to `pixi` for the CI


skore/pyproject.toml
```diff
@@ -60,6 +108,130 @@ artifacts = ["src/skore/ui/static/"]
...
+
+[tool.pixi.feature.dev.dependencies]
+ipykernel = "*"
+ipython = "*"
+jupyterlab = "*"
+
+[tool.pixi.feature.min-dependencies.dependencies]
+# Apply SPEC0 rule:
+# 1. Support for Python versions be dropped 3 years after their initial release.
+# 2. Support for core package dependencies be dropped 2 years after their initial
```
auguste-probabl, line 170:
```suggestion
# 1. Support for Python versions will be dropped 3 years after their initial release.
# 2. Support for core package dependencies will be dropped 2 years after their initial
```

glemaitre:
Yep it is just a POC for the moment. I wanted to have this such that we have a starting point to speak with @auguste-probabl and @thomass-dev

skore/pyproject.toml
```diff
@@ -60,6 +108,130 @@ artifacts = ["src/skore/ui/static/"]
...
+[tool.pixi.feature.doc.tasks]
+build-docs = { cmd = "make html", cwd = "../sphinx" }
+build-docs-quick = { cmd = "make html-noplot", cwd = "../sphinx" }
+clean-docs = { cmd = "rm -rf build/ && rm -rf auto_examples/ && rm -rf reference/api/", cwd = "../sphinx" }
+
+[tool.pixi.feature.lint.tasks]
+lint = { cmd = "pre-commit install && pre-commit run -v --all-files --show-diff-on-failure", cwd = "../" }
+
+[tool.pixi.feature.test.tasks]
+tests = { cmd = "mkdir -p coverage && python -m pytest -n auto src/ tests/ --junitxml=coverage/coverage.xml --cov-config=pyproject.toml --cov | tee coverage/coverage.txt" }
```
rouk1, line 222: Not sure if my comment is pertinent ^^
But... Using pixi as a task runner, allows to write this stuffs in python. Which make the repo windows friendly. Hence maybe we can avoid using rm, mv, mkdir, ...).

rouk1 (changes requested):
I find the pixi DX very nice. I wont approve this PR as I know other people may have strong divergent opinion on this.

To me, to get this merge, you should delete the top level makefile as it is replaced by pixi scripts. And address (or not^^) my comment on non DOS-compatible command usage.

Too bad that pixi can't manage installing a javascript runtime...

skore/pyproject.toml
glemaitre: Yes it is indeed one of the thing that I would like to have and indeed I did not make it DOS-compliant in my first pass.

I'm thinking that `spin` (https://github.com/scientific-python/spin) could help if we don't want to bother for some of the part :)

thomass-dev:
I've just had a really quick look and i find it really more complex. I need to deep-dive into your PR and maybe we should have a pair-session :).

auguste-probabl:
Now that I've seen what `pip-compile.sh` looks like I'd like to push for migrating to a robust solution like pixi. The current solution doesn't work on NixOS, and I presume it won't work on Windows either.

thomass-dev:
I tried `uv` and it can't be used to generate different requirement files for different python version.

glemaitre:
> I will make a PoC with uv, and see the difference.

`pixi` is using `uv` when you declare `pypi-dependencies`. So I would avoid to do an additional PoC :). However, we can make some environment that are pure PyPI and some that are conda-forge to be sure that both works as expected.

thomass-dev:

PR #1333, opened 2025-02-15
docs: Fix formatting of example section in docstring

### LINKED ISSUE #1326: docs: Docstring examples are broken
Doctests are not rendered as code blocks. See for example the latest version of the docs:

Image

Already occurs in version 0.6 of the docs (https://skore.probabl.ai/0.6/generated/skore.Project.html#skore.Project.get_note)

glemaitre:
closes #1326 

Fixing formatting in "Examples" section some method classes.

### DIFF
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -45,12 +45,10 @@ class Project:
 
     Examples
     --------
-    >>>
-    >> import skore
-    >>
-    >> project = skore.Project("my-xp")
-    >> project.put("score", 1.0)
-    >> project.get("score")
+    >>> import skore
+    >>> project = skore.Project("my-xp")
+    >>> project.put("score", 1.0)
+    >>> project.get("score")
     1.0
     """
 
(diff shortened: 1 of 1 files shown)


sylvaincom (approved):
LGTM, thanks

skore/src/skore/project/project.py
```diff
@@ -45,12 +45,10 @@ class Project:
...
-    >>>
-    >> import skore
-    >>
-    >> project = skore.Project("my-xp")
-    >> project.put("score", 1.0)
-    >> project.get("score")
+    >>> import skore
+    >>> project = skore.Project("my-xp")
+    >>> project.put("score", 1.0)
+    >>> project.get("score")
```
thomass-dev, line 51: This was intentionally done to avoid running this doctest.

glemaitre: In this case we need `# +SKIP`

thomass-dev: We can't skip an entire block with only one instruction :(

glemaitre: We would need the following package: https://github.com/scipy/scipy_doctest

I think it allows to skip for an entire block. Also it could be nice because it deals properly with the float numbers check.

sylvaincom: Should I create a separate issue for this?

sylvaincom: Done, here is the issue:

https://github.com/probabl-ai/skore/issues/1347

PR #1334, opened 2025-02-15
feat: Add config manager to manage progress bar

### LINKED ISSUE #1264: Allow me to turn off progress notification
### Describe the bug

Image

In this codeblock you can see that I am trying to wrap around `report.metrics.metrics_report()`. However, it always displays the progress bar. Can we turn this off? It makes sense when called directly but it is in the way when I try to build something that wraps around it.

### Steps/Code to Reproduce

Just call `.metrics.metrics_report` on the cross validation report.

### Expected Behavior

Have a flag to turn it off. 

### Actual Behavior

There is no such flag. 

### Environment

```shell
System:
    python: 3.12.6 (main, Sep  6 2024, 19:03:47) [Clang 15.0.0 (clang-1500.1.0.2.5)]
executable: /Users/vincent/Development/probabl/.venv/bin/python3
   machine: macOS-13.4.1-arm64-arm-64bit

Python dependencies:
        skore: 0.6.1
          pip: None
   setuptools: 75.8.0
    diskcache: 5.6.3
      fastapi: 0.115.7
       joblib: 1.4.2
   matplotlib: 3.10.0
        numpy: 2.2.2
       pandas: 2.2.3
       plotly: 5.24.1
      pyarrow: 19.0.0
         rich: 13.9.4
 scikit-learn: 1.6.1
        skops: 0.11.0
      uvicorn: 0.34.0
Selection deleted
```

glemaitre:
closes #1264 

Add a config management functions (i.e. `get_config`, `set_config` and `config_context`) to globally set up some behaviour. We first intend to disable the progress bar globally using this helper.

From scikit-learn, there are corner case to propagate such configuration in parallel processing. We therefore use the same pattern by implementing `Parallel` and `delayed` and propagate our configuration within those function.

### TODO

- [x] We needs to add a test for the `Parallel` and `delayed` functions
- [x] We could test the parallelisation at the same time as the configuration but the fact that `tests` is not a module, it cannot be pickle and therefore, we cannot really test it.

### DIFF
--- a/skore/src/skore/_config.py
+++ b/skore/src/skore/_config.py
@@ -2,7 +2,7 @@
 
 import threading
 import time
-from contextlib import contextmanager as contextmanager
+from contextlib import contextmanager
 
 _global_config = {
     "show_progress": True,
@@ -123,7 +123,8 @@ def config_context(
 def _set_show_progress_for_testing(show_progress, sleep_duration):
     """Set the value of show_progress for testing purposes after some waiting.
 
-    This function show exist in a Python module to be pickable.
+    This function should exist in a Python module rather than in tests, otherwise
+    joblib will not be able to pickle it.
     """
     with config_context(show_progress=show_progress):
         time.sleep(sleep_duration)
--- a/skore/src/skore/_config.py
+++ b/skore/src/skore/_config.py
@@ -1,6 +1,7 @@
 """Global configuration state and functions for management."""
 
 import threading
+import time
 from contextlib import contextmanager as contextmanager
 
 _global_config = {
@@ -117,3 +118,13 @@ def config_context(
         yield
     finally:
         set_config(**old_config)
+
+
+def _set_show_progress_for_testing(show_progress, sleep_duration):
+    """Set the value of show_progress for testing purposes after some waiting.
+
+    This function show exist in a Python module to be pickable.
+    """
+    with config_context(show_progress=show_progress):
+        time.sleep(sleep_duration)
+        return get_config()["show_progress"]
--- a/skore/src/skore/_config.py
+++ b/skore/src/skore/_config.py
@@ -49,8 +49,6 @@ def set_config(
 ):
     """Set global skore configuration.
 
-    .. versionadded:: 0.19
-
     Parameters
     ----------
     show_progress : bool, default=None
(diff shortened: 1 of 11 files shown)


skore/src/skore/_config.py
```diff
@@ -0,0 +1,130 @@
...
+    try:
+        yield
+    finally:
+        set_config(**old_config)
+
+
+def _set_show_progress_for_testing(show_progress, sleep_duration):
+    """Set the value of show_progress for testing purposes after some waiting.
+
+    This function show exist in a Python module to be pickable.
```
auguste-probabl, line 126:
```suggestion
    This function should exist in a Python module to be pickable.
```
I guess?

skore/tests/unit/test_config.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+@pytest.mark.parametrize("backend", ["loky", "multiprocessing", "threading"])
+def test_config_threadsafe_joblib(backend):
+    """Test that the global config is threadsafe with all joblib backends.
+    Two jobs are spawned and sets assume_finite to two different values.
+    When the job with a duration 0.1s completes, the assume_finite value
+    should be the same as the value passed to the function. In other words,
+    it is not influenced by the other job setting assume_finite to True.
+    """
```
auguste-probabl, line 96:
```suggestion
    """Test that the global config is threadsafe with all joblib backends.
    Two jobs are spawned and each sets `show_progress` to two different values.
    When the job with a duration of 0.1s completes, the `show_progress` value
    should be the same as the value passed to the function. In other words,
    it is not influenced by the other job setting `show_progress` to True.
    """
```

skore/tests/unit/test_config.py
```diff
@@ -0,0 +1,124 @@
...
+    show_progresses = [False, True, False, True]
+    sleep_durations = [0.1, 0.2, 0.1, 0.2]
+
+    with ThreadPoolExecutor(max_workers=2) as e:
+        items = [
+            output
+            for output in e.map(
+                _set_show_progress_for_testing, show_progresses, sleep_durations
+            )
+        ]
```
auguste-probabl, line 122:
```suggestion
        items = list(
            e.map(
                _set_show_progress_for_testing, show_progresses, sleep_durations
            )
        )
```

auguste-probabl (approved):
A very welcome change indeed!

The complexity is worth it.

PR #1336, opened 2025-02-15
feat: Add a method .style to displays

### LINKED ISSUE #1273: enh: displays should expose a higher `.style(...)` API
As discussed in https://github.com/probabl-ai/skore/issues/1261, we should implement a `.style` that could allow to setup the style of the plotting.

glemaitre:
closes #1273 

This PR is adding a method `.style` used to set the plotting style that can be reused each time the `display.plot()`.

### DIFF
--- a/skore/src/skore/sklearn/_plot/style.py
+++ b/skore/src/skore/sklearn/_plot/style.py
@@ -1,0 +1,48 @@
+class StyleDisplayMixin:
+    """Mixin to control the style plot of a display."""
+
+    @property
+    def _style_params(self):
+        """Get the list of available style parameters.
+
+        Returns
+        -------
+        list
+            List of style parameter names (without '_default_' prefix).
+        """
+        prefix = "_default_"
+        suffix = "_kwargs"
+        return [
+            attr[len(prefix) :]
+            for attr in dir(self)
+            if attr.startswith(prefix) and attr.endswith(suffix)
+        ]
+
+    def style(self, **kwargs):
+        """Set the style parameters for the display.
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Style parameters to set. Each parameter name should correspond to a
+            a style attribute passed to the plot method of the display.
+
+        Returns
+        -------
+        self : object
+            Returns the instance itself.
+
+        Raises
+        ------
+        ValueError
+            If a style parameter is unknown.
+        """
+        for param_name, param_value in kwargs.items():
+            default_attr = f"_default_{param_name}"
+            if not hasattr(self, default_attr):
+                raise ValueError(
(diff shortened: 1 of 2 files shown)


skore/src/skore/sklearn/_plot/style.py
```diff
@@ -0,0 +1,48 @@
...
+            If a style parameter is unknown.
+        """
+        for param_name, param_value in kwargs.items():
+            default_attr = f"_default_{param_name}"
+            if not hasattr(self, default_attr):
+                raise ValueError(
+                    f"Unknown style parameter: {param_name}. "
+                    f"The parameter name should be one of {self._style_params}."
+                )
+            setattr(self, default_attr, param_value)
```
auguste-probabl, line 47: I would rather avoid in-place operations if it can be helped. Can we make a copy of `self`?

auguste-probabl (changes requested):

skore/src/skore/sklearn/_plot/style.py
glemaitre: Making a copy could be annoying I think. A report will return a display and cache it. If we make a copy then, we don't anymore of the same object and then it would be weird.

I think it would be fine to change the attribute but maybe there is 2 things that are weird right now:

- the prefix `_default` introduce the concept that it will not change which is actually not what I meant
- `style` could be replace by `set_style` that might be a better name for letting us know that we are going to modify stuff in-place.

auguste-probabl: `set_style` would be a better name for sure if we go with this. Still, I would really prefer a declarative API with immutable objects à-la-Polars, even if there are performance costs.

glemaitre: When it comes to data I can get the point but here what is the benefit for a user to get immutable objects when it comes to plotting?

auguste-probabl: Plots are data -- everything is data!

More seriously, renaming to `set_style` is a good middle-ground.

skore/src/skore/sklearn/_plot/precision_recall_curve.py
auguste-probabl, line 1: Can we merge #1309 first?

glemaitre: Yep go ahead.

auguste-probabl, line 1: can you update this in light of #1309 ?

glemaitre: done

PR #1337, opened 2025-02-15
chore: Add typing to EstimatorReport and CrossValidationReport

glemaitre:
This PR introduces typing into the `EstimatorReport` and `CrossValidationReport` and remove the stubs that were used for that.


glemaitre:
I added typing mainly to the files that I touched in the past. At some point, it would make sense to have `mypy` if we want to enforce typing in a consistent manner.

skore/src/skore/__main__.py
auguste-probabl, line 1: Blocked by #1286

auguste-probabl: Please do the same for ComparisonReport :)

glemaitre:
Added the comparison report. I think that we should add this file quite quickly otherwise I'll spend a bit too much time rebasing :)

glemaitre:
Going to draft mode to actually solve the `mypy` problem right away.

It will cost some beers to the typing fan boys (@rouk1 ;))

glemaitre:
I think it is ready for review. I added `mypy` to the `pre-commit` to ensure that we write something consistent.

Basically, I only added some slight changes like `assert` to help `mypy` (which in practice I dislike). But the code and tests did not change indeed.

glemaitre:
ping @thomass-dev @auguste-probabl @rouk1 Could we merge this asap such that subsequent PR will type checking.

skore/src/skore/cli/color_format.py
```diff
@@ -97,15 +111,16 @@ def format_help(self):
...
     """Custom argument parser for the CLI."""
 
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs, formatter_class=RichColorHelpFormatter)
+    def __init__(self, *args: Any, **kwargs: Any):
+        kwargs["formatter_class"] = RichColorHelpFormatter
+        super().__init__(*args, **kwargs)
 
-    def print_help(self, file=None):
+    def print_help(self, file: Any = None) -> None:
```
auguste-probabl, line 118:
```suggestion
    def print_help(self, file: Any = None) -> NoReturn:
```
or
```suggestion
    def print_help(self, file: Any = None):
```
I guess?

skore/src/skore/sklearn/_base.py
```diff
@@ -27,28 +32,41 @@ def _get_methods_for_help(self):
...
-    def _create_help_panel(self):
+    @abstractmethod
+    def _create_help_tree(self) -> Tree:
+        """Create the help tree."""
+        pass
+
+    @abstractmethod
+    def _get_help_panel_title(self) -> str:
+        """Get the help panel title."""
+        pass
```
auguste-probabl, line 65:
```suggestion
    def _create_help_tree(self) -> Tree:
        """Create the help tree."""

    @abstractmethod
    def _get_help_panel_title(self) -> str:
        """Get the help panel title."""
```

auguste-probabl (dismissed):
Alright, works for me

skore/src/skore/cli/color_format.py
auguste-probabl: Would be nice to have a consistent way of typing functions that don't return anything -- I don't really mind which one between, `-> None`, `-> NoReturn` and nothing

rouk1 (dismissed):
Nice work ! I'm sure this will help catch bug and give confidence to refactor.

skore/src/skore/cli/color_format.py
glemaitre: I have to check but apparently the rule is to return `None` when the function do not have any return but leave the function and `NoReturn` if it does not leave (e.g. `raise` an error inside).

glemaitre: I'll make a pass having this rule in mind then.

glemaitre:
> Nice work ! I'm sure this will help catch bug and give confidence to refactor.

Yep the only thing that I don't like is the fact that you might need to have some `assert` when returning different type which is common when you have a common private API. But let's hope for the best :)

skore/src/skore/cli/color_format.py
glemaitre: I checked and the `None` for the moment are each time for the first case (implicit `return None`). So we should be good to go.

auguste-probabl:
> The comments are addressed. The next step would be to call `mypy` in a `--strict` mode. I have to check what it involves :)

@rouk1 has experience with this, as far as I remember strict mode is generally considered *too* strict for most purposes

rouk1:
> > The comments are addressed. The next step would be to call `mypy` in a `--strict` mode. I have to check what it involves :)
> 
> @rouk1 has experience with this, as far as I remember strict mode is generally considered _too_ strict for most purposes

Not that much to be honest. I know it can be painful especially when you heavily rely on untyped dependencies.

Issue #1320, opened 2025-02-14
Feat(EstimatorReport): Display the feature weights for linear models


MarieSacksick:
### Is your feature request related to a problem? Please describe.

As a Data Scientist, to explain my model and understand the problem I'm trying to solve, I need to check the feature importance by looking at its weights. This should be available linear models only.

### Describe the solution you'd like

```python
df = report.feature_importance.model_weights() # renders a dataframe
display = report.feature_importance.plot.model_weights() # renders a display
display.plot()
```

### Describe alternatives you've considered, if relevant

Later, if the object report contains too many accessors, we will group the feature importance and add a parameter to decide which of the feature importance type we want to display.

### Additional context

part of epic #1314

auguste-probabl:
> report.feature_importance.model_weights() # renders a dataframe
> report.feature_importance.model_weights() # renders a display

There is one method for both use-cases, I guess that's a typo?

Currently you write `report.metrics.[...]`, should we make a separate `feature_importance` accessor or write `report.metrics.feature_importance.model_weights`?

MarieS-WiMLDS:
thanks for the typo, it was everywhere, it's fixed now.  
Yes, it should be a separate accessor.

sylvaincom:
By "linear models" (https://scikit-learn.org/stable/modules/linear_model.html), which models do you mean exactly?

FYI, for a very first iteration, we dealt with LinearRegression, Ridge, and Lasso.

MarieS-WiMLDS:
So cool 🤩   
I was thinking about all these models: https://scikit-learn.org/stable/api/sklearn.linear_model.html. 
I don't know if there is a private method in scikit-learn to recognize them?

sylvaincom:
https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#sphx-glr-auto-examples-inspection-plot-linear-model-coefficient-interpretation-py

We need to add a check to see if the data is normalized or not, otherwise it makes no sense to interpret the coefficients

If the input estimator is a pipeline, and in the pipeline there is a scaler, then it is good.
If the input estimator is just an estimator (no pipeline) or if the pipeline has no scaler, then we must check if the input data is scaled or not.

@glemaitre do you know if there is an internal in scikit-learn to check if the input data is normalized?

sylvaincom:
It would also be nice to have this kind of pandas styling:

[...]

PR #1339, opened 2025-02-17
feat: Display the model coefficients for linear models

sylvaincom:
Introduces a new accessor on `EstimatorReport`, `feature_importance`, which currently has one method, `coefficients`, only available for estimators with a `coef_` attribute (such as linear models).

Demo:
```python
from sklearn.linear_model import LinearRegression
X, y = make_regression(random_state=42)
estimator = LinearRegression().fit(X_train, y_train)
report = EstimatorReport(estimator, X_test=X_test, y_test=y_test)
report.help()
╭─────────────────── Tools to diagnose estimator LinearRegression ───────────────────╮
│ EstimatorReport                                                                      │
│ ├── .metrics                                                                         │
│ │   ├── .accuracy(...)         (↗︎)     - Compute the accuracy score.                 │
│ │   ├── .brier_score(...)      (↘︎)     - Compute the Brier score.                    │
│ │   ├── .log_loss(...)         (↘︎)     - Compute the log loss.                       │
│ │   ├── .precision(...)        (↗︎)     - Compute the precision score.                │
│ │   ├── .precision_recall(...)         - Plot the precision-recall curve.            │
│ │   ├── .recall(...)           (↗︎)     - Compute the recall score.                   │
│ │   ├── .roc(...)                      - Plot the ROC curve.                         │
│ │   ├── .roc_auc(...)          (↗︎)     - Compute the ROC AUC score.                  │
│ │   ├── .custom_metric(...)            - Compute a custom metric.                    │
│ │   └── .report_metrics(...)           - Report a set of metrics for our estimator.  │
│ ├── .feature_importance                                                              │
│ │   └── .coefficients(...)             - Report the coefficients of a regression     │
│ │       estimator.                                                                   │
│ ├── .cache_predictions(...)            - Cache estimator's predictions.              │
│ ├── .clear_cache(...)                  - Clear the cache.                            │
│ └── Attributes                                                                       │
│     ├── .X_test                                                                      │
│     ├── .y_test                                                                      │
│     ├── .estimator_                                                                  │
│     └── .estimator_name_                                                             │
│                                                                                      │
│                                                                                      │
│ Legend:                                                                              │
│ (↗︎) higher is better (↘︎) lower is better                                             │

report.feature_importance.coefficients()
```

Note what happens if the estimator is not a regressor (e.g. a LogisticRegression):
```python
report.feature_importance.help()
╭─ Available feature importance methods ─╮
│ report.feature_importance              │
╰────────────────────────────────────────╯
```

---

Close https://github.com/probabl-ai/skore/issues/1320

Co-authored-by: @auguste-probabl <auguste@probabl.ai>


skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1649,3 +1649,61 @@ def prediction_error(
             display_class=PredictionErrorDisplay,
             display_kwargs=display_kwargs,
         )
+
+    @available_if(_check_is_linear_regression)
+    def model_weights(self):
```
glemaitre, line 1654: We want to have a new accessor called `feature_importance` and not have this one in the `metrics`. So you will need to have a new file `feature_importance_accessor.py`

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1649,3 +1649,61 @@ def prediction_error(
             display_class=PredictionErrorDisplay,
             display_kwargs=display_kwargs,
         )
+
+    @available_if(_check_is_linear_regression)
```
glemaitre, line 1653: we should not use `linear_regression` because it is a specific name for a specific model.

Instead, we should duck-type and look for the `coef_` attribute.

skore/src/skore/utils/_accessor.py
```diff
@@ -13,3 +13,15 @@ def check(accessor):
         return True
 
     return check
+
+
+def _check_is_linear_regression(accessor):
+    """Check if the estimator is a linear regression, Ridge, or Lasso."""
+    from sklearn.linear_model import Lasso, LinearRegression, Ridge
+
+    if isinstance(accessor._parent.estimator_, (LinearRegression, Ridge, Lasso)):
```
glemaitre, line 22: We need to be much more permissive. If someone pass a `Pipeline` then this code is not going to work because `instance` will be to strict. However, the weights are usually defined by exposing a `coef_` parameter. So we need to check for that.

Ultimately, we should not restrain ourself to regression only but we should be able to show the coefficients of classifier like `LogisticRegression` or `RidgeClassifier`.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1649,3 +1649,61 @@ def prediction_error(
...
+        """
+        reg = self._parent.estimator_
+
+        feature_names = (
+            reg.feature_names_in_
+            if hasattr(reg, "feature_names_in_")
+            else range(reg.n_features_in_)
+        )
+
+        intercept = reg.intercept_
```
glemaitre, line 1699: There is an option `fit_intercept=True/False` that will change the behaviour of this thing here.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1649,3 +1649,61 @@ def prediction_error(
...
+        coef = reg.coef_.flatten()
```
glemaitre, line 1700: It does not look right to flatten. When coming to the classifier it will be problematic.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1649,3 +1649,61 @@ def prediction_error(
...
+        8           250.535095
+        9            99.577694
+        """
+        reg = self._parent.estimator_
+
+        feature_names = (
+            reg.feature_names_in_
+            if hasattr(reg, "feature_names_in_")
+            else range(reg.n_features_in_)
+        )
```
glemaitre, line 1697: You might want to have a look at https://github.com/glemaitre/skore/blob/cross_validate_reporter/skore/src/skore/sklearn/_plot.py#L254-L261

to handle the pipeline.

By default I would also create string such that we do have to index with interger

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -1180,3 +1180,72 @@ def test_estimator_has_no_deep_copy():
...
+    result = report.metrics.model_weights()
+
+    expected = pd.DataFrame(
+        data=[
+            -1.776357e-15,
+            6.459172e01,
+            9.865152e01,
+            5.707783e01,
+            6.057748e01,
+            3.560967e01,
```
glemaitre, line 1236: Instead of hard coded the number, it is better to get them from the estimator. It shows in the test itself, what to expect in terms of where the weights come from

glemaitre (commented):
We should anticipate how hard is to do the plotting with only pandas. It is to decide whether or not we should have a display.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: Yes this is planned

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: Yes, we know. For now we went for the simple option first because `coef_` can have different types depending on the model (e.g. for LogisticRegression it's a 2-d array).

skore/src/skore/utils/_accessor.py
auguste-probabl: See above. Point taken about Pipeline though.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: When `fit_intercept=False` the intercept is still defined, and set to 0.0. We thought about adding some kind of note when this is the case, but the UX is not done yet.

skore/src/skore/utils/_accessor.py
glemaitre: A better way would therefore be:

```python
estimator = accessor._parent.estimator_
if is_regressor(estimator) and hasattr(estimator, "coef_"):
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: We will have to jungle I assume here because I could expect scikit-learn to not have the best consistent API, e.g. https://github.com/scikit-learn/scikit-learn/issues/14473

skore/src/skore/utils/_accessor.py
glemaitre: I was looking quickly, and I think that we can manage the multioutput regression easily:

image

would provide the following plot:

image

Looking at the results, I think that we could really default plotting sytle and thus have a display, e.g. add axis, remove the spine, handle the color.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
sylvaincom: done

skore/src/skore/sklearn/_estimator/metrics_accessor.py
sylvaincom: We now have a `_check_is_regressor_coef_task`

skore/src/skore/utils/_accessor.py
sylvaincom: should be good now

skore/src/skore/sklearn/_estimator/metrics_accessor.py
sylvaincom: should be good now

skore/tests/unit/sklearn/test_estimator.py
sylvaincom: should be ok now

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: > By default I would also create string such that we do have to index with interger

What do you mean by this?

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: we use `numpy.atleast_2d` which seems to work quite well. Maybe it will also work for LogisticRegression models

auguste-probabl: Checked locally and it works fine -- but maybe it doesn't make sense. I'll keep my change local for now.

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+
+    def __init__(self, parent: Any) -> None:
+        super().__init__(parent)
+
+    @available_if(_check_is_regressor_coef_task)
+    def model_weights(self) -> pd.DataFrame:
+        """Report the coefficients of a regression estimator.
+
+        Only works for LinearRegression, Ridge, and Lasso scikit-learn estimators.
```
sylvaincom, line 26: Is this still true?

auguste-probabl: If you refer to whether it works for `LogisticRegression`, that's still at the experimental stage until @glemaitre or @MarieSacksick gives their opinion.

MarieSacksick: Sorry I missed it, where do you need me?

auguste-probabl: Currently `model_weights` is only available for *regressors*, whereas the logic technically works for any *linear* model, e.g `LogisticRegression`. However, just because it's technically possible doesn't mean it makes sense, and here my expertise reaches its limits.

MarieSacksick: It doesn't have the same meaning as in regressors, but yes it makes sense (on some conditions, that we should check when we implement the warnings :D).

MarieSacksick: you can find some theoretical elements here: https://christophm.github.io/interpretable-ml-book/logistic.html.  
I don't know how we would do for multiclass though. I would output a feature importance df for each target I guess. I saw some people doing average in some blogs, but it seems like an aweful solution.

sylvaincom:
We have a toy example for example-driven dev at `examples/model_evaluation/plot_feature_importance.py`

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,92 @@
...
+
+# %%
+df_model_weights = linear_regression_report_names.feature_importance.model_weights()
+df_model_weights.style.bar(align="mid", color=["#d65f5f", "#5fba7d"]).format(
+    precision=2
+)
+
+# %%
+# .. warning::
+#   a nice visualization of model weights requires scaling!
```
MarieSacksick, line 77:
```suggestion
#   For a correct interpretation of model weights, the features must have been through scaling.
```

MarieSacksick (commented):
I checked /examples/model_evaluation/plot_feature_importance.py and from this, it seems nice to me!  
I didn't check the code nor several different models.

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
sylvaincom: @auguste-probabl, I mean there are other linear models than just LinearRegression, Ridge, and Lasso

See full list at https://scikit-learn.org/stable/modules/linear_model.html

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+        intercept = np.atleast_2d(estimator.intercept_)
+        coef = np.atleast_2d(estimator.coef_)
+
+        data = np.concatenate([intercept, coef.T])
+
+        df = pd.DataFrame(
+            data=data,
+            index=["Intercept"] + list(feature_names),
```
sylvaincom, line 80: We include the intercept, but is it relevant for model inspection?

sylvaincom:
@thomass-dev do you know why the documentation preview does not include the new example? 🙏 Is it because the example was added after the documentation preview was generated?

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: @MarieSacksick so if I understand correctly our criteria for which estimators are allowed are:
- Has `coef_` attribute
- If a classifier, is a *binary* classifier

@sylvaincom Can you be more explicit? Are you saying "just because it's a linear model doesn't mean we should expose the `model_weights`"?

MarieSacksick: Our criteria for estimators allowed is that they are linear; it seems that this condition is equal to having a `coef_` attribute in scikit-learn indeed.  

We can allow multiclass use-cases. I saw that you started to output a dataframe with several columns, and I think that it's a good idea!  
image

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
MarieSacksick: I think that it is relevant for model inspection yes, it provides info about the value outputed by the model for an element with average value for all features. It's not really feature importance per se, but I would keep it. Maybe later we can add a helper to read it, depending on the model.

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: @MarieSacksick That output is for multi-output regression. I'm not sure if it would translate cleanly to multiclass classification.

MarieSacksick: after-thought: let's drop the multiclass for now. We can iterate to improve this feature after.

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
sylvaincom: Ok, thanks

sphinx/conf.py
```diff
@@ -70,6 +70,7 @@
     "../examples/model_evaluation",
     "../examples/skore_project",
     "../examples/technical_details",
+    "../examples/example_driven_dev",
```
MarieSacksick, line 73: This should be removed, otherwise the example will land in the documentation.

MarieSacksick (changes requested):
Only one comment, otherwise it's good for me!

examples/example_driven_dev/plot_feature_importance.py
```diff
@@ -0,0 +1,121 @@
+"""
+.. _example_feature_importance:
+
+=============================================
+`EstimatorReport`: Get the feature importance
+=============================================
+
+.. warning::
```
glemaitre, line 8: So for the final example, I think that we can create a use-case example on "how to interpret your model". It will be of course small at the first round because we have only linear model but we can go quite far, e.g. https://scikit-learn.org/dev/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#sphx-glr-auto-examples-inspection-plot-linear-model-coefficient-interpretation-py

sylvaincom: Yes! We keep that scikit-learn example in mind!
For merging purposes, I'll remove `examples/example_driven_dev/plot_feature_importance.py`

sphinx/conf.py
sylvaincom: Should be ok now

glemaitre (commented):
I just started by reviewing the example.

Now I'm thinking that `.model_weights()` should return a display. The plan would be that this display has `.plot()` to make the right plotting for the user, as `.frame` property that return the current dataframe.

In addition, I like the current direction for the pandas style and I think that we can make it the default HTML repr of the display. One issue with the style is that it does not return a real dataframe that you can manipulate. That's why, I think that we can use it in the `repr` since it is not an object that is used by the user afterwards and we still have the `.frame` in case someone wants to access the real dataframe.

examples/example_driven_dev/plot_feature_importance.py
glemaitre: Yep, it is great during this development phase.

glemaitre (commented):
We need to extend the test and check what happens when passing a classifier. It tells me that somehow, we might want to validate the shape of `coef_` and `intercept_` to something that we know how to handle and raise an error otherwise.

examples/example_driven_dev/plot_feature_importance.py
```diff
@@ -0,0 +1,121 @@
...
+
+# %%
+estimator_report.help()
+
+# %%
+estimator_report.feature_importance.model_weights()
+
+# %%
+# .. note::
+#   Does it make sense to include the ``interpret`` for model inspection?
```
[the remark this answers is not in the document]
MarieSacksick: I think it's intercept, rather than interpret.

examples/example_driven_dev/plot_feature_importance.py
```diff
@@ -0,0 +1,121 @@
...
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+
+estimator_report.feature_importance.model_weights()
+
+# %%
+df_model_weights = estimator_report.feature_importance.model_weights()
+df_model_weights.style.bar(align="mid", color=["#d65f5f", "#5fba7d"]).format(
```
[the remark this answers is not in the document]
MarieSacksick: For the styling, we can improve it to be a prettier example, but it's not meant to be dealt with in this PR. Let's keep this PR focused to feature importance for linear models, removing the more complex use case of multi-classification. We ca iterate later. The goal is to avoid to start a tunnel, having long reviews, and planning drift.

skore/src/skore/utils/_accessor.py
```diff
@@ -16,3 +18,19 @@ def check(accessor: Any) -> bool:
...
+        parent_estimator.steps[-1][1]
+        if isinstance(parent_estimator, Pipeline)
+        else parent_estimator
+    )
+    if hasattr(estimator, "coef_"):
+        return True
+    raise AttributeError(
+        f"Estimator {accessor._parent.estimator_} is not a supported estimator by "
+        "the function called."
+    )
```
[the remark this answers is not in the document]
auguste-probabl: The argument of `available_if` needs to take `self` as argument and return a boolean, so I think we're good

examples/example_driven_dev/plot_feature_importance.py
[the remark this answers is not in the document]
auguste-probabl: i guess it should read ``intercept``?

auguste-probabl:
> We need to extend the test and check what happens when passing a classifier. It tells me that somehow, we might want to validate the shape of `coef_` and `intercept_` to something that we know how to handle and raise an error otherwise.

We do pass a `LogisticRegression`.  Our strategy for dealing with the non-standardized shapes of `coef_` and `intercept_` has been to use `np.atleast_2d`, which seems to work pretty well.

examples/example_driven_dev/plot_feature_importance.py
```diff
@@ -0,0 +1,121 @@
...
+# ------------------
+
+# %%
+X, y = make_regression(n_targets=2, random_state=0)
+X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0, shuffle=False)
+
+estimator_report = EstimatorReport(
+    LinearRegression(), X_train=X_train, X_test=X_test, y_train=y_train, y_test=y_test
+)
+estimator_report.feature_importance.model_weights()
```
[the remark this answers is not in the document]
auguste-probabl: @MarieSacksick thoughts on this one?

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -1193,3 +1193,220 @@ def test_estimator_has_no_deep_copy():
...
+            LinearRegression(),
+            pd.DataFrame(
+                data=[
+                    -1.776357e-15,
+                    6.459172e01,
+                    9.865152e01,
+                    5.707783e01,
+                    6.057748e01,
+                    3.560967e01,
+                ],
```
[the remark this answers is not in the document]
auguste-probabl: done, we now ignore the data and check only shape/names

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: Made it work for logisticRegression.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: @glemaitre ping

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: Closing as these points are discussed in later comments

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -0,0 +1,105 @@
...
+            columns=(
+                [f"Target #{i}" for i in range(data.shape[1])]
```
glemaitre, line 79: I currently checking more but for classifier, we will have `(n_features,)` or `(n_classes, n_features)`. So we would need to transpose in the second case. Instead of "Target", we should have "Class ..." or the name of of the classes picked up in `estimator.classes_`.

Where we need to search more is on the visualization and the meaning of the weights: a `LogisticRegression` will show the log odd-ratio if I am not wrong why the representation of the weights for a `RidgeClassifier` will have some other meaning. So I'm wondering how to convey this information.

In any case, plotting the weights is OK in both cases. But it will be useful to have a better axis labelling.

examples/example_driven_dev/plot_feature_importance.py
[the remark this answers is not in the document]
glemaitre: OK this make sense then :) I was completely side ways.

examples/example_driven_dev/plot_feature_importance.py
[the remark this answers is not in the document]
glemaitre: So by looking a bit more I think that the dataframe should column named "Target ..." or "Classes ..." and we don't need to mention anything about "Coefficients"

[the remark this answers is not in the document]
MarieSacksick: It's true that I use both without thinking much about it. Let's try to have something uniform indeed.  
In scikit-learn, the term `coefficient` is more used, and `weights` refer more to the importance of a sample.  
To stick to it --> let's use coefficients.  
Therefore the API would become: `estimator_report.feature_importance.coefficients()`

glemaitre:
> We have a test that passes a LogisticRegression. Our strategy for dealing with the non-standardized shapes of coef_ and intercept_ has been to use np.atleast_2d, which seems to work pretty well.

Your test does not check for multiclass. As mentioned https://github.com/probabl-ai/skore/pull/1339#discussion_r1969466598, I think that the transpose in this case will do the trick.

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -1193,3 +1193,184 @@ def test_estimator_has_no_deep_copy():
...
+                    "Feature #4",
+                ],
+                columns=["Target #0", "Target #1", "Target #2"],
+            ),
+        ),
+    ],
+)
+def test_estimator_report_model_weights_numpy_arrays(data, estimator, expected):
+    X, y = data
+    X_train, _, y_train, _ = train_test_split(X, y, test_size=0.2, random_state=42)
```
glemaitre, line 1308: Actually for the test, we don't really care about the `train_test_split`. You could remove the line.

glemaitre: Here and below.

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -1193,3 +1193,184 @@ def test_estimator_has_no_deep_copy():
...
+                    "Feature #2",
+                    "Feature #3",
+                    "Feature #4",
+                ],
+                columns=["Target #0", "Target #1", "Target #2"],
+            ),
+        ),
+    ],
+)
+def test_estimator_report_model_weights_numpy_arrays(data, estimator, expected):
```
glemaitre, line 1306: I would right specific tests for classification and classification where the idea would be to test all known scikit-learn models that exposes the `coef_` and `intercept_` to be sure that in the case of regression (single and multiouput) and classification (single and multioutput). If we succeed to pass all of those, we have a good chance to accept third-party models.

examples/example_driven_dev/plot_feature_importance.py
glemaitre: Yes they can be resolved. But we will need an example in this PR also :). In terms of documentation, we need to add this functionality to the API documentation.

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -1193,3 +1193,184 @@ def test_estimator_has_no_deep_copy():
...
+    report = EstimatorReport(estimator)
+    result = report.feature_importance.model_weights()
+
+    assert_frame_shape_equal(result, expected)
+
+
+def test_estimator_report_model_weights_pandas_dataframe():
+    """If provided, the model weights dataframe uses the feature names, where the
+    estimator is a single estimator (not a pipeline)."""
+    X, y = make_regression(n_features=5, random_state=42)
```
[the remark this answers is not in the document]
auguste-probabl: Done for the regression dataset -- the classificatio dataset isn't used enough for it to make sense imo

skore/tests/unit/sklearn/test_estimator.py
auguste-probabl: done

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: We already transpose currently, so it seems that the current code works properly. I'll add a test for multiclass-classification

auguste-probabl: done

skore/src/skore/utils/_accessor.py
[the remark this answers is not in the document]
auguste-probabl: changed `_check_has_coef` to return a callable -- although to be honest I didn't manage to understand what could go wrong.

examples/example_driven_dev/plot_feature_importance.py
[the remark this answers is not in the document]
auguste-probabl: renamed `model_weights` to `coefficients`

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
glemaitre: Cool, so could we name the column target for regression and classes in the case of classification. For binary classification, I assume that we will show the feature importance related to the log odds ratios of the positive outcome which is `estimator.classes_[1]`.

examples/example_driven_dev/plot_feature_importance.py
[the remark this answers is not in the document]
auguste-probabl: And for the column names, here is what I understand from the responses:
- if multi-output regression, `Target #...`
- if multi-class classification, `Class #...`
- if single-output, we have several options:
    - Keep "Coefficients"
    - Empty string
    - Empty string, and we return a Series instead of a DataFrame 

Can you settle this @MarieSacksick ?

skore/tests/unit/sklearn/test_estimator.py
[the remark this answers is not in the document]
glemaitre: Good with me.

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: this is covered in another comment, so please resolve if everything else is done here

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -1193,3 +1193,201 @@ def test_estimator_has_no_deep_copy():
...
+                columns=["Target #0", "Target #1", "Target #2"],
+            ),
+        ),
+    ],
+)
+def test_estimator_report_coefficients_numpy_arrays(
+    request, data, estimator, expected
+):
+    if isinstance(data, str):
+        data = request.getfixturevalue(data)
```
glemaitre, line 1333: Ok, in this case I would not data the fixture to avoid this line. I did not think about it before.

skore/tests/unit/sklearn/test_estimator.py
```diff
@@ -1193,3 +1193,201 @@ def test_estimator_has_no_deep_copy():
...
+    X, y = data
+
+    estimator.fit(X, y)
+
+    report = EstimatorReport(estimator)
+    result = report.feature_importance.coefficients()
+
+    assert_frame_shape_equal(result, expected)
```
glemaitre, line 1341: What I meant was maybe close to something like:


```python
@pytest.mark.parametrize(
    "data, estimator",
    [
        (
            make_regression(n_features=5, random_state=42),
            LinearRegression()
        ),
        (
            make_classification(n_features=5, random_state=42),
            LogisticRegression()
        ),
        (
            make_regression(n_features=5, random_state=42),
            Pipeline([("scaler", StandardScaler()), ("reg", LinearRegression())])
        ),
        (
            make_regression(n_features=5, n_targets=3, random_state=42),
            LinearRegression()
        ),
    ],
)
def test_estimator_report_model_weights_numpy_arrays(data, estimator):
    X, y = data
    X_train, _, y_train, _ = train_test_split(X, y, test_size=0.2, random_state=42)

    estimator.fit(X_train, y_train)
    report = EstimatorReport(estimator)
    result = report.feature_importance.model_weights()

    n_features = X.shape[1]
    n_targets = y.shape[1] if len(y.shape) > 1 else 1
    expected_rows = n_features + 1  # +1 for intercept
    expected_cols = n_targets

    assert result.shape == (expected_rows, expected_cols)

    expected_index = ["Intercept"] + [f"Feature #{i}" for i in range(n_features)]
    assert result.index.tolist() == expected_index

    if n_targets == 1:
        assert result.columns.tolist() == ["Coefficient"]
    else:
        expected_columns = [f"Target #{i}" for i in range(n_targets)]
        assert result.columns.tolist() == expected_columns
 ```
 
 It is rough generation to show that I don't need to check the entire frame but just the column and index. If we are consistent with column name then it can be simplified most probably. We could probably pass the expected name of columns and shape in the parametrization. So roughly something like:
 
 
```python
@pytest.mark.parametrize(
    "data, estimator, column_base_name, expected_shape",
    [
        (
            make_regression(n_features=5, random_state=42),
            LinearRegression(),
            "Target",
            (6, 1),
        ),
        (
            make_classification(n_features=5, random_state=42),
            LogisticRegression(),
            "Class",
            (6, 1),
        ),
        (
            make_classification(
                n_features=5, n_targets=3, n_clusters_per_class=1, random_state=42
            ),
            LogisticRegression(),
            "Class",
            (6, 3),
        ),
        (
            make_regression(n_features=5, random_state=42),
            Pipeline([("scaler", StandardScaler()), ("reg", LinearRegression())]),
            "Target",
            (6, 1),
        ),
        (
            make_regression(n_features=5, n_targets=3, random_state=42),
            LinearRegression(),
            "Target",
            (6, 3),
        ),
    ],
)
def test_estimator_report_model_weights_numpy_arrays(
    data, estimator, column_base_name, expected_shape
):
    X, y = data
    estimator.fit(X, y)
    report = EstimatorReport(estimator)
    result = report.feature_importance.model_weights()
    assert result.shape == expected_shape

    expected_index = ["Intercept"] + [f"Feature #{i}" for i in range(X.shape[1])]
    assert result.index.tolist() == expected_index

    expected_columns = [f"{column_base_name} #{i}" for i in range(y.shape[1])]
    assert result.columns.tolist() == expected_columns
```

examples/example_driven_dev/plot_feature_importance.py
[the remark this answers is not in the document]
MarieSacksick: `Coefficients` is good!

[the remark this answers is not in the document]
glemaitre: If we go with "Coefficients", then we need a multi-index for the for the multi-output because they the coefficients related to a specific target or class.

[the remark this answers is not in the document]
glemaitre: What I mean is the following:

```
import numpy as np
import pandas as pd

coef_single_output = np.random.randn(5, 1)
coef_multi_output = np.random.randn(5, 2)

df_single_output = pd.DataFrame(
    coef_single_output,
    columns=["Coefficients"],
    index=[f"Feature #{i}" for i in range(5)]
)
df_multi_output = pd.DataFrame(
    coef_multi_output,
    columns=pd.MultiIndex.from_tuples([
        ("Coefficients", "Target #1"),
        ("Coefficients", "Target #2"),
    ]),
    index=[f"Feature #{i}" for i in range(5)]
)
```

Resulting to:

image

image

I think that we should make kind of understandable that the coefficients are with regards to something; the target or the probability of a certain class. In binary class, it is kind of straightforward.

skore/tests/unit/sklearn/test_estimator.py
auguste-probabl: done

skore/tests/unit/sklearn/test_estimator.py
auguste-probabl: done, thanks!

examples/example_driven_dev/plot_feature_importance.py
sylvaincom: Just added a proper example (not for example driven dev, it is meant to stay)

skore/tests/unit/sklearn/test_estimator.py
auguste-probabl: I've made a first attempt. Things to note:
- I only try single-output tasks for now
- About 10 estimators give me strange errors -- I'll try them again when I have the time to go through the docs for each of them.
- I found the estimators by searching for `coef_` in the sklearn docs. Let me know if I have missed some or if some should not be there.

examples/example_driven_dev/plot_feature_importance.py
auguste-probabl: API docs have been updated as well

glemaitre:
Regarding the code, I'm fine with the current set of feature for a first iteration apart from having the multindex such that Coefficients always appear and thus introduce the flat_index parameter. 

In a subsequent PR I think we need to have a display but we will need the refactor first. So we can jump on the MDI with trees.

MarieSacksick (approved):
Thanks :) !

auguste-probabl:
> apart from having the multindex

It has been decided that we stick to single-index until further notice: https://github.com/probabl-ai/skore/pull/1339#discussion_r1970173002

skore/tests/unit/sklearn/test_estimator.py
auguste-probabl: In the interest of time I'll consider this todo in the next iteration

auguste-probabl: see #1364

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,94 @@
...
+estimator_report = EstimatorReport(
+    LinearRegression(),
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+estimator_report.metrics.report_metrics()
+
+# %%
```
glemaitre, line 59: Here, I would expect an analysis of the results. Indeed, the R2 score is not tremendous so any analysis that we have after is depending of the quality of the model. So here, it is is kind of problematic. Could we have some preprocessing that make the model perform better?

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,94 @@
...
+# some metrics:
+
+# %%
+from skore import train_test_split, EstimatorReport
+from sklearn.linear_model import LinearRegression
+
+X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)
+
+estimator_report = EstimatorReport(
+    LinearRegression(),
```
glemaitre, line 51: Do not use a `LinearRegression` but instead a `Ridge`.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,94 @@
...
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+estimator_report.metrics.report_metrics()
```
glemaitre, line 57: We should describe the numbers and what they mean.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,94 @@
...
+
+# %%
+# Now, let us apply the :class:`~skore.EstimatorReport` on a linear regression and get
+# some metrics:
+
+# %%
+from skore import train_test_split, EstimatorReport
+from sklearn.linear_model import LinearRegression
+
+X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)
```
glemaitre, line 48: It looks to me that the warning is a false positive. Since it is shown, we have 2 ways, either we explain why it is a problem and we still continue our example or we don't use the function because we don't have any methodological issue.

Here, I don't think that there is any ordering of i.i.d issue and I would use scikit-learn to make the train-test split.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,94 @@
...
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+scaled_estimator_report.feature_importance.coefficients()
+
+# %%
+# Now, the obtained coefficients can be properly interpreted.
+# After scaling, it appears that the ``HouseAge`` feature is actually quite important.
```
glemaitre, line 87: Instead of mentioning that `HouseAge` is more important, it is rather that the coefficients are in the same range and can be easier to be compared between them. However, now we have no clue regarding the interpretation of weight regarding the original unit. So in terms of interpretation it is quite important to mention this point.

glemaitre (commented):
OK @sylvaincom I made a pass on the example. The first and second messages are kind of outdated and I did a more detailed pass.

examples/example_driven_dev/plot_feature_importance.py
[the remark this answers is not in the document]
glemaitre: > A compromise could be to name the columns Coefficients for target #1, but I'm not convinced as it's quite long

It is exactly the job of the `flat_multi_index` that would create `coefficients_target_1` when you are in a flat setting to make it easy to index.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,94 @@
...
+models and look at some scores that are automatically computed for you.
+
+In practice, once you have fitted an estimator, you want to evaluate it, and
+you also want to inspect it: you can do that with the estimator report's
+:meth:`~skore.EstimatorReport.feature_importance` accessor that we showcase in this
+example.
+"""
+
+# %%
+# Linear models: inspecting the coefficients
```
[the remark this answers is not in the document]
sylvaincom: If I take the employee salary dataset, I use a skrub TableVectorizer?

examples/model_evaluation/plot_feature_importance.py
sylvaincom: done, used scikit-learn's
I believe we will re-do the warnings soon either way

examples/model_evaluation/plot_feature_importance.py
sylvaincom: done

examples/model_evaluation/plot_feature_importance.py
sylvaincom: should be good now

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
+"""
+.. _example_feature_importance:
+
+=====================================================================
+`EstimatorReport`: Inspecting your models with the feature importance
+=====================================================================
+
+In this example, we tackle the California housing dataset where the goal is to perform
+a regression task: predicting house prices based on the number of bedrooms, etc.
```
MarieSacksick, line 9:
```suggestion
a regression task: predicting house prices based on features describing the house, such as the number of bedrooms.
```

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
...
+X = X_y.drop(columns=[target_name])
+y = X_y[target_name]
+X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)
+
+# %%
+# Linear models
+# =============
+
+# %%
+# For our regression task, let us start with linear models.
```
MarieSacksick, line 176: A bit redundant with line 10: https://github.com/probabl-ai/skore/blob/a577cad91663adfb61dc9b66e7fb220aac97339a/examples/model_evaluation/plot_feature_importance.py#L10? And with line 183 too https://github.com/probabl-ai/skore/blob/a577cad91663adfb61dc9b66e7fb220aac97339a/examples/model_evaluation/plot_feature_importance.py#L183?

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
...
+# Before trying any complex model, we start with a simple linear model to have a
+# baseline of what a "good score" is (remember that all scores are relative).
+# Here, we use a Ridge regression along with some scaling and evaluate it using
+# :meth:`skore.EstimatorReport.metrics`:
+
+# %%
+from sklearn.preprocessing import StandardScaler
+from skore import EstimatorReport
+from sklearn.linear_model import Ridge
+from sklearn.pipeline import make_pipeline
```
MarieSacksick, line 192:
```suggestion
from skore import EstimatorReport
from sklearn.preprocessing import StandardScaler
from sklearn.linear_model import Ridge
from sklearn.pipeline import make_pipeline
```

nitpick: group sklearn import in the display, and put more simple import at the top

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
...
+# .. note::
+#   More generally, skore can help you inspect the coefficients of all linear models.
+#   We consider a linear model as defined in `scikit-learn's user guide <https://scikit-learn.org/stable/modules/linear_model.html>`_.
+#   In short, we consider a "linear model" as a scikit-learn compatible estimator that
+#   holds a ``coef_`` attribute (after being fitted).
+
+# %%
+# As we have performed scaling in our pipeline, the above coefficients are in the same
+# range so they can be compared to one another.
+# Indeed, scaling matters for feature importance.
```
MarieSacksick, line 241:
```suggestion
# Indeed, without scaling, the coefficients of the features in a linear model are affected by the scale of the features values, and therefore are not comparable.
```

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
...
+# Let us unscale the coefficients, without forgetting the intercerpt, so that the
+# coefficients can interpreted using the original units:
+
+# %%
+import numpy as np
+
+unscaling_weights = np.insert(ridge_report.estimator_[0].scale_, 0, 1, axis=0).reshape(
+    -1, 1
+)
+ridge_report.feature_importance.coefficients() * unscaling_weights
```
MarieSacksick, line 265: product question here, it's interesting that you have to do this; do you think it could be something useful provided by skore automatically? We have eveything we need in the estimator report to do that.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
...
+    engineered_ridge,
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+engineered_ridge_report.metrics.report_metrics()
+
+# %%
+# We get a much better score!
```
MarieSacksick, line 342: :clap:  

```suggestion
# We get a much better score! As a reminder, we had less than 0.6 in :math:`R^2`,
# and more than 0.73 as RMSE.  
```

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
...
+X_train_transformed = compromised_ridge_report.estimator_[:-1].transform(X_train)
+print("Number of features after feature engineering:", X_train_transformed.shape[1])
+
+# %%
+# Now, let us get some intuition on the results of our grid search by using a
+# `plotly.express.parallel_coordinates
+# <https://plotly.com/python-api-reference/generated/plotly.express.parallel_coordinates.html>`_,
+# as done in
+# `this scikit-learn example
+# <https://scikit-learn.org/stable/auto_examples/model_selection/plot_grid_search_text_feature_extraction.html>`_:
```
MarieSacksick, line 429:
```suggestion
```

Unsure: I'm not convinced that it's worth citing this example.

MarieSacksick (commented):
I just had a look that the new version of the example, it's a very instructive and useful content, congrats!

examples/model_evaluation/plot_feature_importance.py
MarieSacksick: (just to avoid scrolling up for the reader)

sylvaincom: I'll try to do `ComparisonReport` for this

examples/model_evaluation/plot_feature_importance.py
sylvaincom: It's because I copy-pasted their code ahah

MarieSacksick: It's open source and scikit-learn examples are already heavily cited so it's not as if we were hiding our inspiration!

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,492 @@
...
+#   `Common pitfalls in the interpretation of coefficients of linear models <https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html>`_.
+#
+# Here, it appears that the ``MedInc``, ``Latitude``, and ``Longitude`` features are
+# the most important, with regards to the absolute value of other coefficients.
+# This finding is consistent with our previous observations from the *Associations*
+# tab of the table report.
+#
+# However, due to the scaling, we can not interpret the coefficient values
+# with regards to the original unit of the feature.
+# Let us unscale the coefficients, without forgetting the intercerpt, so that the
```
auguste-probabl, line 256:
```suggestion
# Let us unscale the coefficients, without forgetting the intercept, so that the
```

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Yes, definitely!
And for the intercept, I need to correct it: add the mean instead

auguste-probabl: You mean we could look inside the Pipeline for some `StandardScaler`s? Can there be more than one? Does it work the same for multi-output tasks?

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Should be ok now

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,94 @@
...
+
+As shown in :ref:`example_estimator_report`, the :class:`~skore.EstimatorReport` has
+a :meth:`~skore.EstimatorReport.metrics` accessor that enables you to evaluate your
+models and look at some scores that are automatically computed for you.
+
+In practice, once you have fitted an estimator, you want to evaluate it, and
+you also want to inspect it: you can do that with the estimator report's
+:meth:`~skore.EstimatorReport.feature_importance` accessor that we showcase in this
+example.
+"""
```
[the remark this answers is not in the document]
sylvaincom: Should be ok now

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Can we raise a new issue for enh right now please? 🙏 I have too many comments to read 😂

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Should be good now

examples/model_evaluation/plot_feature_importance.py
sylvaincom: It's removed

examples/model_evaluation/plot_feature_importance.py
MarieSacksick: done [here](https://github.com/probabl-ai/skore/issues/1367).

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,546 @@
...
+# %%
+from sklearn.cluster import KMeans
+from sklearn.compose import make_column_transformer
+from sklearn.preprocessing import PolynomialFeatures, SplineTransformer
+
+geo_columns = ["Latitude", "Longitude"]
+
+preprocessor = make_column_transformer(
+    (KMeans(n_clusters=10), geo_columns),
+    remainder="passthrough",
```
sylvaincom, line 326: @glemaitre here, we have `remainder="passthrough"`, which is not the same as in the grid search below, why?

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,546 @@
...
+# the most important, with regards to the absolute value of other coefficients.
+# This finding is consistent with our previous observations from the *Associations*
+# tab of the table report.
+#
+# However, due to the scaling, we can not interpret the coefficient values
+# with regards to the original unit of the feature.
+# Let us unscale the coefficients, without forgetting the intercept, so that the
+# coefficients can be interpreted using the original units:
+# %%
+mean = ridge_report.estimator_[0].mean_.mean()
```
sylvaincom, line 266: @glemaitre is this correct?

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,546 @@
...
+    "columntransformer__remainder__degree": randint(low=1, high=4),
+    "columntransformer__remainder__n_knots": randint(low=2, high=10),
+    "selectkbest__k": randint(low=5, high=100),
+    "ridge__alpha": np.logspace(-5, 5, num=100),
+}
+random_search = RandomizedSearchCV(
+    model,
+    param_distributions=parameter_grid,
+)
+random_search.fit(X_train, y_train)
```
sylvaincom, line 427: @glemaitre I fit this only on the train set, is this ok?

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,546 @@
...
+
+# %%
+import matplotlib.pyplot as plt
+import seaborn as sns
+
+sns.histplot(data=X_y, x=target_name, bins=100)
+plt.show()
+
+# %%
+# There seems to be a threshold-effect for high-valued houses: all houses with a price
```
sylvaincom, line 99: I notice outliers... but do not remove them ; should I add an explanation?

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,546 @@
...
+from skrub import TableReport
+
+TableReport(X_y)
+
+# %%
+# From the table report, we can draw some key observations:
+#
+# - Looking at the *Stats* tab, all features are numerical and there are no
+#   missing values.
+# - Looking at the *Distributions* tab, we can notice that some features have some
```
sylvaincom, line 77: I notice outliers... but do not remove them ; should I add an explanation?

examples/model_evaluation/plot_feature_importance.py
sylvaincom: see new comment

glemaitre:
@auguste-probabl I see that you force-pushed in the history. In the past I mentioned that I did not like it but I never gave the reason. The reason is this side-effect when trying to update the branch that I was using locally.

image

So I can get around it with a hard reset and a pull but it shows that the history is not linear.

examples/model_evaluation/plot_feature_importance.py
glemaitre: Actually, we can let the default to drop. We are consuming all features, so no feature will be dropped anyway. So they are equivalent.

glemaitre: Oh wait we have a single transformer, so the remainder is essential. Let me check the entire file then :).

auguste-probabl:
> @auguste-probabl I see that you force-pushed in the history. In the past I mentioned that I did not like it but I never gave the reason. The reason is this side-effect when trying to update the branch that I was using locally.
> 
> image
> 
> So I can get around it with a hard reset and a pull but it shows that the history is not linear.

Sorry, but with long-lasting branches I prefer rebasing from time to time to re-linearize the history--otherwise the branch gets full of merge commits which make it difficult to see where the branch actually starts. This is my way of working. Aside from resetting, you can also pull with rebase, as git suggests in the image

examples/model_evaluation/plot_feature_importance.py
glemaitre: `remainder="passthrough"` means that you pass the feature as-is without transforming them. The `remainder=SpineTransformer()` will create new features by applying splines to all feature not used otherwise.

So the difference here is that we don't compute spline on the `KMeans` feature while in the later grid-search, we are also computing the spline on the geo feature. I think it is best to have the same processing. So I would quickly try to apply the `SplineTransformer` on all features and not in the `ColumnTransformer` only and check the performance.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+preprocessor = make_column_transformer(
+    (KMeans(n_clusters=20), geo_columns),
+    remainder=SplineTransformer(),
+)
+model = make_pipeline(
+    preprocessor,
+    PolynomialFeatures(degree=1, interaction_only=True),
+    VarianceThreshold(),
+    SelectKBest(k=50),
+    Ridge(),
```
glemaitre, line 418: Let's do a `RidgeCV`:


```suggestion
    RidgeCV(np.logspace(-5, 5, num=100)),
```

Because it will use a GCV under the hood, it will be much cheaper than doing it in the search.
The small drawback is that we have a small data leak but we can discard it.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+from sklearn.feature_selection import SelectKBest, VarianceThreshold
+from sklearn.model_selection import RandomizedSearchCV
+
+preprocessor = make_column_transformer(
+    (KMeans(n_clusters=20), geo_columns),
+    remainder=SplineTransformer(),
+)
+model = make_pipeline(
+    preprocessor,
+    PolynomialFeatures(degree=1, interaction_only=True),
```
glemaitre, line 415:
```suggestion
    PolynomialFeatures(degree=1, interaction_only=True, include_bias=False),
```

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+    VarianceThreshold(),
+    SelectKBest(k=50),
+    Ridge(),
+).set_output(transform="pandas")
+
+parameter_grid = {
+    "columntransformer__kmeans__n_clusters": randint(low=1, high=30),
```
glemaitre, line 422: Let's avoid to have trying to small number of cluster.

```suggestion
    "columntransformer__kmeans__n_clusters": randint(low=10, high=30),
```

examples/model_evaluation/plot_feature_importance.py
sylvaincom: I did just Ridge instead of RidgeCV so that it's easy for me to do the parallel coordinates plot below?

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+# regression task about predicting house prices:
+
+# %%
+import pandas as pd
+from sklearn.datasets import fetch_california_housing
+
+X_load, y_load = fetch_california_housing(return_X_y=True, as_frame=True)
+X_y = pd.concat([X_load, y_load], axis=1)
+target_name = y_load.name
+X_y.head(2)
```
glemaitre, line 44: I think you can remove `return_X_y=True` and it will simplify some of the code

```python
california_housing = fetch_california_housing(as_frame=True)
X_load, y_load = california_housing.data, california_housing.target
```

Then, you also have access at `california_housing.target_name` or `california_housing.frame` for the full dataset containing y inside.

I don't think you have to use `_load`. It is better to name in the subsequent variables to reflect the change that you did.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+#
+# - Looking at the *Stats* tab, all features are numerical and there are no
+#   missing values.
+# - Looking at the *Distributions* tab, we can notice that some features have some
+#   outliers: ``MedInc``, ``AveRooms``, ``AveBedrms``, ``Population``, and ``AveOccup``.
+#   The feature with the largest number of outliers is ``AveBedrms``, probably
+#   corresponding to vacation resorts.
+# - Looking at the *Associations* tab, we observe that the target feature
+#   ``MedHouseVal`` is mostly associated with ``MedInc``, ``Longitude``, and
+#   ``Latitude``, which makes sense.
```
glemaitre, line 83: "which makes sense" is not enough explicit :P. I think that we can go a bit further: we could build the intuition that people with a higher revenue will leave in an area where the median value of house is higher. And we could expect that those areas are located next to each other.

For me the interesting point as well is to mitigate the "which makes sense" with the associations power: it is not super high. So each individual feature cannot predict the target. With what I said before, since there is a correlation between latitude and longitude a some with median income and the location then we might already build an instinct that linear combination might not be enough :)

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+
+# %%
+# Target feature
+# --------------
+#
+# Moreover, the target distribution has a long tail:
```
glemaitre, line 89: I don't think that you need "However" since you start a new section.

examples/model_evaluation/plot_feature_importance.py
glemaitre: They are not outliers :) They are part of the data. We don't know the reason but we should keep them.

For instance, it is super interesting because, when we model with linear model, nothing prevent us to predict prices above $500,000 because linear model can extrapolate. Trees cannot extrapolate and for instance will predict much better this particular effect.

Coming back to linear model, when we do the prediction error plot, then we will have a straight line that represents those data point. So we can again observe this behaviour at this point. In terms of modeling, if I want to have a good model, it would mean that for my particular application, I should wrap the `predict` function and add this clipping to be sure to not increase the error the high price (I would actually be interested to know how better is the model).

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+plt.show()
+
+# %%
+# There seems to be a threshold-effect for high-valued houses: all houses with a price
+# above $500,000 are given the value $500,000.
+
+# %%
+
+# %%
+# Now, let us assess how the median income ``MedInc`` relates to the median house
```
glemaitre, line 105: Here, you can emphasize that you do that because the MedInc was the variable the most correlated with the median value of the house.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+    return fig
+
+
+# %%
+fig = plot_map(X_y, target_name)
+fig
+
+# %%
+# As could be expected, the price of the houses near the ocean is higher, especially
+# around big cities like Los Angeles and San Jose.
```
glemaitre, line 154: you miss San Francisco ;)

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+
+# %%
+# Just before diving into our first model, let us split our data into a train and a
+# test split:
+
+# %%
+from sklearn.model_selection import train_test_split
+
+X = X_y.drop(columns=[target_name])
+y = X_y[target_name]
```
glemaitre, line 169: you could use the original X and y

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+from skore import EstimatorReport
+
+ridge_report = EstimatorReport(
+    make_pipeline(StandardScaler(), Ridge()),
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+ridge_report.metrics.report_metrics()
```
glemaitre, line 202: For me, it is interesting to have this plot just right after:

```python
ridge_report.metrics.prediction_error().plot(kind="actual_vs_predicted")
```

You can show the effect of clipping. You can also see that the model is in trouble to predict with high values and will predict lower associated values.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+
+mu = ridge_report.estimator_[0].mean_
+sigma = ridge_report.estimator_[0].scale_
+
+
+def unscale_coefficients(df, mu, sigma):
+    df.loc["Intercept", "Coefficient"] = df.loc["Intercept", "Coefficient"] - np.sum(
+        df.loc[df.index != "Intercept", "Coefficient"] * mu / sigma
+    )
+    df.loc[df.index != "Intercept", "Coefficient"] = (
```
glemaitre, line 277: It would be easier to read if you store in a variable the column that you try to indexed, e.g.:

```python
mask_intercept_column = df.index == "intercept"
```

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+
+# %%
+import numpy as np
+
+mu = ridge_report.estimator_[0].mean_
+sigma = ridge_report.estimator_[0].scale_
+
+
+def unscale_coefficients(df, mu, sigma):
+    df.loc["Intercept", "Coefficient"] = df.loc["Intercept", "Coefficient"] - np.sum(
```
glemaitre, line 274: I think that you don't need to call "Coefficient" here `df.loc["Intercept"]` should be enough

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+engineered_ridge
+
+# %%
+# Now, let us compute the metrics and compare it to our previous model using
+# a :class:`skore.ComparisonReport`:
+
+# %%
+from skore import ComparisonReport
+
+engineered_ridge_report = EstimatorReport(
```
glemaitre, line 349: If we do:

```python
engineered_ridge_report.metrics.prediction_error().plot(kind="actual_vs_predicted")
```

Then, we can compare with the first plot and we will see that we have trouble to predict the high MedHousVal as in the previous linear case. So it means that some interaction that we created are actually useful to alleviate the problem.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+# %%
+# We get a much better score!
+# However, interpreting the features is harder: indeed, our complex feature engineering
+# introduced a lot of features:
+
+# %%
+print("Initial number of features:", X_train.shape[1])
+
+X_train_transformed = engineered_ridge_report.estimator_[:-1].transform(X_train)
+n_features_engineered = X_train_transformed.shape[1]
```
glemaitre, line 368: Here you trigger a computation while you already have this info:

```python
engineered_ridge_report.estimator_[-1].n_features_in_
```

I think it is good to explain what `[-1]` is doing because not everyone knows that you can slice a pipeline: we isolate the predictor from the pipeline and check the number of feature entering in the linear model.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+selectkbest_ridge = random_search.best_estimator_
+selectkbest_ridge
+
+# %%
+# Let us get the metrics for the best model of our grid search, and compare it with
+# our previous iterations:
+
+# %%
+selectk_ridge_report = EstimatorReport(
+    selectkbest_ridge,
```
glemaitre, line 443: This should work directly

```suggestion
    random_search,
```

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+selectk_features = selectk_ridge_report.estimator_[-1].feature_names_in_
+print(selectk_features)
+
+# %%
+# We can see that, in the best features, according to statistical tests, there are
+# geospatial features (derived from the K-means clustering) and splines on the median
+# income.
+
+# %%
+# And here are the feature importances based on our model:
```
glemaitre, line 482: Usually importances will be singular.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+
+# %%
+df_engineered_selectkbest_coef = selectk_ridge_report.feature_importance.coefficients()
+sort_absolute_values(df_engineered_selectkbest_coef)
+
+# %%
```
glemaitre, line 488: I would skip this parallel coordinate plot I think. To have it we would:

- have many more runs in the randomized search and thus pay the cost of the run (maybe `n_iter=100`) is still feasible thought
- have a clear story about what we can see. Checking quickly, apart of saying that when `alpha` is too large, then the models are bad, there is not much. Maybe a story there is to identify which family of models are fast to fit and provide good results.

glemaitre (commented):
OK, it is my round of review. There is no major changes requested only minor remark. It looks really good.

glemaitre:
By reviewing #1368, I just realised that we can a minimum unit test in `test_accessor.py` for the `check_as_coef`. We test it indirectly with the integration test but since the file exist and already collect some unit test then it would be nice.

examples/model_evaluation/plot_feature_importance.py
sylvaincom: This is "moreover"? But sure, I can remove it

examples/model_evaluation/plot_feature_importance.py
glemaitre: `RidgeCV` will expose an `alpha_` parameter. I also commented that the parallel coordinate plot is maybe not necessary right now.

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Thanks, done

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Thanks, should be good now

glemaitre:
> otherwise the branch gets full of merge commits which make it difficult to see where the branch actually starts.

I would agree if we would not squash before merging. NumPy and SciPy would go this road. Pandas or scikit-learn request squashing before merge and thus would advocate to not rebase (GH was not handling properly the comment history in this case) or force-pushed.

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+# For now, keep in mind that any observations drawn from inspecting the coefficients
+# of this simple Ridge model are made on a model that performs quite poorly, hence
+# must be treated with caution.
+
+# %%
+# To inspect our model, let us use the
+# :meth:`skore.EstimatorReport.feature_importance` accessor:
+
+# %%
+ridge_report.feature_importance.coefficients()
```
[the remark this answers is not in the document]
sylvaincom: Done

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+# However, due to the scaling, we can not interpret the coefficient values
+# with regards to the original unit of the feature.
+# Let us unscale the coefficients, without forgetting the intercept, so that the
+# coefficients can be interpreted using the original units:
+
+# %%
+import numpy as np
+
+mu = ridge_report.estimator_[0].mean_
+sigma = ridge_report.estimator_[0].scale_
```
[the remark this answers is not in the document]
sylvaincom: Thanks, done

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Thanks, done

examples/model_evaluation/plot_feature_importance.py
glemaitre: Whoops. But it is still a new section. I would start the sentence straight away.

examples/model_evaluation/plot_feature_importance.py
sylvaincom: removed

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Done

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Nice catch, thanks

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Thanks

examples/model_evaluation/plot_feature_importance.py
sylvaincom: done

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+print("Number of features after feature engineering:", n_features_engineered)
+
+# %%
+# Let us display the 15 largest absolute coefficients:
+
+# %%
+df_engineered_ridge_coef = engineered_ridge_report.feature_importance.coefficients()
+
+
+def sort_absolute_values(df):
```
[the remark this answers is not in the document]
sylvaincom: Done, thanks

glemaitre:
Feel free to post a message whenever it is ready to have a final quick overview before the merge.

examples/model_evaluation/plot_feature_importance.py
sylvaincom: ✅ Applying spline on all features seems better, thanks

examples/model_evaluation/plot_feature_importance.py
sylvaincom: ✅ did the warnings catching thing

🚧 for now, I still do the fit outside of the estimator report because of the accessors, will re-try it soon

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Thanks

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Thanks

examples/model_evaluation/plot_feature_importance.py
sylvaincom: Thanks

examples/model_evaluation/plot_feature_importance.py
```diff
@@ -0,0 +1,551 @@
...
+# %%
+# As could have been expected, a high salary often comes with a more expensive house.
+# We can also notice the clipping effect of house prices for very high salaries.
+
+# %%
+# Geospatial features
+# -------------------
+
+# %%
+# Let us look into the coordinates of the districts in California, with regards to the
```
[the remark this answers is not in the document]
sylvaincom: Thanks

auguste-probabl:
> @auguste-probabl By reviewing #1368, I just realised that we can a minimum unit test in `test_accessor.py` for the `check_as_coef`. We test it indirectly with the integration test but since the file exist and already collect some unit test then it would be nice.

done!

examples/use_cases/plot_feature_importance.py
```diff
@@ -0,0 +1,582 @@
...
+selectk_ridge_report = EstimatorReport(
+    selectkbest_ridge,
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+comparator = ComparisonReport(
+    reports=[ridge_report, engineered_ridge_report, selectk_ridge_report]
+)
```
glemaitre, line 512: You can also pass a dict here.

glemaitre (approved):
LGTM, once we pass a dict to have a better column naming.

examples/use_cases/plot_feature_importance.py
sylvaincom: Cool, I created a dict that I can append

PR #1345, opened 2025-02-18
docs: Change config_context docs

MarieS-WiMLDS:
From the example given in config_context, I understand that it is not global, as apposed to set_config, but rather a local behavior. Is it correct?  
In any case, the current docs is blurry to me, because it doesn't appear at first sight what is the difference between `set_config` and `config_context`.

### DIFF
--- a/skore/src/skore/_config.py
+++ b/skore/src/skore/_config.py
@@ -30,7 +30,7 @@ def get_config():
 
     See Also
     --------
-    config_context : Context manager for global skore configuration.
+    config_context : Context manager for local skore configuration.
     set_config : Set global skore configuration.
 
     Examples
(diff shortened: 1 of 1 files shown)


auguste-probabl:
> From the example given in config_context, I understand that it is not global, as apposed to set_config, but rather a local behavior. Is it correct?

Correct, it allows to temporarily override the global config

glemaitre:
Actually global would refer to set a variable for the main process and dispatch the same config to sub process or thread. Local is wrong in this case because it would refer to thread local meaning that you might expect the configuration to work in locally for the main process and not being dispatched to other threads that are started afterword.

I understand that there is confusion here about the term. So we should probably be more explicit in the long summary to remove any ambiguity.

glemaitre:
I'll open a new PR trying to have both the Python details and DS language in there.

PR #1349, opened 2025-02-18
docs: Fix the ambuiguity regarding global configuration

glemaitre:
This PR tries to disambiguate what we mean by "global" when it comes to the configuration setting.

In short global refer that we are going to dispatch the configuration to local threads or sub-process.

It is really different from the "local" meaning within the code impacted by the context manager.


thomass-dev (approved):
Now it's clearer, thanks!

PR #1361, opened 2025-02-25
chore: Refactor Displays in reporters

glemaitre:
This PR refactor the displays in the reporter since we have duplicated code in the `ComparisonReport`.


skore/src/skore/sklearn/_plot/metrics/roc_curve.py
```diff
@@ -0,0 +1,865 @@
...
+                colormaps.get_cmap("tab10"), 10 if len(fpr) < 10 else len(fpr)
+            )
+
+            for est_idx, est_name in enumerate(estimator_names):
+                est_color = class_colors[est_idx]
+
+                for class_idx, class_ in enumerate(fpr):
+                    fpr_est_class = fpr[class_][est_idx]
+                    tpr_est_class = tpr[class_][est_idx]
+                    roc_auc_mean = np.mean(roc_auc[class_][est_idx])
```
glemaitre, line 592: @chkoar

Previously we had:

```python
roc_auc_mean = np.mean(roc_auc[class_])
```

So we averaged the statistics of all estimators for a class and not for a specific estimator.

skore/tests/unit/sklearn/plot/test_precision_recall_curve.py
```diff
@@ -1,24 +1,27 @@
...
-from skore import CrossValidationReport, EstimatorReport
+from skore import ComparisonReport, CrossValidationReport, EstimatorReport
 from skore.sklearn._plot import PrecisionRecallCurveDisplay
 from skore.sklearn._plot.utils import sample_mpl_colormap
 
 
 @pytest.fixture
 def binary_classification_data():
-    X, y = make_classification(random_state=42)
+    X, y = make_classification(class_sep=0.1, random_state=42)
```
glemaitre, line 15: Modified the generation to make sure that if we don't compute the mean properly, then we can detect it. With this change, we would have detected the regression.

glemaitre:
@auguste-probabl @thomass-dev It should be ready to be reviewed. I mainly moved the class a split the plotting for each type of reporter. The tests that we had allowed me to be sure that we did not break the backward compatibility. I moved the tests from the comparison module into the same file and adapt them.

skore/src/skore/sklearn/_plot/base.py
```diff
@@ -0,0 +1,16 @@
...
+class Display(Protocol):
+    """Protocol for the public API of display objects."""
+
+    def plot(self, ax: Optional[Axes] = None, **kwargs: Any) -> None:
+        """Plot the display."""
+        ...
+
+    def set_style(self, **kwargs: Any):
+        """Set the style of the display."""
+        ...
```
auguste-probabl, line 16:
```suggestion
    def plot(self, ax: Optional[Axes] = None, **kwargs: Any) -> None:
        """Plot the display."""

    def set_style(self, **kwargs: Any) -> None:
        """Set the style of the display."""
```

skore/src/skore/sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -0,0 +1,793 @@
...
+                    ax=self.ax_,
+                    estimator_name=(
+                        self.estimator_names[0]
+                        if estimator_name is None
+                        else estimator_name
+                    ),
+                    pr_curve_kwargs=pr_curve_kwargs,
+                )
+            )
+        else:  # self.report_type == "comparison-estimator"
```
auguste-probabl, line 648:
```suggestion
        elif self.report_type == "comparison-estimator":
            # ...
        else:
            raise ValueError(f"report_type should be one of ...; got {report_type}")
```

skore/src/skore/sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -0,0 +1,793 @@
...
+        )
+        recall: dict[Union[int, float, bool, str], list[ArrayLike]] = defaultdict(list)
+        average_precision: dict[Union[int, float, bool, str], list[float]] = (
+            defaultdict(list)
+        )
+
+        if ml_task == "binary-classification":
+            for y_true_i, y_pred_i in zip(y_true, y_pred):
+                # assert for mypy that pos_label_validated is not None
+                assert pos_label_validated is not None, (
```
auguste-probabl, line 747: Does it work with
```suggestion
                # assert for mypy that pos_label_validated is not None
                if pos_label_validated is None:
                    raise TypeError(...)
```

skore/src/skore/sklearn/_plot/metrics/precision_recall_curve.py
glemaitre: You are right. We will probably add new one `comparison-cross-validation` or something like that and it might be dangerous.

skore/src/skore/sklearn/_plot/metrics/precision_recall_curve.py
auguste-probabl: Thinking about it again, if all we're doing here is to satisfy mypy, I think something like `typing.cast` would be enough

glemaitre: I have to give a try (not super strong yet with typing ;))

glemaitre:
@auguste-probabl I used a proper `elif` and an `else` to raise the error and added some tests.

skore/src/skore/sklearn/_plot/metrics/roc_curve.py
chkoar: Now it correctly displays a separate value for each curve.

However, in my [example](https://github.com/probabl-ai/skore/pull/1361#1443), it is not particularly useful because I have two classifiers for ten classes, resulting in 20 curves. It would be great to have a built-in option to view the micro or macro-averaged curve for each classifier, using either the `OvR` or `OvO` decomposition. If you agree I could open a separate issue.

glemaitre: I think that the first thing that I had in head to solve this problem (and some other plot) is to break down the plot for cases where we know we have too much information: https://github.com/probabl-ai/skore/issues/1445

Once done, I think that I would prefer to implement the OvO approach but not necessarly rush for the averaging because sometimes people don't understand what average they are looking at and forcing them to see all the curves is a better default in this regards. But it needs to be readable :)

auguste-probabl (approved):
Alright, LGTM

PR #1362, opened 2025-02-25
feat: Automatically pass OTP to HUB

thomass-dev:


skore/src/skore/hub/login.py
```diff
@@ -0,0 +1,87 @@
...
+
+        port = launch_callback_server(callback=callback)
+        authorization_url, device_code, user_code = api.get_oauth_device_login(
+            success_uri=f"http://localhost:{port}"
+        )
+
+        webbrowser.open(authorization_url)
+
+        while access is None or refreshment is None or expires_at is None:
+            sleep(0.5)
```
rouk1, line 45: I suppose that we might have to respect the timeout here aswell.

rouk1 (changes requested):

skore/src/skore/hub/login.py
thomass-dev: Fixed.

rouk1 (approved):
Nice : )

PR #1366, opened 2025-02-26
feat(repr): Give a description to attributes

### LINKED ISSUE #1235: feat: Give more description in the `CrossValidationReport` to access a fold

*I got this feedback from a user test.*

When using the `CrossValidationReport` helper, he did not know how to access the folds, indeed there is no description for the `.estimator_reports_` attribute:

Add a description such as `Access the estimator report for each fold`.

_No response_

### Additional context

_No response_

MarieSacksick:
Closes #1235   

I had to go through regex because from what I understand from inspect, there is no way to access the documentation of an attribute. There is very little risk that the regex matches something before the attribute description because of naming overlap. If we think that this risk is to big, as there is no way to do a test, we should close #1235 as not possible.  

What it does for an EstimatorReport:  
image

What it does for a CrossValidationReport:  
image

What it does for a ComparisonReport:  
image


skore/src/skore/sklearn/_base.py
```diff
@@ -191,8 +202,9 @@ def _create_help_tree(self) -> Tree:
         attributes = self._get_attributes_for_help()
         if attributes:
             attr_branch = tree.add("[bold cyan]Attributes[/bold cyan]")
-            for attr in attributes:
-                attr_branch.add(f".{attr}")
+            for attr_name in attributes:
+                description = self._get_attributes_description(attr_name)
+                attr_branch.add(f".{attr_name.ljust(29)}" f" - {description}")
```
auguste-probabl, line 207:
```suggestion
                attr_branch.add(f".{attr_name.ljust(29) - {description}")
```

skore/src/skore/sklearn/_base.py
```diff
@@ -50,6 +51,16 @@ def _get_method_description(self, method: Any) -> str:
             else "No description available"
         )
 
+    def _get_attributes_description(self, name: str) -> str:
+        """Get the description for each attribute."""
+        if self.__doc__ is None:
+            return "No description available"
+        regex_pattern = rf"{name} : .*?\n\s*(.*?)\."
+        # this regex will look for the first sentence
+        # on the line below the pattern 'attribute_name : '
```
auguste-probabl, line 60:
```suggestion
        # Look for the first sentence on the line below the pattern 'attribute_name : '
        regex_pattern = rf"{name} : .*?\n\s*(.*?)\."
```

skore/src/skore/sklearn/_base.py
```diff
@@ -50,6 +51,16 @@ def _get_method_description(self, method: Any) -> str:
             else "No description available"
         )
 
+    def _get_attributes_description(self, name: str) -> str:
```
auguste-probabl, line 54:
```suggestion
    def _get_attribute_description(self, name: str) -> str:
```

auguste-probabl (commented):
Works for me in principle!

skore/src/skore/sklearn/_base.py
```diff
@@ -50,6 +51,16 @@ def _get_method_description(self, method: Any) -> str:
...
+        """Get the description for each attribute."""
```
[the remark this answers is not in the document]
MarieSacksick: oups, accepted too fast. It's rather on its class docstring than the attribute's docstring itself.

skore/src/skore/sklearn/_base.py
auguste-probabl: doesn't look like my suggestion was applied, is this intentional?

glemaitre: It looks like a reformating where ruff does not merge string ;)

skore/src/skore/sklearn/_base.py
```diff
@@ -50,6 +51,15 @@ def _get_method_description(self, method: Any) -> str:
...
         )
 
+    def _get_attribute_description(self, name: str) -> str:
+        """Get the description of an attribute from its class docstring."""
+        if self.__doc__ is None:
+            return "No description available"
+        # Look for the first sentence on the line below the pattern 'attribute_name : '
+        regex_pattern = rf"{name} : .*?\n\s*(.*?)\."
+        search_result = re.search(regex_pattern, self.__doc__)
+        return search_result.group(1) if search_result else "No description available"
```
glemaitre, line 61: Instead of "No description available", it might be better to have an empty string?

glemaitre (commented):
I would request a test in the different reporter to at least on one attribute just to catch potential regression in the future.

skore/src/skore/sklearn/_base.py
MarieSacksick: We could!  
It's what we have today for methods: https://github.com/probabl-ai/skore/blob/844dd03b3da020661611b9d3f45e3d21867ea5ac/skore/src/skore/sklearn/_base.py#L51.  
Although we can be 99.9% sure that we will always have a description for methods, but not as much for attributes.  
Or I can change in both places.

auguste-probabl: I think "No description available" is good.

auguste-probabl (dismissed):
Thanks

auguste-probabl (approved):

PR #1370, opened 2025-02-28
ci: Fix sphinx pipeline

thomass-dev:
Removes the temporary fix introduced to download `skrub` resources when OpenML was inaccessible.

image

Removes `HOME` envar overload which modified the `pip-cache` directory used in the sphinx build action, which differed from the `pip-cache` directory used in the calling workflow causing [errors](https://github.com/probabl-ai/skore/actions/runs/13583937631/job/37974812004).

image


auguste-probabl (approved):
Alright, thanks!

PR #1371, opened 2025-02-28
ci: Add numpydoc pre-commit hook

### LINKED ISSUE #1338: Use numpydoc validation as a linter for the documentation
To have the most consistent documentation when it comes to the API, we could use the NumPyDoc validator: https://numpydoc.readthedocs.io/en/latest/validation.html

We have such implementation in scikit-learn. It catches discrepancy at the moment where we introduce new documentation.

auguste-probabl:
- Only files in `skore/src` are checked
- Some lints are ignored
  - Even then, there remains a lot of work to make numpydoc happy
  - I didn't ignore the "summary line should be on a new line" rule, but actually this is PEP 257-compliant and we could ignore it

Closes #1338

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -117,6 +117,14 @@ convention = "numpy"
 "src/*" = ["UP007"]
 "tests/*" = ["D"]
 
+[tool.numpydoc_validation]
+checks = [
+  "all",  # Report on all checks, except the below
+  "ES01",  # No "extended summary" section
+  "EX01",  # No "Examples" section
+  "SA01",  # No "See also" section
+]
+
 [tool.typos.default]
 extend-ignore-re = [
   # Line ignore with trailing `# spellchecker:disable-line`
--- a/.github/workflows/backend.yml
+++ b/.github/workflows/backend.yml
@@ -38,7 +38,9 @@ jobs:
 
       - name: Lint
         working-directory: skore/
-        run: pre-commit run --all-files ruff
+        run: |
+          pre-commit run --all-files ruff
+          pre-commit run --all-files numpydoc-validation
 
   backend-lockfiles:
     runs-on: "ubuntu-latest"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,12 @@ repos:
       - id: ruff-format
         files: ^(skore/(src|tests))|(examples)/
 
+  - repo: https://github.com/numpy/numpydoc
+    rev: v1.8.0
+    hooks:
+      - id: numpydoc-validation
+        files: ^(skore/src)/
+
   - repo: https://github.com/pre-commit/mirrors-mypy
     rev: v1.15.0
     hooks:
--- a/examples/skore_project/plot_working_with_projects.py
+++ b/examples/skore_project/plot_working_with_projects.py
@@ -120,7 +120,8 @@ my_project.get("my_string")
 my_project.put(
     "my_string_2",
     (
-        """Hello world!, **bold**, *italic*, `code`
+        """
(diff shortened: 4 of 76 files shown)


thomass-dev:
Surprised that it isn't already effective with the `pyproject.toml` section

```
[tool.ruff.lint.pydocstyle]
convention = "numpy"
```

auguste-probabl:
> Surprised that it isn't already effective/covered by the `pyproject.toml` section

It kind of is, but not completely; see the failed CI run for examples of things that aren't checked by ruff. There is an open issue in ruff to cover all of numpydoc: https://github.com/astral-sh/ruff/issues/8425.

auguste-probabl:
Personally I don't know that it's worth the trouble right now.

skore/pyproject.toml
```diff
@@ -117,6 +117,14 @@ convention = "numpy"
 "src/*" = ["UP007"]
 "tests/*" = ["D"]
 
+[tool.numpydoc_validation]
+checks = [
+  "all",  # Report on all checks, except the below
+  "ES01",  # No "extended summary" section
+  "EX01",  # No "Examples" section
+  "SA01",  # No "See also" section
+]
```
glemaitre, line 126: We have those filters in scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/fef620292973dd25ca206c8bbdff194771c857fc/sklearn/tests/test_docstrings.py#L45

skore/pyproject.toml
```diff
@@ -117,6 +117,14 @@ convention = "numpy"
 "src/*" = ["UP007"]
 "tests/*" = ["D"]
 
+[tool.numpydoc_validation]
+checks = [
+  "all",  # Report on all checks, except the below
+  "ES01",  # No "extended summary" section
+  "EX01",  # No "Examples" section
+  "SA01",  # No "See also" section
```
glemaitre, line 125: Those one are quite important for a nice documentation actually.

glemaitre:
> Personally I don't think that it's worth the trouble right now.

Right now apart from the long summary, examples and see also that find important to have, we only have 3 failures with the rendering of the `help` which would could skip if we see that we cannot support properly.

In terms of rule, we have set that we skip in scikit-learn because numpydoc is picky a bit too much on something that does not necessarily improve the documentation.

Regarding the "necessity", I think that we should pretty soon have a user guide and have proper docstring. Otherwise, it is already a project smell when you don't have a clear documentation explaining with words why a project is useful (a getting_started is not enough). So I would think that we should still fix those issues on the way and not only postpone them releases after releases.

auguste-probabl:
Update with different settings. If we don't ignore "extended summary", "example" or "see also", but we do ignore "summary should be on a different line", we get ~2000 lints. Most of them are superfluous IMO.

Yeah the fine-grained approach that sklearn uses sounds like a good start because we can spread out the change over time, so that we don't have to keep a PR open while we fix 2000 errors.

auguste-probabl:
> > we get ~2000 lints. Most of them are superfluous IMO.
> 
> Is it that we get warning on function that are not part of the public API? The code in scikit-learn, is actually reducing this behaviour for the public code base and I would argue that it is indeed enough.

Yes that's right. I don't know what regex to give to numpydoc to filter things correctly (e.g `EstimatorReport` is defined in `_estimator/report.py` so `_+` is not enough), so we'd need something like sklearn's custom pytest-based system to get something correct (whatever "correct" means for us).

glemaitre:
Yep so it is indeed more involved work than just having the linter on. So we might delay that. We just need to ensure that we keep documenting as good as possible in the meantime ;)

auguste-probabl:
What we could also do is start enforcing rules in CI only on *changed* files, so for a while every PR would include a bit of docs improvement

PR #1372, opened 2025-02-28
fix(docs): Change axis of concat

### LINKED ISSUE #1363: bug: Use axis=1 instead of axis=0 in use case example when concatenating metric report
In the following example:

https://skore.probabl.ai/0.7/auto_examples/use_cases/plot_employee_salaries.html#investigating-the-models

There is an issue because we don't concatenate anymore on the right dimension using pandas. It happens because we transpose the dataframe.

MarieSacksick:
Closes #1363


auguste-probabl (approved):
Thanks!

glemaitre:
You missed l. 303 as well.

PR #1373, opened 2025-02-28
chore: Add stale bot labels only

MarieSacksick:
To help myself in github grooming.

### DIFF
--- a/.github/workflows/label-stale.yml
+++ b/.github/workflows/label-stale.yml
@@ -1,0 +1,4 @@
+- name: label stale issues and PR
+  uses: actions/stale@v9.1.0
+  days-before-stale: 30
+  days-before-close: -1 # never close
--- a/.github/workflows/label-stale.yml
+++ b/.github/workflows/label-stale.yml
@@ -1,4 +1,17 @@
-- name: label stale issues and PR
-  uses: actions/stale@v9.1.0
-  days-before-stale: 30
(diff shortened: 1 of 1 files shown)


auguste-probabl:
IMO issues marked as "stale" reflect badly on project maintainers, as though their project were abandoned. 30 days is a really short window: we only have ~60 issues, but ~30 are older than a month -- that's 50%.

Can I ask the rationale behind this? What would this help you achieve that you could not do otherwise?
For example, if you goal is to stay aware of older issues, you could sort issues by "last-updated"?

MarieSacksick:
I hadn't run the baseline, you're right, these 30 days are short.  
The idea is that I would just have to visit [this link filtering issues](https://github.com/probabl-ai/skore/issues?q=is%3Aissue%20state%3Aopen%20label%3Astale) to see what issues should have an update. It also writes a message in this issue, raising the attention of the author, so I don't have to ping them myself, and they can update it if necessary.

thomass-dev:
Can you limit the scope of the stale labeling to issues?

MarieSacksick:
I found a work-around, I'll use this [gh filtering](https://github.com/probabl-ai/skore/issues?q=is%3Aissue%20state%3Aopen%20updated%3A%3C%40today-30d).  
I'll ask the question in the issue when I don't know to ask the authors and/or participants. If you think it's not a good label to have, then it's a better way :).

Issue #1344, opened 2025-02-18
CI: Release message to internal slack


MarieSacksick:
### Describe the bug

Since early december, the CI part in charge of releasing is not functional anymore.  

### Steps/Code to Reproduce

Do a release and check the logs (for instance for [skore 0.6.1.](https://github.com/probabl-ai/skore/actions/runs/12912252585/job/36008329089)).  

### Expected Behavior

We should have a message in our internal slack.  

### Actual Behavior

Nothing happens, that part of the CI fails (but it doesn't block anything, and this behavior is correct, we don't want a release to be blocked because of slack).   

![Image](https://github.com/user-attachments/assets/a0f8515d-4197-4bf5-87d6-4be909544af6)

### Environment

```shell
N/A
```

MarieS-WiMLDS:
Typically the kind of issues we could tackle inbetween functional developments.

PR #1380, opened 2025-03-04
ci: Fix SlackAPI call

jerome-probabl:
This PR should fix https://github.com/probabl-ai/skore/issues/1344 issue


thomass-dev (changes requested):
Just a message to ensure that i can review before merge.

PR #1381, opened 2025-03-04
chore: Remove skore-ui

rouk1:
Following irl discussion with @MarieSacksick, @thomass-dev & @nbesson-probabl 

It was acted that:

> We don't plan any dev on any Web UI for Lib in the next 6 months as far as company-driven efforts are concerned (that's the decision)
> But it does not mean neither :
> - that we're gonna restart working on an UI for Lib in 6 months (we could have no web UI for Lib for the next 5 years) 
> - nor that a Web UI for Lib in 6+ months would be similar to the current skore-ui

So as things may move largely during those 6 months and given the maintenance effort that this code come with, this PR deletes every UI related stuff from the repo.


auguste-probabl (approved):

thomass-dev:
We should consider removing the views too, which are only relative to the `skore-ui`.

rouk1:
> We should consider removing the views too, which are only relative to the `skore-ui`.

Woops my bad, I will open a follow up PR.

rouk1:
https://github.com/probabl-ai/skore/pull/1383

thomass-dev:
Please take a look at the following PR https://github.com/probabl-ai/skore/pull/1385.

thomass-dev:
The release pipeline was also impacted https://github.com/probabl-ai/skore/pull/1406/files.

PR #1385, opened 2025-03-05
fix: Remove deprecated code after the removing of `skore-ui`

thomass-dev:
Fix after https://github.com/probabl-ai/skore/pull/1381 and https://github.com/probabl-ai/skore/pull/1383.

- Remove broken script `__main__.py` as the cli was removed
- Remove from `pyproject.toml` the broken entrypoint, and the artifacts which came from `skore-ui`
- Remove the npm section in dependabot configuration
- Remove frontend relative GH workflow
- Remove `skore.open` function which is not anymore relevant
- Purge dependencies


auguste-probabl (approved):
ltgm, thanks!

rouk1 (commented):
Sorry I forgot dependencies and that CI workflow.
Regarding the open command I did not removed it on purpose as I thought it was used. Sounds that only constructor is documented. My bad.

Issue #60, opened 2024-07-22
Check that all our requirements are well defined and license-compliant


thomass-dev:
Particular attention to [`jsonschema[format]`](https://python-jsonschema.readthedocs.io/en/latest/validate/#validating-formats) .

tuscland:
Hi! Can you please remove my assignee status?

glemaitre:
> Hi! Can you please remove my assignee status?

Done.

PR #1387, opened 2025-03-05
ci(licenses): Check Python requirements licenses

jerome-probabl:
Using [liccheck](https://pypi.org/project/liccheck/) to check Python dependencies licenses.

Note: I'm using the `paranoid` mode of liccheck so if you use a dependency with a new license you need to add it in `authorized_licenses` section.

https://github.com/probabl-ai/skore/issues/60

### DIFF
--- a/.github/workflows/check-licenses.yaml
+++ b/.github/workflows/check-licenses.yaml
@@ -1,0 +1,35 @@
+name: check-licenses
+
+on:
+  pull_request:
+    paths:
+      - 'skore/requirements.in'
+      - 'skore/pyproject.toml'
+      - '.github/workflows/check-licenses.yaml'
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: "bash"
+
+jobs:
+  check-python-licenses:
+    runs-on: "ubuntu-latest"
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+          cache: 'pip'
+
+      - run: pip install -U pip setuptools liccheck==0.9.2
+
+      - run: pip install ./skore
+
+      - name: Check Skore dependencies with liccheck
+        working-directory: ./skore
+        run: liccheck
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -136,3 +136,33 @@ exclude = ["src/skore/externals/.*", "ci/*", "hatch/*", "tests/*"]
 [[tool.mypy.overrides]]
 module = ["diskcache.*", "sklearn.*"]
 ignore_missing_imports = true
+
+[tool.liccheck]
+authorized_licenses = [
+    "bsd",
+    "new bsd",
+    "bsd license",
+    "new bsd license",
+    "simplified bsd",
+    "apache",
+    "apache 2.0",
+    "apache software license",
+    "gnu lgpl",
+    "lgpl with exceptions or zpl",
+    "isc license",
+    "isc license (iscl)",
+    "mit",
+    "mit license",
+    "Mozilla Public License.*",
+    "python software foundation.*",
+    "The Unlicense.*",
+    "wtfpl",
+    "zpl 2.1",
+]
+unauthorized_licenses = [
+    "\bagpl",
+    "\bgpl",
+]
+as_regex = true
+level = "PARANOID"
+requirement_txt_file = "./requirements.in"


thomass-dev (changes requested):
Thanks, i have some open questions.

.github/workflows/check-licenses.yaml
```diff
@@ -0,0 +1,37 @@
...
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+          cache: 'pip'
+
+      - run: pip install -U pip setuptools
+
+      - run: pip install -r skore/requirements.in
```
thomass-dev, line 29:
```suggestion
      - run: pip install -r ./skore
```

.github/workflows/check-licenses.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Check Python dependencies with liccheck
+        uses: andersy005/gh-action-py-liccheck@main
+        with:
+          strategy-ini-file: .github/actions/liccheck-strategy.ini
+          level: paranoid
+          requirements-txt-file: skore/requirements.in
```
thomass-dev, line 36: Isn't the `pyproject.toml` file supported? If so, please change and try it.

.github/workflows/check-licenses.yaml
thomass-dev, line 29: In addition to my previous comment, we should instead call `pip install ./skore` even if the package itself will be installed with the dependencies.

.github/workflows/check-licenses.yaml
```diff
@@ -0,0 +1,37 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.13'
+          cache: 'pip'
+
+      - run: pip install -U pip setuptools
```
thomass-dev, line 27: What's the need of `setuptools` here?

.github/workflows/check-licenses.yaml
jerome-probabl: it's supported, I tried to use it but it doesn't support this:

```
[tool.hatch.metadata.hooks.custom]
[...]
dependencies = { "file" = "requirements.in" }
```

I also try with `requirement_txt_file =` in `pyproject.toml` but it didn't work.

skore/pyproject.toml
```diff
@@ -136,3 +136,31 @@ exclude = ["src/skore/externals/.*", "ci/*", "hatch/*", "tests/*"]
...
+    "apache software license",
+    "gnu lgpl",
+    "lgpl with exceptions or zpl",
+    "isc license",
+    "isc license (iscl)",
+    "mit",
+    "mit license",
+    "Mozilla Public License.*",
+    "python software foundation.*",
+    "The Unlicense.*",
```
rouk1, line 158: We need [WTFPL](https://en.wikipedia.org/wiki/WTFPL) too : )

skore/pyproject.toml
```diff
@@ -136,3 +136,31 @@ exclude = ["src/skore/externals/.*", "ci/*", "hatch/*", "tests/*"]
...
+    "zpl 2.1",
+]
+unauthorized_licenses = [
+    "\bgpl"
```
rouk1, line 162: Is there a specific reason for this backspace ?
Although I suppose that AGPL is a superset of GPL it should be added to the unauthorized list.

rouk1 (changes requested):

skore/pyproject.toml
jerome-probabl: AGPL added (I'm using `panaroid` mode so if the license isn't accepted it will not work.

thomass-dev (changes requested):
Ensure i can re-review before merge.

glemaitre:
Is `liccheck` maintained? I don't see a release since September 2023.

jerome-probabl:
> Is `liccheck` maintained? I don't see a release since September 2023.

There have been no new commits for over a year, and the same is true for pip-licenses. It seems that none of the tools performing this task are actively maintained.

jerome-probabl:
> Does https://github.com/aboutcode-org/scancode-toolkit not scanning package metadata to get a similar results? I did not check details and complexity of using this tool but in terms of maintenance, it is better maintained.

Thanks I didn't find this one. I will test it today.

thomass-dev:
@jerome-probabl can we close this PR, as it's seems to be outdated?

auguste-probabl:
Closed as out-of-date

Issue #1388, opened 2025-03-05
feat: Retrieve `skrub.TableReport` as is from the local persistency


glemaitre:
The following snippet:

```python
import skore

my_project = skore.Project("../data/my_project", if_exists=True)

# %%
from skrub.datasets import fetch_employee_salaries

datasets = fetch_employee_salaries()
df, y = datasets.X, datasets.y

# %%
from skrub import TableReport

table_report = TableReport(df)
table_report

# %%
my_project.put("Input data summary", table_report)

# %%
xxx = my_project.get("Input data summary")
xxx
```

It looks like `xxx` here is the HTML representation an not the project itself. Did we notice this in the past?

glemaitre:
But the `TableReport` is a python object that I would like to interact with? Storing the HTML representation only does not seems the "what you put, is what you get", isn't it?

thomass-dev:
> "what you put, is what you get"
Originally, it was more "what you put, is most of the time what you get".  You want to pickle the object instead?

thomass-dev:
I was on the equivalent on the remote project, nice catch 👍🏻 .

PR #1389, opened 2025-03-05
feat: Retrieve `skrub.TableReport` as is from the local persistency

thomass-dev:
Actually, if you put a `skrub.TableReport` in a project, you will get its HTML representation instead of the python object.
This PR corrects this behavior to be compliant with "what you put is what you get".

Closes https://github.com/probabl-ai/skore/issues/1388


auguste-probabl (approved):
Nice, thanks!

glemaitre:
Thanks :)

PR #1390, opened 2025-03-05
docs: For feature importance, add MDI and PI in the example on California housing dataset

sylvaincom:
About our new features of feature importance for models.

The example on California housing dataset, since https://github.com/probabl-ai/skore/pull/1339, uses coefficients for linear models.

In this PR, we build on upon this example to add MDI and permutation importance (PI), see https://github.com/probabl-ai/skore/pull/1365.

This PR also updates the README, docs index, and getting started example


examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# %%
+# First, let us interpret our model with regards to the original features.
+# For the visualization, we fix a very low ``max_depth`` so that it will be easy for
+# the human eye to visualize the tree using :func:`sklearn.tree.plot_tree`:
+
+# %%
+from sklearn.tree import plot_tree
+
+plt.figure(dpi=300)
+_ = plot_tree(
```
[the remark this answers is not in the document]
sylvaincom: It's intentional, little trick to avoid having any ugly output as in this screenshot:

Capture d’écran 2025-03-07 à 15 02 04

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# Tree-based models: mean decrease in impurity (MDI)
+# ==================================================
+
+# %%
+# Now, let us look into tree-based models.
+# For feature importance, we inspect their Mean Decrease in Impurity (MDI).
+# The MDI of a feature is the normalized total reduction of the criterion (loss)
+# brought by that feature.
+# The MDI is also known as the Gini importance.
+# The higher the MDI, the more important the feature.
```
glemaitre, line 588: I think that we should state from the start the limitation of the method:

- it should be carefully used when the cardinality between features is large: the MDI will favor high cardinality features. Here we are fine because we have numerical features that would have similar cardinality
- then, we should mention that the MDI here is computed from the training set. Therefore, the model is overfitting then feature importance will also be bias. The tree will tend to partition part of the space that are actually not important and it will inflate the value of the MDI.

And we can link directly to this example: https://scikit-learn.org/dev/auto_examples/inspection/plot_permutation_importance.html#sphx-glr-auto-examples-inspection-plot-permutation-importance-py

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# Indeed, in the above tree visualization, ``MedInc`` is used multiple times for splits
+# and contributes greatly to reducing the squared error at multiple nodes.
+# At the root, it reduces the error from :math:`1.335` to :math:`0.832` and :math:`0.546`
+# in the children.
+
+# %%
+# Random forest
+# -------------
+#
+# Now, let us apply a slightly more elaborate model: a random forest.
```
glemaitre, line 718: You can remove slightly (you have x100 more trees and depth :))

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+
+# %%
+# Without any feature engineering and any grid search,
+# the random forest beats all linear models!
+
+# %%
+# Let us print the number of trees in our random forest:
+
+# %%
+print(f"Number of trees in the forest: {rf_report.estimator_.n_estimators}")
```
glemaitre, line 747: you don't really need this here. You could instead pass `n_estimators` to the forest and it will be more explicit.

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+fig = px.histogram(
+    X_y_plot,
+    x=california_housing.target_names[0],
+    color="AveOccup",
+    category_orders={"AveOccup": bin_order},
+)
+fig
+
+# %%
+# Gradient-boosted tree
```
glemaitre, line 793: I would not put the Gradient boosting here. We already have quite enough.

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+
+# %%
+# Permutation feature importance measures the contribution of each feature to
+# a fitted model's performance.
+# It randomly shuffles the values of a single feature and observes the resulting
+# degradation of the model's score.
+# Permuting a predictive feature makes the performance decrease, while
+# permuting a non-predictive feature does not degrade the performance much.
+
+# %%
```
glemaitre, line 868: I would not do any subsection but instead make the comparison between the models directly in the same section.

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# Model-agnostic: permutation feature importance
+# ==============================================
+
+# %%
+# In the previous sections, we have inspected coefficients that are specific to linear
+# models and the MDI that is specific to tree-based models.
+# In this section, we look into the
+# `permutation importance <https://scikit-learn.org/stable/modules/permutation_importance.html>`_
+# which is model agnostic, meaning that it can be applied to any fitted estimator.
+# In particular, it works for linear models and tree-based ones.
```
glemaitre, line 858: We should also mention that we can compute the permutation on the testing set. Actually I think it would be good to have a plot to compute on the training for the two models and then one figure with the same on the test.

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+    ["mean", "std"], axis=1
+).plot.barh(y="mean", xerr="std")
+plt.tight_layout()
+
+# %%
+# The result of the inspection is the same as with the MDI:
+# the most important features are ``MedInc``, ``Latitude``, ``Longitude``,
+# and ``AveOccup``.
+
+# %%
```
glemaitre, line 922: I would also drop this one. The example is long enough and we might not be able to teach that much new stuff.

glemaitre (commented):
A couple of feedback.

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# Tree-based models: mean decrease in impurity (MDI)
+# ==================================================
+
+# %%
+# Now, let us look into tree-based models.
+# For feature importance, we inspect their Mean Decrease in Impurity (MDI).
+# The MDI of a feature is the normalized total reduction of the criterion (loss)
+# brought by that feature.
+# The MDI is also known as the Gini importance.
```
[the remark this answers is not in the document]
sylvaincom: OK, I'll remove it

I got this from scikit-learn: https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomForestRegressor.html#sklearn.ensemble.RandomForestRegressor.feature_importances_

Do you think it deserves an update?

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+
+# %%
+# Let us start with a simple decision tree.
+
+# %%
+# .. seealso::
+#   For more information about decision trees, see scikit-learn's example on
+#   `Understanding the decision tree structure <https://scikit-learn.org/stable/auto_examples/tree/plot_unveil_tree_structure.html>`_.
+#   In this example, we are tackling a regression task, but it is easier to gain
+#   intuition on decision trees on a classification task.
```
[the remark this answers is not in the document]
sylvaincom: Ok, it's easier if I remove this phrase ahah ; or I need to explain an explanation and it's less concise

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+from sklearn.tree import DecisionTreeRegressor
+
+tree_report = EstimatorReport(
+    DecisionTreeRegressor(random_state=0),
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+dict_reports["Decision tree"] = tree_report
```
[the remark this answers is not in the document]
sylvaincom: Sure, done

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+
+# %%
+# First, let us interpret our model with regards to the original features.
+# For the visualization, we fix a very low ``max_depth`` so that it will be easy for
+# the human eye to visualize the tree using :func:`sklearn.tree.plot_tree`:
+
+# %%
+from sklearn.tree import plot_tree
+
+plt.figure(dpi=300)
```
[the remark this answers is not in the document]
sylvaincom: Ok, I remove this

examples/use_cases/plot_feature_importance.py
sylvaincom: Done

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# The permutation importance is often calculated several times, each time
+# with different permutations of the feature.
+# Hence, we can have measure its variance (or standard deviation).
+# Now, we plot the permutation feature importance along with its standard deviation:
+
+# %%
+df = (
+    ridge_report.feature_importance.feature_permutation()
+    .aggregate(["mean", "std"], axis=1)
+    .plot.barh(y="mean", xerr="std")
```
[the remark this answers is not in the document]
sylvaincom: indeed, should be good now

examples/use_cases/plot_feature_importance.py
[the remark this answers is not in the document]
glemaitre: Uhm interesting. I need to check again. We might have right it in the base class and both the classifier and regressor inherit from it. Let me check to be sure that I don't say something wrong then.

examples/use_cases/plot_feature_importance.py
sylvaincom: You could like something like this?
```python

n_estimators = 100 
rf_report = EstimatorReport(
    RandomForestRegressor(random_state=0, n_estimators= n_estimators),
    X_train=X_train,
    X_test=X_test,
    y_train=y_train,
    y_test=y_test,
)
```

I like fetching the n_estimators so it shows the default value of 100, thus what is recommended, that's already huge

examples/use_cases/plot_feature_importance.py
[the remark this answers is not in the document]
glemaitre: OK so, it people still call it Gini importance even in regression while it is linked to the Gini index. I assume it is one of those thing that make no sense. I would advocate to not mention it here to avoid any confusion :) You have another Gini score or something like that is something completely different. It is hard when the same researcher did several things :)

examples/use_cases/plot_feature_importance.py
sylvaincom: So it would be this?

I) loading data and EDA
II) Linear models
for each model, we do coefficients then permutation importance
III) Tree-based models
for each model, we do MDI then permutation importance
IV) conclusion

glemaitre: Ah no, what I meant is section 4., should be permutation importance but inside this section, I would not make two sections linear models and tree based models. I would play with both model directly.

glemaitre:
@sylvaincom feel free to ping me when you have address the comments. I can give a quick look such that we can merge it and the subsequent PR with the implementation of the feature importance.

examples/use_cases/plot_feature_importance.py
sylvaincom: Done, thanks

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# %%
+tree_report.feature_importance.mean_decrease_impurity().plot.barh(
+    title="Feature importance",
+    xlabel="MDI",
+    ylabel="Feature",
+)
+plt.tight_layout()
+
+# %%
+# The MDI (Mean Decrease in Impurity) is a measure used in decision trees to quantify
```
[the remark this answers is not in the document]
sylvaincom: Ok, I simplified it

examples/use_cases/plot_feature_importance.py
```diff
@@ -571,15 +575,411 @@ def unscale_coefficients(df, feature_mean, feature_std):
...
+# the importance of each feature in making predictions.
+# MDI calculates how much a feature contributes to reducing the impurity (or error)
```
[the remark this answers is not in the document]
sylvaincom: Ok, I simplified it

examples/use_cases/plot_feature_importance.py
sylvaincom: @glemaitre? 🙏

examples/use_cases/plot_feature_importance.py
sylvaincom: @glemaitre what exactly would you drop? 🙏 The decision tree? The random forest? The MLP neural network?

glemaitre: The MLP regressor below.

examples/use_cases/plot_feature_importance.py
glemaitre: I would find easier to access fitted attributes, ending with `_` instead of parameter because usually user specify them.

examples/use_cases/plot_feature_importance.py
sylvaincom: How can I choose to compute the permutation on the train and test sets?

cc @auguste-probabl

examples/use_cases/plot_feature_importance.py
sylvaincom: Done, removed

examples/use_cases/plot_feature_importance.py
sylvaincom: Ok, done

examples/use_cases/plot_feature_importance.py
sylvaincom: I just removed the subsection titles

examples/use_cases/plot_feature_importance.py
sylvaincom: Should be something like:
```python
ridge_report.feature_importance.feature_permutation(data_source="train")
```

PR #1392, opened 2025-03-06
fix: Make ci/pip-compile.sh work on OSX and from anywhere

thomass-dev:


auguste-probabl:
or switch to pixi ? XD

PR #1394, opened 2025-03-07
fix(prediction error plot): Do not cache plot if `random_state` is None

### LINKED ISSUE #1379: Bug when caching `report.metrics.prediction_error()` with `random_state=None`
When calling `report.metrics.prediction_error(random_state=None)`, we use `None` as a key to cache the results. However, `None` should impose no caching in this case.

So we should revisit and do not cache the result when `random_state is None`.

auguste-probabl:
Closes #1379


glemaitre:
Could you fix also the cross-validation report. The diff will be pretty small as well.

auguste-probabl:
> Could you fix also the cross-validation report. The diff will be pretty small as well.

done, also fixed ComparisonReport.

glemaitre (approved):

glemaitre:
LGTM. Thanks @auguste-probabl

PR #1401, opened 2025-03-09
enh: Alternative to get/put

### LINKED ISSUE #1269: rfc (api): Revisit the API of `put` and `get`
In the past, we had many discussions where @GaelVaroquaux or myself complain with the needs to use `.put` and specified the name of an item to store it (and get it).

I came https://github.com/probabl-ai/skore/issues/949 to mention that I wanted my `CrossValidationReport` accessible from `project` such that this operation happen. However, one limitation as mentioned by @auguste-probabl is that native Python type cannot benefit from this API. There is also discussion that you can use the same class into different ways that could be confusing.

In an IRL with @adrinjalali, an API that seems much better would be to have a `dict`/`Bunch` UX to deal with this use case:

```python
project.put("my_int, 1)
```

would be replaced by:

```
project.artifacts.my_int = 1
```

or for a report

```python
report = CrossValidationReport(...)
project.put("estimator_report, report)
```

would be replaced by:

```python
project.artifacts.estimator_report = CrossValidateRerport(...)
```

To access, the latest version of an items:

```python
# pushing to the extreme without storing the artifact in another variable
project.artifacts.estimator_report.metrics.roc.plot()
```

would work.

All of those would work if we consider that `skore.open` set implicitly the "run" for which we store and access the items. We would still need `.get` if one wants to access an item from a different `run` where we come back to the discussion on having `skore.get_metadata` or similar to be able to query the database.

glemaitre:
closes #1269 

Really wip just as an exploration.

### DIFF
--- a/examples/getting_started/plot_skore_getting_started.py
+++ b/examples/getting_started/plot_skore_getting_started.py
@@ -268,9 +268,9 @@ my_project = skore.Project("my_project")
 # directory) using :func:`~skore.Project.put`), with a "universal" key-value convention:
 
 # %%
-my_project.put("my_int", 3)
-my_project.put("df_cv_report_metrics", df_cv_report_metrics)
-my_project.put("roc_plot", roc_plot)
+my_project.artifacts.my_int = 3
+my_project.artifacts.df_cv_report_metrics = df_cv_report_metrics
+my_project.artifacts.roc_plot = roc_plot
 
 # %%
 # .. note ::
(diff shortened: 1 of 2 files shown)


thomass-dev:
@glemaitre I'd advise you to stay at a high level in your exploration, as we will surely evolve the implementation between several packages..

glemaitre:
@thomass-dev it is just a POC to see what Python allows me to do, what is missing and what is the feeling :).

thomass-dev:

auguste-probabl:
Closing as it's likely outdated; feel free to re-open.

glemaitre:
Yep I can make a second POC since we already have the metadata now.

PR #1402, opened 2025-03-10
ci: Ensure mypy checks are valid

thomass-dev:
Missing in https://github.com/probabl-ai/skore/pull/1337, call `mypy` from `pre-commit` in the GH pipelines.


rouk1 (approved):
💪🏻 types

PR #1410, opened 2025-03-11
test: Make it possible to ignore a whole block of doctests

### LINKED ISSUE #1347: docs: In docstrings, enable to ignore block of code when running doctests

Following [this discussion](https://github.com/probabl-ai/skore/pull/1333#discussion_r1959262300), in the docstrings, we want to ignore an entire block of code when running doctests.

Currently, we can only ignore line by line by using `# +SKIP`

### Suggested improvement

Use https://github.com/scipy/scipy_doctest

### Additional context

_No response_

auguste-probabl:
The PR introduces a new test dependency, [xdoctest](https://github.com/Erotemic/xdoctest), a modern rewrite of `doctest`. xdoctest makes it possible to apply a (x)doctest directive to all following lines. Example from [xdoctest's own docs](https://xdoctest.readthedocs.io/en/latest/auto/xdoctest.directive.html):

```python
>>> # An inline directive appears on the same line as a command and
>>> # only applies to the current line.
>>> raise AssertionError('this will not be run (a)')  # xdoctest: +SKIP
>>> print('This line will print: (A)')
>>> print('This line will print: (B)')
>>> # However, if a directive appears on its own line, then it applies
>>> # too all subsequent lines.
>>> # xdoctest: +SKIP()
>>> raise AssertionError('this will not be run (b)')
>>> print('This line will not print: (A)')
>>> # Note, that SKIP is simply a state and can be disabled to allow
>>> # the program to continue executing.
>>> # xdoctest: -SKIP
>>> print('This line will print: (C)')
>>> print('This line will print: (D)')
>>> # This applies to inline directives as well
>>> # xdoctest: +SKIP("an assertion would occur")
>>> raise AssertionError('this will not be run (c)')
>>> print('This line will print: (E)')  # xdoctest: -SKIP
>>> raise AssertionError('this will not be run (d)')
>>> # xdoctest: -SKIP("a reason can be given as an argument")
>>> print('This line will print: (F)')
```

Closes #1347

### DIFF
--- a/skore/ci/requirements/python-3.10/scikit-learn-1.6/test-requirements.txt
+++ b/skore/ci/requirements/python-3.10/scikit-learn-1.6/test-requirements.txt
@@ -189,3 +189,5 @@ urllib3==2.3.0
     # via requests
 virtualenv==20.29.1
     # via pre-commit
+xdoctest==1.2.0
+    # via skore (skore/pyproject.toml)
--- a/skore/ci/requirements/python-3.11/scikit-learn-1.6/test-requirements.txt
+++ b/skore/ci/requirements/python-3.11/scikit-learn-1.6/test-requirements.txt
@@ -184,3 +184,5 @@ urllib3==2.3.0
     # via requests
 virtualenv==20.29.1
     # via pre-commit
+xdoctest==1.2.0
+    # via skore (skore/pyproject.toml)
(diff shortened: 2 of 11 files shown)


auguste-probabl:
`skore/ci/pip-compile.sh` doesn't work on NixOS so please review the PR and if it's approved I'll look for a way to re-generate requirements.

thomass-dev (commented):
The latest version of `xdoctest` is 7 months old, so even if the last commit is a month old, is the package really maintained?

This must be taken in account if it replaces a module from the standard lib.

thomass-dev (approved):
LGTM, but a bit afraid to replace `doctest` from the standard lib if `xdoctest` is not actively maintained.

We should keep a eye on this package, and rollback if needed.

glemaitre:
Arff, I was going to comment. Why not: https://github.com/scipy/scipy_doctest It is maintained by SciPy to which we have a large overlap in terms of development practices?

PR #1429, opened 2025-03-13
chore: Replace the github default blank issue with template

MarieSacksick:


thomass-dev:
Please don't use private link on public PR, thanks!

PR #1432, opened 2025-03-14
chore: Add dependencies for docs

### LINKED ISSUE #1426: docs: Ad `nbformat` to `pyproject.toml` for the docs

executing some examples locally from the raw python file

`nbformat` not installed

### Suggested improvement

install it by default when doing the `make` for contributing

### Additional context

_No response_

MarieSacksick:
Closes #1426   

Add 2 dependencies to be able to run .py examples locally in the IDE.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 install-skore:
-	python -m pip install -e './skore[test,sphinx]'
+	python -m pip install -e './skore[test,sphinx,dev]'
 	pre-commit install
 
 lint:
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -67,6 +67,8 @@ sphinx = [
   "sphinx-tabs",
   "sphinx<8.2.0",
   "sphinx_autosummary_accessors",
+]
+dev = [
   "nbformat",
   "ipykernel",
 ]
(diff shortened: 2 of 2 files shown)


MarieSacksick:
I'd like the confirmation of an software eng to confirm that it's worth it.

auguste-probabl:
do we use nbformat anywhere in our pipeline? I think ruff can everything it does. For ipykernel i'm not sure but i would say it's fine.

MarieSacksick:
We don't need `nbformat` in the pipeline, but for some parts of the examples (using plotly). The code does not fail, it's not blocking, but it displays this:  
image

glemaitre:
Actually, I would consider those as `dev` tools. They are handy in our development workflow but as stated above we don't use it in the doc workflow.

Having `ipython`, `jupyter-lab` or `ipykernel` is kind of handy when developing but it is only optional.

We took this direction in `skrub` (and other projects as well): https://github.com/skrub-data/skrub/blob/main/pyproject.toml#L42-L47

MarieSacksick:
It's a dev exp issue.   

In the contributing guide, we suggest to go by `make install-skore`, so this is what I do with new venv.  

I added a new "sub-menu" in the optional dependencies as advised.

thomass-dev (approved):

PR #1435, opened 2025-03-14
chore: Reorder issues template

MarieSacksick:
Goal: have the blank issue at the bottom.  
It's order alphabetically, therefore it was at the top, creating an incentive to use instead of the other templates, while it's supposed to be a back-up.  

Current:  
image


thomass-dev (changes requested):
Please pad the order with at least two digits, and set the blank-one at 99, in case we will have more than 10 templates :).

MarieSacksick:
Done!  
Although I really hope we will never reach 10 templates, that would be a lot...

thomass-dev (approved):

thomass-dev:
@MarieSacksick can you resolve the merge-conflict? Thanks!

Issue #1391, opened 2025-03-06
enh: Use fast hash function in cache


auguste-probabl:
> I understand that hashes are computed on the full datasets, I was a bit worried that it could be slow for large datasets (ranging from 100MBs to GBs ?) and that the hash step could itself be memory-intensive. Did you benchmark the hash functions that are used, does it scale to bigger datasets ?

> I was thinking that it could be useful later on, to offer users the possibility to use their custom hashing function, especially if some datasets (e.g. pandas dataframe with object column) are not hashable.

I think that we should use a better hash function like XXH3 (https://xxhash.com/). Since the hash compute 30GB/s then it should not be a bottleneck anymore. MD5 is indeed slow.

_Originally posted by @glemaitre in https://github.com/probabl-ai/skore/discussions/1384#discussioncomment-12390675_

auguste-probabl:
I'd like to take the opportunity to do some performance testing with large datasets.

thomass-dev:
I'm interested in the results of your investigation!

glemaitre:
> Would mumurhash (used by scikit-learn, and hence not leading to an extra dependency) be an option?

It should be x3 faster than the MD5 that we have in scikit-learn but x10 slower than the xxh3.

But if we already have a bandwidth of 4 GB/s with murmur, it might be enough. For sure we should bench a bit to know what are the trade-off that we currently have and if the slowness make the tool unusable (which I'm not sure).

GaelVaroquaux:
> But if we already have a bandwidth of 4 GB/s with murmur, it might be enough. For sure we should bench a bit to know what are the trade-off that we currently have and if the slowness make the tool unusable (which I'm not sure).

I'm just really worried about an additional dependency

cakedev0:
We mostly don't hash anything anymore for caching now. So we can maybe close this issue?

The questions around hashing datasets will resurface though, and we will want to remember this link then: https://pandas.pydata.org/docs/reference/api/pandas.util.hash_pandas_object.html#pandas-util-hash-pandas-object

glemaitre:
I think that we can indeed close this issue for the time being. Also, we should really consider whether hashing is a real bottleneck because it appears that we would like to avoid having dependency not easily installable in WASM.

PR #1438, opened 2025-03-14
perf: Benchmark hash functions

auguste-probabl:
This PR aims to study the performance impact of the hash function used for caching.

The cache uses a variation of `joblib.hash`, which had to be generalized to make third party hash algorithms fit.
The hash function can be configured using `skore.set_config/config_context`, as seen in the performance tests.

Additionally, a wrapper around `sklearn.utils.murmurhash3_32` was written to accomodate the `joblib.hash` API.

Results
```
```

To reproduce:
```sh
pip install -e .[test]  # will install xxhash and pytest-benchmark
python -m pytest tests/perf/test_perf.py::test_hashh
```

---

- **vendor joblib.hash**
- **extend vendored joblib.hash to any hash object**
- **change _hash behaviour without passing arguments, using config**
- **benchmark**
- **add xxhash to benchmark**
- **add murmurhash from sklearn to benchmark**

Addresses #1391


auguste-probabl:
Closing this to avoid polluting, but the branch will be kept in my fork for future reference.

PR #1440, opened 2025-03-16
feat: Change the default of `aggregate` parameter for cross-validation report

thomasloux:
Following the corresponding issue, I change the default behaviour of CrossValidationReport to aggregate fold metrics as mean and std, rather than report metrics for all folds.

I choose to use Ellipsis as default value to make the difference with ``aggregate=None`` that outputs metrics for all folds.

I've also updated the tests, most are just column names, 1 is actual values, 2 behaviours in case of interruption (you may want to check these ones, as I am less familiar with the expected behaviour of the object is this setting).

Docstring still needs to be changed.


thomass-dev:
Nitpick: on form, and this is open to debate, i find it more explicit to use a dedicated sentinel `DEFAULT=object()` instead of the ellipsis when `None` is a value that makes sense.

thomasloux:
I wasn't sure how to write this, so of course it can be changed!

glemaitre:
I think that we can pass the default `aggregate=("mean", "std")` in the signature. It means that we can relax the type by accepting list or tuple of literal.

thomasloux:
updated, did not think about this simpler possibility. Sequence of str also makes more sense, as more aggregate options exists in pandas than mean and std

glemaitre:
Now, we should amend a couple of things:

- change the docstring of the metric to reflect the default
- change the output of the "Examples" section of the metric
- make a pass in the example gallery to make sure that the discussion are aligned with what we compute: whenever possible, remove the `aggregate=["mean", "std"]` since it is the default and when we relied on `aggregate=None`, then we should explicitly use it.

glemaitre:
@thomasloux would you be able to make the latest requested changes?

thomasloux:
Hello, I've changed the doctstring. Not sure what you mean by "Examples" section if not the one in the docstring. You may want to check that briefly if you want to check exact formatting (where to truncate results...). For all metrics, I've remove any argument link with ``aggregate`` and report then report the results with new behavior, so: mean and std. But you may want to provide one with aggregate=None to show this specific behaviour.

Additionally, here is the new description of the argument:
aggregate : None or list of str, default=("mean", "std")
            Function to aggregate the scores across the cross-validation splits.
            None will return the scores for each split.

Let me know if you have suggestion on these change!

thomasloux:
Hey, seems like I've forgotten to push a previous commit where I also changed docstrings, sorry. To be sure, don't you want to emphasize the ``aggregate=None`` in docstring as well. I feel like this one is slightly less clear
```
aggregate : {"mean", "std"} or list of such str, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
```

glemaitre:
@thomasloux Indeed, it was the default (and this no need to specify it) before and I forgot to add it explicitly:

```
aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
```

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -100,8 +101,9 @@ def report_metrics(
             Whether to flatten the `MultiIndex` columns. Flat index will always be lower
             case, do not include spaces and remove the hash symbol to ease indexing.
 
-        aggregate : {"mean", "std"} or list of such str, default=None
+        aggregate : None or list of str, default=("mean", "std")
```
glemaitre, line 104: It would be my prefer writing.

```suggestion
        aggregate : {"mean", "std}, list of such str or None, default=("mean", "std")
```

thomasloux: updated all docstring

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -151,9 +152,7 @@ def _compute_metric_scores(
         report_metric_name: str,
         *,
         data_source: DataSource = "test",
-        aggregate: Optional[
-            Union[Literal["mean", "std"], list[Literal["mean", "std"]]]
-        ] = None,
+        aggregate: Optional[Sequence[str]] = None,
```
glemaitre, line 155: The typing seems wrong here. I see that in the code we check if it is a string to make it a sequence so it would mean that previous typing was correct.=

glemaitre (dismissed):
Otherwise LGTM. Thanks @thomasloux

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -337,7 +337,8 @@ def _check_results_single_metric(report, metric, expected_n_splits, expected_nb_
 
     # check that the columns contains the expected split names
     split_names = result.columns.get_level_values(1).unique()
-    expected_split_names = [f"Split #{i}" for i in range(expected_n_splits)]
+    # expected_split_names = [f"Split #{i}" for i in range(expected_n_splits)]
```
glemaitre, line 340: You can remove this line as well.

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -373,7 +374,8 @@ def _check_results_report_metric(
...
```
glemaitre, line 377: and this one as well.

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -337,7 +337,8 @@ def _check_results_single_metric(report, metric, expected_n_splits, expected_nb_
 
     # check that the columns contains the expected split names
     split_names = result.columns.get_level_values(1).unique()
```
glemaitre, line 339: Let's rename this variable `aggregate_names`

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -373,7 +374,8 @@ def _check_results_report_metric(
...
```
glemaitre, line 376: Let's rename this variable `aggregate_names`

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -373,7 +374,8 @@ def _check_results_report_metric(
...
-    expected_split_names = [f"Split #{i}" for i in range(expected_n_splits)]
+    # expected_split_names = [f"Split #{i}" for i in range(expected_n_splits)]
+    expected_split_names = ["mean", "std"]
```
glemaitre, line 378: and rename also this variable `expected_aggregate_names`

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -337,7 +337,8 @@ def _check_results_single_metric(report, metric, expected_n_splits, expected_nb_
...
```
glemaitre, line 341: and rename also this variable `expected_aggregate_names`

skore/tests/unit/sklearn/test_cross_validation.py
glemaitre, line 340: Actually, instead of changing the test here, let's change the value of aggregate:

```python
result = getattr(report.metrics, metric)(aggregate=None)
```

We have other location where we already check the behaviour of requesting the mean and std

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -673,9 +675,12 @@ def test_cross_validation_report_report_metrics_with_scorer(regression_data):
         ]
         for est_rep in report.estimator_reports_
     ]
+    expected_result = np.transpose(expected_result)
```
glemaitre, line 678: let's do the same here, let's request `aggregate=None` to not change the test.

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
glemaitre: Indeed ;)

glemaitre:
I just pushed a small fix in your branch to fix the conflict to be able to run the tests. It looks good to me otherwise.

@thomasloux Next time, you might want to make a branch instead of submitting from your `main` branch.

glemaitre (approved):

glemaitre:
Thanks @thomasloux. The CIs is happy so I am.

Nice job.

PR #1444, opened 2025-03-18
feat(EstimatorReport): Add `fit_time` and `predict_time` metrics

auguste-probabl:
Closes #1241

- **add time measurement context manager**
- **add fit_time metric**
- **add predict_time metric**

PR #1454, opened 2025-03-21
feat(EstimatorReport): Add `timing` metrics

### LINKED ISSUE #1241: feat: Add execution time (`fit_time`/`score_time`) to the `EstimatorReport` and `CrossValidationReport` helpers

Currently (skore v0.6.1), in the helpers of the `EstimatorReport` and the `CrossValidationReport`, we give access to metrics that contain scores, but no execution time.

As a data scientist, when assessing the performance of my model, I would like to have access to the score but also the execution time.

In the helpers, add the fit time and also the score time (prediction / inference).

The fit time should be straightforward. For the score time, we need to have at least one prediction computed to have access to it.

_No response_

### Additional context

_No response_

auguste-probabl:
### Example

```python
>>> from sklearn.datasets import load_breast_cancer
>>> from sklearn.linear_model import LogisticRegression
>>> from skore import train_test_split
>>> from skore import EstimatorReport

>>> data = train_test_split(*load_breast_cancer(return_X_y=True), as_dict=True)
>>> report = EstimatorReport(LogisticRegression(), **data)

>>> # Reports all recorded times
>>> # Here just the fit time
>>> report.metrics.timings()
{'fit_time': 0.012448247987776995}

>>> report.metrics.accuracy()
>>> # Now that we've computed predictions, the key "predict_time_test" will appear
>>> report.metrics.timings()
{'fit_time': 0.012448247987776995, 'predict_time_test': 0.00036852300399914384}

>>> # Also 
>>> report.metrics.report_metrics()
                              LogisticRegression
Metric       Label / Average
Precision    0                          0.979167
             1                          0.936842
Recall       0                          0.886792
             1                          0.988889
ROC AUC                                 0.989099
Brier score                             0.037032
Fit time                                0.012448
Predict time                            0.000311
```

### Design

Contrary to #1444, the timing info is all in `_cache`. I had to update tests to accomodate for this.

### Todo

- [ ] Check coverage

---

Closes #1241
Supersedes #1444

### DIFF
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -14,7 +14,7 @@ from sklearn.base import BaseEstimator
 from sklearn.utils._response import _check_response_method, _get_response_values
 
 from skore.externals._sklearn_compat import is_clusterer
-from skore.utils._measure_time import _measure_time
+from skore.utils._measure_time import MeasureTime
 
 
 class _HelpMixin(ABC):
@@ -395,7 +395,7 @@ def _get_cached_response_values(
         assert isinstance(cached_predictions, np.ndarray)
         return cached_predictions
 
-    with _measure_time() as predict_time:
+    with MeasureTime() as predict_time:
         predictions, _ = _get_response_values(
             estimator,
             X=X,
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -14,6 +14,7 @@ from sklearn.base import BaseEstimator
 from sklearn.utils._response import _check_response_method, _get_response_values
 
 from skore.externals._sklearn_compat import is_clusterer
+from skore.utils._measure_time import _measure_time
 
 
 class _HelpMixin(ABC):
@@ -398,13 +399,16 @@ def _get_cached_response_values(
         assert isinstance(cached_predictions, np.ndarray)
         return cached_predictions
 
-    predictions, _ = _get_response_values(
-        estimator,
-        X=X,
-        response_method=prediction_method,
-        pos_label=pos_label,
-        return_response_method_used=False,
-    )
+    with _measure_time() as predict_time:
+        predictions, _ = _get_response_values(
+            estimator,
+            X=X,
+            response_method=prediction_method,
+            pos_label=pos_label,
+            return_response_method_used=False,
+        )
+
     cache[cache_key] = predictions
+    timings_cache[cache_key + ("predict_time",)] = predict_time()
 
     return predictions
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -321,7 +321,8 @@ class _BaseAccessor(_HelpMixin, Generic[ParentT]):
 
 def _get_cached_response_values(
     *,
-    cache: dict[tuple[Any, ...], ArrayLike],
+    cache: dict[tuple[Any, ...], Any],
+    timings_cache: dict[tuple[Any, ...], Any],
     estimator_hash: int,
     estimator: BaseEstimator,
     X: Union[ArrayLike, None],
@@ -336,6 +337,9 @@ def _get_cached_response_values(
     ----------
     cache : dict
         The cache to use.
+
+    timings_cache : dict
+        The cache to use for time measurement results.
 
     estimator_hash : int
         A hash associated with the estimator such that we can retrieve the data from
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -382,13 +382,12 @@ def _get_cached_response_values(
     else:
         cache_key = (estimator_hash, None, prediction_method, data_source)
 
-    if data_source == "X_y":
-        if data_source_hash is None:
-            # Only trigger hash computation if it was not previously done.
-            # If data_source_hash is not None, we internally computed ourself the hash
-            # and it is trustful
-            data_source_hash = joblib.hash(X)
-        cache_key = cache_key + (data_source_hash,)
+    if data_source == "X_y" and data_source_hash is None:
+        # Only trigger hash computation if it was not previously done.
+        # If data_source_hash is not None, we internally computed ourself the hash
+        # and it is trustful
+        data_source_hash = joblib.hash(X)
+    cache_key = cache_key + (data_source_hash,)
 
     if cache_key in cache:
         cached_predictions = cache[cache_key]
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -405,6 +405,13 @@ def _get_cached_response_values(
         )
 
     cache[cache_key] = predictions
-    cache[cache_key + ("predict_time",)] = predict_time()
+
+    predict_time_cache_key: tuple[Any, ...] = (
+        estimator_hash,
+        data_source,
+        data_source_hash,
+        "predict_time",
+    )
+    cache[predict_time_cache_key] = predict_time()
 
     return predictions
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -370,6 +370,13 @@ def _get_cached_response_values(
         The response values.
     """
     prediction_method = _check_response_method(estimator, response_method).__name__
+
+    if data_source == "X_y" and data_source_hash is None:
+        # Only trigger hash computation if it was not previously done.
+        # If data_source_hash is not None, we internally computed ourself the hash
+        # and it is trustful
+        data_source_hash = joblib.hash(X)
+
     if prediction_method in ("predict_proba", "decision_function"):
         # pos_label is only important in classification and with probabilities
         # and decision functions
@@ -378,16 +385,16 @@ def _get_cached_response_values(
             pos_label,
             prediction_method,
             data_source,
+            data_source_hash,
         )
     else:
-        cache_key = (estimator_hash, None, prediction_method, data_source)
-
-    if data_source == "X_y" and data_source_hash is None:
-        # Only trigger hash computation if it was not previously done.
-        # If data_source_hash is not None, we internally computed ourself the hash
-        # and it is trustful
-        data_source_hash = joblib.hash(X)
-    cache_key = cache_key + (data_source_hash,)
+        cache_key = (
+            estimator_hash,
+            None,
+            prediction_method,
+            data_source,
+            data_source_hash,
+        )
 
     if cache_key in cache:
         cached_predictions = cache[cache_key]
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -377,24 +377,18 @@ def _get_cached_response_values(
         # and it is trustful
         data_source_hash = joblib.hash(X)
 
-    if prediction_method in ("predict_proba", "decision_function"):
+    if prediction_method not in ("predict_proba", "decision_function"):
         # pos_label is only important in classification and with probabilities
         # and decision functions
-        cache_key: tuple[Any, ...] = (
-            estimator_hash,
-            pos_label,
-            prediction_method,
-            data_source,
-            data_source_hash,
-        )
-    else:
-        cache_key = (
-            estimator_hash,
-            None,
-            prediction_method,
-            data_source,
-            data_source_hash,
-        )
+        pos_label = None
+
+    cache_key: tuple[Any, ...] = (
+        estimator_hash,
+        pos_label,
+        prediction_method,
+        data_source,
+        data_source_hash,
+    )
 
     if cache_key in cache:
         cached_predictions = cache[cache_key]
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -323,7 +323,6 @@ class _BaseAccessor(_HelpMixin, Generic[ParentT]):
 def _get_cached_response_values(
     *,
     cache: dict[tuple[Any, ...], Any],
-    timings_cache: dict[tuple[Any, ...], Any],
     estimator_hash: int,
     estimator: BaseEstimator,
     X: Union[ArrayLike, None],
@@ -338,9 +337,6 @@ def _get_cached_response_values(
     ----------
     cache : dict
         The cache to use.
-
-    timings_cache : dict
-        The cache to use for time measurement results.
 
     estimator_hash : int
         A hash associated with the estimator such that we can retrieve the data from
@@ -409,6 +405,6 @@ def _get_cached_response_values(
         )
 
     cache[cache_key] = predictions
-    timings_cache[cache_key + ("predict_time",)] = predict_time()
+    cache[cache_key + ("predict_time",)] = predict_time()
 
     return predictions
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -452,7 +452,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
 
         kwargs are available for compatibility with other metrics.
         """
-        return self._parent._cache.get(self._parent._fit_time_cache_key())
+        return self._parent.fit_time_
 
     def _predict_time(
         self,
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -163,7 +163,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
                     scoring += ["_roc_auc", "_log_loss"]
             else:
                 scoring = ["_r2", "_rmse"]
-            scoring += ["fit_time", "predict_time"]
+            scoring += ["_fit_time", "_predict_time"]
 
         if scoring_names is not None and len(scoring_names) != len(scoring):
             if scoring_was_none:
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -475,14 +475,14 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
             "predict_time",
         )
 
-        return self._parent._cache.get(predict_time_cache_key, default=None)
+        return self._parent._cache.get(predict_time_cache_key, None)
 
     def timings(self) -> dict:
         """Get all measured processing times related to the estimator.
 
-        When an estimator is fitted inside the :class:`~skore.EstimatorReport`, the time to fit is
-        recorded. Similarly, when predictions are computed on some data, the time to
-        predict is recorded. This function returns all the recorded times.
+        When an estimator is fitted inside the :class:`~skore.EstimatorReport`, the time
+        to fit is recorded. Similarly, when predictions are computed on some data, the
+        time to predict is recorded. This function returns all the recorded times.
 
         Returns
         -------
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -448,7 +448,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
         return score
 
     def _fit_time(self, **kwargs) -> Union[float, None]:
-        """Compute time to fit the estimator.
+        """Get time to fit the estimator.
 
         kwargs are available for compatibility with other metrics.
         """
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -478,7 +478,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
         return self._parent._cache.get(predict_time_cache_key, default=None)
 
     def timings(self) -> dict:
-        """Get all measured times.
+        """Get all measured processing times related to the estimator.
 
         When an estimator is fitted inside the :class:`~skore.EstimatorReport`, the time to fit is
         recorded. Similarly, when predictions are computed on some data, the time to
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -172,7 +172,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
                     "The `scoring_names` parameter should be of the same length as "
                     "the `scoring` parameter. In your case, `scoring` was set to None "
                     f"and you are using our default scores that are {len(scoring)}. "
-                    "The list is the following: {scoring}."
+                    f"The list is the following: {scoring}."
                 )
             else:
                 raise ValueError(
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -487,7 +487,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
         -------
         timings : dict
             The recorded times, in seconds,
-            in the form of a dict with some or all of the following keys:
+            in the form of a `dict` with some or all of the following keys:
 
             - "fit_time", for the time to fit the estimator in the train set.
             - "predict_time_{data_source}", where data_source is "train", "test" or
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -489,6 +489,85 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
             )
         return {"fit": fit_time}
 
+    def _predict_time(
+        self,
+        *,
+        data_source: DataSource = "test",
+        X: Optional[ArrayLike] = None,
+        y: Optional[ArrayLike] = None,
+    ) -> dict:
+        """Get the time taken to compute the estimator's `predict` method.
+
+        Will compute the predictions if necessary.
+
+        Examples
+        --------
+        >>> from sklearn.datasets import load_breast_cancer
+        >>> from sklearn.linear_model import LogisticRegression
+        >>> from sklearn.model_selection import train_test_split
+        >>> from skore import EstimatorReport
+        >>> X_train, X_test, y_train, y_test = train_test_split(
+        ...     *load_breast_cancer(return_X_y=True), random_state=0
+        ... )
+        >>> classifier = LogisticRegression(max_iter=10_000)
+        >>> report = EstimatorReport(
+        ...     classifier,
+        ...     X_train=X_train,
+        ...     y_train=y_train,
+        ...     X_test=X_test,
+        ...     y_test=y_test,
+        ... )
+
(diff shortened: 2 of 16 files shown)


skore/src/skore/utils/_measure_time.py
```diff
@@ -0,0 +1,22 @@
...
+    Callable returning a float
+
+    Examples
+    --------
+    >>> with _measure_time() as time_taken:
+    ...     1+1
+    >>> time_taken()  # Note: time_taken is a callable
+    """
+    t1 = t2 = perf_counter()
+    yield lambda: t2 - t1
```
glemaitre, line 21: Because a `lambda`, I think that we are going to get into trouble when we will want to pickle the class where it is used.

Quickly asking to a GPT:

```python
import pickle
from time import perf_counter, sleep
from contextlib import contextmanager

@contextmanager
def catchtime():
    t1 = t2 = perf_counter() 
    yield lambda: t2 - t1
    t2 = perf_counter()

# Example class that uses our context manager
class TimedFunction:
    def __init__(self):
        self.timer = None

    def run(self):
        with catchtime() as t:
            sleep(0.1)  # Simulate work
            self.timer = t  # Store the timer function

    def get_execution_time(self):
        return self.timer()  # Call the lambda

# Create and use the class
my_function = TimedFunction()
my_function.run()
print(f"Execution time: {my_function.get_execution_time():.6f} seconds")

# Try to pickle the object
try:
    pickle_data = pickle.dumps(my_function)
    print("Successfully pickled the object")
except Exception as e:
    print(f"Pickling failed: {type(e).__name__}: {e}")
```

We could refactor by having a real callable:

```python
import pickle
from time import perf_counter, sleep
from contextlib import contextmanager

class TimingResult:
    """A callable class to replace the lambda function."""
    def __init__(self, start, end=None):
        self.start = start
        self.end = end

    def __call__(self):
        """Makes instances callable like functions."""
        return self.end - self.start

@contextmanager
def catchtime():
    """Context manager that yields a callable timing object."""
    result = TimingResult(perf_counter())
    yield result
    result.end = perf_counter()

# Example class that uses our context manager
class TimedFunction:
    def __init__(self):
        self.timer = None

    def run(self):
        with catchtime() as t:
            sleep(0.1)  # Simulate work
            self.timer = t  # Store the timer function

    def get_execution_time(self):
        return self.timer()  # Call the timing result object

# Create and use the class
my_function = TimedFunction()
my_function.run()
print(f"Execution time: {my_function.get_execution_time():.6f} seconds")

# Verify that the object can be pickled
try:
    # Pickle the object
    pickle_data = pickle.dumps(my_function)
    print("Successfully pickled the object!")

    # Unpickle to verify it works completely
    unpickled_function = pickle.loads(pickle_data)
    print(f"Unpickled execution time: {unpickled_function.get_execution_time():.6f} seconds")
except Exception as e:
    print(f"Pickling failed: {type(e).__name__}: {e}")

# Demonstrate that timing is preserved after external delays
sleep(2)  # This delay should not affect the timing result
print(f"Time after delay: {my_function.get_execution_time():.6f} seconds")
```

But then it might be easier to use a class directly:

```python
class Timer:
    """Class-based context manager for timing code execution."""
    def __enter__(self):
        self.start = perf_counter()
        self.end = self.start  # Initialize end time
        return self  # Return self so it can be used in the with statement

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.end = perf_counter()

    def __call__(self):
        """Makes the instance callable to get the elapsed time."""
        return self.end - self.start
```

auguste-probabl: Thanks for the suggestion; the class context manager is also fine. I would like to add a non-regression test but I can't reach the pickling issue you mention:
```python
>>> from sklearn.datasets import make_classification
>>> from sklearn.model_selection import train_test_split
>>> from sklearn.linear_model import LogisticRegression
>>> X, y = make_classification(random_state=42)
>>> X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)
>>> estimator = LogisticRegression().fit(X_train, y_train)
>>> from skore import EstimatorReport
>>> report = EstimatorReport(estimator, X_test=X_test, y_test=y_test)

# It just works
>>> pickle.loads(pickle.dumps(report)).metrics.timing()
{'predict_test': 0.0003976030002377229}
```
Can you think of a way to make the current implementation fail?

auguste-probabl: Discussed IRL: our code won't fail because it doesn't *return* lambdas, but still, the class implementation is entirely free of lambdas and is not more complicated, so might as well avoid that pitfall entirely.

glemaitre (commented):
a very partial review before that I restart one.

skore/src/skore/sklearn/_base.py
```diff
@@ -394,13 +395,16 @@ def _get_cached_response_values(
...
+        predictions, _ = _get_response_values(
+            estimator,
+            X=X,
+            response_method=prediction_method,
+            pos_label=pos_label,
+            return_response_method_used=False,
+        )
+
     cache[cache_key] = predictions
+    cache[cache_key + ("predict_time",)] = predict_time()
```
glemaitre, line 408: I'm thinking that we could simplify the cache key here:

```python
(estimator_hash, data_source, data_source_hash, "predict_time")
```

would be enough. We don't really care that we computed `predict` or `predict_proba` or `decision_function`. Usually the difference between them will be negligeable. This would have the benefit to simplify the search later

skore/src/skore/sklearn/_base.py
```diff
@@ -394,13 +395,16 @@ def _get_cached_response_values(
...
         return cached_predictions
 
-    predictions, _ = _get_response_values(
-        estimator,
-        X=X,
-        response_method=prediction_method,
-        pos_label=pos_label,
-        return_response_method_used=False,
-    )
+    with _measure_time() as predict_time:
```
glemaitre, line 398: Above, I see that we store the hash only when we need. I'm thinking that we could have key that does not change size and add `None` in the case that we don't need to compute the hash. It will also simplify the `find_in_cache` function.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+            # Search for first matching key in cache
+            if data_source_hash is None:
+                for key in self._parent._cache:
+                    if key[-2:] == (data_source, "predict_time"):
+                        return self._parent._cache[key]
+            else:
+                for key in self._parent._cache:
+                    if key[-3:] == (data_source, data_source_hash, "predict_time"):
+                        return self._parent._cache[key]
+            return None
```
glemaitre, line 547: Could you make it a private method of the class. By changing the code in the `_base.py`, I think that we can avoid the `if/else` and have a single branch`. Also, we could create they key instead to have to slice it:

```python
expected_key = (self._parent._estimator_hash, data_source, data_source_hash or None, "predict_time")
if expected_key in self._parent._cache:
    return ...
return None
```

I assume that we will get a performance gain when checking the keys of the `_cache` directly.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+        if predict_time is None:
+            # If the time was not found in the cache, compute predictions
+            _get_cached_response_values(
+                cache=self._parent._cache,
+                estimator_hash=self._parent._hash,
+                estimator=self._parent.estimator_,
+                X=X,
+                response_method="predict",
+                data_source=data_source,
+                data_source_hash=data_source_hash,
```
glemaitre, line 560: It might be worth to add `pos_label=None` to be explicit with a small comment that in binary case, it will default to `self._parent.estimator_.classes_[1]`

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+        y: Optional[ArrayLike] = None,
+    ) -> dict:
+        """Report fit and prediction times, in seconds.
+
+        The times are reported in the form of a dict of the following shape:
+
+        - If kind="fit", a single entry with key "fit"
+        - If kind="predict", a single entry with key "predict_{data_source}"
+          (where data_source is "train", "test" or "X_y_{data_source_hash}")
+        - If kind="auto", two entries: a concatenation of the "fit" and "predict" cases.
```
glemaitre, line 588: I would prefer to have this info below in the description of the return value.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+
+        Fit information might not be available, e.g. if the estimator was fitted outside
+        of the EstimatorReport. In this case, passing kind="fit" will result in an
+        error, and passing kind="auto" is equivalent to passing kind="predict".
```
glemaitre, line 592: I'm fine with this one. Can you mention that `"predict"` will potentially request the computation of the predictions if not available in the local cache.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+            validation target provided when creating the report.
+
+        Returns
+        -------
+        dict
+            The times.
+
+        Raises
+        ------
+        FitInformationUnavailableError
```
glemaitre, line 621: I would do a blend `ValueError`.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: You can keep the small sentence that we prepend the `data_source` info (hash or name). But I would do a single sentence.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+        # Fit information is not available
+        >>> classifier = LogisticRegression(max_iter=10_000).fit(X_train, y_train)
+        >>> report = EstimatorReport(classifier, X_test=X_test, y_test=y_test)
+
+        >>> report.metrics.timing()
+        {'predict_test': ...}
+
+        >>> report.metrics.timing(kind="fit")
+        Traceback (most recent call last):
+        skore.sklearn._estimator.metrics_accessor.FitInformationUnavailableError: ...
```
glemaitre, line 669: I would avoid this part. We don't need to go until raising the error.

skore/src/skore/sklearn/_estimator/report.py
```diff
@@ -144,6 +151,9 @@ def __init__(
 
         self._initialize_state()
 
+        if fit_time is not None:
+            self._cache[("fit_time",)] = fit_time
```
glemaitre, line 155: Maybe prepend the `estimator_hash` for consistency with other keys.

skore/src/skore/sklearn/_estimator/report.py
```diff
@@ -178,11 +188,16 @@ def clear_cache(self) -> None:
         ...     y_test=y_test,
         ... )
         >>> report.cache_predictions()
+        >>> assert len(report._cache) > 1  # cache is not empty
```
glemaitre, line 191: Can you add the info in the long summary that we don't remove this key related to fit.

skore/tests/unit/sklearn/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -13,6 +14,22 @@
 from skore import EstimatorReport
 
 
+@contextlib.contextmanager
+def check_changed(value):
```
glemaitre, line 18: Let's making the context manager specialized to the `cache`: `check_cached_changed` (only about renaming).

glemaitre (commented):
OK this is the remaining of the review. No drastic changes. It looks good.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+        ...     y_test=y_test
+        ... )
+        >>> report.metrics._fit_time()
+        {'fit': ...}
+
+        >>> estimator = LogisticRegression().fit(X_train, y_train)
+        >>> report = EstimatorReport(estimator, X_test=X_test, y_test=y_test)
+        >>> report.metrics._fit_time()
+        Traceback (most recent call last):
+        skore.sklearn._estimator.metrics_accessor.FitInformationUnavailableError: ...
```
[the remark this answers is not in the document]
auguste-probabl: This is meant more as a unit test than documentation. Even private methods can be documented I would say.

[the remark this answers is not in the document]
glemaitre: Oh you should write a small test :) We should not rely on the doctest for the test.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: As above, this is a test that happens to be in the documentation; I think it's good

skore/src/skore/sklearn/_estimator/metrics_accessor.py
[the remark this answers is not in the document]
auguste-probabl: What is the reasoning behind this? We run doctests just like normal tests

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: That function already raises a ValueError when `kind` is unrecognized; I think more specific errors are preferrable.

MarieSacksick (commented):
I have two comments, which are quite open:  
image

1. Shouldn't we add `_time` to the keys, so that the dict is stand-alone? Otherwise, the meaning of the keys are not so clear.
2. When using the parameters `kind = "auto", data_source = "test"`, it outputs the key `fit`. I find this misleading. We could either complete the key to `fit_train`, or change the behavior of auto -> it renders what makes sense according to the data source requested. If None specified, fit_train + predict_test is returned, if train is specified, fit_train + predict_train is returned, and if anything else (test or X_y), only predict_source is returned.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -26,6 +26,13 @@
 DataSource = Literal["test", "train", "X_y"]
 
 
+class FitInformationUnavailableError(Exception):
+    """The report does not contain fit information.
+
+    This can happen when the estimator is fitted before the report is created.
+    """
```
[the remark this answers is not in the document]
auguste-probabl: See below: where it is used, ValueError is already used to mean something else.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -448,6 +455,238 @@ def _compute_metric_scores(
...
+        return {"fit": fit_time}
+
+    def _predict_time(
+        self,
+        *,
+        data_source: DataSource = "test",
+        X: Optional[ArrayLike] = None,
+        y: Optional[ArrayLike] = None,
+    ) -> dict:
+        """Get the time taken to compute the estimator's `predict` method.
```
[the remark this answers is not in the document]
auguste-probabl: see discussion above

skore/src/skore/utils/_measure_time.py
```diff
@@ -0,0 +1,23 @@
+from time import perf_counter
+
+
+class _measure_time:
```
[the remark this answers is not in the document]
auguste-probabl: Updated doctest and renamed class.

auguste-probabl:
> Shouldn't we add `_time` to the keys, so that the dict is stand-alone? Otherwise, the meaning of the keys are not so clear.

I agree, I'll change it

> When using the parameters `kind = "auto", data_source = "test"`, it outputs the key `fit`. I find this misleading. We could either complete the key to `fit_train`, or change the behavior of auto -> it renders what makes sense according to the data source requested. If None specified, fit_train + predict_test is returned, if train is specified, fit_train + predict_train is returned, and if anything else (test or X_y), only predict_source is returned.

Fair point. `fit_train` is kind of redundant, so I guess the second solution sounds better to me. But in all honesty, it now seems like these two metrics shouldn't be together, if putting them together is confusing...

skore/src/skore/sklearn/_base.py
auguste-probabl: it simplified things so much that i removed `find_in_cache`!

glemaitre:
Ok, so let's go back to expose two functions. We would still have the ambuigity when calling

```python
report.metrics.report_metrics(scoring=["fit_time", "score_time"])
```

It might be less confusing, because we don't expose `kind="auto"`, it will be implicit that we return those infos only if we can retrieve them.

MarieSacksick:
Another comment came to my mind, about the `predict_X_y_hash`. This key is not easy to handle. As a user, I can't automate that piece of code. I get that remembering the hash is required for the cache, but for the output, since the function can only take one (X,y) at a time, there is no confusion possible. Therefore, could we drop the hash in the key?  
The downside would be that if we use several times the X_y data source parameter, then the outputs could get mixed.

glemaitre:
> Therefore, could we drop the hash in the key?

Actually, we are going to report a single float so we don't key anymore :) But on the principle I agree. When it will come to the `report_metrics` and that we will have a column name `predict_time` would be enough because we don't allow to get several `data_source`. So it should not be ambiguous. But we can iron the details when we come with this implementation.

MarieSacksick:
> Ok, so let's go back to expose two functions.

To make sure I understand correctly this potential solution: does it mean to drop the `report.metrics.timing`? and the report_metrics would become the only way to retrieve it?

skore/src/skore/sklearn/_estimator/metrics_accessor.py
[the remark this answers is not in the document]
auguste-probabl: now that fit_time is on its own I made it return None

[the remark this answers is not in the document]
glemaitre: OK. I would have raise but `None` is maybe more permissive. Let's start with that.

auguste-probabl:
> does it mean to drop the `report.metrics.timing`? and the report_metrics would become the only way to retrieve it?

No, we're back to having two methods `report.metrics.fit_time` and `report.metrics.predict_time`. They'll be added to `report_metrics`, but probably not in this PR.

glemaitre:
So last small addition for this PR: could you add the function in the API rst file. In addition, it would be nice to expose it in an example to have at least one location where we call directly the two functions.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -444,6 +444,144 @@ def _compute_metric_scores(
...
+                pos_label=None,
+            )
+
+            predict_time = self._parent._cache.get(predict_time_cache_key)
+
+            if predict_time is None:
+                raise Exception(
+                    "Prediction was computed, but predict time was not found. "
+                    "This error should never be reached."
+                )
```
glemaitre, line 581: Let's make it an assertion since it is exactly what the assert means rather than a general Exception.

skore/src/skore/utils/_measure_time.py
```diff
@@ -0,0 +1,24 @@
...
+
+class MeasureTime:
+    """Measure the time to go through the context, in seconds.
+
+    Examples
+    --------
+    >>> from time import sleep
+    >>> with MeasureTime() as time_taken:
+    ...     sleep(0.01)
+    >>> assert 0 <= time_taken() <= 1 # Note: time_taken is a callable
```
glemaitre, line 12: When I request a test, usually I refer to adding a dedicated test file ;)
While doctest could be handy for some testing, it is quite a bad practice to rely on them for testing purposing. I know it mades sometimes trivial test file but it is better.

Slightly more on the topic: NumPy complained to scikit-learn when we rely on such doctest failures because we are really detecting implementation details. Usually it resort to testing only the latest build with docstring and not on all platforms. So it is one additional reason that it is safer to have a small file with the test.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -444,6 +444,144 @@ def _compute_metric_scores(
...
+        0.0...
+
+        >>> estimator = LogisticRegression().fit(X_train, y_train)
+        >>> report = EstimatorReport(estimator, X_test=X_test, y_test=y_test)
+        >>> report.metrics.fit_time()
+        None
+        """
+        return self._parent._cache.get(self._parent._fit_time_cache_key())
+
+    def predict_time(
```
glemaitre, line 483: Since the `fit_time` and `predict_time` function does not use the shared method to compute metric, it might be good to add two small tests in the `test_report.py` to check the API surface of those methods.

The general tests are already good but will not check the type of output and so on.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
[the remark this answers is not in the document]
glemaitre: This is resolved since they are public :)

auguste-probabl:
> could you add the function in the API rst file. 

Done

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl: added tests

skore/src/skore/sklearn/_estimator/metrics_accessor.py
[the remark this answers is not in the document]
glemaitre: I touch about it IRL but just commenting before to have this thread resolved.

It is more a conceptual reason: we want to run doctest to ensure that our documentation run and it is still up to date but not really want to check functional aspects of the code. The latest is expected to in its own unit/functional tests. The fact that the docstests still test some functionality related to the code is rather a side effect than a something that we rely on.

auguste-probabl:
> it would be nice to expose it in an example

Done

skore/src/skore/utils/_measure_time.py
auguste-probabl: done

glemaitre:
So let me summarize some IRL discussions:

- we can go with a single `timing` function. The aim of the function is to collect whatever timing that have been measured. The function will not try to filter anything.
- in terms of implementation, we can keep the `fit_time` and `predict_time` functions private because in the future, if someone write `report.metrics.report_metrics(scoring="timing")`, then we know that we can use both private functions and more precisely `_predict_time` and the `data_source` parameter to filter the useful data.

auguste-probabl:
> The aim of the function is to collect whatever timing that have been measured. The function will not try to filter anything.

By filter you mean compute?

auguste-probabl:
Should I remove `fit_time` and `predict_time` from `report_metrics`?

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -444,6 +447,73 @@ def _compute_metric_scores(
...
+            self._parent._hash,
+            data_source,
+            data_source_hash,
+            "predict_time",
+        )
+
+        return self._parent._cache.get(predict_time_cache_key)
+
+    def timings(self) -> dict:
+        """Get all measured times.
```
glemaitre, line 480:
```suggestion
        """Get all processing times measured related to estimator.
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -444,6 +447,73 @@ def _compute_metric_scores(
...
+
+        When an estimator is fitted inside the EstimatorReport, the time to fit is
+        recorded. Similarly, when predictions are computed on some data, the time to
+        predict is recorded. This function returns all the recorded times.
+
+        Returns
+        -------
+        timings : dict
+            The recorded times, in seconds,
+            in the form of a dict with some or all of the following keys:
```
glemaitre, line 490:
```suggestion
            in the form of a `dict` with some or all of the following keys:
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -444,6 +447,73 @@ def _compute_metric_scores(
...
+            )
+
+        predict_time_cache_key = (
+            self._parent._hash,
+            data_source,
+            data_source_hash,
+            "predict_time",
+        )
+
+        return self._parent._cache.get(predict_time_cache_key)
```
glemaitre, line 477: I know it is the default but it might be easier to have it such that we don't have any question if `get` is supposed to raise or just return `None`.

```suggestion
        return self._parent._cache.get(predict_time_cache_key, default=None)
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -444,6 +447,73 @@ def _compute_metric_scores(
...
+        return self._parent._cache.get(self._parent._fit_time_cache_key())
+
+    def _predict_time(
+        self,
+        *,
+        data_source: DataSource = "test",
+        data_source_hash: Optional[int] = None,
+        X: Optional[ArrayLike] = None,
+        y: Optional[ArrayLike] = None,
+    ) -> Union[float, None]:
```
glemaitre, line 464:
```suggestion
    ) -> Union[float, None]:
        """Get prediction time if it has been already measured."""
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -444,6 +447,73 @@ def _compute_metric_scores(
                 return score[0]
         return score
 
+    def _fit_time(self, **kwargs) -> Union[float, None]:
+        """Compute time to fit the estimator.
```
glemaitre, line 451: Instead of "Compute", it would be best to say "Get" since we don't compute it.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -161,6 +163,7 @@ def report_metrics(
                     scoring += ["_roc_auc", "_log_loss"]
             else:
                 scoring = ["_r2", "_rmse"]
+            scoring += ["fit_time", "predict_time"]
```
glemaitre, line 166: I would have expected to pass the private function "_fit_time" and "_predict_time" but I need to check why is it working :)

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl:
```suggestion
        """Get all measured processing times related to the estimator.
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: OK so we do some processing afterwards to get the right function. Maybe let's be inline why the previous function such that we have the name corresponding to the actual private function.

skore/src/skore/sklearn/_estimator/report.py
```diff
@@ -160,6 +170,9 @@ def _initialize_state(self) -> None:
     def clear_cache(self) -> None:
         """Clear the cache.
 
+        Note that the cache might not be empty after this method is run as some
+        values need to be kept, such as the fit time.
```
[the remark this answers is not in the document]
glemaitre: It will also reduce some of boiler code like the utils.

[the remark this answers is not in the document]
auguste-probabl: Sure, I'm fine with that!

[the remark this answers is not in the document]
auguste-probabl: done

glemaitre (commented):
A partial review.

glemaitre (approved):
So the implementation is fine with me. Next step is to implement the `timings` functions for both the cross-validation and comparison reports.

Issue #1453, opened 2025-03-21
feat(train_test_split): Make a wrapper that returns a dict of X_train, y_train, X_test, y_test


auguste-probabl:
### Is your feature request related to a problem? Please describe.

Writing the full `X_train, y_train, X_test, y_test = train_test_split()` is tedious and error-prone:

```python
X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)
# or is it X_train, y_train, X_test, y_test? I can't remember :)
report = EstimatorReport(estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test)
```

### Describe the solution you'd like

Why not wrap the logic so that it returns a dict:
```python
train_test_data = custom_train_test_split(X, y)
# or
train_test_data = pack_data(sklearn_train_test_split(X, y))

>>> train_test_data
{"X_train": ..., "X_test": ..., "y_train": ..., "y_test": ...}
```

That makes it possible to write
```python
EstimatorReport(estimator, **train_test_data)
```

auguste-probabl:
Following #1460 I've hit https://github.com/probabl-ai/skore/pull/1460#issuecomment-2750426087: you have to pass arguments by keyword rather than by position.
I find this too restrictive, which will make people avoid `as_dict` altogether. I think we can reasonably say that in 99% of uses of `train_test_split` the first positional argument will be `X` and the second will be `y`.

So I'd advocate for something like:
```python
>>> skore.train_test_split(X, as_dict=True)
{"X_train": ..., "X_test": ...}

>>> skore.train_test_split(X, y, as_dict=True)
{"X_train": ..., "X_test": ..., "y_train": ..., "y_test": ...}

>>> skore.train_test_split(X, y, z=z, as_dict=True)
{"X_train": ..., "X_test": ..., "y_train": ..., "y_test": ..., "z_train": ..., "z_test": ...}
```

auguste-probabl:
Closing in favor of #1544

PR #1460, opened 2025-03-24
feat(train_test_split): Add optional argument to return dict

nkapila6:
# What does it solve?
Adds enhancement based on #1453.

# Design choice
I deviate from what is asked in #1453 and add a boolean argument `return_dict` that returns a dictionary if `True`. If required, I can change it into a wrapper instead but this should not break any code base per se.

# Todo
What was done?
- [x] Implemented return dictionary.
- [x] Updated doc string.
- [x] Tests - Would you require them?
- [x] If all is okay, let me know so I can add in an example in the doc-string.
- [x] Tested with EstimatorReport as per #1453.

# Example
```python
import numpy as np

def main():
    X, y = np.arange(10).reshape((5, 2)), range(5)
    
    # default
    X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.33, random_state=42)

    # with my_dict true
    my_dict = train_test_split(X,y, test_size=0.33,random_state=42, return_dict=True)

    print(f"X_train: {X_train}\n X_test:{X_test}\n y_train:{y_train}\n y_test:{y_test}")
    print(my_dict)

    from sklearn.neural_network import MLPClassifier
    mlp = MLPClassifier()
    report = EstimatorReport(mlp, **my_dict)
    print(report.metrics)

if __name__ == "__main__":
    main()
```

# Output
```cmd
╭────────────────────── HighClassImbalanceTooFewExamplesWarning ───────────────────────╮
│ It seems that you have a classification problem with at least one class with fewer   │
│ than 100 examples in the test set. In this case, using train_test_split may not be a │
│ good idea because of high variability in the scores obtained on the test set. We     │
│ suggest three options to tackle this challenge: you can increase test_size, collect  │
│ more data, or use skore's cross_validate function.                                   │
╭───────────────────────────────── ShuffleTrueWarning ─────────────────────────────────╮
│ We detected that the `shuffle` parameter is set to `True` either explicitly or from  │
│ its default value. In case of time-ordered events (even if they are independent),    │
│ this will result in inflated model performance evaluation because natural drift will │
│ not be taken into account. We recommend setting the shuffle parameter to `False` in  │
│ order to ensure the evaluation process is really representative of your production   │
│ release process.                                                                     │
╭────────────────────── HighClassImbalanceTooFewExamplesWarning ───────────────────────╮
│ It seems that you have a classification problem with at least one class with fewer   │
│ than 100 examples in the test set. In this case, using train_test_split may not be a │
│ good idea because of high variability in the scores obtained on the test set. We     │
│ suggest three options to tackle this challenge: you can increase test_size, collect  │
│ more data, or use skore's cross_validate function.                                   │
╭───────────────────────────────── ShuffleTrueWarning ─────────────────────────────────╮
│ We detected that the `shuffle` parameter is set to `True` either explicitly or from  │
│ its default value. In case of time-ordered events (even if they are independent),    │
│ this will result in inflated model performance evaluation because natural drift will │
│ not be taken into account. We recommend setting the shuffle parameter to `False` in  │
│ order to ensure the evaluation process is really representative of your production   │
│ release process.                                                                     │
X_train: [[4 5]
 [0 1]
 [6 7]]
 X_test:[[2 3]
 [8 9]]
 y_train:[2, 0, 3]
 y_test:[1, 4]
{'X_train': array([[4, 5],
       [0, 1],
       [6, 7]]), 'X_test': array([[2, 3],
       [8, 9]]), 'y_train': [2, 0, 3], 'y_test': [1, 4]}
/Users/nkapila6/Code/rands/.venv/lib/python3.10/site-packages/sklearn/neural_network/_multilayer_perceptron.py:691: ConvergenceWarning: Stochastic Optimizer: Maximum iterations (200) reached and the optimization hasn't converged yet.
  warnings.warn(
EstimatorReport(estimator=MLPClassifier(), ...)
╭─────────── skore.EstimatorReport.metrics ───────────╮
│ Get guidance using the report.metrics.help() method │
╰─────────────────────────────────────────────────────╯
```

## With `skore.EstimatorReport` in Debug Console (breakpoint in __init__ method).
image

### DIFF
--- a/examples/model_evaluation/plot_estimator_report.py
+++ b/examples/model_evaluation/plot_estimator_report.py
@@ -47,7 +47,12 @@ pos_label, neg_label = "allowed", "disallowed"
 # and a validation set.
 from skore import train_test_split
 
-X_train, X_test, y_train, y_test = train_test_split(df, y, random_state=42)
+# If you have many dataframes to split on, you can always ask train_test_split to return a dictionary.
+# Remember, it needs to be passed as a keyword argument!
+my_dict = train_test_split(X=df, y=y, random_state=42, return_dict=True)
+
+# Since we only have 2 dataframes to split, we can simply unpack as follows.
+X_train, X_test, y_train, y_test = my_dict.values()
 
 # %%
 # By the way, notice how skore's :func:`~skore.train_test_split` automatically warns us
--- a/examples/model_evaluation/plot_estimator_report.py
+++ b/examples/model_evaluation/plot_estimator_report.py
@@ -49,10 +49,7 @@ from skore import train_test_split
 
 # If you have many dataframes to split on, you can always ask train_test_split to return a dictionary.
 # Remember, it needs to be passed as a keyword argument!
-my_dict = train_test_split(X=df, y=y, random_state=42, return_dict=True)
-
-# Since we only have 2 dataframes to split, we can simply unpack as follows.
-X_train, X_test, y_train, y_test = my_dict.values()
+split_data = train_test_split(X=df, y=y, random_state=42, as_dict=True)
 
 # %%
 # By the way, notice how skore's :func:`~skore.train_test_split` automatically warns us
@@ -67,7 +64,9 @@ X_train, X_test, y_train, y_test = my_dict.values()
 # So let's create a classifier for our task and fit it on the training set.
 from skrub import tabular_learner
 
-estimator = tabular_learner("classifier").fit(X_train, y_train)
+estimator = tabular_learner("classifier").fit(
+    split_data["X_train"], split_data["y_train"]
+)
 estimator
 
 # %%
@@ -82,9 +81,7 @@ estimator
 # detect that our estimator is already fitted and will not fit it again.
 from skore import EstimatorReport
 
-report = EstimatorReport(
-    estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
-)
+report = EstimatorReport(estimator, **split_data)
 
 # %%
 #
@@ -188,7 +185,10 @@ report.metrics.log_loss(data_source="train")
 
 start = time.time()
 metric_report = report.metrics.report_metrics(
-    data_source="X_y", X=X_test, y=y_test, pos_label=pos_label
+    data_source="X_y",
+    X=split_data["X_test"],
+    y=split_data["y_test"],
+    pos_label=pos_label,
 )
 end = time.time()
 metric_report
@@ -205,7 +205,10 @@ print(f"Time taken to compute the metrics: {end - start:.2f} seconds")
 # %%
 start = time.time()
 metric_report = report.metrics.report_metrics(
-    data_source="X_y", X=X_test, y=y_test, pos_label=pos_label
+    data_source="X_y",
+    X=split_data["X_test"],
+    y=split_data["y_test"],
+    pos_label=pos_label,
 )
 end = time.time()
 metric_report
@@ -247,7 +250,7 @@ def operational_decision_cost(y_true, y_pred, amount):
 import numpy as np
 
 rng = np.random.default_rng(42)
-amount = rng.integers(low=100, high=1000, size=len(y_test))
+amount = rng.integers(low=100, high=1000, size=len(split_data["y_test"]))
 
 # %%
 #
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -70,7 +70,8 @@ def train_test_split(
         If not None, data is split in a stratified fashion, using this as the
         class labels.
     return_dict : bool, default is False
-        If True, returns a dictionary with keys instead of a list. Dictionary keys are in format ``X_train``, ``X_test``, ``y_train``, and ``y_test``.
+        If True, returns a Dictionary with keys values ``X_train``, ``X_test``,
+        ``y_train``, and ``y_test`` instead of a List.
 
     Returns
     -------
@@ -82,7 +83,9 @@ def train_test_split(
         arrays passed positionally, in the order they were passed, then ``X`` if it
         was passed, then ``y`` if it was passed.
 
-        If return_dict=True: Dictionary with keys ``X_train``, ``X_test``, ``y_train``, and ``y_test``, each containing respective split data.
+        If return_dict=True: Dictionary with keys
+        ``X_train``, ``X_test``, ``y_train``, and ``y_test``,
+        each containing respective split data.
 
     Examples
     --------
@@ -183,14 +186,11 @@ def train_test_split(
             )
 
     if return_dict:
-        result = {
-            'X_train': output[0],
-            'X_test': output[1]
-            }
+        result = {"X_train": output[0], "X_test": output[1]}
 
-        if len(output)>2:
-            result['y_train'] = output[2]
-            result['y_test'] = output[3]
+        if len(output) > 2:
+            result["y_train"] = output[2]
+            result["y_test"] = output[3]
         output = result
-        
+
     return output
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -75,7 +75,7 @@ def train_test_split(
     Returns
     -------
     splitting : list or dict
-        If return_dict=False (default): List containing train-test split of inputs.
+        If return_dict=False (the default): List containing train-test split of inputs.
         The length of the list is twice the number of arrays passed, including
         the ``X`` and ``y`` keyword arguments. If arrays are passed positionally as well
         as through ``X`` and ``y``, the output arrays are ordered as follows: first the
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -70,9 +70,13 @@ def train_test_split(
     stratify : array-like, optional
         If not None, data is split in a stratified fashion, using this as the
         class labels.
-    return_dict : bool, default is False
+    as_dict : bool, default is False
         If True, returns a Dictionary with keys values ``X_train``, ``X_test``,
-        ``y_train``, and ``y_test`` instead of a List.
+        ``y_train``, and ``y_test`` instead of a List. Requires data to be
+        passed as keyword arguments.
+    **keyword_arrays : array-like, optional
+        Additional array-like arguments passed by keyword. Each keyword should
+        correspond to an array-like object.
 
     Returns
     -------
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -25,7 +25,7 @@ def train_test_split(
     random_state: Optional[Union[int, RandomState]] = None,
     shuffle: bool = True,
     stratify: Optional[ArrayLike] = None,
-    return_dict: bool = False,
+    as_dict: bool = False,
     **keyword_arrays: ArrayLike,
 ):
     """Perform train-test-split of data.
@@ -131,7 +131,7 @@ def train_test_split(
         new_arrays.append(y)
         keys += ["y"]
 
-    if return_dict and arrays:
+    if as_dict and arrays:
         raise ValueError(
             "When return_dict=True, arrays must be passed as keyword arguments.\n"
             "Example: train_test_split(X=X, y=y, sw=sample_weight, return_dict=True)"
@@ -204,7 +204,7 @@ def train_test_split(
                 )
             )
 
-    if return_dict:
+    if as_dict:
         result = {}
         for i, k in enumerate(keys):
             train, test = i * 2, i * 2 + 1
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -26,6 +26,7 @@ def train_test_split(
     shuffle: bool = True,
     stratify: Optional[ArrayLike] = None,
     return_dict: bool = False,
+    **arr,
 ):
     """Perform train-test-split of data.
 
@@ -122,10 +123,24 @@ def train_test_split(
     import sklearn.model_selection
 
     new_arrays = list(arrays)
+    keys = []
     if X is not None:
         new_arrays.append(X)
+        keys += ["X"]
     if y is not None:
         new_arrays.append(y)
+        keys += ["y"]
+
+    if return_dict and arrays:
+        raise ValueError(
+            "When return_dict=True, arrays must be passed as keyword arguments.\n"
+            "Example: train_test_split(X=X, y=y, sw=sample_weight, return_dict=True)"
+        )
+    elif arr:
+        if X is None and y is None:
+            arrays = tuple(arr.values())  # if X and y is not passed but other variables
+        keys += list(arr.keys())
+        new_arrays += list(arr.values())
 
     output = sklearn.model_selection.train_test_split(
         *new_arrays,
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -140,7 +140,8 @@ def train_test_split(
             "When return_dict=True, arrays must be passed as keyword arguments.\n"
             "Example: train_test_split(X=X, y=y, sw=sample_weight, return_dict=True)"
         )
-    elif keyword_arrays:
+
+    if keyword_arrays:
         if X is None and y is None:
             arrays = tuple(
                 keyword_arrays.values()
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -201,11 +201,10 @@ def train_test_split(
             )
 
     if return_dict:
-        result = {"X_train": output[0], "X_test": output[1]}
-
-        if len(output) > 2:
-            result["y_train"] = output[2]
-            result["y_test"] = output[3]
+        result = {}
+        for i, k in enumerate(keys):
+            train, test = i * 2, i * 2 + 1
+            result[f"{k}_train"], result[f"{k}_test"] = output[train], output[test]
         output = result
 
     return output
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -183,11 +183,14 @@ def train_test_split(
             )
 
     if return_dict:
-        output = {
+        result = {
             'X_train': output[0],
-            'X_test': output[1],
-            'y_train': output[2],
-            'y_test': output[3]
+            'X_test': output[1]
             }
+
+        if len(output)>2:
+            result['y_train'] = output[2]
+            result['y_test'] = output[3]
+        output = result
         
     return output
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -25,6 +25,7 @@ def train_test_split(
     random_state: Optional[Union[int, RandomState]] = None,
     shuffle: bool = True,
     stratify: Optional[ArrayLike] = None,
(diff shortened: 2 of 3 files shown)


MarieSacksick:
I like this option where the user only has to add a parameter! We might benefit a `pack_data` function for people who use scikit-learn `train_test_split` function, but it can be done in another PR later, in order to keep each PR as unitary as possible.    

Can you please:
- fix linting (you can check in the contributing guide [here](https://skore.probabl.ai/dev/contributing.html#linting))
- add a test (at least to test that when the param `return_dict` is `True`, that the output is a dict indeed)
- add some doc -> in the doc string, and use this new function in the docs, for instance [here](https://github.com/probabl-ai/skore/blob/main/examples/model_evaluation/plot_estimator_report.py#L50-81)

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -177,4 +182,12 @@ class labels.
                 )
             )
 
+    if return_dict:
+        output = {
+            'X_train': output[0],
+            'X_test': output[1],
+            'y_train': output[2],
+            'y_test': output[3]
```
auguste-probabl, line 190: We should check that `output[2]` and `output[3]` exist, in case the user didn't pass `y` (for example when working on an unsupervised learning task)

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -68,17 +69,21 @@ def train_test_split(
...
         class labels.
+    return_dict : bool, default is False
+        If True, returns a dictionary with keys instead of a list. Dictionary keys are in format ``X_train``, ``X_test``, ``y_train``, and ``y_test``.
 
     Returns
     -------
-    splitting : list
-        List containing train-test split of inputs.
+    splitting : list or dict
+        If return_dict=False (default): List containing train-test split of inputs.
```
auguste-probabl, line 78:
```suggestion
        If return_dict=False (the default): List containing train-test split of inputs.
```

auguste-probabl (commented):
Thank you for the PR! This is on me when writing the issue, but I'm not sure that `skore.train_test_split` will keep existing, thus the other solution would be preferable: a function that takes the output of `train_test_split` and returns a dict.

MarieSacksick:
> Thank you for the PR! This is on me when writing the issue, but I'm not sure that skore.train_test_split will keep existing, thus the other solution would be preferable: a function that takes the output of train_test_split and returns a dict.  

Yes, we might remove skore.train_test_split but we don't know when. Until then, I prefer the solution where it's just an additional parameter. Otherwise the user would have to chain 2 function, it's not great for them.  
We can create the `pack_data` function later, to replace it. Let's keep #1453 open after this PR to remember.

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: Added in [7412270](https://github.com/probabl-ai/skore/pull/1460/commits/7412270c6c71f69f92d65975f2a37fed60134b3e).

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: Added in [ec5c5b8](https://github.com/probabl-ai/skore/pull/1460/commits/ec5c5b8ddc7c1ed9d62135180dd5ed325a44c5f2)

glemaitre:
One thing that is limiting here is that we only return `X` and `y` and ignore any other parameter. For instance, we could have:

```python
X, y = make_classification()
sample_weight = np.random.randn(y.shape)
X_train, X_test, y_train, y_test, sample_weight_train, sample_weight_test = train_test_split(
    X, y, sample_weight
)
```

In this case, we would need to infer the input variable names and then suffix to add it in the dictionary. Just as a demonstration, it would be a potential way to go forward:

```python
import inspect
from sklearn.model_selection import train_test_split as sklearn_train_test_split

def infer_train_test_split(*arrays, **options):
    """
    Split arrays into random train and test subsets and return as a dictionary,
    automatically inferring variable names from the calling code.
    """
    # Try to get the variable names by inspecting the caller's local variables
    frame = inspect.currentframe().f_back
    locals_dict = frame.f_locals

    # Find variable names by comparing identity with passed arrays
    var_names = []
    for arr in arrays:
        found = False
        for name, value in locals_dict.items():
            if value is arr:  # Check object identity
                var_names.append(name)
                found = True
                break
        if not found:
            var_names.append(f"array_{len(var_names)}")

    # Call the original sklearn train_test_split
    split_arrays = sklearn_train_test_split(*arrays, **options)

    # Create the dictionary with the desired naming
    result = {}
    for i, name in enumerate(var_names):
        result[f"{name}_train"] = split_arrays[i*2]
        result[f"{name}_test"] = split_arrays[i*2+1]

    return result
```

with the following expected output:

```python
import numpy as np

X, y = make_classification()
sample_weight = np.random.randn(y.shape[0])

# The function will automatically detect variable names
split_data = infer_train_test_split(X, y, sample_weight, test_size=0.2)

# Access results using original variable names with suffixes
print(split_data.keys())
# Output: dict_keys(['X_train', 'X_test', 'y_train', 'y_test', 'sample_weight_train', 'sample_weight_test'])
```

Here, the fact that we use frame inspection is potentially brittle. We could potentially use the `**kwargs` to get outside of this mess and force the user to pass the array by explicitly naming data arrays:

```python
from sklearn.model_selection import train_test_split as sklearn_train_test_split

def keyword_only_train_test_split(*, test_size=0.25, random_state=None, **data_dict):
    """
    Split data into train and test sets.

    Parameters:
    -----------
    test_size : float, default=0.25
        Proportion of the dataset to include in the test split.
    random_state : int, default=None
        Controls the shuffling applied to the data.
    **data_dict : dict
        Keyword arguments containing the data to split.

    Returns:
    --------
    dict
        Dictionary with keys matching input keys plus '_train' or '_test' suffix.
    """
    # Extract arrays from the data dictionary
    arrays = list(data_dict.values())
    names = list(data_dict.keys())

    # Call the original sklearn train_test_split
    split_arrays = sklearn_train_test_split(
        *arrays, test_size=test_size, random_state=random_state
    )

    # Create a dictionary with the desired naming
    result = {}
    for i, name in enumerate(names):
        result[f"{name}_train"] = split_arrays[i*2]
        result[f"{name}_test"] = split_arrays[i*2+1]

    return result
```

nkapila6:
Wow, I totally overlooked that!

The second approach using a dict as input and output does feel more robust.

nkapila6:
I'm not sure how robust `varname` would be.. https://pwwang.github.io/python-varname/

Another way would be to ask the user to send in variable names? If `return_dict` is true, `arrays` is a Tuple? I wonder if that could be too complicated.

glemaitre:
If `return_dict=True`, then we might want to disallow to pass stuff by position and only go by keyword (raising a proper error message) and then use the keyword name for the output. Meaning that we stick to scikit-learn implementation with `return_dict=False` otherwise we force a new API:

```python
output = train_test_split(X=X, y=y, sw=sample_weight, return_dict=True)
```

and 

```python
output = train_test_split(X, y, sample_weight, return_dict=True)
```

would raise an error.

@auguste-probabl @MarieSacksick what are your thoughts around this idea.

auguste-probabl:
Passing arguments by keyword is always good in my book.

Thanks for your reactivity @nkapila6!

nkapila6:
All done!

# without dict
image

# with dict
image

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -180,3 +180,41 @@ def test_train_test_split_kwargs():
...
+
+
+def test_train_test_split_dict_kwargs():
+    """Passing data without keyword arguments with return_dict=True
+    should raise ValueError."""
+
+    X = [[1]] * 20
+    y = [0] * 10 + [1] * 10
+
+    with pytest.raises(ValueError):
```
auguste-probabl, line 192: Just to be sure that the right ValueError is raised:
```suggestion
    with pytest.raises(ValueError, match="When return_dict=True, arrays must be passed as keyword arguments"):
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -180,3 +180,41 @@ def test_train_test_split_kwargs():
...
+    """Check 2 keys are is returned if `return_dict` argument is True for
+    unsupervised case."""
+
+    X = [[1]] * 20
+    output = train_test_split(X=X, random_state=0, return_dict=True)
+    assert len(output.keys()) == 2
+
+
+def test_train_test_split_check_dict_no_X_no_y():
+    """Check if a dict is returned when X, y are None and a third kwarg is passed."""
```
auguste-probabl, line 214:
```suggestion
    """Check if a dict is returned when X, y are None and a third kwarg is passed."""
    """If the input is a keyword argument and `X` and `y` are None, then the result is a dict with 2 keys."""
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -180,3 +180,41 @@ def test_train_test_split_kwargs():
...
+
+    z = [[1]] * 20
+    output = train_test_split(z=z, test_size=0.33, random_state=0, return_dict=True)
+    keys = output.keys()
+    assert len(keys) == 2
+    assert list(keys) == ["z_train", "z_test"]
```
auguste-probabl, line 220: Since we check the list, checking the length is redundant
```suggestion
    keys = output.keys()
    assert list(keys) == ["z_train", "z_test"]
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -25,6 +25,8 @@ def train_test_split(
     random_state: Optional[Union[int, RandomState]] = None,
     shuffle: bool = True,
     stratify: Optional[ArrayLike] = None,
+    return_dict: bool = False,
+    **arr,
```
auguste-probabl, line 29:
```suggestion
    **arrays,
```

auguste-probabl, line 29: Can you add a description of `**arr`/`**arrays` in the docstring?

auguste-probabl: ah no, `arrays` is already used. Maybe `keyword_arrays`?

nkapila6: Arrays is already used. Hence, used arr 😓

nkapila6: Sure, I'll change it to `keyword_arrays`.

auguste-probabl: I guess you're planning on doing it anyway

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/9b1a875b869f49ed0bcccdd80777c19502345226

skore/src/skore/sklearn/train_test_split/train_test_split.py
auguste-probabl: yes I realized that after commenting ^^'

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -180,3 +180,41 @@ def test_train_test_split_kwargs():
...
+    """Check dict is returned if `return_dict` argument is True"""
+    X = [[1]] * 20
+    y = [0] * 10 + [1] * 10
+    output = train_test_split(X=X, y=y, random_state=0, return_dict=True)
+    assert type(output) is dict
+
+
+def test_train_test_split_check_dict_unsupervised_case():
+    """Check 2 keys are is returned if `return_dict` argument is True for
+    unsupervised case."""
```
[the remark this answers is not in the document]
nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/9b1a875b869f49ed0bcccdd80777c19502345226

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -180,3 +180,41 @@ def test_train_test_split_kwargs():
...
+
+    X = [[1]] * 20
+    y = [0] * 10 + [1] * 10
+
+    with pytest.raises(ValueError):
+        train_test_split(X, y, random_state=0, return_dict=True)
+
+
+def test_train_test_split_check_dict():
+    """Check dict is returned if `return_dict` argument is True"""
```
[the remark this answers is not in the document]
nkapila6: Well noted. It does seem simpler, I shall adopt this if-then approach too 😄 

Added in https://github.com/probabl-ai/skore/pull/1460/commits/9b1a875b869f49ed0bcccdd80777c19502345226

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/9b1a875b869f49ed0bcccdd80777c19502345226

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/9b1a875b869f49ed0bcccdd80777c19502345226

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -47,7 +47,12 @@
 # and a validation set.
 from skore import train_test_split
 
-X_train, X_test, y_train, y_test = train_test_split(df, y, random_state=42)
+# If you have many dataframes to split on, you can always ask train_test_split to return a dictionary.
+# Remember, it needs to be passed as a keyword argument!
+my_dict = train_test_split(X=df, y=y, random_state=42, return_dict=True)
+
+# Since we only have 2 dataframes to split, we can simply unpack as follows.
+X_train, X_test, y_train, y_test = my_dict.values()
```
auguste-probabl, line 55: `return_dict=True` isn't super useful if the user just unpacks right after. The goal of the feature is to instead refer to the data by `my_dict["X_train"]`, etc.

The benefit comes when creating the `EstimatorReport` below: the call becomes
```python
report = EstimatorReport(estimator, **my_dict)
```

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -47,7 +47,12 @@
 # and a validation set.
 from skore import train_test_split
 
-X_train, X_test, y_train, y_test = train_test_split(df, y, random_state=42)
+# If you have many dataframes to split on, you can always ask train_test_split to return a dictionary.
+# Remember, it needs to be passed as a keyword argument!
+my_dict = train_test_split(X=df, y=y, random_state=42, return_dict=True)
```
auguste-probabl, line 52:
```suggestion
split_data = train_test_split(X=df, y=y, random_state=42, return_dict=True)
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -68,17 +70,24 @@ def train_test_split(
     stratify : array-like, optional
         If not None, data is split in a stratified fashion, using this as the
         class labels.
+    return_dict : bool, default is False
+        If True, returns a Dictionary with keys values ``X_train``, ``X_test``,
+        ``y_train``, and ``y_test`` instead of a List.
```
auguste-probabl, line 75: Now that I think about it, I think `as_dict` would be a better name for the argument. `return_xxx` usually indicates that something will be returned *in addition* to the usual return values, for example in `sklearn.model_selection.cross_validate(return_train_score=True)`, whereas `as_xxx` is used to specify the return type, like `sklearn.datasets.load_iris(as_frame=True)`

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/3d5fc9c8e810ec318edfe65934286b00bb8d8642

nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/9208ca291de12d1c2cb1704ecd76e87801947c65

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/4fa5326d3bb948fb9e28aaa6259b2fd1ab9525f3

examples/model_evaluation/plot_estimator_report.py
nkapila6: Added in https://github.com/probabl-ai/skore/pull/1460/commits/ee62b8eecd33f79ae6353f2af811c15d7d987fe0

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -114,10 +127,26 @@ class labels.
...
     if y is not None:
         new_arrays.append(y)
+        keys += ["y"]
+
+    if as_dict and arrays:
+        raise ValueError(
+            "When return_dict=True, arrays must be passed as keyword arguments.\n"
+            "Example: train_test_split(X=X, y=y, sw=sample_weight, return_dict=True)"
+        )
+    elif keyword_arrays:
```
auguste-probabl, line 143: Since we exit the function in the above `if`
```suggestion
    
    if keyword_arrays:
```

nkapila6: How did I miss that? 😞
 
Added in https://github.com/probabl-ai/skore/pull/1460/commits/21fc68c8bf9a8f6122711e9c44abd0e0e5100363

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -68,17 +70,28 @@ def train_test_split(
...
         the ``X`` and ``y`` keyword arguments. If arrays are passed positionally as well
         as through ``X`` and ``y``, the output arrays are ordered as follows: first the
         arrays passed positionally, in the order they were passed, then ``X`` if it
         was passed, then ``y`` if it was passed.
 
+        If return_dict=True: Dictionary with keys
+        ``X_train``, ``X_test``, ``y_train``, and ``y_test``,
+        each containing respective split data.
+
     Examples
```
auguste-probabl, line 95: Would you mind adding an example of `as_dict` to the docstring examples? Something like
```python
>>> train_test_split(X=X, y=y, sample_weights=sample_weights, as_dict=True, random_state=0)
{'X_train': ..., 'X_test': ..., 'y_train': ..., 'y_test': ..., 'sample_weights_train': ..., 'sample_weights_test': ...}
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -68,17 +70,28 @@ def train_test_split(
     stratify : array-like, optional
         If not None, data is split in a stratified fashion, using this as the
         class labels.
+    as_dict : bool, default is False
+        If True, returns a Dictionary with keys values ``X_train``, ``X_test``,
+        ``y_train``, and ``y_test`` instead of a List. Requires data to be
+        passed as keyword arguments.
+    **keyword_arrays : array-like, optional
+        Additional array-like arguments passed by keyword. Each keyword should
+        correspond to an array-like object.
```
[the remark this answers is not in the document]
nkapila6: Added in 5924940966ea1a39b5e5c4031d432a5c2b576229

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: Added in 5924940966ea1a39b5e5c4031d432a5c2b576229

auguste-probabl (approved):
Looks great! Thanks for your patience.

thomass-dev:
As @auguste-probabl has validated, I'll merge as soon as it's green.

PR #1461, opened 2025-03-24
chore: Replace `_register_accessor` with assignment

auguste-probabl:
`_register_accessor` is mysterious, but its implementation is simply an assignment with an extra check.

It turns out that this check makes it impossible to run single doctests; e.g.
```
# with _register_accessor

pytest src/skore/sklearn/_estimator/report.py::EstimatorReport.clear_cache:0
...
<frozen importlib._bootstrap_external>:999: in exec_module
    ???
<frozen importlib._bootstrap>:488: in _call_with_frames_removed
    ???
src/skore/sklearn/_estimator/__init__.py:8: in <module>
    _register_accessor("metrics", EstimatorReport)(_MetricsAccessor)
src/skore/externals/_pandas_accessors.py:44: in decorator
    raise ValueError(
E   ValueError: registration of accessor <class 'skore.sklearn._estimator.metrics_accessor._MetricsAccessor'> under name 'metrics' for type <class 'skore.sklearn._estimator.report.EstimatorReport'> is overriding a preexisting attribute with the same name.
...
```

```
# without _register_accessor

pytest src/skore/sklearn/_estimator/report.py::EstimatorReport.clear_cache:0
1 passed in 1.49s
```

I had to add `cast`s to satisfy mypy.

### DIFF
--- a/skore/src/skore/externals/_pandas_accessors.py
+++ b/skore/src/skore/externals/_pandas_accessors.py
@@ -36,18 +36,3 @@ class Accessor:
             # we're accessing the attribute of the class, i.e., Dataset.geo
             return self._accessor
         return self._accessor(obj)
-
-
-def _register_accessor(name, cls):
-    def decorator(accessor):
-        if hasattr(cls, name):
-            raise ValueError(
-                f"registration of accessor {accessor!r} under name "
-                f"{name!r} for type {cls!r} is overriding a preexisting "
-                f"attribute with the same name."
-            )
-        setattr(cls, name, Accessor(name, accessor))
-        cls._accessors.add(name)
-        return accessor
-
-    return decorator
--- a/skore/src/skore/sklearn/_comparison/__init__.py
+++ b/skore/src/skore/sklearn/_comparison/__init__.py
@@ -1,7 +1,9 @@
-from skore.externals._pandas_accessors import _register_accessor
+from typing import cast
+
+from skore.externals._pandas_accessors import Accessor
 from skore.sklearn._comparison.metrics_accessor import _MetricsAccessor
 from skore.sklearn._comparison.report import ComparisonReport
 
-_register_accessor("metrics", ComparisonReport)(_MetricsAccessor)
+ComparisonReport.metrics = cast(_MetricsAccessor, Accessor("metrics", _MetricsAccessor))
 
 __all__ = ["ComparisonReport"]
--- a/skore/src/skore/sklearn/_comparison/report.py
+++ b/skore/src/skore/sklearn/_comparison/report.py
@@ -13,7 +13,7 @@ from skore.sklearn._estimator.report import EstimatorReport
 from skore.utils._progress_bar import progress_decorator
 
 if TYPE_CHECKING:
-    from skore.sklearn._estimator.metrics_accessor import _MetricsAccessor
+    from skore.sklearn._comparison.metrics_accessor import _MetricsAccessor
 
 
 class ComparisonReport(_BaseReport, DirNamesMixin):
--- a/skore/src/skore/sklearn/_cross_validation/__init__.py
+++ b/skore/src/skore/sklearn/_cross_validation/__init__.py
@@ -1,9 +1,13 @@
-from skore.externals._pandas_accessors import _register_accessor
+from typing import cast
+
+from skore.externals._pandas_accessors import Accessor
 from skore.sklearn._cross_validation.metrics_accessor import _MetricsAccessor
 from skore.sklearn._cross_validation.report import (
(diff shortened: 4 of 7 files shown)


skore/src/skore/sklearn/_comparison/__init__.py
```diff
@@ -1,7 +1,9 @@
-from skore.externals._pandas_accessors import _register_accessor
```
thomass-dev, line 1: Don't you need to remove the corresponding file/code?

thomass-dev (commented):
I think you're right, this seems much clearer.
What do you think about assigning the accessors directly in the report class definition, for instance in its constructor?

skore/src/skore/sklearn/_comparison/__init__.py
auguste-probabl: done, thanks

skore/src/skore/externals/_pandas_accessors.py
auguste-probabl, line 1: maybe we can wait for @glemaitre 's review

skore/src/skore/externals/_pandas_accessors.py
```diff
@@ -36,18 +36,3 @@ def __get__(self, obj, cls):
...
-def _register_accessor(name, cls):
-    def decorator(accessor):
-        if hasattr(cls, name):
-            raise ValueError(
-                f"registration of accessor {accessor!r} under name "
-                f"{name!r} for type {cls!r} is overriding a preexisting "
-                f"attribute with the same name."
-            )
-        setattr(cls, name, Accessor(name, accessor))
-        cls._accessors.add(name)
```
auguste-probabl, line 50: I can't figure out what this line does

glemaitre (commented):
I would probably keep the current implementation because it is borrow from `pandas`: https://github.com/pandas-dev/pandas/blob/v2.2.3/pandas/core/accessor.py

If we don't have a bug, I would not necessarely change it.

skore/src/skore/externals/_pandas_accessors.py
glemaitre: `_accessors` is a set so it means that you add a new string to the set.

glemaitre:
> It turns out that this check makes it impossible to run single doctests

Can you run:

```
pytest src -k EstimatorReport
```

As far I remember, it works in this case. What does not work is when you access the `EstimatorReport`. I would not be surprise that the accessor was not yet created or something like that.

auguste-probabl:
> Can you run:
> 
> ```
> pytest src -k EstimatorReport
> ```
> 
> As far I remember, it works in this case. What does not work is when you access the `EstimatorReport`.

Just tested and yes it works.

> I would not be surprise that the accessor was not yet created or something like that.

I put the full error message above: `ValueError: registration of accessor <class 'skore.sklearn._estimator.metrics_accessor._MetricsAccessor'> under name 'metrics' for type <class 'skore.sklearn._estimator.report.EstimatorReport'> is overriding a preexisting attribute with the same name.`

skore/src/skore/externals/_pandas_accessors.py
```diff
@@ -36,18 +36,3 @@ def __get__(self, obj, cls):
             # we're accessing the attribute of the class, i.e., Dataset.geo
             return self._accessor
         return self._accessor(obj)
-
-
-def _register_accessor(name, cls):
-    def decorator(accessor):
-        if hasattr(cls, name):
-            raise ValueError(
```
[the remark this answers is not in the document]
auguste-probabl: Can you give an example? The only time that I see this error is when trying to run a single doctest, as seen above.

auguste-probabl:
> I would probably keep the current implementation because it is borrow from `pandas`: https://github.com/pandas-dev/pandas/blob/v2.2.3/pandas/core/accessor.py
> 
> If we don't have a bug, I would not necessarely change it.

I agree that it's a [Chesterton fence](https://en.wikipedia.org/wiki/G._K._Chesterton#Chesterton's_fence) situation. On the other hand I didn't change things that much compared to the pandas code, and the current implementation causes problems, as seen above.

skore/src/skore/externals/_pandas_accessors.py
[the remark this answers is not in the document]
glemaitre: For instance, we have the method `cache_predictions`. If we try to register an accessor named `cache_predictions`, it will conflict with it and raise this error.

It is indeed not clear to me why with the single doctest, it is raising an error.

auguste-probabl:
Closing as this is not that important. I might come back to it when I can argue for it better.

PR #1462, opened 2025-03-25
fix: Make `EstimatorReport.compute_metric_scores` return metrics in native types

thomass-dev:
Make `EstimatorReport.compute_metric_scores` return metrics in native types, for consistency across functions.
Allow metrics to be JSONable as is.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -322,12 +322,12 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
                     index = pd.Index([metric_name], name="Metric")
                     score_array = np.array(score).reshape(-1, 1)
             elif self._parent._ml_task in ("regression", "multioutput-regression"):
-                if isinstance(score, np.ndarray):
+                if isinstance(score, list):
                     index = pd.MultiIndex.from_arrays(
                         [[metric_name] * len(score), list(range(len(score)))],
                         names=["Metric", "Output"],
                     )
-                    score_array = score.reshape(-1, 1)
+                    score_array = np.array(score).reshape(-1, 1)
                 else:
                     index = pd.Index([metric_name], name="Metric")
                     score_array = np.array(score).reshape(-1, 1)
@@ -1151,7 +1151,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
         multioutput: Union[
             Literal["raw_values", "uniform_average"], ArrayLike
         ] = "raw_values",
-    ) -> Union[float, np.ndarray]:
+    ) -> Union[float, list]:
         """Compute the R² score.
 
         Parameters
@@ -1183,7 +1183,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
 
         Returns
         -------
-        float or ndarray of shape (n_outputs,)
+        float or list of len(n_outputs)
             The R² score.
 
         Examples
@@ -1224,7 +1224,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
         multioutput: Union[
             Literal["raw_values", "uniform_average"], ArrayLike
         ] = "raw_values",
-    ) -> Union[float, np.ndarray]:
+    ) -> Union[float, list]:
         """Private interface of `r2` to be able to pass `data_source_hash`.
 
         `data_source_hash` is either an `int` when we already computed the hash
@@ -1244,8 +1244,8 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
             self._parent._ml_task == "multioutput-regression"
             and multioutput == "raw_values"
         ):
-            assert isinstance(result, np.ndarray), (
-                "The R² score should be a numpy.ndarray, got "
+            assert isinstance(result, list), (
+                "The R² score should be a list, got "
                 f"{type(result)} with value {result}."
             )
             return result
@@ -1268,7 +1268,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
         multioutput: Union[
             Literal["raw_values", "uniform_average"], ArrayLike
         ] = "raw_values",
-    ) -> Union[float, np.ndarray]:
+    ) -> Union[float, list]:
         """Compute the root mean squared error.
 
         Parameters
@@ -1300,7 +1300,7 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
 
         Returns
         -------
-        float or ndarray of shape (n_outputs,)
+        float or list of len(n_outputs)
             The root mean squared error.
 
         Examples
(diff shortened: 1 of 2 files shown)


skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1180,7 +1183,7 @@ def r2(
 
         Returns
         -------
-        float or ndarray of shape (n_outputs,)
+        float or list of len(n_outputs)
```
auguste-probabl, line 1186:
```suggestion
        float or list of len `n_outputs`
```

thomass-dev: Double back-quotes with RST?

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1420,7 +1423,7 @@ def custom_metric(
 
         Returns
         -------
-        float, dict, or ndarray of shape (n_outputs,)
+        float, dict, or list of len(n_outputs)
```
auguste-probabl, line 1426:
```suggestion
        float, dict, or list of length ``n_outputs``
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -428,25 +435,21 @@ def _compute_metric_scores(
...
-            return {
-                label: float(score_label)
-                for label, score_label in zip(self._parent._estimator.classes_, score)
-            }
+            return dict(
+                zip(
+                    map(cast_numpy_to_native, self._parent._estimator.classes_),
+                    map(cast_numpy_to_native, score),
+                )
+            )
```
glemaitre, line 449:
```suggestion
                zip(
                    self._parent._estimator.classes_.tolist(),
                    score.tolist(),
                )
            )
```

glemaitre: The scikit-learn API forces `classes_` to be a numpy array always and we check specifically that `score` is a numpy array.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -428,25 +435,21 @@ def _compute_metric_scores(
...
         elif isinstance(score, np.ndarray):
-            return score[0] if score.size == 1 else score
-        return score
+            return score[0] if score.size == 1 else score.tolist()
+        return cast_numpy_to_native(score)
```
glemaitre, line 452: You cannot have any numpy array at this stag normally. Otherwise, you would return the line above.

glemaitre (commented):
What is the reason to expect native types here? I see a minimum cost of memory and non-contiguous memory so I'm just curious what it would simplify on the hub side?

skore/src/skore/sklearn/_estimator/metrics_accessor.py
thomass-dev: The problem here is no the list, but the type of each element in the list that can be int64 etc.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
thomass-dev: Here, score can be float64 etc.

thomass-dev:
> What is the reason to expect native types here? I see a minimum cost of memory and non-contiguous memory so I'm just curious what it would simplify on the hub side?

As you can see in the diff, i haven't changed the type of the cached values to keep the advantage of numpy objects memory cost.
But, i changed the return of the metrics to allow direct JSON serialization. Without direct benefit for user to retrieve a numpy object, i think it's a good idea to return metrics as native objects.

For instance, if the return of the metric is a dict with the classes as key, the key can be numpy objects (int64 etc).
Even with `orjson`, you can't serialize this metric in state. You need to manually cast all the keys of your dict.

```python
In [15]: recall = report.metrics.recall(); print(recall)
    ...: orjson.dumps(recall, option=orjson.OPT_SERIALIZE_NUMPY | orjson.OPT_NON_STR_KEYS)
{np.int64(0): 1.0, np.int64(1): 1.0}
TypeError                                 Traceback (most recent call last)
Cell In[15], line 2
      1 recall = report.metrics.recall(); print(recall)
----> 2 orjson.dumps(recall, option=orjson.OPT_SERIALIZE_NUMPY | orjson.OPT_NON_STR_KEYS)

TypeError: Dict key must a type serializable with OPT_NON_STR_KEYS
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: So here, we could do `return float(score)`?

glemaitre:
Oh so the problem is not only the arrays, it is really the `dtype`. Let me make another pass with this in head.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
thomass-dev: Yes.
But as i have added a more generic function that can convert any atomic object from numpy to native, i have removed the `if isinstance(score, np.float)`.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: If you don't want to think to much, you can follow the scikit-learn convention (that required hours of debates :)): https://scikit-learn.org/dev/developers/contributing.html#guidelines-for-writing-docstrings

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: I would prefer to be explicit instead to be generic but I will have a fresh look.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
glemaitre: > but the type of each element in the list classes_ or score that can be np.float64,np.int64 etc.

`.tolist()` should convert automatically to native type:

- NumPy integers (int8, int16, int32, int64) → Python int
- NumPy floats (float32, float64) → Python float
- NumPy booleans → Python bool
- NumPy complex numbers → Python complex

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -428,25 +435,21 @@ def _compute_metric_scores(
             )
 
             score = metric_fn(y_true, y_pred, **kwargs)
-            if isinstance(score, np.floating):
-                # convert np.float64 and np.float32 to float for consistency across
-                # functions
-                score = float(score)
```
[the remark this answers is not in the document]
thomass-dev: > So here, I wanted to persist in Python native in the cache because it is what people are working on the scikit-learn side with the array-api. So I would prefer to keep it this way instead of postponing.

I though the opposite, it does indeed simplify the changes.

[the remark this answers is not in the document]
glemaitre: > I though the opposite, it does indeed simplify the changes.

For the short story, converting to Python native type will trigger some lazy computation and is therefore used without to request an explicit `.compute()` in some framework from the top of the head.

glemaitre:
The failure is just docstring so we will make it consistent now. So LGTM.

glemaitre (dismissed):

glemaitre (approved):

skore/src/skore/sklearn/_estimator/metrics_accessor.py
thomass-dev: In convention with scikit-learn, changed to 

> list of length ``n_outputs``

glemaitre:
Nice. Thanks @thomass-dev

PR #1474, opened 2025-03-27
chore: Rename scatter_kwargs and line_kwargs in PredictionErrorDisplay

### LINKED ISSUE #1458: chore: In `PredictionErrorDisplay`, we should rename the dictionary to rename the parameters

Right now, we have two parameters in `PredictionErrorDisplay` allowing to change the rendering in matplotlib: `scatter_kwargs` and `line_kwargs`.

The meaning of those keywords are related to the `matplotlib` functionality rather than their purpose. It is different from the `RocCurveDisplay` and `PrecisionRecallDisplay`. So we could rename at least `line_kwargs` to reflect that it is the line of a perfect model. For `scatter_kwargs`, we should make sure it is related to the observations, samples, or data points.

glemaitre:
closes #1458 

This PR rename:

- `scatter_kwargs` to `data_points_kwargs`
- `line_kwargs` to `perfect_model_kwargs`

to align to the same naming semantic than with precision-recall or roc curve.


auguste-probabl (approved):
Works for me, thanks!

PR #1481, opened 2025-03-28
feat: Implement timings for CrossValidationReport and ComparisonReport

### LINKED ISSUE #1479: feat: Add `timings` function for both `CrossValidationReport` and `ComparisonReport`

We recently added the `timings` function into the `EstimatorReport`. We should extend the support to the `CrossValidationReport` and `ComparisonReport` as well.

glemaitre:
closes #1479 

Add `timings` method to `CrossValidationReport` and `ComparisonReport`.

### DIFF
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -243,6 +243,29 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             self._parent._cache[cache_key] = results
         return results
 
+    def timings(self) -> pd.DataFrame:
+        """Get all measured processing times related to the different estimators.
+
+        The index of the returned dataframe is the name of the processing time. When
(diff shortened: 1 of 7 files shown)


skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -243,6 +243,29 @@ def _compute_metric_scores(
...
+        """Get all measured processing times related to the different estimators.
+
+        The index of the returned dataframe is the name of the processing time. When
+        the estimators were not used to predict, no timings regarding the prediction
+        will be present.
+
+        Returns
+        -------
+        pd.DataFrame
+            A dataframe with the processing times.
```
auguste-probabl, line 256: Can you add an example, if only to make the review faster?

glemaitre: Good point.

auguste-probabl (approved):
Good stuff.

PR #1482, opened 2025-03-28
chore: Delegate availability methods to underlying EstimatorReport

### LINKED ISSUE #1472: chore: CrossValidationReport methods should delegate if they are available to underlying reports

As a follow-up to https://github.com/probabl-ai/skore/pull/1471, in `CrossValidationReport`, we should delegate the availability of methods based on the availability on the list of underlying `EstimatorReport`.

So we should revisit how `available_if` is done right now.

glemaitre:
closes #1472

Delegate the availability of all different methods from `CrossValidationReport` depending if they are available into the `EstimatorReport` from the first CV split.

### DIFF
--- a/skore/src/skore/utils/_accessor.py
+++ b/skore/src/skore/utils/_accessor.py
@@ -2,6 +2,10 @@ from typing import Any, Callable
 
 from sklearn.pipeline import Pipeline
 
+########################################################################################
+# Accessor related to `EstimatorReport`
+########################################################################################
+
 
 def _check_supported_ml_task(supported_ml_tasks: list[str]) -> Callable:
     def check(accessor: Any) -> bool:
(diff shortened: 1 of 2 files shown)


skore/src/skore/utils/_accessor.py
```diff
@@ -36,6 +36,31 @@ def check(accessor: Any) -> bool:
     return check
 
 
+def _check_estimator_report_has_method(
```
auguste-probabl, line 39: This check is specific to CrossValidationReports, can you differentiate it from the others?

auguste-probabl (dismissed):

skore/src/skore/utils/_accessor.py
glemaitre: Would you prefer that we pass explicitly the report as well:

```python
_check_report_has_method(
    self._parent.estimator_reports_[0],
    "metrics",
    "accuracy",
)
```

?

auguste-probabl: Sure, I meant mostly just to put it at the end of  `_accessor.py` and start a section called
```
### CrossValidationReport-specific checks
```
or something but if we can do it with code then all the better.

glemaitre: Uhm actually it is tricky to pass the report because in the `available_if` we don't have access to the `self` at this point.

auguste-probabl (approved):

thomass-dev:
Thanks!

PR #1484, opened 2025-03-28
fix: Avoid pandas warning by using float('nan') instead of None

glemaitre:
While working on #1483, I realised that we are raising a pandas warning when a concatenation is done with fitted estimator in `report_metrics` because `fit_time=None` and `None` does not have the same `dtype` than other dataframe (`float`).

To avoid such issue, we can use `float("nan")` whenever we don't have the information and keep the `dtype` consistent. It will avoid the pandas warning as well.

In addition, I found a failure in #1483 in which, for some reason, the value of the hash of the data is changing. Therefore, I'm not hard-coding it anymore and instead computing on the go the value.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -1,4 +1,5 @@
 import inspect
+import math
 from collections.abc import Iterable
 from functools import partial
 from typing import Any, Callable, Literal, Optional, Union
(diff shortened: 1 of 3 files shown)


auguste-probabl (changes requested):
I don't want to make our code less explicit because of pandas warnings -- in particular I would rather have as few null values as possible. Can we not instead cast things to float when creating the dataframe (https://github.com/probabl-ai/skore/blob/3ef7a28165d31229c5586f28dd40225e7707f152/skore/src/skore/sklearn/_estimator/metrics_accessor.py#L333)?

glemaitre:
OK so I kept the semantic with the `None` and add an optional `cast` parameter such that we call internally `_fit_time`, we get a `np.nan`

auguste-probabl (approved):
After IRL discussion, this solution seems like a good middle-ground. It's simple, it only touches private APIs and it avoids having to add special cases in the dataframe generation.

PR #1488, opened 2025-03-29
chore: Private method to list methods and arguments to be called for serialization

### LINKED ISSUE #1450: feat(report): Retrieve programmatically all the predefined metrics and plots functions

As a user, i want to retrieve programmatically all the predefined metrics and plots functions that i can call from a report. 
This way, i can dynamically compute all metrics and plots in a for-loop, and without knowing explicitly the available metrics.
In addition, i want to be able to retrieve the list of metrics separately from the plots one.

Currently as a user, the only way to do such is to get all the public member of the related metrics accessor using inspection, and filter the unrelated methods (`help`, `custom_metrics` etc). To separate the functions used to compute scalar metrics from the plots, there is no easy solution for the user without running it: he can inspect the signature of the function to retrieve typing and filter only on float, but its not robust.

glemaitre:
closes #1450 

This PR shows how we could provide a private method to precompute all necessary information for `skore-hub`.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
@@ -625,3 +625,16 @@ class _FeatureImportanceAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
             class_name="skore.EstimatorReport.feature_importance",
             help_method_name="report.feature_importance.help()",
         )
+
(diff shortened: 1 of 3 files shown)


glemaitre:
It is pretty much a draft but it shows one potential way of doing what was asked. ping @thomass-dev

glemaitre:
I assume that I can close that since we merged the new version of the `put`.

Issue #1417, opened 2025-03-12
feat(EstimatorReport): Give access to predictions


sylvaincom:
### Is your feature request related to a problem? Please describe.

As requested by @adrinjalali, in the [inspecting example](https://skore.probabl.ai/0.8/auto_examples/use_cases/plot_feature_importance.html), at the end of the section on linear models, it would be nice to plot the map (Latitude, Longitude), and for each sample, the prediction error, in order to see where our model does bad predictions.

### Describe the solution you'd like

For that, I would need the prediction error for each sample, for the estimator report.

A suggestion by @glemaitre: As we have `cache_predictions`, we could have a `get_predictions` that uses the cache

glemaitre:
There two different things here:

- being able to get the prediction: we could add `get_predictions` to be able to get predictions that uses the cache effectively
- getting the prediction error: since we have `report.metrics.prediction_errors()` returns a display and could provide access to the raw prediction errors as an attribute because we need them to make the plotting.

sylvaincom:
I think it would be nice to have the 2 options you propose:
1. `get_predictions` is useful in time series where you take one univariate time series and compare your predictions to the actual values, to see for example what your model takes into account well or what the model is missing (extreme peaks, seasonal effects, etc)
2. `get_prediction_errors` is useful for the use case mentioned on the original issue: on a map, see where the districts where the prediction error is high

Combining these two options, we could use `get_prediction_errors` to see which samples your model predicts the worse, then plot these worse samples using `get_predictions` to do the plotting stated in option 1, to see where you could improve your model

glemaitre:
`get_predictions` is general to any model while `get_predictions_errors` does only make sens in regression. That is why, it is better to go through the report object: `report.metrics.prediction_errors().residuals_`

adrinjalali:
Even though `get_prediction_errors` might not make sense, `get_misclassifications` might make sense. The method can even accept a "error" function to calculate a custom error on misclassification in custom usecases.

glemaitre:
It might be linked to a getting the `confusion_matrix` and have a display to get the different info in classification setting.

MarieSacksick:
For the record, adding a request from a user on discord: https://discord.com/channels/1275821367324840119/1303409800872333415/1353712759140909067.  

![Image](https://github.com/user-attachments/assets/b7088194-933b-4963-b1f5-7354d99b88af)

MarieSacksick:
Let's focus this issue on providing the predictions.  
[...]

Issue #1359, opened 2025-02-25
enh: Add a `.frame()` method to Displays


glemaitre:
Right now, our displays have a single method `.plot` (and potentially `.set_style`). I most cases, it would be great to expose a `.frame` property returning a dataframe containing all the information allowing to make the plot.

For instance, for the feature importance, it could be a dataframe with the name of the features as row and the importance as column. For the ROC curve, it could be the fpr, tpr, and the associated thresholds.

glemaitre:
Actually, I think that we could make it a method such that we can offer the possibility to have name easy to index vs. name to make the rendering nice, in the same spirit of the `flat_index`.

MarieSacksick:
# TODO

Let's list our displays:  

## EstimatorReport

- [ ] PrecisionRecallCurveDisplay

| y_true | y_pred |
| --- | --- |
| values | values |

- [ ] RocCurveDisplay

For binary classification, we would return a dataframe:  
| fpr | tpr | threshold |
| --- | --- | --- |
| values | values | values |

For multiclass, we only have ovr:  

| fpr | tpr | threshold | method | class |
| --- | --- | --- | --- | --- |
| values | values | values | OvR | values |  

- [ ] PredictionErrorDisplay

For binary classification, we would return a dataframe:  
| precision | recall | threshold |
| --- | --- | --- |
| values | values | values |

For multiclass, we only have ovr:  

| precision | recall | threshold | method | class |
| --- | --- | --- | --- | --- |
| values | values | values | OvR | values |  

## CrossValidationReport

- [ ] PredictionErrorDisplay

| y_true | y_pred | fold_id |
| --- | --- | --- |
| values | values | id |

- [ ] PrecisionRecallCurveDisplay

For binary classification, we would return a dataframe:  
| precision | recall | threshold | fold_id | 
| --- | --- | --- | --- |
| values | values | values | id |

For multiclass, we only have ovr:  

| precision | recall | threshold | method | class | fold_id |  
| --- | --- | --- | --- | --- | --- |
| values | values | values | OvR | values | id |


- [ ] RocCurveDisplay

For binary classification, we would return a dataframe:  
| fpr | tpr | threshold | fold_id | 
| --- | --- | --- | --- |
| values | values | values | id |

For multiclass, we only have ovr:  

| fpr | tpr | threshold | method | class | fold_id | 
| --- | --- | --- | --- | --- | --- |
| values | values | values | OvR | values |  id |  

## ComparisonReport

[...]

PR #1489, opened 2025-03-29
feat: Add `get_predictions` method to all reporters

glemaitre:
Partially address #1417 by implementing `EstimatorReport.get_predictions` such that one can get the predictions from a model and eventually from the cache if it was already computed before.

### DIFF
--- a/skore/src/skore/sklearn/_comparison/report.py
+++ b/skore/src/skore/sklearn/_comparison/report.py
@@ -6,6 +6,7 @@ from typing import TYPE_CHECKING, Any, Literal, Optional, Union
 
 import joblib
 import numpy as np
+from numpy.typing import ArrayLike
 
 from skore.externals._pandas_accessors import DirNamesMixin
 from skore.sklearn._base import _BaseReport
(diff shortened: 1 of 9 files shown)


skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -267,6 +268,88 @@ def cache_predictions(
             )
             progress.update(main_task, advance=1, refresh=True)
 
+    def get_predictions(
```
auguste-probabl, line 271: Why not
```suggestion
    def predict(
```

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -267,6 +268,88 @@ def cache_predictions(
...
+        self,
+        *,
+        data_source: Literal["train", "test", "X_y"],
+        response_method: str,
```
auguste-probabl, line 275:
```suggestion
        response_method: Union[Literal["predict", "predict_proba", "decision_function"], str],
```

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -267,6 +268,88 @@ def cache_predictions(
...
+        Parameters
+        ----------
+        data_source : {"test", "train", "X_y"}, default="test"
+            The data source to use.
+
+            - "test" : use the test set provided when creating the report.
+            - "train" : use the train set provided when creating the report.
+            - "X_y" : use the provided `X` and `y` to compute the metric.
+
+        response_method : str
```
auguste-probabl, line 292:
```suggestion
        response_method : {"predict", "predict_proba", "decision_function"} or some other string
```

glemaitre: Actually it cannot be other strings ;)

skore/src/skore/sklearn/_comparison/report.py
glemaitre: I would reserve this to a scikit-learn estimator. Potentially the output is not exactly the same than `predict` (because we handle `pos_label` for the user). So I would prefer to make it quite separated

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl:
```suggestion
        response_method: Literal["predict", "predict_proba", "decision_function"],
```
since there cannot be another choice

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: Right. I just prefer to avoid function names like `return_output`, but I can't think of a better name.

glemaitre: I see. Since we have `cache_predictions`, I thought that it was nice to have the `get_` but I see your point. (I don't really like the `get_`, `set_`, and those functions in general.

auguste-probabl (dismissed):

MarieSacksick:
Why do you say that it's only a partial answer to #1417?  It feels to me that it answers completely.  
For the other part of this issue, i.e. to have the errors, it's somehow a duplicate of https://github.com/probabl-ai/skore/issues/1359 and should be addressed there.

glemaitre:
In #1417, there is the part about classification error in which I think it could be addressed through a confusion matrix display. I don't mind that we consolidate a single issue about this. I just did not want to close the issue since there was 2 topics in the #1417.

PR #1492, opened 2025-03-30
fix: Support joblib < 1.4 by using list container as output

### LINKED ISSUE #1452: bug: Parallel.__init__() got an unexpected keyword argument 'return_as' when using CrossValidationReport
### Describe the bug

I tried running `CrossValidationReport` on runtime with older `joblib`, which results in an error.

The reason and solution is described [here](https://github.com/GFZ/arosics/issues/39#issuecomment-2034163564)

Specifying constraints [here](https://github.com/probabl-ai/skore/blob/9c438211f0a05237804e37bb0f22d32491cf356c/skore/pyproject.toml#L27) would solve the issue.

### Steps/Code to Reproduce

`joblib==1.2.0`
`skore==0.7.1`

### Expected Behavior

no error

### Actual Behavior

`Parallel.__init__() got an unexpected keyword argument 'return_as'`

### Environment

```shell
`joblib==1.2.0`
`skore==0.7.1`
```

glemaitre:
closes #1452 

This PR introduces a way to be compatible with `joblib < 1.4` where `return_as` is not implemented. In this case, we can only use `list`.

The issue with using `list` is that we will be in degraded mode for the progress bars.

The proposal here is to:

- warn at import about the old `joblib` version and ask potentially to update it.
- disable the progress bar.

### DIFF
--- a/skore/src/skore/__init__.py
+++ b/skore/src/skore/__init__.py
@@ -1,11 +1,14 @@
 """Configure logging and global settings."""
 
 import logging
+import warnings
 
+import joblib
 from rich.console import Console
 from rich.theme import Theme
 
 from skore._config import config_context, get_config, set_config
+from skore.externals._sklearn_compat import parse_version
 from skore.project import Project
 from skore.sklearn import (
     ComparisonReport,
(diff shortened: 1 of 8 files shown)


auguste-probabl (dismissed):
Nice.

auguste-probabl (approved):

PR #1494, opened 2025-03-30
docs: Use left-out dataset for the external case in example

### LINKED ISSUE #1216: docs: `plot_cache_mechanism.py` example should have actual external data

`technical_details/plot_cache_mechanism.py` Sphinx example

We have this:

Image

However, the external data is `X_test` and `y_test` on which the `EstimatorReport` already previously did the predictions? This is just to simulate external data? In that case, I find it unclear for a user? cc @glemaitre

### Suggested improvement

Should we have actual external data?

### Additional context

_No response_

glemaitre:
closes #1216 

This PR will make two splits instead of a single one to have a completely left-out set to not have any confusion when we present the `data_source="X_y"` principle.

### DIFF
--- a/examples/technical_details/plot_cache_mechanism.py
+++ b/examples/technical_details/plot_cache_mechanism.py
@@ -64,6 +64,10 @@ model
 from skore import train_test_split
 
 X_train, X_test, y_train, y_test = train_test_split(df, y, random_state=42)
+# Let's keep a completely left-out dataset
+X_train, X_external, y_train, y_external = train_test_split(
+    X_train, y_train, random_state=42
+)
 
 # %%
 # Caching the predictions for fast metric computation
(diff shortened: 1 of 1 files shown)


examples/technical_details/plot_cache_mechanism.py
```diff
@@ -64,6 +64,10 @@
 from skore import train_test_split
 
 X_train, X_test, y_train, y_test = train_test_split(df, y, random_state=42)
+# Let's keep a completely left-out dataset
+X_train, X_external, y_train, y_external = train_test_split(
+    X_train, y_train, random_state=42
+)
```
auguste-probabl, line 70: I think the name "validation set" is appropriate here.
```suggestion
# Let's keep a validation dataset
X_train, X_val, y_train, y_val = train_test_split(
    X_train, y_train, random_state=42
)
```
or at least 
```suggestion
# Let's keep a completely separate dataset
X_train, X_external, y_train, y_external = train_test_split(
    X_train, y_train, random_state=42
)
```

auguste-probabl (approved):

PR #1495, opened 2025-03-30
chore: Rename random_state to seed to highlight int-only support

### LINKED ISSUE #1397: api: Replace random_state by seed to follow SPEC7
I think it would nice to use SPEC7 and use `seed` instead of `random_state` and accept only integers: https://scientific-python.org/specs/spec-0007/

The advantage for us is that we can safely use the caching when dealing with integers while it starts to be tricky with `np.random.RandomState` because you would need to check the state of the random number generator.

So I think that we introduce `random_state` and limited it to an integer but we could change the name to `seed` that is indeed a more appropriate name.

glemaitre:
closes #1397 

This PR proposes to rename `random_state` handled by `skore` to `seed` since we can only support integral when it comes to caching.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
@@ -297,7 +297,7 @@ class _FeatureImportanceAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
         n_repeats: int = 5,
         max_samples: float = 1.0,
         n_jobs: Optional[int] = None,
-        random_state: Optional[int] = None,
+        seed: Optional[int] = None,
         flat_index: bool = False,
     ) -> pd.DataFrame:
         """Report the permutation feature importance.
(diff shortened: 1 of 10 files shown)


skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -297,7 +297,7 @@ def permutation(
         n_repeats: int = 5,
         max_samples: float = 1.0,
         n_jobs: Optional[int] = None,
-        random_state: Optional[int] = None,
```
auguste-probabl, line 300: I am starting to think that `seed` should not be `None` by default, as a matter of good practice (+ it makes cache-ing useless, even though it's a nice feature for skore). Can you point to any discussions in sklearn about this @glemaitre ?

auguste-probabl (approved):
Nice, much cleaner!

skore/src/skore/sklearn/_estimator/feature_importance_accessor.py
glemaitre: The problem is that `None` is not a bad choice in terms of statistical analysis. You experimentation should be robust to those. The seed is only nice for pure reproducibility. So it would be a bit weird to not make it the default since that it methodologically it is kind of correct.

auguste-probabl: I don't think robustness and reproducibility are contradictory: you can check that a model is robust by trying it for many explicit seeds (like `1,2,3,4,5`). The question is, how would you then teach a DS this good practice...

PR #1496, opened 2025-03-30
fix: Check if private metrics are available for an estimator

### LINKED ISSUE #1396: enh: Restrain the possible set of predefined metrics based on estimator when calling `report_metrics`
Right now, when creating an `EstimatorReport` and call `report.metrics.report_metrics`, we don't check that the metric requested when it is a string is a metric available for the type of estimator.

We could improve the user experience by validating that the requested metric is meaningful for the underlying estimator. Maybe, `hasattr(self, metric_name)` would be enough because we only define metric depending on the ML task.

glemaitre:
closes #1396 

The private metric functions should be decorate as the public one such that we don't provide an unsupported metric for a class of estimators via the `report.metrics.report_metrics` method.

Note that one test was actually wrong. We needed to at least train the classifier to infer the proper ML task.

### DIFF
--- a/skore/src/skore/sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_estimator/metrics_accessor.py
@@ -1,6 +1,7 @@
 import inspect
 from collections.abc import Iterable
 from functools import partial
+from operator import attrgetter
 from typing import Any, Callable, Literal, Optional, Union
 
 import joblib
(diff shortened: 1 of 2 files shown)


glemaitre:
Indeed, it makes sense and it is a easier to understand.

auguste-probabl (approved):

PR #1497, opened 2025-03-30
fix: Add classification type for _find_ml_task when y is unknown

glemaitre:
It is a bit weird to get an `unknown` type for `_find_ml_task` by providing an unfitted classifier and no target. We could instead return `classification`. We cannot infer better but it is more inline with whatever task we have at hand.

### DIFF
--- a/skore/src/skore/sklearn/find_ml_task.py
+++ b/skore/src/skore/sklearn/find_ml_task.py
@@ -138,7 +138,8 @@ def _find_ml_task(y, estimator=None) -> MLTask:
             else:
                 # fallback on the target
                 if y is None:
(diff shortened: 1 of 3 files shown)


auguste-probabl (approved):
Why not, we'll see with usage if this could lead to issues.

glemaitre:
Potentially, it could allow to relax some of the metrics availability (when a metric support both binary and multiclass, then we can as well accept `classification`).

glemaitre:
I had a borderline example that was this case. I refrained to implement it because I'm not sure that we should get in practice. But the getting `_ml_task == "unknown"` was the suprising part :)

PR #1502, opened 2025-03-31
fix: Make sure that average != None return a float for metric

### LINKED ISSUE #1501: bug: Precision and recall does not work when `average != None`

Currently there is an assertion checking for `float` or `dict`. However, in case that `average != None` is not taken into account and a dictionary is expected which is not the case.

glemaitre:
closes #1501 

Fixing the regression described in #1501 where averaged score were falsely expected to be a `dict` instead of a `float`.


auguste-probabl (approved):
Nice work.

PR #1475, opened 2025-03-27
feat(skore): Replace `skore/project` by `skore-local-project` or `skore-hub-project`

thomass-dev:
Part of #1436 .

- Replace `project` by a wrapper over `skore-local-project` and `skore-hub-project`
- Replace `project.get` by `project.reports.metadata` (new search feature)
- Remove deprecated `persistence/` and related tests
- Fix missing imports in tests
- Update documentation and examples

PR #1510, opened 2025-03-31
feat: Parallel coordinate plot to filter metadata

glemaitre:
This a PoC to see what we can do with parallel coordinate plot from plotly and a couple of widget to easily filter the dataframe containing the metadata.

So the design right now if the following:

image

In terms of code, I currently do:

```python
# %%
df = MetaDataFrame(data)
df

# We expect a user to make some selection and then we have the following

# %%
# Show the query string. Not a must have but maybe useful
df.get_selection_query()

# %%
# Get a pandas DataFrame and all pandas functionality
# we could do that on the original display but it will show the parallel coordinate plot
df.to_frame(filter=True).sort_values(by="accuracy_score", ascending=False)

# %%
# Get a comparison report directly
# We can get a list or a dict of EstimatorReport as well.
comparison_report = df.reports(filter=True)
comparison_report.metrics.report_metrics()

# %%
roc_display = comparison_report.metrics.roc()
roc_display.plot()
_ = roc_display.ax_.legend(bbox_to_anchor=(1.05, 1), loc="upper left")
```


nbesson-probabl:
I'm a bit surprised to see there's something else than performance metrics as // plot dimension here.
Isn't the scope of the first iteration to only filter on those perf metrics ?

glemaitre:
> Isn't the scope of the first iteration to only filter on those perf metrics ?

Basically, it will be defined by what I'll get from the metadata given in https://github.com/probabl-ai/skore/pull/1475. Here, they are only simulated.

glemaitre:
Couple of additional thoughts:

- I think that we should use a proper scoring rules (log-loss for classification and RMSE for regression). Because:
  - With classification and thresholded metrics, we have issue with the decision threshold and usually we never look at a specific metrics but a couple of them
  - a proper scoring rules provides a good idea to know if methodologically speaking a model is well-specified. Usually, one wants to choose one of those
  - it always a single scalar (in the single output case)
  - they are always computed by default
- I don't think that business metric should be part of this first filtering because they might be different depending on the dataset and right now, it is not clear to me that we have a single problem when we filter.
- Business metrics of thresholded metrics analysis can be differed to the analysis when getting the `ComparisonReport` where we can provide much better information regarding the methodology and interpretation to have.

nbesson-probabl:
> > Isn't the scope of the first iteration to only filter on those perf metrics ?
> 
> Basically, it will be defined by what I'll get from the metadata given in #1475. Here, they are only simulated. The one thing that I think could come for free is the learner because it is known as `estimator_name_` in the report. The scaler and encoder steps are just to think out of the box.

Ok ! Thank you

nbesson-probabl:
@glemaitre 

> The selector on the dataset is a must have because a ComparisonReport can only be built on EstimatorReport with the same (X_test, y_test).

Agreed, yet as discussed with @MarieSacksick I think it should be in a selector (as the ML Task) instead of as a dimension of the // plot itself.
I don't believe we'll ever be able to properly "name" a dataset from its properties, and I'd rather have a selector based on a complex presentation (number of entries / number of columns, whatever is important :sweat_smile: ) than any kind of name
So it's probably safer to select it "out" of the // plot !

glemaitre:
> I don't believe we'll ever be able to properly "name" a dataset from its properties, and I'd rather have a selector based on a complex presentation

If it is in a dropdown menu, would you expect to just put some hashes? 

I have probably 2 thoughts that make me think that a dimension is not that bad:

- I can see connection between dataset and learners. Potentially, it allow me to guess/recall what model I tried on which dataset and make me infer which one to quickly select. The dropdown menu does not allow to see this relationship
- I could imagine that we could have an HTML tooltip showing the output of `X.info()` when hovering on the dataset id shown in the dimension. In this case, it would be easier to have it as a dimension than in the a dropdown menu.

nbesson-probabl:
> If it is in a dropdown menu, would you expect to just put some hashes?

Nope nope, a dropdown of complex representations (as what you would put in a tooltip)

> I could imagine that we could have an HTML tooltip showing the output of X.info() when hovering on the dataset id shown in the dimension. In this case, it would be easier to have it as a dimension than in the a dropdown menu.

As discussed with @MarieSacksick, this UX seems quite painful, it's very unusual to allow filtering / selection on options that are not displayed by default. It somehow means that you need to hover on each item to know which ones are worth selecting.
And often in the UX field, "unusual" options must be used only if we have solid arguments in favor of it. 

> I can see connection between dataset and learners. Potentially, it allow me to guess/recall what model I tried on which dataset and make me infer which one to quickly select. The dropdown menu does not allow to see this relationship

Well, you and @MarieSacksick are the experts on this matter.
But as it's a somehow degraded UX (with tooltips), I'd rather have a **selector** for the majority of use cases, and if we display a specific list of dimensions that justifies to add dataset as a dimension, only in this case add it in the // plot too.
Does that makes sense ?

thomass-dev:
Closed in favor of https://github.com/probabl-ai/skore/pull/1560.

PR #1512, opened 2025-04-02
feat: Compare `CrossValidationReport`s

### LINKED ISSUE #1414: feat(ComparisonReport): Be able to compare several `CrossValidationReport`

Currently, in skore v0.8, the `ComparisonReport` can compare several `EstimatorReport` which is great. However, it can not compare several `CrossValidationReport`, which would be very useful to users.

Be able to compare several `CrossValidationReport`

_No response_

### Additional context

_No response_

auguste-probabl:
This PR makes it possible to pass a list of `CrossValidationReport`s to the `ComparisonReport` (which currently only accepts `EstimatorReport`s).

For now, plots like `Comparison.metrics.roc_curve` are not supported for ComparisonReports holding CrossValidationReports.

---

Example:
```python
>>> from sklearn.datasets import make_classification
>>> from sklearn.linear_model import LogisticRegression
>>> from skore import ComparisonReport, CrossValidationReport
>>> X, y = make_classification(random_state=42)
>>> estimator_1 = LogisticRegression()
>>> estimator_2 = LogisticRegression(C=2)  # Different regularization
>>> report_1 = CrossValidationReport(estimator_1, X, y)
>>> report_2 = CrossValidationReport(estimator_2, X, y)
>>> report = ComparisonReport({"model1": report_1, "model2": report_2})
>>> report.metrics.report_metrics(aggregate=None)
                                                    Value
Metric       Label / Average Estimator Split
Precision    0               model1    Split #0  1.000000
                                       Split #1  1.000000
                                       Split #2  1.000000
                                       Split #3  1.000000
                                       Split #4  0.909091
...                                                   ...
Predict time                 model2    Split #0  0.000185
                                       Split #1  0.000177
                                       Split #2  0.000187
                                       Split #3  0.000177
                                       Split #4  0.000188

[80 rows x 1 columns]
>>> report.metrics.report_metrics()
                                            mean       std
Metric       Label / Average Estimator
Precision    0               model1     0.981818  0.040656
                             model2     1.000000  0.000000
             1               model1     0.981818  0.040656
                             model2     0.981818  0.040656
Recall       0               model1     0.980000  0.044721
                             model2     0.980000  0.044721
             1               model1     0.980000  0.044721
                             model2     1.000000  0.000000
ROC AUC                      model1     0.996000  0.008944
                             model2     0.996000  0.008944
Brier score                  model1     0.022532  0.013130
                             model2     0.018731  0.014566
Fit time                     model1     0.001959  0.000559
                             model2     0.002603  0.000185
Predict time                 model1     0.000198  0.000014
                             model2     0.000183  0.000005
```

---

Closes #1414 

Todo:
- [x] Coverage
- [x] mypy
- [ ] Docs probably
- [x] Fix bug with progress bars not getting cleaned up
- [x] Format tables properly according to #1414 
- [x] ~Make plots work~ deferred
  - [x] Add NotImplementedError to plots
- [x] Test what happens when `aggregate` is passed to `ComparisonReport[EstimatorReport]`
- [x] Test what happens when `X, y` is passed to `ComparisonReport[CrossValidationReport]`
- [x] Add doctests to `ComparisonReport` showcasing `ComparisonReport[CrossValidationReport]`
- [x] Add timings
- [x] Check if some estimators are named the same, if so make them different (do it when initializing `report.report_names_`)

### DIFF
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -296,37 +296,17 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
                 aggregate = [aggregate]
 
             if "Label / Average" in df.columns:
-                by = ["Metric", "Label / Average", "Estimator"]
+                index = ["Metric", "Label / Average"]
             else:
-                by = ["Metric", "Estimator"]
+                index = ["Metric"]
 
-            df = df.groupby(by=by, sort=False)["Value"].agg(func=aggregate)
-
-            df = df.reset_index()
-
-            metric_order = df["Metric"].unique()
-
-            df["metric_order_index"] = df["Metric"].apply(
-                lambda x: list(metric_order).index(x)
+            df = df.pivot_table(
+                index=index,
+                columns="Estimator",
+                values="Value",
+                aggfunc=aggregate,
+                sort=False,
             )
-
-            if "Label / Average" in df.columns:
-                by = ["metric_order_index", "Label / Average", "Estimator"]
-            else:
-                by = ["metric_order_index", "Estimator"]
-
-            df = (
-                df.sort_values(by=by)
-                .drop("metric_order_index", axis=1)
-                .reset_index(drop=True)
-            )
-
-            if "Label / Average" in df.columns:
-                new_index = ["Metric", "Label / Average", "Estimator"]
-            else:
-                new_index = ["Metric", "Estimator"]
-
-            df = df.set_index(new_index)
         else:
             # Make sure the Split column goes [0, 1, 2, 0, 1, 2]
             # Rather than [0, 0, 1, 1, 2, 2]
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -645,6 +645,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             data_source=data_source,
             X=X,
             y=y,
+            aggregate=aggregate,
         )
 
     @available_if(
@@ -760,6 +761,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             y=y,
             pos_label=pos_label,
             scoring_kwargs={"average": average},
+            aggregate=aggregate,
         )
 
     @available_if(
@@ -876,6 +878,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             y=y,
             pos_label=pos_label,
             scoring_kwargs={"average": average},
+            aggregate=aggregate,
         )
 
     @available_if(
@@ -954,6 +957,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             data_source=data_source,
             X=X,
             y=y,
+            aggregate=aggregate,
         )
 
     @available_if(
@@ -1073,6 +1077,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             X=X,
             y=y,
             scoring_kwargs={"average": average, "multi_class": multi_class},
+            aggregate=aggregate,
         )
 
     @available_if(
@@ -1153,6 +1158,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             data_source=data_source,
             X=X,
             y=y,
+            aggregate=aggregate,
         )
 
     @available_if(
@@ -1245,6 +1251,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             X=X,
             y=y,
             scoring_kwargs={"multioutput": multioutput},
+            aggregate=aggregate,
         )
 
     @available_if(
@@ -1337,6 +1344,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             X=X,
             y=y,
             scoring_kwargs={"multioutput": multioutput},
+            aggregate=aggregate,
         )
 
     def custom_metric(
@@ -1455,6 +1463,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
             X=X,
             y=y,
             scoring_names=[metric_name] if metric_name is not None else None,
+            aggregate=aggregate,
         )
 
     ####################################################################################
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -252,71 +252,22 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
 
             return df
 
-        def reshape_results(df):
-            """Put data in the correct format.
-
-            - Index is "Metric", optionally "Label / Average", "Split"
-            - Columns are "Estimator"
-
-            Examples
-            --------
-            >>> # xdoctest: +SKIP
-            >>> df
-                                                    Split #0  Split #1  Split #2
-            Estimator Metric       Label / Average
-            m1        Precision    0                1.000000  1.000000  1.000000
-                                   1                1.000000  1.000000  1.000000
-                      ROC AUC                       1.000000  1.000000  1.000000
-            m2        Precision    0                1.000000  1.000000       NaN
-                                   1                1.000000  0.941176       NaN
-                      ROC AUC                       1.000000  0.996324       NaN
-            >>> reshape_results(df)
-            Estimator                                   m1        m2
-            Metric      Label / Average Split
-            Precision   0               Split #0  1.000000  1.000000
-                                        Split #1  1.000000  1.000000
-                                        Split #2  1.000000       NaN
-                        1               Split #0  1.000000  1.000000
-                                        Split #1  1.000000  0.941176
-                                        Split #2  1.000000       NaN
-            ROC AUC                     Split #0  1.000000  1.000000
-                                        Split #1  1.000000  0.996324
-                                        Split #2  1.000000       NaN
-            """
-            splits = df.columns
-
+        def melt_df(df):
             df_reset = df.reset_index()
 
-            metric_order = df_reset["Metric"].unique()
-
-            # Melt the Split columns into rows
             if "Label / Average" in df_reset.columns:
-                id_vars = ["Estimator", "Metric", "Label / Average"]
+                id_vars = ["Metric", "Label / Average", "Estimator"]
             else:
-                id_vars = ["Estimator", "Metric"]
-            melted_df = pd.melt(
+                id_vars = ["Metric", "Estimator"]
+
+            melted = pd.melt(
                 df_reset,
                 id_vars=id_vars,
-                value_vars=splits,
                 var_name="Split",
                 value_name="Value",
             )
 
-            # Now pivot to have models as columns
-            if "Label / Average" in melted_df.columns:
-                index = ["Metric", "Label / Average", "Split"]
-            else:
-                index = ["Metric", "Split"]
-            result_df = pd.pivot_table(
-                melted_df,
-                index=index,
-                columns="Estimator",
-                values="Value",
-            )
-
-            result_df = result_df.reindex(metric_order, level="Metric")
-
-            return result_df
+            return melted
 
         results = individual_results.copy()
 
@@ -330,20 +281,78 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         else:
             favorability = None
 
-        df_model_name_in_index = pd.concat(
-            [
-                add_model_name_to_index(df, estimator_name)
-                for df, estimator_name in zip(results, estimator_names)
-            ]
-        )
+        dfs_model_name_in_index = [
+            add_model_name_to_index(df, estimator_name)
+            for df, estimator_name in zip(results, estimator_names)
+        ]
+
+        dfs_melted = [melt_df(df) for df in dfs_model_name_in_index]
+
+        df = pd.concat(dfs_melted, axis=0)
 
         if aggregate:
             if isinstance(aggregate, str):
                 aggregate = [aggregate]
 
-            df = df_model_name_in_index.aggregate(func=aggregate, axis=1)
+            if "Label / Average" in df.columns:
+                by = ["Metric", "Label / Average", "Estimator"]
+            else:
+                by = ["Metric", "Estimator"]
+
+            df = df.groupby(by=by, sort=False)["Value"].agg(func=aggregate)
+
+            df = df.reset_index()
+
+            metric_order = df["Metric"].unique()
+
+            df["metric_order_index"] = df["Metric"].apply(
+                lambda x: list(metric_order).index(x)
+            )
+
+            if "Label / Average" in df.columns:
+                by = ["metric_order_index", "Label / Average", "Estimator"]
+            else:
+                by = ["metric_order_index", "Estimator"]
+
+            df = (
+                df.sort_values(by=by)
+                .drop("metric_order_index", axis=1)
+                .reset_index(drop=True)
+            )
+
+            if "Label / Average" in df.columns:
+                new_index = ["Metric", "Label / Average", "Estimator"]
+            else:
+                new_index = ["Metric", "Estimator"]
+
+            df = df.set_index(new_index)
         else:
-            df = reshape_results(df_model_name_in_index)
+            # Make sure the Split column goes [0, 1, 2, 0, 1, 2]
+            # Rather than [0, 0, 1, 1, 2, 2]
+
+            metric_order = df["Metric"].unique()
+
+            df["metric_order_index"] = df["Metric"].apply(
+                lambda x: list(metric_order).index(x)
+            )
+
+            if "Label / Average" in df.columns:
+                by = ["metric_order_index", "Label / Average", "Estimator", "Split"]
+            else:
+                by = ["metric_order_index", "Estimator", "Split"]
+
+            df = (
+                df.sort_values(by=by)
+                .drop("metric_order_index", axis=1)
+                .reset_index(drop=True)
+            )
+
+            new_index = []
+            for col in df.columns:
+                if col == "Value":
+                    break
+                new_index.append(col)
+            df = df.set_index(new_index)
 
         if favorability is not None:
             df["Favorability"] = favorability
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -237,21 +237,15 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
                     require="sharedmem",
                 )
             )
-            if self._parent._reports_type == "EstimatorReport":
-                kwargs = dict(
-                    data_source=data_source,
-                    X=X,
-                    y=y,
-                    **metric_kwargs,
-                )
-            else:  # "CrossValidationReport"
-                kwargs = dict(
-                    data_source=data_source,
-                    X=X,
-                    y=y,
-                    aggregate=None,
-                    **metric_kwargs,
-                )
+
+            kwargs = dict(
+                data_source=data_source,
+                X=X,
+                y=y,
+                **metric_kwargs,
+            )
+            if self._parent._reports_type == "CrossValidationReport":
+                kwargs["aggregate"] = None
 
             generator = parallel(
                 joblib.delayed(getattr(report.metrics, report_metric_name))(**kwargs)
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -111,6 +111,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -480,6 +481,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -602,6 +604,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -715,6 +718,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -832,6 +836,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -914,6 +919,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -1033,6 +1039,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -1115,6 +1122,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -1207,6 +1215,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -1300,6 +1309,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
@@ -1406,6 +1416,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
+            Ignored when comparison is between EstimatorReports.
 
         Returns
         -------
--- a/skore/src/skore/sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_comparison/metrics_accessor.py
@@ -112,7 +112,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
-            Ignored when comparison is between EstimatorReports.
+            Ignored when comparison is between :class:`~skore.EstimatorReport`s.
 
         Returns
         -------
@@ -299,7 +299,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
-            Ignored when comparison is between EstimatorReports.
+            Ignored when comparison is between :class:`~skore.EstimatorReport`s.
 
         Returns
         -------
@@ -422,7 +422,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
-            Ignored when comparison is between EstimatorReports.
+            Ignored when comparison is between :class:`~skore.EstimatorReport`s.
 
         Returns
         -------
@@ -536,7 +536,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
-            Ignored when comparison is between EstimatorReports.
+            Ignored when comparison is between :class:`~skore.EstimatorReport`s.
 
         Returns
         -------
@@ -654,7 +654,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
-            Ignored when comparison is between EstimatorReports.
+            Ignored when comparison is between :class:`~skore.EstimatorReport`s.
 
         Returns
         -------
@@ -737,7 +737,7 @@ class _MetricsAccessor(_BaseAccessor, DirNamesMixin):
         aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
             Function to aggregate the scores across the cross-validation splits.
             None will return the scores for each split.
-            Ignored when comparison is between EstimatorReports.
+            Ignored when comparison is between :class:`~skore.EstimatorReport`s.
 
         Returns
         -------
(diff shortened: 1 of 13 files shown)


skore/src/skore/sklearn/_cross_validation_comparison/metrics_accessor.py
```diff
@@ -0,0 +1,1402 @@
...
+
+    @available_if(
+        _check_supported_ml_task(
+            supported_ml_tasks=["binary-classification", "multiclass-classification"]
+        )
+    )
+    def accuracy(
+        self,
+        *,
+        data_source: DataSource = "test",
```
glemaitre, line 368: We should expose an `aggregate` parameters that should aggregate the results from each cross-validation estimator.

auguste-probabl: done

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -61,9 +62,7 @@ def report_metrics(
         pos_label: Optional[Union[int, float, bool, str]] = None,
         indicator_favorability: bool = False,
         flat_index: bool = False,
-        aggregate: Optional[
-            Union[Literal["mean", "std"], Sequence[Literal["mean", "std"]]]
-        ] = ("mean", "std"),
+        aggregate: Optional[Aggregate] = ("mean", "std"),
```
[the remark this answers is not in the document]
auguste-probabl: Fair point. I checked `help(CrossValidationComparisonReport.metrics)` and it looks like it unfolds all the definitions:
```
Help on class _MetricsAccessor in module skore.sklearn._cross_validation_comparison.metrics_accessor:

class _MetricsAccessor(skore.sklearn._base._BaseAccessor, skore.externals._pandas_accessors.DirNamesMixin)
 |  _MetricsAccessor(parent: skore.sklearn._cross_validation_comparison.report.CrossValidationComparisonReport) -> None
 |
 |  Accessor for metrics-related operations...
 |
 |  Methods defined here:
 |
 |  ...
 |
 |  accuracy(self, *, data_source: Literal['test', 'train'] = 'test', aggregate: Union[Literal['mean', 'std'], collections.abc.Sequence[Lite
ral['mean', 'std']], NoneType] = ('mean', 'std')) -> pandas.core.frame.DataFrame
 |      Compute the accuracy score.
 |      ...
```

As for our own docs we don't show these types.

auguste-probabl:
Discussed IRL: The code common between `ComparisonReport` and `CrossValidationComparisonReport` should be factored in a `BaseComparisonReport`.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -85,82 +86,170 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
 
-        test_dataset_hashes = {
-            joblib.hash((report.X_test, report.y_test))
-            for report in reports
-            if not ((report.X_test is None) and (report.y_test is None))
-        }
-        if len(test_dataset_hashes) > 1:
-            raise ValueError("Expected all estimators to have the same testing data.")
+        if len(set(id(report) for report in reports_list)) < len(reports_list):
+            raise ValueError("Expected reports to be distinct objects")
```
auguste-probabl, line 190: This new constraint results from #1536; indeed, if the same report is passed twice, when we reset the progress bar at the end of the first computation (see below), when the second computation starts, the progress object is set to `None`, whereas it should be set to the ComparisonReport's progress object.

https://github.com/probabl-ai/skore/blob/836f171ece041d9b48434319bf03b1c152d2c186/skore/src/skore/utils/_progress_bar.py#L85-L86

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -85,82 +86,170 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
-        self.estimator_reports_ = reports
+        if isinstance(reports_list[0], CrossValidationReport):
+            reports_type = "CrossValidationReport"
+        elif isinstance(reports_list[0], EstimatorReport):
+            reports_type = "EstimatorReport"
+        else:
+            raise TypeError(
+                "Report type is undetermined. "
+                "This error should have been caught during validation."
+            )
```
auguste-probabl, line 213: In theory this is an unreachable statement

skore/src/skore/utils/_progress_bar.py
auguste-probabl, line 1: #1536

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl, line 1: #1535

skore/tests/unit/sklearn/comparison/estimator/test_comparison.py
```diff
@@ -65,7 +67,7 @@ def test_comparison_report_without_testing_data(binary_classification_model):
     estimator, X_train, _, y_train, _ = binary_classification_model
     estimator_report = EstimatorReport(estimator, X_train=X_train, y_train=y_train)
 
-    report = ComparisonReport([estimator_report, estimator_report])
```
auguste-probabl, line 68: This is necessary because of the new constraint that all reports must be distinct

skore/src/skore/sklearn/_comparison/report.py
auguste-probabl: Note that right now this constraint is only really useful when comparing CVReports, not when comparing EstimatorReports. I didn't put it in the if-block because I can imagine that one day EstimatorReports might have more progress bars of their own.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -85,82 +86,170 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
     >>> report = ComparisonReport(
     ...     {"model1": estimator_report_1, "model2": estimator_report_2}
     ... )
-    ...
+
+    >>> from skore import CrossValidationReport
+    >>> report_1 = CrossValidationReport(estimator_1, X, y)
+    >>> report_2 = CrossValidationReport(estimator_2, X, y)
+    >>> report = ComparisonReport([report_1, report_2])
+    >>> report = ComparisonReport({"model1": report_1, "model2": report_2})
```
auguste-probabl, line 97: Currently the only docstring where we showcase comparing CrossValidationReports.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -2,20 +2,23 @@
...
 from skore.sklearn._estimator.report import EstimatorReport
 from skore.utils._progress_bar import progress_decorator
 
 if TYPE_CHECKING:
     from skore.sklearn._estimator.metrics_accessor import _MetricsAccessor
 
+    ReportType = Literal["EstimatorReport", "CrossValidationReport"]
+
 
 class ComparisonReport(_BaseReport, DirNamesMixin):
```
auguste-probabl, line 23: Maybe it would make sense to type it more strongly, e.g. `ComparisonReport[EstimatorReport]`.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
+    >>> from sklearn.datasets import make_classification
+    >>> from sklearn.linear_model import LogisticRegression
+    >>> from skore import ComparisonReport, CrossValidationReport
+    >>> X, y = make_classification(random_state=42)
+    >>> estimator_1 = LogisticRegression()
+    >>> estimator_2 = LogisticRegression(C=2)  # Different regularization
+    >>> report_1 = CrossValidationReport(estimator_1, X, y)
+    >>> report_2 = CrossValidationReport(estimator_2, X, y)
+    >>> report = ComparisonReport([report_1, report_2])
+    >>> report = ComparisonReport({"model1": report_1, "model2": report_2})
```
auguste-probabl, line 104: The only docstring where we showcase `ComparisonReport[CVReport]`

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
+        for report_name in report_names_:
+            indexes_of_report_names = [
+                index for index, name in enumerate(report_names_) if name == report_name
+            ]
+            if len(indexes_of_report_names) == 1:
+                # report name appears only once
+                continue
+            for n, index in enumerate(indexes_of_report_names, start=1):
+                result[index] = f"{report_name}_{n}"
+        return result
```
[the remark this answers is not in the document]
glemaitre: Just a quick benchmark even if we deal with really small list in practice:

```python
names = np.random.choice([f"{i}" for i in range(100)], size=10000).tolist()
```

```
In [19]: %%time
    ...: _deduplicate_report_names(names)
CPU times: user 2.26 s, sys: 5.99 ms, total: 2.26 s
Wall time: 2.26 s
```

and the equivalent of the one above:

```
In [20]: %%time
    ...: deduplicate_with_suffix_inplace(names)
CPU times: user 6.51 ms, sys: 69 μs, total: 6.58 ms
Wall time: 6.59 ms
```

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -46,7 +49,7 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
 
     Attributes
     ----------
-    estimator_reports_ : list of :class:`~skore.EstimatorReport`
+    reports_ : list of :class:`~skore.EstimatorReport`
```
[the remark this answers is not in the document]
glemaitre: @auguste-probabl what do you think to have `report_` to always be a `dict` tha also include the names instead to have 2 distinct parameters?

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
+            list[EstimatorReport],
+            dict[str, EstimatorReport],
+            list[CrossValidationReport],
+            dict[str, CrossValidationReport],
+        ],
+    ) -> tuple[
+        Union[list[EstimatorReport], list[CrossValidationReport]],
+        list[str],
+        ReportType,
+    ]:
```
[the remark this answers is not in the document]
glemaitre: Since we return values, it could be cool to have an extended docstring actually.

skore/tests/unit/sklearn/comparison/estimator/test_comparison.py
```diff
@@ -590,6 +617,25 @@ def test_estimator_report_report_metrics_indicator_favorability(
...
+        estimator,
+        X_train=X_train,
+        y_train=y_train,
+        X_test=X_test,
+        y_test=y_test,
+    )
+
+    report = ComparisonReport([estimator_report, copy(estimator_report)])
+    assert_allclose(
+        report.metrics.report_metrics(aggregate="mean"),
```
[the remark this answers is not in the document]
glemaitre: Yep I see that we cannot do better. So you can ignore my comment.

glemaitre (commented):
A first round of review before to go in the metrics.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -314,6 +560,7 @@ def accuracy(
         data_source: DataSource = "test",
         X: Optional[ArrayLike] = None,
         y: Optional[ArrayLike] = None,
+        aggregate: Optional[Aggregate] = ("mean", "std"),
```
glemaitre, line 563: It looks like you did not pass the `aggregate` parameter to `report_metrics` for several metrics.

skore/src/skore/sklearn/_comparison/report.py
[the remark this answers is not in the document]
auguste-probabl: no strong feelings; i guess it would make ComparisonReport a bit more different from CVReport.

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
-        ml_tasks = {report: report._ml_task for report in reports}
+        ml_tasks = {report: report._ml_task for report in reports_list}
         if len(set(ml_tasks.values())) > 1:
             raise ValueError(
                 f"Expected all estimators to have the same ML usecase; got {ml_tasks}"
             )
 
         if report_names is None:
-            self.report_names_ = [report.estimator_name_ for report in reports]
+            deduped_report_names = ComparisonReport._deduplicate_report_names(
```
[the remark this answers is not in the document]
auguste-probabl: I prefer making functions as pure as possible

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
+        Union[list[EstimatorReport], list[CrossValidationReport]],
+        list[str],
+        ReportType,
+    ]:
         if not isinstance(reports, Iterable):
             raise TypeError(f"Expected reports to be an iterable; got {type(reports)}")
 
         if len(reports) < 2:
-            raise ValueError("At least 2 instances of EstimatorReport are needed")
+            raise ValueError("Expected at least 2 reports to compare")
```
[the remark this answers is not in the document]
auguste-probabl: to be honest i don't think this error is warranted; it should work with 1 estimator, and if the list is empty then the error should be different

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
 
         report_names = (
             list(map(str, reports.keys())) if isinstance(reports, dict) else None
         )
```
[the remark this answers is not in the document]
auguste-probabl: be conservative in what you accept?

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
-        if not all(isinstance(report, EstimatorReport) for report in reports):
-            raise TypeError("Expected instances of EstimatorReport")
+        reports_list = list(reports.values()) if isinstance(reports, dict) else reports
+
+        reports_type: ReportType
+        if all(isinstance(report, EstimatorReport) for report in reports_list):
+            reports_list = cast(list[EstimatorReport], reports_list)
+            reports_type = "EstimatorReport"
+
+            test_dataset_hashes = {
```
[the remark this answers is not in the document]
auguste-probabl: same, this constraint of "same test data" is not completely necessary to me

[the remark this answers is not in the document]
auguste-probabl: do you want to just do the change?
```suggestion
            test_dataset_hashes = {
                joblib.hash(report.y_test) for report in reports_list
                if report.y_test is not None)
            }
```

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -88,77 +91,161 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
...
+        -----
+        We check that the estimator reports can be compared:
+        - all reports are estimator reports,
+        - all estimators are in the same ML use case,
+        - all estimators have non-empty X_test and y_test,
+        - all estimators have the same X_test and y_test.
+        """
+        self.reports_, self.report_names_, self._reports_type = (
+            ComparisonReport._validate_reports(reports)
+        )
```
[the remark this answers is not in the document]
auguste-probabl: i don't disagree, but i don't want to give away the power of pure functions XD

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: ^^'

skore/src/skore/sklearn/_comparison/report.py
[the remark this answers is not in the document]
auguste-probabl: added comment

skore/src/skore/sklearn/_comparison/report.py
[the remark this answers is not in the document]
auguste-probabl: add "got..."

skore/src/skore/sklearn/_comparison/report.py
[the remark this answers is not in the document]
auguste-probabl: thanks!

skore/src/skore/sklearn/_comparison/report.py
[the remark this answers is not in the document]
glemaitre: The naming is more meaningful for the `ComparisonReport` because we can have heterogeneous report. Let's go this direction and support a dictionary. I think that it will simplified the code in some place.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+        if flat_index and isinstance(results.index, pd.MultiIndex):
+            results.index = flatten_multi_index(results.index)
+        return results
+
+    @staticmethod
+    def _combine_estimator_results(
+        individual_results: list[pd.DataFrame],
+        estimator_names: list[str],
+        indicator_favorability: bool,
+    ) -> pd.DataFrame:
```
glemaitre, line 178: What would be useful is to have a docstring to show kind of the input and we expect as output.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
         return results
 
+    @staticmethod
+    def _combine_cross_validation_results(
+        individual_results: list[pd.DataFrame],
+        estimator_names: list[str],
+        indicator_favorability: bool,
+        aggregate: Optional[Aggregate],
+    ) -> pd.DataFrame:
+        """Combine a list of dataframes.
```
glemaitre, line 203:
```suggestion
        """Combine a list of dataframes provided by `CrossValidationReport`s.
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+
+            - every dataframe has the form:
+                - index: Index "Metric", or MultiIndex ["Metric", "Label / Average"]
+                - columns: MultiIndex with levels ["Estimator", "Splits"] (can be
+                  unnamed)
+            - all dataframes have the same metrics
+
+            The dataframes are not required to have the same number of columns (splits).
+
+        estimator_names : list of str of len (len(results))
```
glemaitre, line 220:
```suggestion
        estimator_names : list of str of length `len(results)`
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+            else:
+                new_index = ["Metric", "Estimator"]
+            df = df.reset_index().set_index(new_index)
+
+            # Then drop the model from the columns
+            df.columns = df.columns.droplevel(0)
+
+            return df
+
+        def melt_df(df):
```
glemaitre, line 249: add a small docstring with the transform that we do.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+            melted = pd.melt(
+                df_reset,
+                id_vars=id_vars,
+                var_name="Split",
+                value_name="Value",
+            )
+
+            return melted
+
+        results = individual_results.copy()
```
glemaitre, line 266: Can we avoid the copy here?

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+        indicator_favorability : bool
+            Whether to keep the Favorability column.
+
+        aggregate : Aggregate
+            How to aggregate the resulting dataframe.
+        """
+
+        def add_model_name_to_index(df: pd.DataFrame, model_name: str) -> pd.DataFrame:
+            """Move the model name from the column index to the table index."""
+            df = copy.copy(df)
```
glemaitre, line 232: Do we need to make a copy here?

I'm under the impression that we applied a sequence of transformation but that none of the transformation at one stage depends on the original dataframe. So if I'm not wrong, it means that we don't need a copy.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+
+            # Put the model name as a column
```
glemaitre, line 234: I think we don't need a comment here because the next line is explicit.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+            df["Estimator"] = model_name
+
+            # Put the model name into the index
```
glemaitre, line 237: The same here.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+            if "Label / Average" in df.index.names:
+                new_index = ["Metric", "Label / Average", "Estimator"]
+            else:
+                new_index = ["Metric", "Estimator"]
+            df = df.reset_index().set_index(new_index)
+
+            # Then drop the model from the columns
```
glemaitre, line 244:
```suggestion
            # Then drop the `estimator_name` from the columns index
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+            favorability = None
+
+        dfs_model_name_in_index = [
+            add_model_name_to_index(df, estimator_name)
+            for df, estimator_name in zip(results, estimator_names)
+        ]
+
+        dfs_melted = [melt_df(df) for df in dfs_model_name_in_index]
+
+        df = pd.concat(dfs_melted, axis=0)
```
glemaitre, line 285: I would not mind to chain the function to avoid having intermediate variable names.
```suggestion
        df = pd.concat([
            melt_df(add_model_name_to_index(df, estimator_name))
            for df, estimator_name in zip(results, estimator_names)
        ], axis=0)
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: Also we could call it `melt` since it will be call as `melt(df)`.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+
+        if aggregate:
```
glemaitre, line 287: let's put a small example of a the expected output when `aggregate is not None` in a comment.

And the same when it is `None` with a small comment on why we do so.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+            )
+        else:
+            # Make sure the Split column goes [0, 1, 2, 0, 1, 2]
+            # Rather than [0, 0, 1, 1, 2, 2]
+
+            metric_order = df["Metric"].unique()
+
+            df["metric_order_index"] = df["Metric"].apply(
+                lambda x: list(metric_order).index(x)
+            )
```
glemaitre, line 311: I think that we can leverage the categorical dtype from pandas:

```
            df["Metric"] = df["Metric"].astype(
                pd.CategoricalDtype(df["Metric"].unique(), ordered=True)
            )
            df["metric_order_index"] = df["Metric"].cat.codes
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -157,14 +164,175 @@ def report_metrics(
...
+                .drop("metric_order_index", axis=1)
+                .reset_index(drop=True)
+            )
+
+            new_index = []
+            for col in df.columns:
+                if col == "Value":
+                    break
+                new_index.append(col)
+            df = df.set_index(new_index)
```
glemaitre, line 329: Isn't it equivalent to:

```suggestion
            df = df.set_index(df.columns.drop("Value"))
```

skore/tests/unit/sklearn/comparison/estimator/test_comparison.py
```diff
@@ -351,36 +357,27 @@ def test_comparison_report_metrics_repr(binary_classification_model):
...
-    estimator_report = EstimatorReport(
-        estimator,
-        X_train=X_train,
-        y_train=y_train,
-        X_test=X_test,
-        y_test=y_test,
-    )
-
-    comp = ComparisonReport([estimator_report, estimator_report])
+    _, X_train, X_test, _, y_test = binary_classification_model
```
glemaitre, line 363:
```suggestion
    _, _, X_test, _, y_test = binary_classification_model
```

skore/tests/unit/sklearn/comparison/estimator/test_comparison.py
```diff
@@ -663,21 +646,15 @@ def test_comparison_report_plots(
...
-        estimator,
-        X_train=X_train,
-        y_train=y_train,
-        X_test=X_test,
-        y_test=y_test,
-    )
-
-    comp = ComparisonReport([estimator_report, estimator_report])
+    if plot_ml_task == "binary_classification":
+        estimator, _, X_test, _, y_test = binary_classification_model
```
glemaitre, line 653:
```suggestion
        _, _, X_test, _, y_test = binary_classification_model
```

skore/tests/unit/sklearn/comparison/estimator/test_comparison.py
```diff
@@ -663,21 +646,15 @@ def test_comparison_report_plots(
...
+        comp = report_classification
+    else:
+        estimator, _, X_test, _, y_test = regression_model
```
glemaitre, line 656:
```suggestion
        comp = report_classification
    else:
        _, _, X_test, _, y_test = regression_model
```

skore/tests/unit/sklearn/comparison/cross_validation/test_report_metrics.py
```diff
@@ -0,0 +1,199 @@
...
+        [
+            CrossValidationReport(DummyRegressor(), X, y),
+            CrossValidationReport(DummyRegressor(), X, y, cv_splitter=3),
+        ]
+    )
+
+    return report
+
+
+def test_different_split_numbers(report):
```
glemaitre, line 49: Actually, it would be great to a small description for each to know exactly what we test. It is usually self explicit but sometimes it is nice to know that we got some shortcuts.

skore/tests/unit/sklearn/comparison/cross_validation/test_report.py
```diff
@@ -0,0 +1,156 @@
+import re
```
glemaitre, line 1: Right now those, are kind of tested in the `test_cross_validation.py`. Do you think that we could quickly reorganize the test in a module and split the tests. We can do that once we completely agree that all tests are fine.

skore/tests/unit/sklearn/comparison/cross_validation/test_report.py
```diff
@@ -0,0 +1,156 @@
...
+
+
+@pytest.fixture
+def classification_data():
+    X, y = make_classification(class_sep=0.1, random_state=42)
+    return X, y
+
+
+@pytest.fixture
+def cv_report_classification(classification_data):
```
glemaitre, line 18: So I'm thinking that some of the test here could be just a parametrization potentially in the `test_cross_validation.py`. We should check that.
Usually, we only have a gentle assertion that should be fine for both `EstimatorReport` and `CrossValidationReport`.

glemaitre (commented):
A couple of more comments on the part that I did not watch before but it looks pretty good.

skore/src/skore/sklearn/_comparison/report.py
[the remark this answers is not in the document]
auguste-probabl: sure

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: The tests pass, but removing `copy` means that the function wlil have a side-effect (it modifies `individual_results`). Since `individual_results` is not huge I think it's worth keeping it pure. Maybe I can add a warning?

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: I added `copy` mostly for my sanity. Since the dataframes are not big, I think it's worth it. TBH this whole `combine` function feels quite dirty to me, but I don't know how to make it simpler.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: added doctests

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: I agree, but the current phrasing makes debugging quite a bit easier. This whole function is ugly to me.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: added doctest

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: `melt(df_with_name(df, estimator_name))` would be a nicer naming :)

I'm not convinced about the debugging because you will fail inside one of the function.

skore/src/skore/sklearn/_comparison/utils.py
```diff
@@ -0,0 +1,293 @@
...
+    ... )
+                           mean                                 std
+    Estimator DummyClassifier_1 DummyClassifier_2 DummyClassifier_1 DummyClassifier_2
+    Metric
+    Accuracy                ...               ...               ...               ...
+    """
+
+    def add_model_name_to_index(df: pd.DataFrame, model_name: str) -> pd.DataFrame:
+        """Move the model name from the column index to the table index."""
+        df = copy.copy(df)
```
glemaitre, line 192: If you really want to keep the copy.

```suggestion
        df = df.copy()
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: Can you provide me a minimum example where "Value" is not the last column?

glemaitre: In case that we really have this case, we can do:

```python
df = df.set_index(df.columns[:df.get_loc["Value"]])
```

skore/src/skore/sklearn/_comparison/metrics_accessor.py
glemaitre: It is weird that it is outdated but I would prefer this version. The list conversion and the index check looks not optimal.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -198,9 +373,15 @@ def _compute_metric_scores(
...
 
-        total_estimators = len(self._parent.estimator_reports_)
+        total_estimators = len(self._parent.reports_)
         progress.update(main_task, total=total_estimators)
 
+        if self._parent._reports_type == "CrossValidationReport":
+            # Ensure that there is only one Progress instance.
+            # EstimatorReports don't have `_parent_progress`
+            for report in self._parent.reports_:
+                report._parent_progress = progress
```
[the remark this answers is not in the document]
auguste-probabl: This makes `tests/unit/sklearn/comparison/estimator/test_comparison.py::test_estimator_report_cleaned_up` fail. The reason is the following:
- EstimatorReports do have a `_parent_progress` attribute; so before, they were skipped completely, but now they are not
- When a ComparisonReport computation is run, this results in running `EstimatorReport.metrics.report_metrics`
- `EstimatorReport.metrics.report_metrics` does not have a `progress_decorator`, therefore the internal EstimatorReports' _parent_progress will not get cleaned up at the end of the computation
- Since the `_parent_progress` are not cleaned up, the EstimatorReports are not pickle-able.

Possible solutions:
- Revert to old behaviour
- Add progress decorator to `EstimatorReport.metrics.report_metrics`
- Move cleaning up logic somewhere else (e.g. implement `EstimatorReport.__reduce__`)

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: applied suggestions. we can also debug with `breakpoint()`. Also renamed function

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: Removed copy, added note

skore/src/skore/sklearn/_comparison/metrics_accessor.py
auguste-probabl: > Can you provide me a minimum example where "Value" is not the last column?

No, all tests pass. I applied your first suggestion

skore/src/skore/sklearn/_comparison/utils.py
auguste-probabl: removed copy

skore/src/skore/sklearn/_comparison/metrics_accessor.py
```diff
@@ -211,45 +392,70 @@ def _compute_metric_scores(
                     require="sharedmem",
                 )
             )
-            generator = parallel(
-                joblib.delayed(getattr(report.metrics, report_metric_name))(
+            if self._parent._reports_type == "EstimatorReport":
```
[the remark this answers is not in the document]
auguste-probabl: done

skore/src/skore/utils/_testing.py
auguste-probabl, line 1: I can factor this out in another PR actually

skore/tests/unit/sklearn/comparison/cross_validation/test_report.py
auguste-probabl: move some tests to `test_report_common.py`

skore/src/skore/sklearn/_comparison/report.py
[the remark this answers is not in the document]
auguste-probabl: Discussed IRL: We would need to figure out whether users would prefer a dict. In that case, the current attributes can be made private so that nothing is broken.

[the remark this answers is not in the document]
auguste-probabl: issue opened.

skore/tests/unit/sklearn/comparison/cross_validation/test_report.py
auguste-probabl: moved tests to their own `cross_validation` folder; splitting might come later.

skore/src/skore/sklearn/_comparison/metrics_accessor.py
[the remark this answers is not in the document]
auguste-probabl: Discussed IRL: the `__reduce__`/`__set_state__` solution should not be preferred; we should rather ensure that every operation on a `Report` leaves it in a well-defined state.

glemaitre (approved):
Let's merge now. @auguste-probabl could you open an issue to track the improvement/refactoring for the progress bar to not forget.

PR #1513, opened 2025-04-02
docs: Extend contributing guide

MarieSacksick:
Following some feedback from contributors.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -56,8 +56,18 @@ Install dependencies and setup pre-commit with:
     make install-skore
 
 
-Pull request format
--------------------
+Choosing an issue
+-----------------
+
+If you are starting to contribute to open-source, you can start by an [issue tagged `good first issue`](https://github.com/probabl-ai/skore/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22).
+
+The implementation of some issues are not very detailed. You can either propose a solution, or choose only the issues with a Status "Ready".
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.rst
```diff
@@ -56,8 +56,18 @@ Install dependencies and setup pre-commit with:
...
-Pull request format
--------------------
+Choosing an issue
+-----------------
+
+If you are starting to contribute to open-source, you can start by an [issue tagged `good first issue`](https://github.com/probabl-ai/skore/issues?q=is%3Aissue%20state%3Aopen%20label%3A%22good%20first%20issue%22).
+
+The implementation of some issues are not very detailed. You can either propose a solution, or choose only the issues with a Status "Ready".
+
+You can also contribute to the code base without a specific issue related; in that case, please be precise in the description of your pull request.
```
thomass-dev, line 66: The common approach is to create an issue before any implementation.
It allows maintainers to discuss high-level issues on the issue, before reviewing the implementation.

For my information, why not following this practice?

MarieSacksick: I had in mind small stuff for which you don't want to open an issue (this PR for instance, or it could also be spelling correction).

MarieSacksick: if you think it's bad practice, we can remove this sentence, and worst case scenario, people will take 2 minutes to open a useless issue.

thomass-dev: I think it's a good practice to create an issue before the implementation, so IMHO, you should remove this sentence or change it to reflect this.

thomass-dev (approved):

PR #1520, opened 2025-04-04
feat: Turn report_metrics of ComparisonReport into Displays

MarieSacksick:
As part of issue https://github.com/probabl-ai/skore/issues/1782, turn the report_metrics of ComparisonReport into a display. The plot of this display will be a pair plot to visualize one metric against another.  

Add documentation.  

This is part of the narrative "I have several models, how can I choose the best one?". The user will need:  
- a comparison report
- and plots to compare.

blocked by https://github.com/probabl-ai/skore/pull/1788.


glemaitre:
From speaking with @GaelVaroquaux, it looks like this plot should be more general meaning that it should be a pairwise plot (comparing 2 scores). Time vs score is probably a good default but we should make it possible to tweak via the parameters.

skore/src/skore/sklearn/_plot/metrics/pair_plot.py
```diff
@@ -0,0 +1,220 @@
...
+
+    @classmethod
+    def from_metrics(
+        cls,
+        metrics,
+        perf_metric_x,
+        perf_metric_y,
+        data_source=None,
+    ):
+        """Create a confusion matrix display from predictions.
```
auguste-probabl, line 113: Is this up-to-date?

skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -396,6 +397,56 @@ def get_predictions(
...
+            The positive class when it comes to binary classification. When
+            `response_method="predict_proba"`, it will select the column corresponding
+            to the positive class. When `response_method="decision_function"`, it will
+            negate the decision function if `pos_label` is different from
+            `estimator.classes_[1]`.
+
+        Returns
+        -------
+        A matplotlib plot.
+
```
auguste-probabl, line 436:
```suggestion
```

skore/src/skore/sklearn/_plot/metrics/pair_plot.py
MarieSacksick: oups, no! It's fixed.

glemaitre (commented):
I think that we should think about the API question before to go further. It would have an impact on how the display will be created.

If we have a display with `report_metrics`, it only means that we will share a common display and the `kind` parameters will decide to show a representation or another.

MarieSacksick:
Closing - fresh start in https://github.com/probabl-ai/skore/pull/1816.

PR #1526, opened 2025-04-10
test: Make sure to close figures after doctests

auguste-probabl:
This suppresses a warning in tests about too many open matplotlib figures.


thomass-dev (commented):
Do you think we can factorize the `pyplot` fixture from the tests?

auguste-probabl:
> Do you think we can factorize the `pyplot` fixture from the tests?

Yes probably, by moving `conftest.py` to the top-level

auguste-probabl:
I can't manage to reproduce the problem, and it doesn't show up in CI, so I think I'll just close this for now.

PR #1528, opened 2025-04-10
refactor(types): Create a `Aggregate` type

auguste-probabl:


skore/src/skore/sklearn/types.py
```diff
@@ -20,6 +20,9 @@
 PositiveLabel = Union[int, float, bool, str]
 
 
+Aggregate = Union[Literal["mean", "std"], Sequence[Literal["mean", "std"]]]
```
thomass-dev, line 23: Nitpick

```suggestion
Aggregate = Union[Literal["mean", "std"], tuple[Literal["mean"], Literal["std"]]]
```

thomass-dev (commented):
LGTM, just a nitpick to fail on `aggregate=("mean", "std", "mean", "mean")`.

skore/src/skore/sklearn/types.py
auguste-probabl: That's not quite correct, as `("mean",)` should pass; also even if it's a bit weird, passing the same aggregation function several times does work:
```
>> report.metrics.report_metrics(aggregate=('mean', 'std', 'mean'))
                             LogisticRegression
                                           mean       std      mean
Metric       Label / Average
Precision    0                         0.938920  0.042185  0.938920
             1                         0.948367  0.024254  0.948367
Recall       0                         0.910377  0.046696  0.910377
             1                         0.963530  0.027874  0.963530
ROC AUC                                0.989126  0.006658  0.989126
Brier score                            0.043530  0.003416  0.043530
Fit time                               0.087589  0.017375  0.087589
Predict time                           0.000230  0.000009  0.000230
```

So I think we can leave it like this.

thomass-dev (approved):

PR #1542, opened 2025-04-12
docs: Pass format to `pd.to_datetime`

### LINKED ISSUE #1504: doc: do not use pandas.to_datetime without format argument

The examples https://github.com/probabl-ai/skore/blob/0.8.1/examples/getting_started/plot_skore_getting_started.py and https://github.com/probabl-ai/skore/blob/0.8.1/examples/model_evaluation/plot_train_test_split.py.

We use pandas.to_datetime without using the format argument, which is not good regarding date validation. In this case pandas uses some heuristics to try to guess the proper format and apply it to the whole column. But it is not reliable, it can make mistakes, and also, the data might not be clean (e.g. dates with different formats, invalid dates, etc) and require ad-hoc inspection. Better practice is to know ahead of time the expected format and enforce it with the format argument, and then explicitly add proper error management on top in case some dates do not comply (e.g. by using the other arguments of to_datetime).

### Suggested improvement

fill the `format` argument.

### Additional context

_No response_

ankith26:
Closes #1504.

I simply inspected the `X_employee["date_first_hired"]` series by printing it, and saw that it uses the `mm/dd/yyyy` format. I have updated both examples to use the corresponding format argument.

The linked issue also mentions `skrub.ToDatetime` but I am not clear on what exactly is to be done with that. It looks like that API is only used in one example: `examples/use_cases/plot_employee_salaries.py`. If anyone can clarify what needs to be done I would love to take it up in this PR or a future PR.

### DIFF
--- a/examples/getting_started/plot_skore_getting_started.py
+++ b/examples/getting_started/plot_skore_getting_started.py
@@ -232,7 +232,9 @@ from skrub.datasets import fetch_employee_salaries
 
 dataset_employee = fetch_employee_salaries()
 X_employee, y_employee = dataset_employee.X, dataset_employee.y
-X_employee["date_first_hired"] = pd.to_datetime(X_employee["date_first_hired"])
+X_employee["date_first_hired"] = pd.to_datetime(
+    X_employee["date_first_hired"], format="%m/%d/%Y"
+)
 X_employee.head(2)
 
 # %%
--- a/examples/model_evaluation/plot_train_test_split.py
+++ b/examples/model_evaluation/plot_train_test_split.py
@@ -217,7 +217,7 @@ from skrub.datasets import fetch_employee_salaries
 
 dataset = fetch_employee_salaries()
 X, y = dataset.X, dataset.y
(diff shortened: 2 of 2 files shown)


sylvaincom (approved):
Hi @ankith26, thanks for your PR.

I believe that the following works since the dates (e.g., "09/22/1986") are in a standard MM/DD/YYYY format that `pd.to_datetime()` can automatically parse:
```python
X_employee["date_first_hired"] = pd.to_datetime(X_employee["date_first_hired"])
```

However, for better performance and clarity, we can specify the format as you suggested:
```python
X_employee["date_first_hired"] = pd.to_datetime(X_employee["date_first_hired"], format="%m/%d/%Y")
```

For these 2 examples, [`skrub.ToDatetime`](https://skrub-data.org/stable/reference/generated/skrub.ToDatetime.html) does not feel mandatory (there is no scikit-learn pipeline).

sylvaincom (dismissed):
Please update your branch

ankith26:
Thanks for the approval. I updated the branch with a rebase.

sylvaincom (approved):
LGTM, thanks!

PR #1552, opened 2025-04-15
feat: Add confusion matrix to EstimatorReport

### LINKED ISSUE #1522: Add confusion matrix to Estimator Report

https://blog.ippon.fr/2025/03/28/a-la-decouverte-de-skore-le-petit-nouveau-de-la-famille-scikit-learn/

> En cas de classification à plus de 2 classes, les métriques, les graphiques et la synthèse montrent les résultats pour chaque classe, ce qui est bien pratique ! Ma seule déception est de ne pas avoir accès à la matrice de confusion. Mais peut-être sera-t-elle ajoutée dans une prochaine version ?

in 🇬‍🇧 : 

> When classifying more than two classes, the metrics, graphs, and summary show the results for each class, which is very useful! My only disappointment is not having access to the confusion matrix. But perhaps it will be added in a future version?

Add confusion matrix in estimator report.  

API suggestion: `estimator_report.metrics.confusion_matrix`.  
it could be a display, meaning that the default output would be a plot, and doing `estimator_report.metrics.confusion_matrix.table()`, I would receive a dataframe.

_No response_

### Additional context

_No response_

waridrox:
Closes #1522.

As originally suggested by @MarieSacksick, to add the confusion matrix in estimator report with the API suggestion:
`estimator_report.metrics.confusion_matrix`

When implementing however, the default plotting part, I noticed the implementation in `matplotlib`, the user might tend to use the `.plot()` function naturally, which would mean to not display the plot in `skore` by default but rather to show the plot only if the user uses the above syntax with the `.plot()` method.

For showing dataframe, one can resolve to `.frame()` as originally proposed.

Also added basic test for the implementation.

Let me know your thoughts regarding the same, open to suggestions and brainstorming :)

## Implementation
Core class: `ConfusionMatrixDisplay`
Initialization: Accepts a precomputed confusion matrix and parameters (e.g., normalize, display_labels).

Methods:
`.plot()`: Renders the confusion matrix as a Matplotlib plot.
`.frame()`: Returns the confusion matrix as a pandas.DataFrame.

Accessor Method: `report.metrics.confusion_matrix()`

_____

Parameters:

- data_source: "test" (default), "train", or "X_y".
- normalize: "true", "pred", "all", or None.
- display_labels: Custom class labels.
- values_format: Formatting for matrix values.
- Returns: ConfusionMatrixDisplay object (requires explicit .plot() to render).

_____

No Auto-Plotting: Users must call .plot() to generate the visualization (matches scikit-learn’s pattern). Ref - https://scikit-learn.org/stable/visualizations.html
Data Source Handling: Uses data_source to fetch predictions from cached train/test data or custom X/y.
Normalization: Validates normalize parameter ("true", "pred", "all").
_____

Testing validation scenarios:

- Binary and multiclass classification
- Normalization modes
- Custom labels
- Iinvalid labels
- Data source switching (test/train/X_y)


skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,250 @@
...
+        n_classes = cm.shape[0]
+
+        if self.normalize:
+            with np.errstate(all="ignore"):
+                if self.normalize == "true":
+                    cm = cm / cm.sum(axis=1, keepdims=True)
+                elif self.normalize == "pred":
+                    cm = cm / cm.sum(axis=0, keepdims=True)
+                else:
+                    cm = cm / cm.sum()
```
auguste-probabl, line 120:
```suggestion
        with np.errstate(all="ignore"):
            if self.normalize == "true":
                cm = cm / cm.sum(axis=1, keepdims=True)
            elif self.normalize == "pred":
                cm = cm / cm.sum(axis=0, keepdims=True)
            elif self.normalize == "all":
                cm = cm / cm.sum()
            else:  # None
                pass
```

auguste-probabl:
Thanks for this!

waridrox:
Confirmed the changes requested, the tests pass :)

examples/model_evaluation/plot_confusion_matrix.py
```diff
@@ -0,0 +1,101 @@
...
+# Using different display options
+# ==============================
+#
+# We can customize the confusion matrix display with different options:
+
+# %%
+# Using custom display labels
+report.metrics.confusion_matrix(
+    display_labels=["Setosa", "Versicolor", "Virginica"], cmap="Blues"
+)
```
auguste-probabl, line 89: The docs are failing to build because of this line; see [stack trace](https://github.com/probabl-ai/skore/actions/runs/14499598970/job/40708229706?pr=1552)

waridrox: Oops, I noticed I used the cmap, I'll add it as an optional parameter...

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,251 @@
...
+        Axes with confusion matrix.
+
+    text_ : ndarray of shape (n_classes, n_classes), dtype=matplotlib Text or \
+            None
+        Array of matplotlib text elements containing the values in the
+        confusion matrix.
+
+    """
+
+    def __init__(
```
auguste-probabl, line 57:
```suggestion
    @StyleDisplayMixin.style_plot
    def __init__(
```
This makes the style of all skore plots consistent.

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,251 @@
...
+        self,
+        confusion_matrix,
+        *,
+        display_labels=None,
+        include_values=True,
+        cmap="viridis",
```
auguste-probabl, line 63: Generally we try to follow the rule that plot-specific settings should be arguments of the `plot` method, rather than attributes of the whole class. So `cmap` and `colorbar` should be moved.

waridrox:
Checked for the docs build locally, changed the implementation slightly to move `cmap` and `colorbar` to plot method as suggested.

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,242 @@
...
+        Figure containing the confusion matrix.
+
+    ax_ : matplotlib Axes
+        Axes with confusion matrix.
+
+    text_ : ndarray of shape (n_classes, n_classes), dtype=matplotlib Text or \
+            None
+        Array of matplotlib text elements containing the values in the
+        confusion matrix.
+
```
auguste-probabl, line 48:
```suggestion
```

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,242 @@
...
+    @StyleDisplayMixin.style_plot
+    def __init__(
+        self,
+        confusion_matrix,
+        *,
+        display_labels=None,
+        include_values=True,
+        normalize=None,
+        values_format=None,
+        ax=None,
```
auguste-probabl, line 60:
```suggestion
```
It's already an argument of `plot`

waridrox: Nice catch! Actually it was implemented in the `constructor` method, so I moved it to `plot` method.

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,244 @@
...
+            values_format=values_format,
+        )
+
+        # If ax param is provided, plot with the provided axes
+        if ax is not None:
+            disp.plot(ax=ax)
+
+        return disp
+
+    def table(self):
```
auguste-probabl, line 228: Discussed with @glemaitre : that method should be called `frame`, to make it clear that the output is a DataFrame (as opposed to e.g. an array).

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,244 @@
...
+        conditions or all the population. If None, confusion matrix will not be
+        normalized.
+
+    values_format : str, default=None
+        Format specification for values in confusion matrix. If None, the format
+        specification is 'd' or '.2g' whichever is shorter.
+
+    ax : matplotlib axes, default=None
+        Axes object to plot on. If None, a new figure and axes is created.
+
```
auguste-probabl, line 35:
```suggestion
```

sylvaincom:
Hi @waridrox, thanks a lot for your time. Could you please comment the parent issue so that I can assign you to it?

https://github.com/probabl-ai/skore/issues/1522

waridrox:
> Hi @waridrox, thanks a lot for your time. Could you please comment the parent issue so that I can assign you to it?
> 
> #1522

Sure, thanks @sylvaincom 🚀

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
waridrox: Made the changes over a new commit, waiting for a review

skore/src/skore/sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -0,0 +1,241 @@
...
+            display_labels=display_labels,
+            include_values=include_values,
+            normalize=normalize,
+            values_format=values_format,
+        )
+
+        # If ax param is provided, plot with the provided axes
+        if ax is not None:
+            disp.plot(ax=ax)
+
```
auguste-probabl, line 222:
```suggestion
```
I don't think this function should take `ax`, or run `plot`; `from_predictions` should just be a helper to create a `ConfusionMatrixDisplay`.

waridrox: Hmm, makes sense 👍🏻

auguste-probabl:
@glemaitre Can we have your opinion on the docs + API?
Otherwise this looks good to me.

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
```diff
@@ -0,0 +1,212 @@
...
+    report = EstimatorReport(
+        estimator,
+        X_train=X_train,
+        y_train=y_train,
+        X_test=X_test,
+        y_test=y_test,
+    )
+
+    with pytest.raises(ValueError):
+        display = report.metrics.confusion_matrix(display_labels=["Only One Label"])
```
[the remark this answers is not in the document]
waridrox: Umm, isn't this kinda confusing tho? I mean, if `confusion_matrix` raises a `ValueError` (which is what we expect), the test would fail before reaching the `pytest.raises` context. So the `pytest.raises` context would then have no code inside it to execute...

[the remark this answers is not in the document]
auguste-probabl: Right, then remove the `display.plot()` line: we only want to verify that `confusion_matrix` fails

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
```diff
@@ -207,6 +207,6 @@ def test_confusion_matrix_display_class_error(
         y_test=y_test,
     )
 
-    with pytest.raises(ValueError):
-        display = report.metrics.confusion_matrix(display_labels=["Only One Label"])
-        display.plot()
+    display = report.metrics.confusion_matrix(display_labels=["Only One Label"])
```
auguste-probabl, line 210: Isn't this supposed to raise a ValueError?

examples/model_evaluation/plot_confusion_matrix.py
```diff
@@ -0,0 +1,102 @@
+"""
+.. _example_confusion_matrix:
+
+===============================================================
+Using Confusion Matrix in EstimatorReport
+===============================================================
```
sylvaincom, line 6:
```suggestion
=====================================
Using the confusion matrix of the EstimatorReport
=====================================
```

sylvaincom (commented):
Thanks for this PR.

I wouldn't have a specific example just for this feature (otherwise we end up with too many examples). Instead, try modifying an existing example where the confusion matrix would make sense.

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
waridrox: No because this is just initialisation of the object. This is a design choice in the library.

```python
@StyleDisplayMixin.style_plot
def __init__(
    self,
    confusion_matrix,
    *,
    display_labels=None,
    include_values=True,
    normalize=None,
    values_format=None,
):
    self.confusion_matrix = confusion_matrix
    self.display_labels = display_labels # <--- stores without any validation
 ```

It allows for creation of the object with whatever settings we want, and errors will only be raised when we call the plot function. That's why, I was checking for the `.plot()` function earlier.

In the `plot` method, we check for the existence for `display_labels` and only proceed if they exist:
```python
if self.display_labels is None:
    self.display_labels = np.arange(n_classes)
elif len(self.display_labels) != n_classes:
    raise ValueError(
        f"display_labels must have length equal to number of classes "
        f"({n_classes}), got {len(self.display_labels)}"
    )
```

waridrox:
> I wouldn't have a specific example just for this feature (otherwise we end up with too many examples). Instead, please try modifying an existing example to include the confusion matrix if it makes sense.

Ended up adding the usage in `estimator report`, over a bunch of examples, lmk if you're fine with that...

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
auguste-probabl: Alright thanks. Didn't you originally want to check that `display.plot()` raises an error? Did you remove the check intentionally?

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,45 @@ def operational_decision_cost(y_true, y_pred, amount):
 # %%
 # As expected, since we need to recompute the predictions, it takes more time.
 
+# %%
+# Visualizing confusion matrix
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
sylvaincom, line 387:
```suggestion
# Visualizing the confusion matrix
# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
waridrox: Well, you suggested to get rid of the plot method, so...I can add that back alongside the assertion with the `display` object.

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,45 @@ def operational_decision_cost(y_true, y_pred, amount):
...
+#
+# Another useful visualization for classification tasks is the confusion matrix,
+# which shows the counts of correct and incorrect predictions for each class.
+# Let's see how one can use the confusion matrix display functionality:
+
+# %%
+# Basic confusion matrix
+cm_display = report.metrics.confusion_matrix()
+cm_display.plot()
```
sylvaincom, line 396:
```suggestion
cm_display.plot()
plt.show()
```

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,45 @@ def operational_decision_cost(y_true, y_pred, amount):
...
+
+# %%
+# We can normalize the confusion matrix to get percentages instead of raw counts.
+# Here we normalize by true labels (rows):
+cm_display = report.metrics.confusion_matrix(normalize="true")
+cm_display.plot(cmap="Blues")
```
sylvaincom, line 402:
```suggestion
cm_display.plot(cmap="Blues")
plt.show()
```

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,45 @@ def operational_decision_cost(y_true, y_pred, amount):
 # %%
 # As expected, since we need to recompute the predictions, it takes more time.
 
+# %%
+# Visualizing confusion matrix
+# ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+#
+# Another useful visualization for classification tasks is the confusion matrix,
+# which shows the counts of correct and incorrect predictions for each class.
+# Let's see how one can use the confusion matrix display functionality:
```
sylvaincom, line 391:
```suggestion
```

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,45 @@ def operational_decision_cost(y_true, y_pred, amount):
...
+
+# %%
+# Basic confusion matrix
```
sylvaincom, line 394:
```suggestion
# Let's first start with a basic confusion matrix:
```

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,45 @@ def operational_decision_cost(y_true, y_pred, amount):
...
+# %%
+# We can also normalize by predicted labels (columns):
+cm_display = report.metrics.confusion_matrix(normalize="pred")
+cm_display.plot(cmap="Greens")
+
+# %%
+# Or we can normalize by the total number of samples:
+cm_display = report.metrics.confusion_matrix(normalize="all")
+cm_display.plot(cmap="Reds")
+
```
sylvaincom, line 413:
```suggestion
# %%
# More plotting options are available, check out the API on the confusion matrix for more information.
```

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,45 @@ def operational_decision_cost(y_true, y_pred, amount):
...
+# %%
+# We can customize the display labels:
+cm_display = report.metrics.confusion_matrix(display_labels=["Disallowed", "Allowed"])
+cm_display.plot()
+
+# %%
+# The confusion matrix can also be exported as a pandas DataFrame for further analysis:
```
sylvaincom, line 420:
```suggestion
# Finally, the confusion matrix can also be exported as a pandas DataFrame for further analysis:
```

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
auguste-probabl: Okay, this is all a big misunderstanding. What I meant by my comment was:
When you write `pytest.raises(ValueError)`, there should only be *one* statement inside the context: the one that you expect to fail.

waridrox: Okay, so that would be the `plot()` method, but in order to invoke the `plot` method, we need the `display` object which is of type `ConfusionMatrixDisplay`

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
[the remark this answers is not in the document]
auguste-probabl: Here, there were two statements inside the `pytest.raises`:
- `display = ...` and
- `display.plot()`

That makes it unclear which of the two statements is supposed to raise an error.
- If the first one raises an error, then the execution will stop, so the second statement will never be run, so it should not be there at all.
- If the second one raises an error, then the first one should not be inside the context, but right before.

auguste-probabl:
@waridrox Can you un-draft your PRs if they are ready for review?

skore/tests/unit/sklearn/plot/test_confusion_matrix.py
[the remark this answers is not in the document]
waridrox: Well, in that case, the `display = ...` statement should exist above the error exception block. I'll add back the `display.plot()` method in the `raises` context.

auguste-probabl:
Also please sign your commits

waridrox:
my bad, pushed changes of another branch here 😅🙏🏻 reverted

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,38 @@ def operational_decision_cost(y_true, y_pred, amount):
...
+# %%
+# We can normalize the confusion matrix to get percentages instead of raw counts.
+# Here we normalize by true labels (rows):
+cm_display = report.metrics.confusion_matrix(normalize="true")
+cm_display.plot(cmap="Blues")
+plt.show()
+
+# %%
+# More plotting options are available, check out the API on the confusion matrix for more information.
+# %%
```
sylvaincom, line 407:
```suggestion
```

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -382,6 +382,38 @@ def operational_decision_cost(y_true, y_pred, amount):
...
+# We can customize the display labels:
+cm_display = report.metrics.confusion_matrix(display_labels=["Disallowed", "Allowed"])
+cm_display.plot()
```
sylvaincom, line 410:
```suggestion
cm_display.plot()
plt.show()
```

sylvaincom:
@waridrox why do you force push?

waridrox:
My bad @sylvaincom, I amended the changes in the previous hash since they were trivial. So had to rewrite upstream history in this branch 😅

auguste-probabl (approved):
Thanks for this!

PR #1564, opened 2025-04-18
feat: Enable data_source=X_y in CrossValidationReport

### LINKED ISSUE #1500: Allow `data_source="X_y"` for `CrossValidationReport`

For the moment, we don't support the case `data_source="X_y"` for `CrossValidationReport`. Actually, I think that there potential use cases to make predict all the estimators from the cross-validation on the same set to know the variance or misclassification samples.

So I think that we could go ahead and implement the feature and at least support the metric for the moment. When it comes to plotting, we need to have specific representation for this case most probably.

glemaitre:
closes #1500 

Allow to pass `data_source="X_y"` to make underlying estimator predict on the same set with `CrossValidationReport`.


auguste-probabl (approved):
OK, works for me

PR #1566, opened 2025-04-19
fix: Fixes search and parallel coordinate widget

glemaitre:
This should fix a couple of bugs observe in https://github.com/probabl-ai/skore/pull/1475

The demo I tried is:

```python
from skore.project import open

if "project" not in globals():
    project = open("remote", name="robin", tenant="Probabl")

metadata = project.metadata()

# %%
metadata

# %%
# Select a subset of reports on the parallel coordinate plots
reports = metadata.reports()

# %%
from skore import ComparisonReport

report = ComparisonReport(reports)

# %%
report.metrics.report_metrics()
```

### DIFF
--- a/skore-remote-project/src/skore_remote_project/project/metadata.py
+++ b/skore-remote-project/src/skore_remote_project/project/metadata.py
@@ -1,12 +1,12 @@
 from __future__ import annotations
 
 import typing
+
 import pandas as pd
 
 from .. import item as item_module
 from ..client.client import AuthenticatedClient
 from .widget import ModelExplorerWidget
-
 
 if typing.TYPE_CHECKING:
     from typing import Union
@@ -109,8 +109,7 @@ class Metadata(pd.DataFrame):
 
     def _repr_html_(self):
         """Display the interactive plot and controls."""
-        if not hasattr(self, "_plot_widget"):
-            self._plot_widget = ModelExplorerWidget(dataframe=self)
+        self._plot_widget = ModelExplorerWidget(dataframe=self)
         self._plot_widget.display()
         return ""
 
@@ -136,11 +135,12 @@ class Metadata(pd.DataFrame):
         if not hasattr(self, "_plot_widget"):
             return None
 
+        self._plot_widget.update_selection()
         selection = self._plot_widget.current_selection.copy()
         query_parts = []
 
         task = selection.pop("ml_task")
-        query_parts.append(f"ml_task == '{task}'")
+        query_parts.append(f"ml_task.str.contains('{task}')")
 
         dataset = selection.pop("dataset")
         query_parts.append(f"dataset == '{dataset}'")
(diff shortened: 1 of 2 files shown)


skore-remote-project/src/skore_remote_project/project/metadata.py
```diff
@@ -109,8 +109,7 @@ def dto(response):
 
     def _repr_html_(self):
         """Display the interactive plot and controls."""
-        if not hasattr(self, "_plot_widget"):
```
glemaitre, line 112: I think it is better to always recreate the update the widget with the latest data.

skore-remote-project/src/skore_remote_project/project/metadata.py
```diff
@@ -136,11 +135,12 @@ def query_string_selection(self) -> Union[str, None]:
         if not hasattr(self, "_plot_widget"):
             return None
 
+        self._plot_widget.update_selection()
```
glemaitre, line 138: The selection was not update before to catch it.

skore-remote-project/src/skore_remote_project/project/metadata.py
```diff
@@ -136,11 +135,12 @@ def query_string_selection(self) -> Union[str, None]:
...
         selection = self._plot_widget.current_selection.copy()
         query_parts = []
 
         task = selection.pop("ml_task")
-        query_parts.append(f"ml_task == '{task}'")
+        query_parts.append(f"ml_task.str.contains('{task}')")
```
glemaitre, line 143: It is about the inconsistence between the ML task `binary-classification` and the `classification` string. Right now, we don't try to go fine grain and I'm not sure that we should go this way.

We have to loop with the Hub if it is the same or not.

skore-remote-project/src/skore_remote_project/project/widget.py
```diff
@@ -380,6 +382,9 @@ def _add_jitter_to_categorical(
         np.ndarray
             Array of encoded categorical values with jitter applied.
         """
+        if categorical_series.cat.categories.size == 1:
+            return categorical_series.cat.codes.to_numpy()
```
glemaitre, line 386: exit early when we have a single category, otherwise the results is really weird :)

skore-remote-project/src/skore_remote_project/project/metadata.py
thomass-dev: I'm not sure: it's quite strange to do a first selection, that is reset on a new repr.

skore-remote-project/src/skore_remote_project/project/metadata.py
thomass-dev: Oh i miss to update this part too during the merge of the two parts... well spotted.

thomass-dev (approved):

PR #1570, opened 2025-04-20
fix: Accept X and y as positional argument with as_dict=True in train_test_split

### LINKED ISSUE #1544: Passing all datasets by keyword makes it annoying to use `as_dict=True`
See https://github.com/probabl-ai/skore/issues/1453 for context.

> Following #1460 I've hit https://github.com/probabl-ai/skore/pull/1460#issuecomment-2750426087: you have to pass arguments by keyword rather than by position.
> I find this too restrictive, which will make people avoid `as_dict` altogether. I think we can reasonably say that in 99% of uses of `train_test_split` the first positional argument will be `X` and the second will be `y`.
> 
> So I'd advocate for something like:
> ```python
>>> skore.train_test_split(X, as_dict=True)
> {"X_train": ..., "X_test": ...}
> 
>>> skore.train_test_split(X, y, as_dict=True)
> {"X_train": ..., "X_test": ..., "y_train": ..., "y_test": ...}
> 
>>> skore.train_test_split(X, y, z=z, as_dict=True)
> {"X_train": ..., "X_test": ..., "y_train": ..., "y_test": ..., "z_train": ..., "z_test": ...}
> ```

amltarek:
closes #1544 

I fixed the issue by updating the handling of keyword arguments when as_dict=True is used. Now, if all datasets are passed as keywords, the function directly returns them as a dictionary without extra processing. This makes the behavior more intuitive and avoids redundancy. I also tested the fix through test_train_split_test.py to ensure it works correctly.
Screenshot from 2025-04-20 22-46-40


skore/src/skore/sklearn/train_test_split/train_test_split.py
auguste-probabl, line 1: Can you explain what led you to delete the function's documentation?

auguste-probabl (changes requested):
Can you add some tests? It would also help to demonstrate what the new behaviour looks like.

skore/src/skore/sklearn/train_test_split/train_test_split.py
amltarek: I deleted it to make the code clearer to me and forgot to put it back,but I returned it in my latest commit.

amltarek:
@auguste-probabl 
I added the tests and restored the function's documentation. Could you check it and let me know if there are any updates I can make?

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -167,21 +177,20 @@ class labels.
...
         y_labels = np.unique(y)
-        y_test = (
-            output[3] if keyword_arrays else output[-1]
-        )  # when more kwargs are given
+        y_test = output[3] if as_dict else output[-1]
     else:
         y_labels = None
         y_test = None
 
+    # Determine the ML task based on y
```
auguste-probabl, line 193:
```suggestion
```
`_find_ml_task(y)` is sufficiently clear IMO

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -221,3 +222,77 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+def test_train_test_split_as_dict_raises_if_positional_args():
+    """Raise error if as_dict=True and inputs are not keyword arguments."""
+    X = np.arange(10).reshape(10, 1)
+    y = np.arange(10)
+
+    with pytest.raises(
+        ValueError,
+        match="When as_dict=True, arrays must be passed as keyword arguments",
+    ):
+        train_test_split(X, y, test_size=0.2, as_dict=True)
```
auguste-probabl, line 264: This is precisely what I don't want :) #1544 says: "if as_dict=True, and the user passes arrays by position, interpret the arrays as being X and y, don't raise an error"

amltarek: @auguste-probabl I thought about it as @glemaitre said we should implement this case to ensure it doesn't produce an error and that we receive the expected dictionary. Should I edit it or keep it as it is?

auguste-probabl: To reiterate, the whole point of #1544 is: I want to be able to pass arrays by position, even with `as_dict=True`; when I do, I would expect skore to understand that the first array is X and the second is y. I don't want the ValueError. Therefore this test must be changed.

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -137,27 +137,37 @@ class labels.
...
+
     if X is not None:
         new_arrays.append(X)
-        keys += ["X"]
+        keys.append("X")
     if y is not None:
         new_arrays.append(y)
-        keys += ["y"]
-
-    if as_dict and arrays:
```
nkapila6, line 147: Why do you remove this? If both are true, it will cause a conflict.

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -137,27 +137,37 @@ class labels.
...
-                keyword_arrays.values()
-            )  # if X and y is not passed but other variables
+            if not keyword_arrays:
+                raise ValueError(
+                    "When as_dict=True, arrays must be passed as keyword arguments"
+                )
+
+            new_arrays = list(keyword_arrays.values())
+
+        if X is not None:
```
nkapila6, line 157: Why is this repeated inside and outside the if? It is redundant.

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -137,27 +137,37 @@ class labels.
...
+            new_arrays.append(X)
+            keys.append("X")
+        if y is not None:
```
nkapila6, line 160: Why is this repeated inside and outside the if? It is redundant.

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -167,21 +177,20 @@ class labels.
         stratify=stratify,
     )
 
-    if X is None:
-        X = arrays[0] if len(arrays) == 1 else arrays[-2]
+    if X is None and len(arrays) >= 1:
```
nkapila6, line 180: What happens in the case `len(arrays)==1`?

nkapila6 (changes requested):

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -221,3 +222,91 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+    assert set(result.keys()) == {"X_train", "X_test", "y_train", "y_test"}
+    assert len(result["X_train"]) == 7
+    assert len(result["X_test"]) == 3
+
+
+def test_train_test_split_only_X():
+    X = [[1]] * 20
+    result = train_test_split(X=X, random_state=0, as_dict=True)
+    assert "X_train" in result
+    assert "X_test" in result
```
auguste-probabl, line 293:
```suggestion
```
Duplicate of `test_train_test_split_check_dict_unsupervised_case`

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -183,17 +184,17 @@ def test_train_test_split_kwargs():
 
 
 def test_train_test_split_dict_kwargs():
-    """Passing data without keyword arguments with return_dict=True
-    should raise ValueError."""
+    """Passing data with positional arguments and as_dict=True should work."""
```
auguste-probabl, line 187:
```suggestion
    """When passing data with positional arguments and as_dict=True, the first argument will
    be interpreted as `X` and the second one as `y`."""
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
auguste-probabl, line 1: Can you add a test with
```python
arr1 = [[1]] * 20
arr2 = [0] * 10 + [1] * 10
train_test_split(arr2, X=arr1, as_dict=True)
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -221,3 +222,91 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+        test_size=0.5,
+        as_dict=True,
+        random_state=42,
+    )
+
+    expected_keys = {"X_train", "X_test", "y_train", "y_test", "z_train", "z_test"}
+
+    assert all(key in result for key in expected_keys)
+    assert result["z_train"].shape[0] == 5
+    assert result["z_test"].shape[0] == 5
```
auguste-probabl, line 274: This feels redundant with the previous test

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -221,3 +222,91 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+def test_train_test_split_as_dict_with_mixed_input_types():
+    """Ensure train_test_split handles a mix of array-like types with as_dict=True."""
+    X = [[i] for i in range(10)]
+    y = np.arange(10)
+
+    result = train_test_split(X=X, y=y, test_size=0.3, as_dict=True, random_state=1)
+
+    assert set(result.keys()) == {"X_train", "X_test", "y_train", "y_test"}
+    assert len(result["X_train"]) == 7
+    assert len(result["X_test"]) == 3
```
auguste-probabl, line 286: Seems redundant with the previous test

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -221,3 +222,91 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+def test_train_test_split_as_dict_with_positioned_args():
+    X = [[1]] * 20
+    y = [0] * 10 + [1] * 10
+
+    result = train_test_split(X, y, test_size=0.2, as_dict=True, random_state=0)
+
+    assert "X_train" in result
+    assert "X_test" in result
+    assert "y_train" in result
+    assert "y_test" in result
```
auguste-probabl, line 312: Duplicate of `test_train_test_split_dict_kwargs`

amltarek:
I have removed all the duplicate functions. Can you check the code? @auguste-probabl

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -221,3 +232,39 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+    weights = np.ones(10)
+
+    result = train_test_split(
+        X=X,
+        y=y,
+        sample_weights=weights,
+        test_size=0.2,
+        as_dict=True,
+        random_state=0,
+    )
```
auguste-probabl, line 251: This test feels a bit redundant, but 
```
train_test_split(
    X,
    y,
    sample_weights=weights,
    ...
)
```
(i.e. a mix of positional and keyword arguments) would be interesting.

auguste-probabl:
@amltarek Can you resolve comments when it's clear that you have addressed them? It helps review your code.

auguste-probabl:
Also please sign your commits

skore/src/skore/sklearn/train_test_split/train_test_split.py
auguste-probabl, line 1: Could you also add an example in the doctests (line 134) to showcase the behaviour?
Something like
```python
>>> train_test_split(X, y, as_dict=True)
{'X_train': ..., 'X_test': ..., 'y_train': ..., 'y_test': ...}
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
amltarek: this is tested through two functions named( test_train_test_split_check_dict()) and test_train_test_split_dict_kwargs().

auguste-probabl: It's a bit different: right now we test either all arguments passed by keyword, or all arguments passed by position. I'd like to also test the combination of both (one array passed by position, one array passed by keyword).

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -221,3 +227,11 @@ def test_train_test_split_check_dict_no_X_no_y():
...
     keys = output.keys()
     assert list(keys) == ["z_train", "z_test"]
+
+
+def test_empty_input():
+    """Tests that passing empty lists for X and y raises a ValueError."""
+    X = []
+    y = []
+    with pytest.raises(ValueError):
+        train_test_split(X, y)
```
auguste-probabl, line 237: Actually I don't think this test is needed; this behaviour is not specific to our function, but rather that sklearn's.

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -132,32 +132,42 @@ class labels.
     {'X_train': ..., 'X_test': ...,
      'y_train': ..., 'y_test': ...,
      'sample_weights_train': ..., 'sample_weights_test': ...}
+
+    >>> # When using positional arguments and as_dict=True
+    >>> # the first argument is assumed to be X, the second y
+    >>>  train_test_split(
+    ...     [[1], [2], [3], [4]], [0, 1, 0, 1], as_dict=True, random_state=0
```
auguste-probabl, line 139:
```suggestion
    ...     [[1], [2], [3], [4]], [0, 1, 0, 1], as_dict=True
```
No need for `random_state` since we don't check the output arrays. You can also remove `random_state` in the previous doctest

auguste-probabl:
Closed due to inactivity

PR #1574, opened 2025-04-22
fix: Raise an error when `CrossValidationReport` fails before at least one estimator is fitted

### LINKED ISSUE #1561: fix: If cross-validation in CrossValidationReport fails before at least one estimator is fitted, raise an error

Right now, [errors during cross-validation don't crash the program, but are converted to warnings.](https://github.com/probabl-ai/skore/blob/main/skore/src/skore/sklearn/_cross_validation/report.py#L200)

This behaviour should not occur if the resulting CrossValidationReport contains zero estimators; if this happens, we should raise an error.

_Originally posted by @glemaitre in [#1553](https://github.com/probabl-ai/skore/issues/1553#issuecomment-2812356176)_

YoussefMAllam:
Fixed #1561
I added an if condition at the end of the "crossvaldiationreport" funcrion to check if there are 0 estimators and raise an error.
I also added a test for it in the file "testcrossvalidation.py" where i created a broken estimator to test the case where no estimators are fitted


skore/src/skore/sklearn/_cross_validation/report.py
```diff
@@ -222,6 +222,11 @@ def _fit_estimator_reports(self) -> list[EstimatorReport]:
                     border_style="cyan",
                 )
             )
+            if len(estimator_reports) == 0:
+                raise RuntimeError(
+                    "Cross-validation failed: no estimators were successfully fitted. "
+                    "Please check your data, estimator, or cross-validation setup."
+                ) from None
```
thomass-dev, line 229: I think this should be at the top of the `except` block, and the exception should be raised from the caught exception to keep the information from the original traceback. I don't think in this particular case that we need to `console.print` a custom message before raising the exception.

thomass-dev (changes requested):

skore/src/skore/sklearn/_cross_validation/report.py
glemaitre: I think that we should raise form `e` here. We can then indicate in the error message the user to check the error from which we raised because it could be more informative that the just requesting to check the data, estimator, or cross-validation.

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -988,3 +988,21 @@ def test_cross_validation_timings(
         "Predict time test",
     ]
     assert timings.columns.tolist() == expected_columns
+
+
+class BrokenEstimator(BaseEstimator, ClassifierMixin):
```
glemaitre, line 993: Mixin goes on the left for the inheritance to work properly.

```suggestion
class BrokenEstimator(ClassifierMixin, BaseEstimator):
```

skore/tests/unit/sklearn/test_cross_validation.py
```diff
@@ -988,3 +988,21 @@ def test_cross_validation_timings(
...
+
+
+def test_cross_validation_report_with_zero_estimators():
+    X, y = make_classification(n_samples=100, n_features=10, random_state=42)
+
+    with pytest.raises(
+        RuntimeError,
+        match="Cross-validation failed: no estimators were successfully fitted",
+    ):
+        CrossValidationReport(BrokenEstimator(), X, y)
```
glemaitre, line 1008: Let's make it more compact.

```suggestion
    err_msg = "Cross-validation failed: no estimators were successfully fitted"
    with pytest.raises(RuntimeError, match=err_msg):
        CrossValidationReport(BrokenEstimator(), X, y)
```

skore/tests/unit/sklearn/test_cross_validation.py
YoussefMAllam: Apologies, my bad.
Resolved

skore/src/skore/sklearn/_cross_validation/report.py
```diff
@@ -213,6 +213,11 @@ def _fit_estimator_reports(self) -> list[EstimatorReport]:
                     "might not contain all the expected results. "
                     f"Traceback: \n{e}"
                 )
+            if len(estimator_reports) == 0:
+                raise RuntimeError(
+                    "Cross-validation failed: no estimators were successfully fitted. "
+                    "Please check your data, estimator, or cross-validation setup."
+                ) from e
```
thomass-dev, line 220: Nitpick: can you move this block before the construction of the message (between 201 and 203)? It should be more readable.

thomass-dev (changes requested):
Just a nitpick, otherwise LGTM. Thanks.

Edit: can you please fix the merge-conflict?

glemaitre (dismissed):

skore/src/skore/sklearn/_cross_validation/report.py
```diff
@@ -198,31 +201,55 @@ def _fit_estimator_reports(self) -> list[EstimatorReport]:
...
+                "all estimators could be fitted; CrossValidationReport object "
+                "might not contain all the expected results.\n"
+                f"Traceback: \n{msg_traceback}"
+            )
+            estimator_reports = [
+                report
+                for report in estimator_reports
+                if not isinstance(report, Exception)
+            ]
+        elif any(isinstance(report, KeyboardInterrupt) for report in estimator_reports):
```
thomass-dev, line 234: Nitpick: 

What about if two exceptions `KeyboardInterrupt` and `Exception` in the same run?
Instead of an exclusion `elif`, don't shouldn't we do two consecutive `if`?

skore/src/skore/sklearn/_cross_validation/report.py
```diff
@@ -198,31 +201,55 @@ def _fit_estimator_reports(self) -> list[EstimatorReport]:
...
-            from skore import console  # avoid circular import
+        for report in generator:
+            estimator_reports.append(report)
+            progress.update(task, advance=1, refresh=True)
+
+        warn_msg = None
+        if all(
+            isinstance(report, (KeyboardInterrupt, Exception))
+            for report in estimator_reports
+        ):
```
thomass-dev, line 212: Nitpick, in case in the future we change the behaviour of `_generate_estimator_report`: 

```suggestion
        if not any (
            isinstance(report, EstimatorReport)
            for report in estimator_reports
        ):
```

skore/src/skore/sklearn/_cross_validation/report.py
glemaitre: Why not, let's give the priority to the `KeyboardInterrupt`.

skore/src/skore/sklearn/_cross_validation/report.py
```diff
@@ -198,31 +201,55 @@ def _fit_estimator_reports(self) -> list[EstimatorReport]:
...
+                "all estimators could be fitted; CrossValidationReport object "
+                "might not contain all the expected results.\n"
+                f"Traceback: \n{msg_traceback}"
+            )
+            estimator_reports = [
+                report
+                for report in estimator_reports
+                if not isinstance(report, Exception)
+            ]
+        elif any(isinstance(report, KeyboardInterrupt) for report in estimator_reports):
```
glemaitre, line 234:
```suggestion
        if any(isinstance(report, KeyboardInterrupt) for report in estimator_reports):
```

thomass-dev (approved):
Thanks @YoussefMAllam for your initiative and @glemaitre for the follow-up.
Nice work!

PR #1606, opened 2025-04-28
ci: Factorize lint jobs

thomass-dev:
**Context:**

Linting must be run on `/skore/`, `/skore-remote-project/` and `/examples/`.

```yml
      - name: Lint
        working-directory: skore-remote-project/
        run: |
          pre-commit run --all-files ruff
          pre-commit run --all-files mypy
```
```yml
      - name: Lint
        working-directory: skore/
        run: |
          pre-commit run --all-files ruff
          pre-commit run --all-files mypy
```

These snippets will not only run on `/skore/` or `/skore-remote-project`, but both on `/`.
So the `ruff` and `mypy` linting will be executed twice in integrality.
Wherever `pre-commit run --all-files` is executed, it will be execute from the root of the repository.

---

**Solution:**

One solution is to fix the command to run linting only on the wanted repository.
```yml
      - name: Lint
        working-directory: skore-remote-project/
        run: |
          git ls-files | xargs pre-commit run ruff --files
          git ls-files | xargs pre-commit run mypy --files
```
```yml
      - name: Lint
        working-directory: skore/
        run: |
          git ls-files | xargs pre-commit run ruff --files
          git ls-files | xargs pre-commit run mypy --files
```
But it implies to add an extra workflow to run `ruff` and `mypy` on `examples/`.
IMO it's not a good idea, so decided to simplify the whole linting process.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -46,3 +46,5 @@ jobs:
           pre-commit run --all-files trailing-whitespace
           pre-commit run --all-files typos
           pre-commit run --all-files nbstripout
+          pre-commit run --all-files ruff
+          pre-commit run --all-files mypy
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -24,10 +24,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/lint.yml
auguste-probabl, line 1: Can you take the opportunity to pin the actions versions?

thomass-dev: Good point, thanks.

auguste-probabl (approved):

PR #1607, opened 2025-04-28
feat: Add `skore-remote-project`

thomass-dev:
Add a new package used to exchange with `skore hub`.


rouk1 (approved):
Looks good to me ! 
Maybe a short explanation on how it will work in the future could be nice.

PR #1617, opened 2025-04-30
docs: Avoid failure when pytorch load weights from TextEncoder with parallelism

glemaitre:
It is a workaround on the failure that is sometimes happening when request `transform` from `TextEncoder` in parallel.

In short, there is a failure in the weights loading from PyTorch where some tensors are loaded on a meta device (only metadata) and does not load the weight. I did not yet found the root cause of the failure why the weights are not loaded properly.

In the meanwhile, we can avoid this issue by storing the weights. We should monitor that loading several time the language model does not make blow up the RAM on the GitHub actions. Othwerwise, we need to deactivate the parallelism most probably.

### DIFF
--- a/examples/use_cases/plot_employee_salaries.py
+++ b/examples/use_cases/plot_employee_salaries.py
@@ -112,7 +112,7 @@ from sklearn.ensemble import HistGradientBoostingRegressor
 from sklearn.pipeline import make_pipeline
 
 model = make_pipeline(
-    TableVectorizer(high_cardinality=TextEncoder()),
+    TableVectorizer(high_cardinality=TextEncoder(store_weights_in_pickle=True)),
     HistGradientBoostingRegressor(),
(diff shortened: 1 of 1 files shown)


thomass-dev:
Nice!!! There is an error in the sphinx pipeline :(.

glemaitre:
Apparently the `.put` does not work properly. I'll check what happen.

thomass-dev:
I'm also looking.

thomass-dev:
Fixed in https://github.com/probabl-ai/skore/pull/1625/.

thomass-dev:
@glemaitre with this fix, the pickled report weights from 8mo to >1go.

thomass-dev (approved):

glemaitre:
It also mean that I need to debug the pytorch bug :)

PR #1633, opened 2025-05-02
fix: Use process instead of thread when caching predictions

glemaitre:
This PR forces to use process-based parallelism instead of thread-based parallelism.

Thread-based parallelism was required such that each thread could update the same cache dictionary. However, with model such as `TextEncoder`, those models could potentially load weights from the disk at start and it is not thread safe.

### DIFF
--- a/skore/src/skore/sklearn/_base.py
+++ b/skore/src/skore/sklearn/_base.py
@@ -330,8 +330,13 @@ def _get_cached_response_values(
     pos_label: Optional[PositiveLabel] = None,
     data_source: Literal["test", "train", "X_y"] = "test",
     data_source_hash: Optional[int] = None,
-) -> NDArray:
+) -> list[tuple[tuple[Any, ...], Any, bool]]:
     """Compute or load from local cache the response values.
+
+    Be aware that the predictions are not loaded from the cache, they will not be added
+    to it. The reason is that we want to be able to run this function in parallel
+    settings in a thread-safe manner. The update should be done outside of this
+    function.
 
     Parameters
     ----------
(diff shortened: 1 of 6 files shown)


thomass-dev (approved):

skore/src/skore/sklearn/_base.py
auguste-probabl, line 1: This makes things quite a bit more complicated. I'm not super familiar with distributed systems primitives, but it seems like an [RLock](https://docs.python.org/3/library/threading.html#threading.RLock) could make this simpler?

glemaitre: Using `RLock` does not look like a good idea. I only used thread to have access to the same memory but otherwise I wanted to use process-based.

Here the only change is that `get_cached_response_methods` only to a get but does make a `set` into the cache.

So in fact, we were a bit abusing the naming. Here, we really get and we have the `set` outside. Potentially, we could create a function that write the assignment.

skore/src/skore/sklearn/_base.py
```diff
@@ -330,9 +330,14 @@ def _get_cached_response_values(
     pos_label: Optional[PositiveLabel] = None,
     data_source: Literal["test", "train", "X_y"] = "test",
     data_source_hash: Optional[int] = None,
-) -> NDArray:
+) -> list[tuple[tuple[Any, ...], Any, bool]]:
     """Compute or load from local cache the response values.
 
+    Be aware that the predictions are not loaded from the cache, they will not be added
+    to it. The reason is that we want to be able to run this function in parallel
```
auguste-probabl, line 337:
```suggestion
    Be aware that the predictions will be loaded from the cache if present, but they will not be added
    to it. The reason is that we want to be able to run this function in parallel
```

thomass-dev:
What is the status on this PR? I'm fine with the proposed solution for now.

glemaitre:
OK I solved the conflict and the linter problem. It should be good to go.
As said, we will probably revisit the issue when implementing a real cache mechanism.

auguste-probabl (approved):
Okay

thomass-dev (approved):

PR #1642, opened 2025-05-05
feat(skore-estimator-report): Add experiments accessor

thomass-dev:
Some minor improvements to prepare for the arrival of the new `search` feature:

-  Add `follow_redirects` instruction in each HTTPX call,
-  Add experiments accessor.

### DIFF
--- a/skore-remote-project/src/skore_remote_project/item/skore_estimator_report_item.py
+++ b/skore-remote-project/src/skore_remote_project/item/skore_estimator_report_item.py
@@ -128,7 +128,7 @@ class Metadata:
             verbose_name: str,
             data_source: str,
             greater_is_better: bool,
-            position: Union[int | None],
+            position: Union[int, None],
             /,
         ) -> Union[Metric, None]:
             if hasattr(self.report.metrics, name):
@@ -149,9 +149,9 @@ class Metadata:
         def timing(
             name: str,
             verbose_name: str,
-            data_source: Union[str | None],
+            data_source: Union[str, None],
             greater_is_better: bool,
-            position: Union[int | None],
+            position: Union[int, None],
             /,
         ) -> Union[Metric, None]:
             timings = self.report.metrics.timings()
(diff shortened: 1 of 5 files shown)


skore-remote-project/src/skore_remote_project/item/skore_estimator_report_item.py
auguste-probabl, line 1: aside: `position` should be `Optional[int]` I guess?

skore-remote-project/src/skore_remote_project/project/project.py
```diff
@@ -123,7 +138,7 @@ def put(
 
         with AuthenticatedClient(raises=True) as client:
             client.post(
-                f"projects/{self.tenant}/{self.name}/items/",
+                "/".join(("projects", self.tenant, self.name, "items")),
```
auguste-probabl, line 141: The previous impl was clearer

skore-remote-project/src/skore_remote_project/item/skore_estimator_report_item.py
thomass-dev: You mean, `Optional[int]` over `Union[int, None]`?

thomass-dev: If so, i prefer to use `Union[int, None]` over `Optional` when its not a parameter, especially when there is no default.

skore-remote-project/src/skore_remote_project/project/project.py
thomass-dev: It is only to avoid trailing slash, and to harmonize the construction of url everywhere.

skore-remote-project/src/skore_remote_project/item/skore_estimator_report_item.py
auguste-probabl: I wasn't precise enough: the current signature is `Union[int | None]`; it should be  `Optional[int]` or `Union[int, None]` or `int | None`.

rouk1 (approved):
Appart from a little comment LGTM !

skore-remote-project/src/skore_remote_project/item/skore_estimator_report_item.py
thomass-dev: Oh, there is a `|` ! Thanks, i will fix it.

skore-remote-project/src/skore_remote_project/project/project.py
thomass-dev: @auguste-probabl fixed in 38cb7a32d81f62fe331b2875a5e0fcee62cc2865.

augustebaum (approved):
LGTM

auguste-probabl (approved):
LGTM

PR #1662, opened 2025-05-06
refactor: Remove `_parent_progress` attribute

### LINKED ISSUE #1620: refactor: Merge `_parent_progress` into `_progress_info`

It seems that the `_parent_progress` attribute of `*Report` objects is redundant with the `_progress_info` attribute. It should be removed.

waridrox:
Closes #1620

Refactored to use a single attribute `_progress_info` for all progress bar handling, eliminating the redundant `_parent_progress` attribute.

- Removed initialization of `_parent_progress` from all report classes: EstimatorReport, CrossValidationReport, ComparisonReport.
- Modified the `progress_decorator` to check only for `_progress_info` attribute instead of checking for both attributes (`_parent_progress` and `_progress_info`).
- When sharing a progress bar from parent to child reports, we have this now:
```python
child._progress_info = {"current_progress": progress}
```
instead of this before:
```python
child._parent_progress = progress
```

### DIFF
--- a/skore/src/skore/utils/_progress_bar.py
+++ b/skore/src/skore/utils/_progress_bar.py
@@ -45,8 +45,10 @@ def progress_decorator(
 
             desc = description(self_obj) if callable(description) else description
 
-            if getattr(self_obj, "_parent_progress", None) is not None:
-                progress = self_obj._parent_progress
+            created_progress = False
+
+            if getattr(self_obj, "_progress_info", None) is not None:
+                progress = self_obj._progress_info["current_progress"]
             else:
                 progress = Progress(
                     SpinnerColumn(),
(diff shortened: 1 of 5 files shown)


skore/src/skore/utils/_progress_bar.py
```diff
@@ -87,7 +90,7 @@ def wrapper(*args: Any, **kwargs: Any) -> T:
                 has_errored = True
                 raise
             finally:
-                if self_obj._parent_progress is None:
+                if created_progress:
```
auguste-probabl, line 93: I think it should be `if not created_progress`; please don't add a new variable though in any case.

waridrox: Removed `created_progress` and instead computed `had_progress_info` boolean which gets the `_progress_info` attribute from the object, returning `None` if it doesn't exist.

waridrox: The `created_progress` flag actually tracks whether the current function call created a new progress bar or not, instead of using an existing one. When `created_progress = True`, it means that the specific function created a new progress bar.

`if not created_progress` would actually fail here because it would stop the progress bars that were passed down from parent objects, prematurely terminating them while they're still needed by their parent creators.

auguste-probabl: Thanks for the explanation; I understand better now. The variable actually improves readability!

auguste-probabl (approved):
Nice!

Issue #1515, opened 2025-04-02
enh: Add DummyRegressor or DummyClassifier performance as baseline in `report_metrics`


glemaitre:
### What would you like to say?

related to #1025 

We could add the statistical performance of a dummy model in the `report_metrics` function. We could control it with a keyword `add_baseline_model=True/False`. It helps at putting into perspective some results.

sylvaincom:
This is great, +1

sylvaincom:
Maybe the name `add_baseline_model` could be named `add_dummy_model`? Because a "baseline" could be a "strong baseline" depending on the usage of that word

alexliap:
@glemaitre @sylvaincom hello, you could assign it to me, seems simple enough.

sylvaincom:
Hi @alexliap, thanks for looking into this, let us know if you need any help.

alexliap:
@sylvaincom hello thnx for the assignment.

I see that every report has its own metrics accessor which implements a `report_metrics()`. so do you want the dummy model in every report or just the `EstimatorReport`?

sylvaincom:
Hi @alexliap.

> I see that every report has its own metrics accessor which implements a report_metrics(). so do you want the dummy model in every report or just the EstimatorReport?

this could be in every report cc @glemaitre @MarieSacksick @augustebaum if you have insights to share?

alexliap:
@MarieSacksick @sylvaincom Now that i think of it should we implement it in Comparison report? I would be kinda hacky if i did it, having seen the code. Also it doesnt seem logical to be in that kind of a report. 
What are you thoughts? (FYI I have completed it for `EstimatorReport` and `CrossValidationReport`, only tests remain)

sylvaincom:
Hi @alexliap thanks a lot for your time.
Could you open a PR with your current code please? You can mark it as a draft PR if it's not ready for review, that way you can get early feedback.
Yes, this could also be in the comparison report.

auguste-probabl:
If you have a working solution for the other reports, you can open a first PR without ComparisonReport (and a second one later)

auguste-probabl:
Fixed by #1663

MarieSacksick:
Following https://github.com/probabl-ai/skore/pull/1663#discussion_r2152943919 and https://github.com/probabl-ai/skore/pull/1663#discussion_r2153235464, issue to be discussed.

sylvaincom:
Related idea:

[...]

PR #1663, opened 2025-05-07
feat: Add DummyRegressor or DummyClassifier performance as baseline in report_metrics

alexliap:
This PR aims to tackle [Issue 1515](https://github.com/probabl-ai/skore/issues/1515).

1) `EstimatorReport`: 
- added argument `add_dummy_model` as bool with False default
- added description about the new argument
- by the time there is no dummy classifier for multiclass scenarios (at least that i know of), for now i have add a `pass` keyword but we can also insert an Error
- The strategies for the dummy estimators are hardcoded but it can change with an extra argument i guess.
- I created an extra private method called `_build_partial_report` in order to simplify/shorten the `report_metrics` execution, because it was hard to follow.
2) `CrossValidationReport`: 
- added argument `add_dummy_model` as bool with False default
- added description about the new argument

The bulk of work is in `EstimatorReport`, while `CrossValidationReport` takes advantage of those changes.
Tests are really basic, maybe they can be integrated into other tests with parametrization. In the tests i am simply testing if there is an extra column for the dummy model & if the column name is the one predefined.

Example Usage:
```python

X, y = make_classification(random_state=42, n_classes=2, n_features=10, n_informative=3, n_samples=10_000)

estimator = LogisticRegression()

report = EstimatorReport(estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test)

df = report.metrics.report_metrics(add_dummy_model=True)

print(df)
```
Result
```bash
                                          LogisticRegression  Dummy Baseline
Metric           Label / Average
Precision        0                          0.875102        0.000000
                 1                          0.870588        0.505200
Recall           0                          0.866613        0.000000
                 1                          0.878860        1.000000
ROC AUC                                     0.948748        0.500000
Brier score                                 0.089642        0.249973
Fit time (s)                                0.008062             NaN
Predict time (s)                            0.000315             NaN
```


thomass-dev:

auguste-probabl:
You should update one more time, we're having CI troubles. Apologies for the inconvenience.

alexliap:
It continues to raise errors about docstings in non test files.

auguste-probabl:
> It continues to raise errors about docstings in non test files.

That's strange, your changes should not impact those doctests. What's more, these warnings originate from `skore.train_test_split`, so they should appear when running `train_test_split` (not a few lines below the call). I'll take a closer look when I have the chance.

alexliap:
> So I ran a doctest manually:
>
> On main:
>
> ```
> >>> from sklearn.datasets import load_diabetes
> >>> from sklearn.linear_model import Ridge
> >>> from skore import train_test_split
> >>> from skore import EstimatorReport
> >>> X, y = load_diabetes(return_X_y=True)
> >>> split_data = train_test_split(X=X, y=y, random_state=0, as_dict=True)
> >>> regressor = Ridge()
> >>> report = EstimatorReport(regressor, **split_data)
> >>> report.metrics.r2()
> ╭───────────────────────────────── ShuffleTrueWarning ─────────────────────────────────╮
> │ We detected that the `shuffle` parameter is set to `True` either explicitly or from  │
> │ its default value. In case of time-ordered events (even if they are independent),    │
> │ this will result in inflated model performance evaluation because natural drift will │
> │ not be taken into account. We recommend setting the shuffle parameter to `False` in  │
> │ order to ensure the evaluation process is really representative of your production   │
> │ release process.                                                                     │
> ╰──────────────────────────────────────────────────────────────────────────────────────╯
> 0.3569596077458861
> ```
>
> on this branch:
>
> ```
> >>> from sklearn.datasets import load_diabetes
> >>> from sklearn.linear_model import Ridge
> >>> from skore import train_test_split
> >>> from skore import EstimatorReport
> >>> X, y = load_diabetes(return_X_y=True)
> >>> split_data = train_test_split(X=X, y=y, random_state=0, as_dict=True)
> >>> regressor = Ridge()
> >>> report = EstimatorReport(regressor, **split_data)
> >>> report.metrics.r2()
> ╭───────────────────────────────── ShuffleTrueWarning ─────────────────────────────────╮
> │ We detected that the `shuffle` parameter is set to `True` either explicitly or from  │
> │ its default value. In case of time-ordered events (even if they are independent),    │
> │ this will result in inflated model performance evaluation because natural drift will │
> │ not be taken into account. We recommend setting the shuffle parameter to `False` in  │
> │ order to ensure the evaluation process is really representative of your production   │
> │ release process.                                                                     │
> ╰──────────────────────────────────────────────────────────────────────────────────────╯
> (0.3569596077458861, None)
> ```
>
> So I think the "ShuffleTrueWarning" can be ignored. The issue is that r2() should return a float, not a tuple.

yes it returns a tuple because i have set it so. Basically the second value is the performance metric from the dummy model if it exists. On this specific example where add_dummy_model equals False by default, the r2() is None.

auguste-probabl:
> yes it returns a tuple because i have set it so.

Sorry if I was unclear: if this is the intended behavior, then the doctest needs to be *changed* to reflect this.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl, line 1: It seems like this commit overwrites things in `main` that should not be

alexliap: oh, must have messed something up in the conflict resolve

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -103,18 +105,15 @@ def report_metrics(
...
             the same length as the `scoring` parameter.
 
         scoring_kwargs : dict, default=None
             The keyword arguments to pass to the scoring functions.
 
-        pos_label : int, float, bool, str or None, default=_DEFAULT
-            The label to consider as the positive class when computing the metric. Use
-            this parameter to override the positive class. By default, the positive
-            class is set to the one provided when creating the report. If `None`,
-            the metric is computed considering each class as a positive class.
```
auguste-probabl: Again, this change should not be here

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -420,6 +400,91 @@ class is set to the one provided when creating the report. If `None`,
...
+            metrics_kwargs : dict
+                Keyword arguments for the respective metrics function.
+
+            pos_label : Optional[PositiveLabel], default=None
+                The positive class.
+
+        Returns
+        -------
+            tuple[NDArray, Union[pd.Index, pd.MultiIndex, list[str], None]]
+                A small chunk of the report.
```
auguste-probabl: indent is wrong

skore/tests/unit/sklearn/estimator/test_estimator.py
```diff
@@ -613,7 +614,8 @@ def test_estimator_report_metrics_binary_classification(
     report = EstimatorReport(estimator, X_test=X_test, y_test=y_test)
     assert hasattr(report.metrics, metric)
     result = getattr(report.metrics, metric)()
-    assert isinstance(result, float)
+    assert isinstance(result[0], float)
+    assert result[1] is None  # no dummy model so 2nd value is None
```
auguste-probabl: Now that I think about it, I don't think the dummy model value should be returned by individual metric functions, because it's unclear what the second value is (and it makes more sense for it to be just a float); please only make the change for `report_metrics`.

auguste-probabl (changes requested):
This still needs some work. It looks like the PR is getting mixed up with other changes that are now in `main`; consider rebasing to clean up the history. Otherwise it's hard to review.

skore/src/skore/sklearn/_estimator/metrics_accessor.py
alexliap: how? everything is linted/formatted by precommit. anyway i will check other docstrings

auguste-probabl: Unfortunately [ruff does not apply numpydoc rules yet](https://github.com/astral-sh/ruff/issues/8425), and the `numpydoc` tool is a bit too strict for our purposes, so there is not much automatic checking for docstrings yet.

skore/tests/unit/sklearn/estimator/test_estimator.py
alexliap: because of the fact that `report_metrics` and separate metric methods all use the `_compute_metric_scores` method should i add an extra argument, e.g. `include_dummy_model`, where it is going to be False in metric methods by default and in `report_metrics` it will take the value from this methods argument?

auguste-probabl: I think the individual metric methods `precision`, etc. should not be touched at all. If you really want to just compute precision with dummy model you can run
```python
report.metrics.report_metrics(['precision'], add_dummy_model=True)
```

alexliap: yes but both `precision()` and `report_metrics()`  use `_compute_metric_scores`  under the hood, so i think there should be an argument that should differentiate them, if i understand correctly. 

I can make it so that `_compute_metric_scores`  returns only the 1st value if the second is None, i guess
So that the method return either tuple[float, float] or float

auguste-probabl: That's right, I think `add_dummy_model` should be an argument of `_compute_metric_scores`. Doing this will allow you to stop using `self._dummy_model` attribute to pass that information through, which will simplify things

skore/tests/unit/sklearn/cross_validation/test_cross_validation.py
```diff
@@ -1137,3 +1137,31 @@ def test_estimator_report_precision_recall_pos_label_overwrite(
...
+
+    report = CrossValidationReport(estimator, X=X, y=y)
+
+    accessor = report.metrics
+
+    metrics_df = accessor.report_metrics(add_dummy_model=True)
+
+    # check that there are 2 columns (one for selected model and one for the dummy one)
+    assert len(set(metrics_df.columns.get_level_values(level=0))) == 2
+    assert "Dummy Baseline" in set(metrics_df.columns.get_level_values(level=0))
```
auguste-probabl:
```suggestion
@pytest.parametrize("data", ["binary_classification_data", "regression_data"])
def test_cross_validation_report_metrics_dummy_model(request, data):
    estimator, X, y = request.getfixturevalue(data)
    report = CrossValidationReport(estimator, X=X, y=y)

    metrics_df = report.metrics.report_metrics(add_dummy_model=True)
    assert "Dummy Baseline" in metrics_df.columns
```

skore/src/skore/sklearn/_estimator/metrics_accessor.py
auguste-probabl, line 1: There was a misunderstanding: I don't think it is necessary for every individual metric function (`precision`, `brier_score`, etc.) to gain the `add_dummy_model` parameter. Users can always fallback to `report_metrics` (now called `summarize`) if they want the dummy model. @glemaitre Any thoughts on this?

alexliap: In the public methods, e.g. `precision`, `accuracy` there is no such argument, i only added it to the private ones e.g. `_precision`, `_accuracy` with false as default such that when the public ones are called it only shows the metric for the respective model and not the dummy one. I don't see how the argument won't exist in these methods by the time the pipeline goes like this

Example:
`report_metrics` -> `_precision` -> `_compute_metrics`

auguste-probabl: Oops, I hadn't realized that the changes were on the private methods. In that case, that's fine by me!

glemaitre: Arff, I don't like the complexity that we introduce here. My thought would be that we are actually building a `ComparisonReport` or adding an `EstimatorReport` to a `ComparisonReport`. I have to think more how we could reduce the boilerplate here.

glemaitre: For an `EstimatorReport`, I think that we could trigger the computation of the metric with the following (rough implementation):

```python
        if add_dummy_model:
            from sklearn.dummy import DummyClassifier, DummyRegressor

            if "classification" in self._parent._ml_task:
                dummy_estimator = DummyClassifier()
            else:
                dummy_estimator = DummyRegressor()
            dummy_report = EstimatorReport(
                dummy_estimator,
                X_train=self._parent.X_train,
                y_train=self._parent.y_train,
                X_test=self._parent.X_test,
                y_test=self._parent.y_test,
            )
            summary = dummy_report.metrics.summarize(
                data_source=data_source,
                X=X,
                y=y,
                scoring=scoring,
                scoring_names=scoring_names,
                scoring_kwargs=scoring_kwargs,
                pos_label=pos_label,
            )
```

then, we can concatenate the results of `summary` with the existing dataframe. I would foresee the same pattern for the `CrossValidationReport`. For the `ComparisonReport`, I am wondering if we should always add a dummy `EstimatorReport` or `CrossValidationReport` and show the results only if requested. Otherwise, we can have a similar pattern in `summarize` itself.

@auguste-probabl do you have new thoughts with this perspective?

auguste-probabl: I'm not super sure about using EstimatorReport to implement EstimatorReport :thinking:

glemaitre: Yep it looks like like a 1-level recursion (I don't like recursion). But it really avoid touching any code and treat the `DummyEstimator` as any other estimator.

alexliap: ok the way I see it, it is either too "ugly" or gets to complex at some stage to add a dummy estimator at a report that is not the `comparison` one.

The simplest way i see to make the comparison between an estimator and a dummy one is to just make a `ComparisonReport` where one of the reports is from a dummy one. Something like this

```python
from skore import ComparisonReport, EstimatorReport

estimator_1 = LogisticRegression()
estimator_report_1 = EstimatorReport(estimator_1, **split_data)
estimator_2 = DummyClassifier()
estimator_report_2 = EstimatorReport(estimator_2, **split_data)

report = ComparisonReport([estimator_report_1, estimator_report_2])

report.metrics.report_metrics()
```
In order to avoid that, we can add an argument only in comparison report where instead of declaring a new report for the dummy estimator we add it behind the scenes.

```python
from skore import ComparisonReport, EstimatorReport

estimator = LogisticRegression()
estimator_report = EstimatorReport(estimator, **split_data)

report = ComparisonReport([estimator_report], add_dummy_model=True)

report.metrics.report_metrics()
```

auguste-probabl: I think you're right, the option should only be on ComparisonReport. That's what that object is for

glemaitre: I think I more comfortable with this proposal. Since we are discussing repurposing the feature, @auguste-probabl what about renaming `add_dummy_model` to `add_baseline`.

- It is a shorter name
- I would like it to accept `"dummy"`, `"tabular_learner"`, `None` or any scikit-learn estimator.

`"dummy"` would create a model as in this PR, `"tabular_learner"` will call `skrub.tabular_learner("classification"/"regression")` based on the underlying `_ml_task` and any estimator in case one need to tweak the model parameter.

The name of this report will be named `"Baseline"`.

@auguste-probabl what do you think?

auguste-probabl: sounds good to me

alexliap: one last thing, should i make it a different pr after all or not?

glemaitre: @alexliap As you wish and what is easier for you. In case, you start a new PR, just link it as well to this PR since an important discussion happen. But starting fresh might be easier to follow for people landing on the new PR.

alexliap: @auguste-probabl after checking out how this is going to be implemented, I found out that if the baseline estimator is going to be defined when the `summarize()` method is called, then we would somehow need to also provide a dataset. ComparisonReport works with the assumption that the respective reports have the same test data, so no data extra data is needed (watch example below from docs).

Example:
```python
from skore import ComparisonReport, EstimatorReport

estimator_1 = LogisticRegression(max_iter=10000, random_state=42)
estimator_report_1 = EstimatorReport(estimator_1, **split_data)
estimator_2 = LogisticRegression(max_iter=10000, random_state=43)
estimator_report_2 = EstimatorReport(estimator_2, **split_data)
comparison_report = ComparisonReport(
    [estimator_report_1, estimator_report_2]
)
comparison_report.metrics.report_metrics(
    scoring=["precision", "recall"],
    pos_label=1,
).frame()
```

So what I propose is this:
When the `add_baseline` argument is True, the user would also have to provide `X` and `y` in summarize method and also change the `data_source` to `X_y`. 

There is one issue with that: when the respective reports are `EstimatorReport`'s and not `CrossValidationReport`'s, `X_y` would be the training dataset (well some of it in reality because of splitting) for the baseline model and as a result training metrics would be reported, which is not fair for the other estimators. When using `CrossValidationReport`, it would be kind of "ok", i guess, because the results would be cross validated.

```python
from sklearn.datasets import make_classification, make_regression
from sklearn.linear_model import LogisticRegression, LinearRegression
from sklearn.tree import DecisionTreeClassifier, DecisionTreeRegressor
from skore import ComparisonReport, EstimatorReport, CrossValidationReport

X, y = make_regression(random_state=42, n_samples=5_000, n_informative=3, n_features=6)

estimator_1 = LinearRegression()
estimator_report_1 = EstimatorReport(estimator_1, **split_data)

estimator_2 = DecisionTreeRegressor()
estimator_report_2 = EstimatorReport(estimator_2, **split_data)


X, y = make_classification(random_state=42, n_samples=1_000, n_classes=2, n_informative=2, n_features=6)

comparison_report = ComparisonReport(
    [estimator_report_1, estimator_report_2]
)
print(comparison_report.metrics.summarize(
    add_baseline=True,
    data_source="X_y",
    X=X,
    y=y
).frame())
```
```bash
Estimator         LinearRegression  DecisionTreeRegressor  Baseline                                                                          
Metric
R²                  -125358.437759         -104732.738585 -0.001067
RMSE                    177.027487             161.810049  0.500258
Fit time (s)              0.001060               0.021628  0.000269
Predict time (s)          0.000195               0.000332  0.000072
```

what do you think?

alexliap: @auguste-probabl @glemaitre the more i think about it, it seems more sensible to just create a report using a dummy model and then adding to a comparison report along with all the other reports, like the comparison report was meant to be used. maybe this feature doesn't need to be implemented after all, because it can already be tackled from the existing code, relatively easily

glemaitre: yep, I was thinking that we could add the `add_baseline` directly into the `__init__` of the `ComparisonReport`. But we force only `y_test` to be the same for all models. Since that even the `DummyEstimator` are using a `y_train`, then we cannot really pick-up a `y_train` from the other estimator because they can be different. It is even more complicated when `X` is needed because in this case, each estimator can have different `X` even if `y_test` is the same.

So I need to put more thoughts about it to better define what would be really useful and if we can or cannot implement it.

Thanks @alexliap for thinking about the problem, it is super valuable even if don't end-up implementing the feature. For sure, we should document in an example including some baseline.

alexliap: thank you, no worries, do you want me to close this PR?

PR #1664, opened 2025-05-07
feat(skore-remote-project): Allow to put only `skore.EstimatorReport`

thomass-dev:
Allow to put only `skore.EstimatorReport`, as long as we can't get them back or display them in the hub.


rouk1 (approved):
Ok LGTM.

If we remove all other kind of "items" I suggest to implicitly put estimator reports and mark the put method a private. what do you think ?

thomass-dev:
> Ok LGTM.
> 
> If we remove all other kind of "items" I suggest to implicitly put estimator reports and mark the put method a private. what do you think ?

Implicit/explicit will be the next step after the release of the `search` feature 😄 .

PR #1665, opened 2025-05-07
feat: Add `Makefile/make-install-skore-lts-cpu` target

thomass-dev:
Following https://github.com/probabl-ai/skore/pull/1618:

- Add a new target to install `skore` on LTS CPU,
- Make `test`/`sphinx` optional dependencies self-sufficient.

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -34,9 +34,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-polars = ["polars"]
-polars-lts-cpu = ["polars-lts-cpu"]
-test = [
+test-base = [
   "altair>=5,<6",
   "pillow",
   "plotly>=5,<6",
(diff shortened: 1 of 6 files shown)


skore/pyproject.toml
auguste-probabl, line 1: This feels too complex for what it does. I'd say the current implementation did its job well

thomass-dev: These changes are useful for compiling lockfiles without worrying about optional dependencies other than `test`.
I find it clearer to have only one optional dependency to install to run tests.
Currently, for those who don't use the `Makefile`, it's a bit obscure that you need to install polars to run the tests.

thomass-dev: > These changes are useful for compiling lockfiles without worrying about optional dependencies other than test.

Otherwise we have to check explicitly for each package the existency of the `polars` optional dependencies.
We have to compile lockfiles with `test` and `polars` if exists. If not, only with `test`.

PR #1667, opened 2025-05-07
refactor(plots): Make internal plot methods non-static

auguste-probabl:
This removes redundancy.


glemaitre (approved):
Yep this makes sense.

PR #1668, opened 2025-05-07
refactor(_get_display): Make `y` data passed to `_get_display` more structured

auguste-probabl:
This addresses some [primitive obsession](https://refactoring.guru/smells/primitive-obsession) in preparation for plotting ComparisonReports of CrossValidationReports.

### DIFF
--- a/skore/src/skore/sklearn/types.py
+++ b/skore/src/skore/sklearn/types.py
@@ -1,7 +1,7 @@
 """Types between parts of the sklearn module."""
 
 from collections.abc import Iterator, Sequence
-from typing import Any, Callable, Literal, Protocol, Union
+from typing import Any, Callable, Literal, Optional, Protocol, TypedDict, Union
 
 from numpy.typing import ArrayLike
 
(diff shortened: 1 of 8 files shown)


skore/src/skore/sklearn/types.py
```diff
@@ -23,6 +23,17 @@
...
 
 
+class YPlotData(TypedDict):
+    """Response values, either `y_true` or `y_pred`.
+
+    Used for passing to Display classes.
+    """
+
+    estimator_name: str
+    split_number: Optional[int]
```
glemaitre, line 33: instead of number, isn't it rather `split_index`?

skore/src/skore/sklearn/types.py
```diff
@@ -23,6 +23,17 @@
 Aggregate = Union[Literal["mean", "std"], Sequence[Literal["mean", "std"]]]
 
 
+class YPlotData(TypedDict):
```
glemaitre, line 26: I'm wondering if we should not use a `dataclass` then. Whenever calling the instance, we can access with `.attribute` instead of `["attribute"]`. And they are data :)

glemaitre (approved):

glemaitre:
Looks good. Let's merge this one.

PR #1674, opened 2025-05-09
docs: Fail on warnings during building

thomass-dev:
After resolving all current warnings, use the `--fail-on-warning` flag during documentation building to fail on warnings.


glemaitre:
While we are annoyed with the warnings, they are not necessarly useless, it is just that they are hard to understand. Not having usually would avoid some discrepancy in the documentation.

The `nit-picky` mode is probably too much thought.

thomass-dev:
> Do we really want to enforce this? Because sphinx may pop up many warnings quite useless..

There are too many after-the-fact PRs that solve sphinx warnings that could impact the rendering. IMO, this shows flaws in the racket.

thomass-dev:
> [!caution]
> This PR must be up-to-date before being merged.

auguste-probabl:
I'd like to try setting the more restrictive option. If we can manage to remove all warnings once, then we should be able to keep it that way

PR #1680, opened 2025-05-12
ci: Change workflow to release `skore|skore-(local|remote)-project`

thomass-dev:
Change release workflow to release `skore|skore-(local|remote)-project` package.

Tags must be prefixed by package:
- `p-a-c-k-a-g-e/X.Y.Z`     
- `p-a-c-k-a-g-e/X.Y.Z-rc.A`

Examples of valid tags: `skore/1.0.0`, `skore-local-project/0.0.1`.

---

Various:
- name artifacts in `release.yml` using UUID to allow parallel releases,
- deploy documentation in `sphinx.yml` only on `skore` tags (and not RC).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,7 +1,6 @@
+# Release a package.
 #
-# p-a-c-k-a-g-e/X.Y.Z
-# p-a-c-k-a-g-e/X.Y.Z-rc.1
-#
+# Package versions are of the form `<package-name>/X.Y.Z` or `<package-name>/X.Y.Z-rc.1`
 
 name: release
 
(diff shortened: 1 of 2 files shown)


thomass-dev:
Fyi @MarieSacksick .

.github/workflows/release.yml
```diff
@@ -1,88 +1,107 @@
+#
+# p-a-c-k-a-g-e/X.Y.Z
+# p-a-c-k-a-g-e/X.Y.Z-rc.1
+#
```
auguste-probabl, line 4:
```suggestion
# Release a package.
# Package versions are of the form
# <package-name>/X.Y.Z
# or
# <package-name>/X.Y.Z-rc.1
```

auguste-probabl (dismissed):
Just a suggestion but otherwise looks great!

.github/workflows/release.yml
thomass-dev: This is not `package versions`, but tags.

PR #1682, opened 2025-05-13
fix(roc-curve): Show chance level in the legend

auguste-probabl:
The chance level line label now appears in the legend.

Before/After:

---

Depends on #1668


glemaitre (approved):

glemaitre:
Thanks for the fix.

PR #1666, opened 2025-05-07
ci: Factorize `skore*.yml` in `pytest.yml`

thomass-dev:
Factorize `skore*.yml` workflows in `pytest.yml`  to run jobs on all packages defined by `vars.PACKAGES`, managed in the GitHub interface.

PR #1687, opened 2025-05-14
ci: Add Python 3.13 to the supported versions

auguste-probabl:


thomass-dev:
You don't want to support scikit-learn 1.4 in python 3.13?

ci/pip-compile.sh
```diff
@@ -34,6 +34,9 @@ COMBINATIONS[2]='3.11;1.6'
 COMBINATIONS[3]='3.12;1.4'
 COMBINATIONS[4]='3.12;1.5'
 COMBINATIONS[5]='3.12;1.6'
+COMBINATIONS[6]='3.13;1.4'
```
thomass-dev, line 37: This combination is not tested.

thomass-dev (changes requested):
I'm not sure what you want to do with python 3.13 ; scikit-learn 1.4.

ci/pip-compile.sh
auguste-probabl: Yup, forgot to remove it. Thanks

sylvaincom (commented):
Please update the badge in the README into this one:

https://pypi-camo.freetls.fastly.net/9a75e2027e0191a6196fa9700de7792e0ee02b3f/68747470733a2f2f696d672e736869656c64732e696f2f707970692f707976657273696f6e732f7363696b69742d6c6561726e2e737667

ci/pip-compile.sh
thomass-dev: Please remove the corresponding lockfiles.

auguste-probabl:
> Also, we should only support 4 versions of python: we should remove the explicit support of python 3.9 in favor of 3.13.

This will be done in a future PR

auguste-probabl:
> @auguste-probabl if we don't want to support `scikit-learn 1.4` on `python 3.13` for now, please add an indication for the user in the README.

Makes sense. When I tried Python 3.9 I remember getting an error to do with numpy2, however the numpy docs say 3.9 should be supported: https://numpy.org/devdocs/release/2.0.0-notes.html#numpy-2-0-python-api-removals
So several options:
- Make it work with 3.9
- Add a message now and make it work with 3.9 later
- Don't make it work with 3.9

sylvaincom:
> Also, we should only support 4 versions of python: we should remove the explicit support of python 3.9 in favor of 3.13.

In that case, we should also update the badge in the README 🙏

rouk1 (approved):
Looks good to me.
3.9 will be EOL in October this year we should drop support for it soon.

.github/workflows/skore.yml
```diff
@@ -91,7 +91,7 @@ jobs:
       fail-fast: false
```
thomass-dev, line 91: The workflows of other packages should also be updated.
Can we wait for https://github.com/probabl-ai/skore/pull/1666?

auguste-probabl: @thomass-dev can you take another look?

thomass-dev (approved):

PR #1698, opened 2025-05-15
feat: Use hub production API url as default

rouk1:
API client now uses production URL & fix mypy issue.


auguste-probabl (approved):
LGTM

PR #1703, opened 2025-05-15
test: Test the parallel coordinate widget

glemaitre:
Adding test for the `ModelExplorerWidget` class.

### DIFF
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -3,8 +3,8 @@ from __future__ import annotations
 import re
 import sys
 import types
-import pandas as pd
 
+import pandas as pd
 
 if sys.version_info < (3, 10):
     from importlib_metadata import entry_points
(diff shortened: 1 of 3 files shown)


skore/src/skore/project/project.py
```diff
@@ -60,13 +60,14 @@ def get(id: str) -> EstimatorReport:  # hide underlying functions from user
...
             """"""
             metadata = pd.DataFrame(self.__project.reports.metadata(), copy=False)
-            metadata["learner"] = pd.Categorical(metadata["learner"])
-            metadata.index = pd.MultiIndex.from_arrays(
-                [
-                    pd.RangeIndex(len(metadata)),
-                    pd.Index(metadata.pop("id"), name="id", dtype=str),
-                ]
-            )
+            if not metadata.empty:
```
thomass-dev, line 63: Fmi, this snippet fails without this `if` condition on empty dataframe?

glemaitre: Yep because the dataframe is really empty and does not have any column. So `df["learner"]` raises a key error. I think having an empty frame is fine. It makes it explicit that there is nothing in the project. We just need to have some test covering this case.

thomass-dev: I think it was covered before, and i added a regression during my iterations. Fine with this solution!

thomass-dev, line 63: I have fixed the failing test by reverting one of my latest changes.
You have to update your branch, and move this add from` project.py` to `metadata.py`.

thomass-dev: Please update your branch, your fix is now integrated. Thanks!

thomass-dev (approved):

PR #1701, opened 2025-05-15
feat(Comparison[CVReport]): Add precision-recall curve

auguste-probabl:
Depends on: #1669

PR #1709, opened 2025-05-16
feat(prediction-error): Support Comparison[CVReport]

auguste-probabl:
Depends on #1701


skore/src/skore/sklearn/_plot/metrics/prediction_error.py
```diff
@@ -149,26 +139,57 @@ def _validate_data_points_kwargs(
         ValueError
             If the format of `data_points_kwargs` is invalid.
         """
+        if self.report_type == "estimator":
+            allow_single_dict = True
+            n_curves = 1
```
glemaitre: I think that we abuse a bit the term `n_curves` here. Maybe `n_scatter_groups`?

glemaitre (commented):
It looks good. A single renaming to avoid confusion.

skore/src/skore/sklearn/_plot/metrics/prediction_error.py
auguste-probabl: I mean, a "scatter group" is just a curve without the lines. Also `n_curves` is consistent with the other plots, which makes the logic a bit easier to grok. I don't really mind either way.

auguste-probabl: renamed

glemaitre:
LGTM. Thanks @auguste-probabl

PR #1717, opened 2025-05-19
docs: Fix documentation following the change in "what you put is what you get"

sylvaincom:
This PR builds on #1475 which introduces a breaking change in the API: it is no longer possible to `put` any kind of python objects, only estimator reports from skore.

Changes made by the current PR:
- deleting the 2 examples about the skore project: _Tracking items_ and _Working with projects_, hence removing the whole subsection
- In _Simplified experiment reporting_, remove logic to do with `skore.Project` since it no longer supports `put`-ing CrossValidationReports
- added the screenshots on the examples, based on https://github.com/probabl-ai/skore/pull/1722

TODO (can be done in a separate PR)
- the Sphinx interlink to our API does not work


examples/getting_started/plot_quick_start.py
```diff
@@ -73,18 +74,58 @@
...
+
+report_get[0].metrics.report_metrics(pos_label=1)
+
+# %%
+# .. note::
+#   If rendered in a Jupyter notebook, ``metadata`` would render an interactive
+#   parallel coordinate plot to search for your preferred model based on some metrics.
+#   Here is a screenshot:
+#
+#   .. image:: https://media.githubusercontent.com/media/probabl-ai/skore/16bf794f929239830a9ff809f930c1bd22318417/sphinx/_static/images/screenshot_quick_start.png
```
glemaitre, line 127: Since `html_static_path` is defined in `conf.py`, we should be able to link in the following manner.

```suggestion
#   .. image:: /images/screenshot_quick_start.png
```

sylvaincom: My answer / remark is related to https://github.com/probabl-ai/skore/pull/1722#issuecomment-2892304585

glemaitre: I don't think that we should document such details in the read me. So I don't think that it should be taken into account.

README.md
```diff
@@ -67,52 +67,33 @@ You can find information on the latest version [here](https://anaconda.org/conda
```
thomass-dev, line 67: Can you please add detail about how to work with local/hub project.

```bash
# local
pip install skore 
# or local + hub
pip install skore[hub]
```
```python
import skore

skore.Project("<project>") # local
skore.Project("hub://<tenant>/<name>") # hub 
```

thomass-dev (approved):
Many thanks for your help!

PR #1718, opened 2025-05-19
fix: Make sure to select properly learners

glemaitre:
There is currently a bug when selecting model using the `Learner` dimension on the parallel coordinate plot.

It is fixing the problem.


thomass-dev (approved):

auguste-probabl (approved):
Merging in the interest of time.

PR #1722, opened 2025-05-19
docs: Add screenshots of parallel coordinates plots

sylvaincom:
For https://github.com/probabl-ai/skore/pull/1717, I need the following screenshots.

### DIFF
--- a/sphinx/_static/images/screenshot_getting_started.png
+++ b/sphinx/_static/images/screenshot_getting_started.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:624e6a91f8ef47361e751c69169c6a60aee87c323fedde45fe1e21b6af48ca88
+size 136602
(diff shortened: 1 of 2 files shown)


thomass-dev (approved):

glemaitre:
Is it normal that it target the `main` branch?

sylvaincom:
Yes, so that the static file is available online right now?

glemaitre:
> For me to use the URL in my PR

I'm not sure to understand. You should use relative path and you could have that in your PR. What I'm surprised is that this feature should be available only in the `search-feature` branch.

sylvaincom:
In the past, relative paths were my way to go by default, but it happened that a relative path of a GIF worked on the README.md but not when rendered to the PyPi index page (it broke as the file could not be found). Since then, I've been using absolute paths. But yes, I could go back to relative paths if I only need to worry about README.md?

PR #1725, opened 2025-05-20
fix: Assign the proper length sequence

glemaitre:
Fixes the failure in `test_widget.py` by assigning a sequence of the same size.


glemaitre:
For the other failure, we should add `anywidget` as a dependency.

glemaitre:
Uhm otherwise, it could be linked to: https://github.com/plotly/plotly.py/issues/5027

thomass-dev (approved):

PR #1735, opened 2025-05-21
chore: Make _query_string_selection private

glemaitre:
I don't think that we need this method to be exposed to the user for the moment.
We can always make it public if we get a feature request.

### DIFF
--- a/skore/src/skore/project/metadata.py
+++ b/skore/src/skore/project/metadata.py
@@ -37,9 +37,6 @@ class Metadata(pd.DataFrame):
         Return the reports referenced by the metadata object from the project.
         If a query string selection exists, it will be automatically applied before, to
         filter the reports to return. Otherwise, returns all reports.
-    _query_string_selection() -> Union[str | None]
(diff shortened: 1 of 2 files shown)


skore/src/skore/project/metadata.py
```diff
@@ -37,7 +37,7 @@ class Metadata(pd.DataFrame):
         Return the reports referenced by the metadata object from the project.
         If a query string selection exists, it will be automatically applied before, to
         filter the reports to return. Otherwise, returns all reports.
-    query_string_selection() -> Union[str | None]
+    _query_string_selection() -> Union[str | None]
         Generate a ``pandas`` query string based on user selections in the parallel
         coordinates plot.
```
thomass-dev, line 42:
```suggestion
```

thomass-dev (changes requested):
LGTM, except you should remove the corresponding docstr from the class docstr.

thomass-dev (approved):

Issue #1736, opened 2025-05-21
bug: I can't put a report where the estimator doesn't have a predict_proba


MarieSacksick:
### Describe the bug

I can't put a report where the estimator doesn't have a predict_proba.

### Steps/Code to Reproduce

```
# %%
from skore import EstimatorReport
from sklearn.datasets import make_classification
from skore import train_test_split
X, y = make_classification(n_samples=1000, n_features=9, random_state=1)
X = pd.DataFrame(data=X, columns=["1", "2", "3", "4", "5", "6", "7", "8", "9"])
X_train, X_test, y_train, y_test = train_test_split(X,y)
clf = svm.SVC(kernel='rbf', C=2, random_state=0)
est_rep = EstimatorReport(estimator=clf, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test)

# %%
import skore
project = skore.Project("project")

# %%
project.put("est_report", est_rep)
```

### Expected Behavior

It should put, the user shouldn't to worry about the fact that some models don't have predict_proba.

### Actual Behavior

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[29], line 1
----> 1 project.put("est_report", est_rep)

File ~/Documents/skore/skore/src/skore/project/project.py:209, in Project.put(self, key, report)
    204 if not isinstance(report, EstimatorReport):
    205     raise TypeError(
    206         f"Report must be a `skore.EstimatorReport` (found '{type(report)}')"
    207     )
--> 209 return self.__project.put(key=key, report=report)

File ~/anaconda3/envs/skore_test/lib/python3.12/site-packages/skore_local_project/project.py:215, in Project.put(self, key, report)
    202             return float(getattr(report.metrics, name)(data_source="test"))
    203     return None
    205 self.__metadata_storage[uuid4().hex] = {
    206     "project_name": self.name,
    207     "run_id": self.run_id,
    208     "key": key,
    209     "artifact_id": pickle_hash,
    210     "date": datetime.now(timezone.utc).isoformat(),
    211     "learner": report.estimator_name_,
    212     "dataset": joblib.hash(report.y_test),
    213     "ml_task": report._ml_task,
    214     "rmse": metric("rmse"),
--> 215     "log_loss": metric("log_loss"),
    216     "roc_auc": metric("roc_auc"),
    217     # timings must be calculated last
    218     "fit_time": report.metrics.timings().get("fit_time"),
[...]

PR #1738, opened 2025-05-21
fix: Log loss only available with estimator exposing predict_proba

glemaitre:
closes https://github.com/probabl-ai/skore/issues/1736

This PR is making sure that we expose the `log_loss` only for estimator exposing a `predict_proba` method.


thomass-dev (approved):
Efficient! Thanks.

PR #1740, opened 2025-05-21
docs(skore): Add Project.reports methods to the documentation

glemaitre:
Change `Project.reports` from `types.SimpleNamespace` to `pandas` style accessor, to let sphinx documents:

- `Project.reports.get`
- `Project.reports.metadata`

### DIFF
--- a/skore/src/skore/project/__init__.py
+++ b/skore/src/skore/project/__init__.py
@@ -1,9 +1,5 @@
 """Alias top level function and class of the project submodule."""
 
-from skore.externals._pandas_accessors import _register_accessor
 from skore.project.project import Project
-from skore.project.reports import _ReportsAccessor
-
-_register_accessor("reports", Project)(_ReportsAccessor)
 
 __all__ = ["Project"]
--- a/skore/src/skore/project/__init__.py
+++ b/skore/src/skore/project/__init__.py
@@ -1,5 +1,9 @@
 """Alias top level function and class of the project submodule."""
 
-from .project import Project
+from skore.externals._pandas_accessors import _register_accessor
+from skore.project.project import Project
+from skore.project.reports import _ReportsAccessor
+
+_register_accessor("reports", Project)(_ReportsAccessor)
 
 __all__ = ["Project"]
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -4,18 +4,21 @@ from __future__ import annotations
 
 import re
 import sys
-import types
+from typing import TYPE_CHECKING
 
 if sys.version_info < (3, 10):
     from importlib_metadata import entry_points
 else:
     from importlib.metadata import entry_points
 
-from ..sklearn._estimator.report import EstimatorReport
-from .metadata import Metadata
+from skore.externals._pandas_accessors import DirNamesMixin
+from skore.sklearn._estimator.report import EstimatorReport
+
+if TYPE_CHECKING:
+    from skore.project.reports import _ReportsAccessor
 
 
-class Project:
+class Project(DirNamesMixin):
     r"""
     API to manage a collection of key-report pairs.
 
@@ -146,6 +149,7 @@ class Project:
     """
 
     __HUB_NAME_PATTERN = re.compile(r"hub://(?P<tenant>[^/]+)/(?P<name>.+)")
+    reports: _ReportsAccessor
 
     def __init__(self, name: str, **kwargs):
         if not (PLUGINS := entry_points(group="skore.plugins.project")):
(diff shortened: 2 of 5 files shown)


skore/src/skore/project/project.py
```diff
@@ -208,19 +212,5 @@ def put(self, key: str, report: EstimatorReport):
 
         return self.__project.put(key=key, report=report)
 
-    @property
-    def reports(self):
```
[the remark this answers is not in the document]
glemaitre: I'll try in another PR. We should really avoid the mangled attribute. They are so much annoying for no particular reason.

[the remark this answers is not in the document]
glemaitre: Actually I don't know what you mean. Something like that?

```python
class _ReportAccessor:
    ...
    
    
class A:

  def __init__(self, ...):
      self.reports = _ReportAccessor()
```

skore/src/skore/project/reports.py
```diff
@@ -0,0 +1,21 @@
+"""Module to manage the persisted reports."""
+
+from skore.externals._pandas_accessors import DirNamesMixin
+from skore.project import Project
+from skore.project.metadata import Metadata
+from skore.sklearn import EstimatorReport
```
thomass-dev, line 6: To avoid circular imports with my other comments:

```suggestion
from __future__ import annotations

from typing import TYPE_CHECKING

from skore.externals._pandas_accessors import DirNamesMixin

if TYPE_CHECKING:
    from skore.project import Project
    from skore.project.metadata import Metadata
    from skore.sklearn import EstimatorReport
```

skore/src/skore/project/project.py
```diff
@@ -146,6 +149,10 @@ class Project:
     """
 
     __HUB_NAME_PATTERN = re.compile(r"hub://(?P<tenant>[^/]+)/(?P<name>.+)")
+    reports: _ReportsAccessor
+    _Project__project: Any
+    _Project__mode: str
+    _Project__name: str
```
thomass-dev, line 155: Is this mandatory for mypy? Because i had no error.

skore/src/skore/project/__init__.py
```diff
@@ -1,5 +1,9 @@
 """Alias top level function and class of the project submodule."""
 
-from .project import Project
+from skore.externals._pandas_accessors import _register_accessor
+from skore.project.project import Project
+from skore.project.reports import _ReportsAccessor
+
+_register_accessor("reports", Project)(_ReportsAccessor)
```
thomass-dev, line 7: Can you move this line to the end of the `project.py` file? IMO, it should me more readable.

thomass-dev (changes requested):
Thanks for the fix.
I'm sad that `types.SimpleNamespace` doesn't work with sphinx, but in retrospect, that's normal. It can't introspect an object created dynamically on call of the property, at runtime...

skore/src/skore/project/project.py
glemaitre: Yep `mypy` was complaining in the accessor file. In this case, I think `mypy` does not understand the mangled attribute.

skore/src/skore/project/__init__.py
glemaitre: I can do that but I wanted just to recall that we did the exact opposite when creating the report :) (I think @rouk1 mentioned it). Somehow, we can think that we create potentially different accessors and assemble them in the `__init__.py`.

Are you still certain that you want to go this road?

thomass-dev: I find quite confusing to have to read both `__init__` and `project.py` to understand what a project is composed of.

> Somehow, we can think that we create potentially different accessors and assemble them in the __init__.py.

Why not assembling the project accessors in the project file?

glemaitre: I find it quite nice that the `__init__.py` is the place to do that. Otherwise, you need to go down the file in `project.py`.

But I made the changes.

skore/src/skore/project/project.py
```diff
@@ -4,18 +4,22 @@
...
     from importlib.metadata import entry_points
 
-from ..sklearn._estimator.report import EstimatorReport
-from .metadata import Metadata
+from skore.externals._pandas_accessors import DirNamesMixin, _register_accessor
+from skore.project.reports import _ReportsAccessor
+from skore.sklearn._estimator.report import EstimatorReport
 
+if TYPE_CHECKING:
+    from skore.project.reports import _ReportsAccessor
```
thomass-dev, line 19: This import is duplicate.

thomass-dev: Otherwise, LGTM, i will merge after that.

thomass-dev (dismissed):
Nice!

thomass-dev (approved):

PR #1741, opened 2025-05-21
chore: Fix warning in sphinx about duplicate object

glemaitre:
This PR tries to solve the issues regarding the warning raised by `sphinx`, e.g.

```
/Users/xxx/skore/src/skore/project/project.py:docstring of skore.Project.reports:1: WARNING: duplicate object description of skore.Project.reports, other instance in api/skore.Project.reports, use :no-index: for one of them
```

It solves as well a couple rendering issues spotted when debugging. It remains a single warning due to the `skore.Project` class but the current solution in conjunction with the changes proposed in https://github.com/probabl-ai/skore/pull/1740 will solve the issue.

### DIFF
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -30,33 +30,33 @@ class Project:
     Two mutually exclusive modes are available and can be configured using the ``name``
     parameter of the constructor:
 
-    - mode : hub
+    .. rubric:: Local mode
 
-        If the ``name`` takes the form of the URI ``hub://<tenant>/<name>``, the project
-        is configured to the ``hub`` mode to communicate with the ``skore hub``.
+    Otherwise, the project is configured to the ``local`` mode to be persisted on
+    the user machine in a directory called ``workspace``.
 
-        A tenant is a ``skore hub`` concept that must be configured on the ``skore hub``
-        interface. It represents an isolated entity managing users, projects, and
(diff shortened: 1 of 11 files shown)


skore/src/skore/project/project.py
```diff
@@ -30,33 +30,33 @@ class Project:
     Two mutually exclusive modes are available and can be configured using the ``name``
     parameter of the constructor:
 
-    - mode : hub
+    .. rubric:: Local mode
 
-        If the ``name`` takes the form of the URI ``hub://<tenant>/<name>``, the project
-        is configured to the ``hub`` mode to communicate with the ``skore hub``.
+    Otherwise, the project is configured to the ``local`` mode to be persisted on
```
thomass-dev, line 35: This sentence no longer makes sense after your reordering.
It should be changed or the ordering should be revert.

IMO it's clearer (as before) to say at first that the `hub` mode derived from the form of the name.
And if not, it's the `local` mode.

thomass-dev (changes requested):
Thanks for the fix and for the slight rewriting of the docstr.
Just a thought to address.

skore/src/skore/project/project.py
glemaitre: I reverted.

thomass-dev (approved):
Great, thanks!

PR #1744, opened 2025-05-22
enh: Add a .frame() method to Displays

### LINKED ISSUE #1359: enh: Add a `.frame()` method to Displays

sylvaincom:
Fix #1359 

TODO:
- add unit tests

### DIFF
--- a/skore/src/skore/sklearn/_plot/metrics/prediction_error.py
+++ b/skore/src/skore/sklearn/_plot/metrics/prediction_error.py
@@ -684,3 +684,15 @@ class PredictionErrorDisplay(StyleDisplayMixin, HelpDisplayMixin):
             ml_task=ml_task,
             report_type=report_type,
         )
+
+    def frame(self):
+        """Return the precision error computations as a dataframe.
+
+        Returns
+        -------
+        frame : pandas.DataFrame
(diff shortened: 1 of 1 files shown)


glemaitre (commented):
I think that the actual implementation will can reuse the refactor from @auguste-probabl. This will probably simplify some code. However, it is great that we already have some tests.

We will need to update the `Display` protocol to expose that we have `.frame` method.

thomass-dev:

sylvaincom:
Good call, I think the `to_frame(...)` method is nice, as it resonates with the method from pandas and is thus quite natural for a DS

sylvaincom:
(I just attempted to resolve some conflicts, I'll get back to it when I get back)

sylvaincom:
Superseded by https://github.com/probabl-ai/skore/pull/1787

PR #1745, opened 2025-05-22
fix: Make sure to not mutate cache during Parallel processing

glemaitre:
We see some flakyness in https://github.com/probabl-ai/skore/actions/runs/15188333231/job/42714438658?pr=1593

I'm thinking that it could be due to this code: since we have a generator, we can modify `self._cache` before/during that it is consume by `Parallel`, making the pickling of the task to fail.

Let's see if it works.


glemaitre:
@auguste-probabl how many times did you witness the failure? Just to know how likely that the fix is solving the problem. I could not reproduce locally.

thomass-dev (approved):
LGTM.

Otherwise, if you really want to update cache on the fly... you have to manage a lock by yourself and even a SharedMemory if each task have to modify itself the cache...

Issue #1702, opened 2025-05-15
feat(CrossValidationReport): Add grouping of predictions from cross validation


foster999:
### Is your feature request related to a problem? Please describe.

I'm using plotting a ROC as an example here, but would like to be able to do this when calculating metrics also.

I'd like to be able to plot a ROC for:
1. Each repeat of a repeated cross validation splitter. Each curve would include one prediction for every sample in the training dataset (k-folds for that repeat grouped together). I feel that this is more informative (and concise) than showing one curve per individual fold in the repeated case and provides a good visualisation of model instability
2. All predictions from across all splits. This provides a good estimate for the ROC that would be produced by the model training process under cross validation. And aids in selecting a threshold for classification in the final model.

### Describe the solution you'd like

To be able to specify the grouping on demand, to use both types:
```
report = skore.CrossValidationReport(
    pipeline,
    X,
    y,
    cv_splitter=RepeatedStratifiedKFold(n_splits=10, n_repeats=8),
    n_jobs=-1
)
repeat_roc_plot = report.metrics.roc(by="repeat")
repeat_roc_plot.plot()

total_roc_plot = report.metrics.roc(by="all")
total_roc_plot.plot()
```

### Additional context

I'm happy to contribute to this if others consider it to be useful and good practice :)

glemaitre:
> Each repeat of a repeated cross validation splitter. Each curve would include one prediction for every sample in the training dataset (k-folds for that repeat grouped together). I feel that this is more informative (and concise) than showing one curve per individual fold in the repeated case and provides a good visualisation of model instability

It is difficult to follow. @foster999 would you mind to provide some pseudo code or a minimal array on how you want to group the predictions.

> All predictions from across all splits. This provides a good estimate for the ROC that would be produced by the model training process under cross validation. And aids in selecting a threshold for classification in the final model.

Actually we have to be extra careful when it comes to get a single ROC. This paper is quite nice: https://openreview.net/pdf?id=FByH3qL87G. In conclusion, there not a single good method because each of them have they drawbacks. Maybe one way would be to have a keyword `average` that could be a string "pooling"/"vertical"/"thresholds" or None that should be the default.

foster999:
Thanks for pointing me to the paper, I'll take a look and reflect on whether it is suitable for my use case. I'm looking to estimate performance of a model from the model training process, so I think would be closest to their "Comparing classifiers" case.

[...]

PR #1750, opened 2025-05-23
feat(CrossValidationReport): Add threshold averaging for roc plot

foster999:
Implements the threshold averaging method for ROC curve averaging (see #1702) 

Includes simple test for averaging logic


foster999:
@glemaitre do you have any thoughts on the questions I've included above?

Regarding showing how much variation there is, I had something like this in mind: 
example_average_roc

I generally have a large number of ROC curves in the average, so wouldn't want to show a legend for each one

foster999:
> [automated comment] Please update your PR with main, so that the `pytest` workflow status will be reported.

Updated to use new format for passing data to displays. I think approval might be needed to get the workflows running?

skore/src/skore/sklearn/_plot/metrics/roc_curve.py
```diff
@@ -900,8 +1026,63 @@ def _compute_data_for_display(
...
+                        cls._threshold_average(
+                            xs=all_fprs,
+                            ys=all_tprs,
+                            thresholds=all_thresholds,
+                        )
+                    )
+                else:
+                    raise TypeError(
+                        "'threshold' is the only supported option for `average`,"
+                        f"but got {average} instead"
```
auguste-probabl, line 1054:
```suggestion
                        'average must be "threshold" or None, '
                        f"got {average}"
```

skore/src/skore/sklearn/_plot/metrics/roc_curve.py
```diff
@@ -900,8 +1026,63 @@ def _compute_data_for_display(
...
+                ):
+                    roc_curve_records.append(
+                        {
+                            "estimator_name": y_true_i.estimator_name,
+                            "split_index": None,
+                            "label": pos_label_validated,
+                            "threshold": threshold,
+                            "fpr": fpr,
+                            "tpr": tpr,
+                            "average": "threshold",
```
auguste-probabl, line 1068: I'm on the fence about "average" being a column in the DataFrame, but I think it makes sense as it changes the interpretation of the values in the DataFrame.

skore/src/skore/sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -549,6 +549,7 @@ def _compute_data_for_display(
         cls,
         y_true: Sequence[YPlotData],
         y_pred: Sequence[YPlotData],
+        average: Optional[Literal["threshold"]] = None,
```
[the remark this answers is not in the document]
foster999: Thanks for taking a look! I added the parameter here but not the functionality - all three display classes currently use the same interface for `_compute_data_for_display`, but I only want to make the addition to the ROC display. I'll update this to just use a different interface for `RocCurveDisplay`

auguste-probabl:
It looks like you're still working on this PR; if so, can you set it to draft?

foster999:
> It looks like you're still working on this PR; if so, can you set it to draft?

Thanks for taking another look, just changed to draft now

I could do with advice on a couple of points please:

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -1211,18 +1216,34 @@ def _get_display(
...
-                ml_task=self._parent._ml_task,
-                data_source=data_source,
-                **display_kwargs,
-            )
+            if display_class == RocCurveDisplay:
+                display_class = cast(type[RocCurveDisplay], display_class)
+                display = display_class._compute_data_for_display(
+                    y_true=y_true,
+                    y_pred=y_pred,
+                    average=average,
```
foster999, line 1224: Mypy is still complaining about this line, even after casting:
`Unexpected keyword argument "average" for "_compute_data_for_display" of "PrecisionRecallCurveDisplay`

Without casting it shows the same error twice. Any suggestions on why?

skore/src/skore/sklearn/_plot/metrics/roc_curve.py
```diff
@@ -387,6 +388,115 @@ def _plot_cross_validated_estimator(
...
+            f"\n(Positive label: {self.pos_label})"
+            if self.pos_label is not None
+            else ""
+        )
+
+        query = f"label == {self.pos_label!r} & average == '{average_type}'"
+        average_roc_curve = self.roc_curve.query(query)
+        average_roc_auc = self.roc_auc.query(query)["roc_auc"].item()
+
+        line_kwargs_validated = _validate_style_kwargs({}, {})
```
foster999, line 464: I'm not sure how best to take style kwargs for the average line. Other lines slice using `split_idx`, but this is `None` for the average line

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
auguste-probabl: Hm, the cast looks correct to me... maybe [reveal_type](https://mypy.readthedocs.io/en/latest/common_issues.html#displaying-the-type-of-an-expression) could help?

foster999: > Hm, the cast looks correct to me... maybe [reveal_type](https://mypy.readthedocs.io/en/latest/common_issues.html#displaying-the-type-of-an-expression) could help?

Thanks, though it looks how I would expect after casting:
`Revealed type is "type[skore.sklearn._plot.metrics.roc_curve.RocCurveDisplay]"`

skore/src/skore/sklearn/_plot/metrics/roc_curve.py
auguste-probabl: Yeah, it might be time to make `roc_curve_kwargs` more specific. Something like
```python
Kwargs = dict[str, Any]
oneOrMore[T] = Union[T, list[T]]

class RocCurveKwargs(TypedDict):
	splits: oneOrMore[Kwargs]
	average: Optional[Kwargs]
```

glemaitre:
@foster999 The design change quite a lot since this PR has been open.

However, I'm convinced that the threshold average is actually what we should have here. It is the only way to have a correspondence between the PR and the ROC curve.

@foster999 do you wish to update the PR. Otherwise, I can take over.

foster999:
@glemaitre if the design is more settled now I would be keen to take another look!

I'll hopefully have chance during the week

foster999:
@glemaitre I've updated this given the new package structure. It's not currently implemented for PR curves, but would be simple to add if it would be useful?

skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -1086,14 +1099,121 @@ def roc(
...
+                report_type="cross-validation",
+            )
+            return average_display
+        else:
+            split_indices = range(len(self._parent.reports_))
+            display = RocCurveDisplay._concatenate(
+                child_displays,
+                report_type=self._parent._report_type,
+                column_data={"split": list(split_indices)},
+            )
```
auguste-probabl: can you put the simple case first?

foster999: Of course, updated now 👍

skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -1086,14 +1099,118 @@ def roc(
...
+            return display
 
+        if average != "threshold":
+            raise ValueError('average must be "threshold" or None.')
+        if self._parent._ml_task != "binary-classification":
+            raise ValueError(
+                "Averaging is not implemented for multi class classification"
+            )
+
```
auguste-probabl, line 1117:
```suggestion
            raise NotImplementedError(
                "Averaging is not implemented for multiclass classification"
            )
```

auguste-probabl (changes requested):
The PR is looking great, thanks very much!

The only thing is I'm getting warnings when I run this with pandas 2:
```py
>>> from sklearn.ensemble import RandomForestClassifier
>>> from skore import evaluate
>>> import sklearn
>>>
>>> X, y = sklearn.datasets.make_classification(n_samples=1000, return_X_y=True)
>>>
>>> model = RandomForestClassifier()
>>> report = evaluate(model, X, y, splitter=5)
>>> report.metrics.roc(average="threshold").plot()

# /home/auguste/Desktop/work/skore/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py:1183: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
#   roc_curve_df = pd.concat([roc_curve_df, avg_curve_df], ignore_index=True)
# /home/auguste/Desktop/work/skore/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py:1190: FutureWarning: The behavior of DataFrame concatenation with empty or all-NA entries is deprecated. In a future version, this will no longer exclude empty or all-NA columns when determining the result dtypes. To retain the old behavior, exclude the relevant entries before the concat operation.
#   roc_auc_df = pd.concat([roc_auc_df, avg_auc_df], ignore_index=True)
```

There are no warnings with pandas 3.

PR #1751, opened 2025-05-23
feat(plots)!: Remove the ability to pass `ax`

auguste-probabl:
Starting from https://github.com/probabl-ai/skore/pull/1669, it is sometimes possible to draw plots onto (a `ndarray` of) several Axes rather than just a single Axes. In fact, starting from https://github.com/probabl-ai/skore/pull/1720, the same data might be drawn on one or the other depending on user parameters.

In this context, allowing the user to pass `ax` might add too much complexity compared to how useful it might be; it is considered an advanced use-case which is difficult to motivate.

This PR thus removes the ability to pass user-created `ax`.

### DIFF
--- a/examples/use_cases/plot_employee_salaries.py
+++ b/examples/use_cases/plot_employee_salaries.py
@@ -262,21 +262,11 @@ comparator.metrics.report_metrics(
 )
 
 # %%
-# Finally, we can even get the individual :class:`~skore.EstimatorReport` for each fold
-# from the cross-validation to make further analysis.
-# Here, we plot the actual vs predicted values for each fold.
-from itertools import zip_longest
+# Finally, we can even get a deeper understanding by analyzing each fold in the
+# :class:`~skore.CrossValidationReport`.
+# Here, we plot the actual-vs-predicted values for each fold.
 import matplotlib.pyplot as plt
 
-fig, axs = plt.subplots(ncols=2, nrows=3, figsize=(12, 18))
-for split_idx, (ax, estimator_report) in enumerate(
-    zip_longest(axs.flatten(), linear_model_report.estimator_reports_)
-):
-    if estimator_report is None:
-        ax.axis("off")
-        continue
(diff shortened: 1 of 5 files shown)


thomass-dev (dismissed):
LGTM, for technical POV.

Keep in mind that such changes can impact projects plugins, and can be considered as breaking-change.
As long as we haven't e2e tests, it can be hard to detect.

Currently we use `display.figure_` in `skore-hub-project`, so there is no impact to remove `ax`🙏🏻 .

auguste-probabl:
Looks like one of our examples depends on this, however this use-case can be addressed using `CrossValidationReport.metrics.prediction_error()`.

auguste-probabl:
> Keep in mind that such changes can impact projects plugins, and can be considered as breaking-change. As long as we haven't e2e tests, it can be hard to detect.

Note that the PR is tagged as breaking (`feat(...)!`)

glemaitre:
@auguste-probabl you need to update the examples as well.

@thomass-dev It is breaking change for the user but for the plugin, it should be fine because we internally still expose `disp.ax_` as well. We just reduce the API surface to constrain the user-side.

glemaitre (commented):
LGTM. We only needs to fix the example.

glemaitre (dismissed):

auguste-probabl:
> @auguste-probabl you need to update the examples as well.

It's in progress

thomass-dev:
> @thomass-dev It is breaking change for the user but for the plugin, it should be fine because we internally still expose `disp.ax_` as well. We just reduce the API surface to constrain the user-side.

Yes, that's what I said 👍🏻  : for the plugin, there's no impact, but we should be careful with such changes without e2e testing.
During my first iteration on `skore-hub-project`, i passed an `ax` that i managed to the metrics.

glemaitre (approved):

glemaitre:
Bypassing the rule. The failure HTTP 403 is a known issue from the figshare. Let's not retrigger the build documentation once again.

PR #1753, opened 2025-05-23
docs: Update conf.py

sylvaincom:
In https://github.com/probabl-ai/skore/pull/1717, we forgot to update `conf.py` since we deleted the examples in the skore project section

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -69,7 +69,6 @@ subsections_order = [
     "../examples/getting_started",
     "../examples/use_cases",
     "../examples/model_evaluation",
-    "../examples/skore_project",
     "../examples/technical_details",
 ]
 


.gitignore
```diff
@@ -186,6 +186,7 @@ tramp
 sphinx/build/
 sphinx/auto_examples/
 sphinx/reference/api/
+sphinx/api/
```
glemaitre, line 189: It is not normal to have this folder. Looking at the content of the folder, it is because I mess up the path in one of the rst file. Could you modify it as per the following:


```diff
diff --git a/sphinx/reference/project.rst b/sphinx/reference/project.rst
index 9246fc14..f32f5420 100644
--- a/sphinx/reference/project.rst
+++ b/sphinx/reference/project.rst
@@ -14,7 +14,7 @@ These functions and classes are meant for managing a Project and its reports.
 .. rubric:: Methods
 
 .. autosummary::
-   :toctree: ../api/
+   :toctree: api/
    :template: class_methods_no_index.rst
 
     Project.put
@@ -22,7 +22,7 @@ These functions and classes are meant for managing a Project and its reports.
 .. rubric:: Reports
 
 .. autosummary::
-   :toctree: ../api/
+   :toctree: api/
    :template: autosummary/accessor.rst
 
    Project.reports
```

Right now, I would suspect some cross-reference to be broken because of the wrong location.

thomass-dev:

PR #1754, opened 2025-05-24
docs: Document the skore.project.metadata.Metadata class

glemaitre:
Right now, we don't document `skore.project.metadata.Metadata`.

In short, it means that no one is aware of the API of the output of `project.reports.metadata()` and what to do with it.

I avoid to list the inherited methods from the pandas dataframe since the import in this case would not work. Instead I modify the documentation to refer to the original documentation in the docstring.

### DIFF
--- a/skore/src/skore/project/metadata.py
+++ b/skore/src/skore/project/metadata.py
@@ -18,25 +18,21 @@ class Metadata(DataFrame):
     """
     Metadata and metrics for all reports persisted in a project at a given moment.
 
-    A metadata object is an extended ``pandas.DataFrame``, containing all the metadata
-    and metrics of the reports that have been persisted in a project. It implements a
-    custom HTML representation, that allows user to filter its reports using a parallel
-    coordinates plot. In a Jupyter Notebook, this representation automatically replaces
-    the standard ``pandas.DataFrame`` one and displays an interactive plot.
+    A metadata object is an extended :class:`pandas.DataFrame`, containing all the
+    metadata and metrics of the reports that have been persisted in a project. It
+    implements a custom HTML representation, that allows user to filter its reports
+    using a parallel coordinates plot. In a Jupyter Notebook, this representation
+    automatically replaces the standard :class:`pandas.DataFrame` one and displays an
+    interactive plot.
 
     The parallel coordinates plot is interactive, such the user can select a filter path
     and retrieve the corresponding reports.
 
     Outside a Jupyter Notebook, the metadata object can be used as a standard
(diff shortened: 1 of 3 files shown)


skore/src/skore/project/metadata.py
```diff
@@ -18,25 +18,21 @@ class Metadata(DataFrame):
...
-    using the standard ``pandas.DataFrame`` functions.
+    :class:`pandas.DataFrame` object. This means that it can be questioned, divided
+    etc., using the standard :class:`pandas.DataFrame` functions.
 
-    Methods
-    -------
-    reports(filter=True) -> list[skore.sklearn.EstimatorReport]
-        Return the reports referenced by the metadata object from the project.
-        If a query string selection exists, it will be automatically applied before, to
-        filter the reports to return. Otherwise, returns all reports.
```
thomass-dev, line 39: This method is important. Why removing it from the class documentation?

thomass-dev (changes requested):

glemaitre:
Because sphinx already do this job for you. It will list all methods and properties of a class on the html documentation.

sphinx/reference/project.rst
```diff
@@ -33,3 +36,26 @@ These functions and classes are meant for managing a Project and its reports.
...
+allow you filter and retrieve the reports.
+
+.. autosummary::
+   :toctree: api/
+   :template: class_without_inherited_members.rst
+
+   project.metadata.Metadata
+
+.. autosummary::
+   :toctree: ../api/
```
glemaitre, line 54:
```suggestion
   :toctree: api/
```

sphinx/reference/project.rst
```diff
@@ -33,3 +36,26 @@ These functions and classes are meant for managing a Project and its reports.
 
    Project.reports.get
    Project.reports.metadata
+
+Skore project's metadata
+^^^^^^^^^^^^^^^^^^^^^^^^
+
+When calling :meth:`Project.reports.metadata`, a :class:`Metadata` object is returned.
+This object is a :class:`pandas.DataFrame` with a specific HTML representation to
+allow you filter and retrieve the reports.
```
glemaitre, line 45:
```suggestion
When calling :meth:`Project.reports.metadata`, a
:class:`~skore.project.metadata.Metadata` object is returned. This object is a
:class:`pandas.DataFrame` with a specific HTML representation to allow you filter and
retrieve the reports.
```

thomass-dev:
> Because sphinx already do this job for you. It will list all methods and properties of a class on the html documentation.

But `help(report)` is impacted by this removing.

PR #1767, opened 2025-05-26
docs: Update plot_skore_getting_started.py

sylvaincom:


examples/getting_started/plot_skore_getting_started.py
```diff
@@ -265,15 +265,14 @@
 # =======================
 #
 # Another key feature of skore is its :class:`~skore.Project` that allows us to store
-# and retrieve items of many types.
+# and retrieve skore reports.
```
MarieSacksick, line 268:
```suggestion
# and retrieve EstimatorReports.
```

for now, it's only estimator reports.

examples/getting_started/plot_skore_getting_started.py
```diff
@@ -287,66 +286,116 @@
...
-report_get[0].metrics.report_metrics(pos_label=1)
+# %%
+reports_get[0].metrics.timings()
 
-# sphinx_gallery_start_ignore
-temp_dir.cleanup()
-# sphinx_gallery_end_ignore
+# %%
+# But what if instead of having stored only 2 estimators reports, we had a few
+# hundreds?
```
MarieSacksick, line 340: even before a few hundreds, with only dozens it will be useful!

MarieSacksick (commented):
The title of this example is `getting started`. All the metadata stuff seems quite complex to me, and not (yet?) core to be presented in the very first notebook.

sylvaincom:
Hi @MarieSacksick, thanks for the feedback, I'll look into it.

Please note that the title here is "getting started" and not "quick start", there is another example for "quick start"

examples/getting_started/plot_skore_getting_started.py
sylvaincom: Sure, I'll make some changes

examples/getting_started/plot_skore_getting_started.py
MarieSacksick: well, its misleading, as for now, we don't support the other types.

MarieSacksick:
sorry I mixed the two names; the idea is the same: this metadata thing has nothing to do in an introductive example.

sylvaincom:
Ok sure, in the getting started guide, I'll just leave the put and the retrieval without going into details 

And I'll create a separate notebook in a skore project section about the search feature

Sounds good?

MarieSacksick:
What do you mean by retrieval? since the `get` now requires an id, how do you plan to document it?

sylvaincom:
I'd keep the current subsection "Retrieving our stored reports" but making it shorter ; I'm talking about this subsection:



The subsection about "Searching through our stored reports" would be removed from the getting started and appear in a separate new example about the skore project

Issue #1626, opened 2025-04-30
feat(navigation): Iterate to add pipeline to parallel coordinate plot


MarieSacksick:
### Is your feature request related to a problem? Please describe.

We would like to be able to extend the navigation with parallel coordinate plots with pipelins.

### Describe the solution you'd like

Pipelines are slides in "groups", and each group is a coordinate.

waridrox:
Hi @MarieSacksick, can I work on this? 
Maybe we can proceed something along the following lines:
- Create `Group(s)` like `Stages` and `Metrics`.
- Within each `Group`, we can create individual coordinates for each pipeline stage or metric as suggested.
- Further, we can visually separate these groups with divider lines and labels when plotting to cater to the parallel plot implementation.

What are your thoughts on this?
```text
                    Pipeline Comparison: Stages and Metrics

┌───────────────┬─────────┬────────────┐ │ ┌──────────────────────────┐
│ STAGES GROUP  │         │            │ │ │      METRICS GROUP       │
│               │         │            │ │ │                          │
├───────────────┼─────────┼────────────┤ │ ├────────┬────────┬────────┤
│ Standard      │ PCA     │ Classifier │ │ │Accuracy │Precision│ ROC AUC│
│ Scaler        │         │            │ │ │         │         │        │
└───────┬───────┴───┬─────┴──────┬─────┘ │ └───┬─────┴─────┬───┴────────┘
        │            │            │      │     │           │
 Pipeline 1 ─────────┴────────────┴──────┼─────┼───────────┴───────────┐
        │            │            │      │     │                       │
 Pipeline 2 ─────────┴────────────┴──────┼─────┴───────────────────────┤
        │            │            │      │                             │
 Pipeline 3 ─────────┴────────────┘      └─────────────────────────────┘
```

thomass-dev:
> is this issue ready to work upon

It's not, sorry. It depends on pull requests that are actually WIP.

MarieSacksick:
not planned anymore.

PR #1771, opened 2025-05-26
feat: Split pipeline in substeps

MarieSacksick:
As first step of the issue https://github.com/probabl-ai/skore/issues/1626, to split the pipeline in sub-steps.

### DIFF
--- a/skore/src/skore/project/metadata.py
+++ b/skore/src/skore/project/metadata.py
@@ -49,8 +49,7 @@ class Metadata(DataFrame):
 
         Notes
         -----
-        This function is not intended @staticmethod
-        for direct use. Instead simply use the accessor
+        This function is not intended for direct use. Instead simply use the accessor
         :meth:`skore.Project.reports.metadata`.
         """
         metadata = DataFrame(project.reports.metadata(), copy=False)


skore/src/skore/project/metadata.py
```diff
@@ -37,6 +38,109 @@ class Metadata(DataFrame):
...
+            type_ = type(step)
+            if type_ == Pipeline:
+                itemized_pipeline = Metadata._explore_sk_pipeline(
+                    step, itemized_pipeline, depth=depth + 1
+                )
+            elif type_ == FeatureUnion:
+                itemized_pipeline = Metadata._explore_sk_feature_union(
+                    step, itemized_pipeline, depth=depth + 1
+                )
+            elif str(type_) == "<class 'skrub._table_vectorizer.TableVectorizer'>":
```
MarieSacksick, line 71: I'm not sure it's how it should be done, but I see no other way without importing skrub, which we don't want to have mandatory in our imports.

skore/src/skore/project/metadata.py
```diff
@@ -37,6 +38,116 @@ class Metadata(DataFrame):
...
+        type_ = type(step)
+        if type_ == Pipeline:
+            itemized_pipeline = Metadata._explore_sk_pipeline(
+                step, itemized_pipeline, depth=depth + 1
+            )
+        elif type_ == FeatureUnion:
+            itemized_pipeline = Metadata._explore_sk_feature_union(
+                step, itemized_pipeline, depth=depth + 1
+            )
+        elif str(type_) == "<class 'skrub._table_vectorizer.TableVectorizer'>":
```
MarieSacksick, line 65: I'm not sure it's how it should be done, but I see no other way without importing skrub, which we don't want to have mandatory in our imports.

skore/tests/unit/project/test_metadata.py
```diff
@@ -298,3 +301,12 @@ def mock_update_selection(*args, **kwargs):
             "or (log_loss >= 0.550000 and log_loss <= 0.550000)) "
             "and learner.isin(['learner1', 'learner3'])"
         )
+
```
auguste-probabl, line 304: can you add a test for the skrub case? use the [coverage comment](https://github.com/probabl-ai/skore/pull/1771#issuecomment-2911720568) to see which lines are not covered.

skore/src/skore/project/metadata.py
```diff
@@ -37,6 +40,135 @@ class Metadata(DataFrame):
...
+
+        sklearn_regex = re.compile(r"sklearn\.(\w+)\.(\w+)")
+        skrub_regex = re.compile(r"skrub\.(\w+)\.(\w+)")
+
+        # pipe_steps: dict[str, list[str]] = dict()
+        pipe_steps = defaultdict(list)
+        for estimator_class in class_names:
+            sklearn_match = sklearn_regex.search(estimator_class)
+            skrub_match = skrub_regex.search(estimator_class)
+            if sklearn_match is not None:
```
auguste-probabl, line 161: nit: This would be the use-case for the walrus (`:=`) operator:
```suggestion
            if (sklearn_match := sklearn_regex.search(estimator_class)) is not None:
```

skore/src/skore/project/metadata.py
```diff
@@ -37,6 +40,135 @@ class Metadata(DataFrame):
...
+        type_ = type(step)
+        if type_ == Pipeline:
+            itemized_pipeline = Metadata._explore_sk_pipeline(
+                step, itemized_pipeline, depth=depth + 1
+            )
+        elif type_ == FeatureUnion:
+            itemized_pipeline = Metadata._explore_sk_feature_union(
+                step, itemized_pipeline, depth=depth + 1
+            )
+        elif str(type_) == "<class 'skrub._table_vectorizer.TableVectorizer'>":
```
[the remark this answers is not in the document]
thomass-dev: Made the same comment 🤣 .

auguste-probabl:
> I think that we can simplify the implementation [by using get_params]

That sounds like a good idea. `get_params` also accepts a `deep` parameter to recover nested estimators!

skore/src/skore/project/metadata.py
```diff
@@ -37,6 +40,135 @@ class Metadata(DataFrame):
...
+            sklearn_match = sklearn_regex.search(estimator_class)
+            skrub_match = skrub_regex.search(estimator_class)
+            if sklearn_match is not None:
+                module, class_name = sklearn_match.groups()
+                pipe_steps[module].append(class_name)
+            elif skrub_match is not None:
+                module, class_name = skrub_match.groups()
+                pipe_steps[module].append(class_name)
+            else:
+                pipe_steps["other"].append(estimator_class)
```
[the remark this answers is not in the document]
MarieSacksick: indeed, it's cleaner!

skore/src/skore/project/metadata.py
MarieSacksick: This is the typical use-case. But I find it more difficult to read.

skore/tests/unit/project/test_metadata.py
MarieSacksick: done! Also one for homemade estimators.

skore/src/skore/project/metadata.py
```diff
@@ -2,9 +2,12 @@
```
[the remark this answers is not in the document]
MarieSacksick: It was the goal to split the feature in several parts to avoid having a PR too long to review actually. I agree with you that in the mean time, it somehow means that we have some dead code, and this part of the features turns out to be quite light.  
@glemaitre will take over in another PR for the widget. As this PR is useless without the next step, we can avoid merging it before the next one is ready, if it seems better @thomass-dev.

skore/pyproject.toml
```diff
@@ -45,6 +45,7 @@ test-base = [
   "pytest-randomly",
   "pytest-xdist",
   "scikit-learn<1.7",
+  "skrub",
```
auguste-probabl, line 48: Can you clarify why this is necessary? skrub is already in our test dependencies

MarieSacksick: Is it? I found it in the dependencies for documentation, but not in tests.  
Before adding it in commit https://github.com/probabl-ai/skore/pull/1771/commits/db115549fb2c7ca297c46c874fef3e577ebd4646, the CI was failing at pytest step, with error `ModuleNotFoundError: No module named 'skrub'`, as if it was not in the test dependencies yet.   
So I must say I'm puzzled.

skore/src/skore/project/metadata.py
```diff
@@ -37,6 +40,55 @@ class Metadata(DataFrame):
 
     _metadata = ["project"]
 
+    @staticmethod
+    def _find_estimators(pipeline: Pipeline) -> dict[str, list[str]]:
```
thomass-dev: Nitpick: 

This function seems to be useful even outside of the metadata module, even for the user.
Why not put it in dedicated module in `skore/sklearn`?

thomass-dev: With this move, i will find it normal to merge the PR.

skore/src/skore/project/metadata.py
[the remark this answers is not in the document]
thomass-dev: See https://github.com/probabl-ai/skore/pull/1771#discussion_r2120551416.

skore/src/skore/project/metadata.py
MarieSacksick: I took the example of _find_ml_task rather than _train_test_split, as find_estimator seems quite straightforward.

MarieSacksick: done!

skore/src/skore/project/metadata.py
```diff
@@ -49,7 +49,8 @@ def factory(project, /):
 
         Notes
         -----
-        This function is not intended for direct use. Instead simply use the accessor
+        This function is not intended @staticmethod
+        for direct use. Instead simply use the accessor
```
thomass-dev: Can you revert please? :)

thomass-dev (approved):

glemaitre:
Actually, I don't think that the strategy defining the key is right. I think that we should maintain ourself a specific list of keys such as "scaler", "encoder", etc. and have either a rule parsing the name that add the estimators in those categories or a dictionary defining the categories.

Parsing based on the name of the module is going to be brittle and I don't see it maintainable on the long run.

MarieSacksick:
@glemaitre we planned an issue for next sprint to define the new specification https://github.com/probabl-ai/skore/issues/1809. Let's discuss it on Wednesday.

PR #1783, opened 2025-05-29
feat (EstimatorReport): Add plotly as plotting backend

### LINKED ISSUE #1079: feat(skore): Add plotly as plotting backend for `Display.plot`

*This was also requested in a user interview. A DS consultant needs plots with interaction to answer questions from clients during a presentation.*

`EstimatorReport` has some plotting features, but only in matplotlib for now. Hence, in the skore UI, the display would not be neat.

We should add plotly as a plotting backend. Kind of what Pandas allows: [Pandas Plotting Backend in Python](https://plotly.com/python/pandas-backend/).

_No response_

### Additional context

_No response_

Muhammad-Rebaal:
Closes #1079

Hi @sylvaincom !

I've implemented `Plotly` as an alternative plotting backend for `EstimatorReport` . The implementation adds global backend configuration functions and Plotly versions of visualization displays while maintaining compatibility with all report types and preserving existing Matplotlib functionality.  Let me know that the PR is in the right direction and looking forward to enhance it more under your guidance.

Thank You!


skore/src/skore/sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -115,23 +134,21 @@ class PrecisionRecallCurveDisplay(
     def __init__(
         self,
         *,
-        precision: dict[Any, list[ArrayLike]],
-        recall: dict[Any, list[ArrayLike]],
-        average_precision: dict[Any, list[float]],
-        estimator_names: list[str],
```
auguste-probabl, line 121: These lines should not change, I think your branch is out-of-date

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -1744,6 +1746,7 @@ def roc(
         X: Optional[ArrayLike] = None,
         y: Optional[ArrayLike] = None,
         pos_label: Optional[PositiveLabel] = None,
+        backend: Optional[PlottingBackend] = None,
```
auguste-probabl, line 1749: `backend` should be a parameter of each Display's `plot()` method instead, since it is only involved when computing the actual plot.

auguste-probabl (changes requested):
Thanks for taking on this work! You need to reset your branch because there are too many changes that shouldn't be there.

Muhammad-Rebaal:
Hi @auguste-probabl !
Thanks for reviewing , will open a new PR soon which doesn't go under conflicts because it would be quicker to raise a new clean one instead of cleaning it which would result more conflicts
Thank You !

PR #1785, opened 2025-05-31
feat: Rename estimator_reports_ to reports_ in CrossValidationReport

### LINKED ISSUE #1760: Rename `estimator_reports_` in `CrossValidationReport`

I think that we could align the API of the `CrossValidationReport` and the `ComparisonReport`. It means that in `CrossValidationReport`, we should rename `estimator_reports_` into `reports_`.

I found myself having to switch some code when switching between the two reports while it would have been nice to not have to do so.

Muhammad-Rebaal:
Closes #1760 

Hi @MarieSacksick , @glemaitre  !

I've renamed `estimator_reports_` to `reports_` throughout the `CrossValidationReport` class, aligning the API with `ComparisonReport`.


MarieSacksick:
hello @Muhammad-Rebaal!  
It looks like there are some places where you have to switch `estimator_reports_` to `reports_` (cf the failing tests). Can you complete the changes please?

Muhammad-Rebaal:
> hello @Muhammad-Rebaal! It looks like there are some places where you have to switch `estimator_reports_` to `reports_` (cf the failing tests). Can you complete the changes please?

Thanks @MarieSacksick for the feedback, 
I've updated it with all the test passed

image

glemaitre:
I could reproduce the failure.

xref: https://github.com/probabl-ai/skore/issues/1790

It looks like it is related to the progress bar not being cleaned properly in the previous stage and we still have a lock that cannot be pickled.

glemaitre:
Actually the failure is legit. It is related to the code of the progress bar:

https://github.com/probabl-ai/skore/blob/2e4a94966abdec7cfaea5b760c8f2ed56442b984/skore/src/skore/utils/_progress_bar.py#L69-L75

Before renaming this attribute, we were not assigning a progress bar to the the `EstimatorReport` and thus did not have a lock. Now, we do that. I have to think about the problem but right now, we should detect if we are going to execute something in a parallel context leading to a pickling issue.

glemaitre (changes requested):
A temporary work around would be to add a parameter `allow_nested` to the progress bar decorator and to not allow the processing when we know that we have an underlying parallel processing.

skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -227,21 +228,30 @@ def _compute_metric_scores(
...
+                    results.append(result)
+                    progress.update(main_task, advance=1, refresh=True)
+            except (pickle.PicklingError, TypeError):
+                # Fall back to sequential execution if parallel fails
+                results = []
+                for report in self._parent.reports_:
+                    method = getattr(report.metrics, report_metric_name)
+                    result = method(data_source=data_source, X=X, y=y, **metric_kwargs)
+                    results.append(result)
+                    progress.update(main_task, advance=1, refresh=True)
```
auguste-probabl: There could be PicklingErrors or TypeErrors for other reasons, so I'd like to avoid `try-catch` here. Can you implement the solution suggested in https://github.com/probabl-ai/skore/pull/1785#pullrequestreview-2890906620?

Muhammad-Rebaal: Yeah sure ! Thanks for reviewing ☺️

skore/tests/unit/sklearn/cross_validation/test_cross_validation.py
```diff
@@ -954,20 +954,26 @@ def test_cross_validation_report_custom_metric(binary_classification_data):
...
+        captured = capsys.readouterr()
+        assert all(word in captured.out for word in error_message.split(" "))
+
+        # Also run the metrics inside the same context to maintain threading
+        result = report.metrics.custom_metric(
+            metric_function=accuracy_score,
+            response_method="predict",
+        )
+
+    # Assertions should work outside the context
```
auguste-probabl: why does this test have to change?

Muhammad-Rebaal: Actually the original code was encountering a pickling error when using `joblib's` default parallel backend. As `joblib` backend uses `multiprocessing` which requires pickling.
Whereas `threading` backend uses threads instead of processes, avoiding the need for pickling. If you'd say to put it back so I'd happy to do so.

glemaitre: Oups, I should have explained better:

```diff
diff --git a/skore/src/skore/utils/_progress_bar.py b/skore/src/skore/utils/_progress_bar.py
index c1c1df77..4fe7012e 100644
--- a/skore/src/skore/utils/_progress_bar.py
+++ b/skore/src/skore/utils/_progress_bar.py
@@ -16,6 +16,7 @@ DescriptionType = Union[str, Callable[..., str]]
 
 def progress_decorator(
     description: DescriptionType,
+    allow_nested: bool = True,
 ) -> Callable[[Callable[..., T]], Callable[..., T]]:
     """Decorate class methods to add a progress bar.
 
@@ -28,6 +29,11 @@ def progress_decorator(
         The description of the progress bar. If a callable, it should take the
         self object as an argument and return a string.
 
+    allow_nested : bool, default=True
+        Whether to allow nested progress bars. Disabling progress bars is useful when
+        the nested reports will need to be pickled and that we cannot pass the
+        progress bar containing a lock.
+
     Returns
     -------
     decorator : function
@@ -68,7 +74,7 @@ def progress_decorator(
 
             # assigning progress to child reports
             reports_to_cleanup: list[Any] = []
-            if hasattr(self_obj, "reports_"):
+            if allow_nested and hasattr(self_obj, "reports_"):
                 for report in self_obj.reports_:
                     if hasattr(report, "_progress_info"):
                         report._progress_info = {"current_progress": progress}
```

We would this patch and turn `allowed_nested=False` when we have an underlying parallel processing that will try to pickle the `EstimatorReport` and thus make the pickling of the parent progress bar fail.

So we need to had this patch, test it in the unit test and then make sure to change the decorator for the case that it was failing in the documentation. Also, we should had a non-regression test in the test file that trigger the problem (which not only rely on the documentation execution).

Muhammad-Rebaal: Hi @glemaitre !

No Worries, I've updated it and thanks for the detailed review.

glemaitre: Could you also add unit test in `test_progress_bar.py` for testing the behaviour that you just added and add a non-regression test that trigger the error.

Muhammad-Rebaal: Yeah sure.

Muhammad-Rebaal:
Hi @glemaitre , @MarieSacksick !

I'm closing this PR as it gets a little messy and it would take a lot of time to review for you as a maintainer , also I added the requested features with this new PR [#1838](https://github.com/probabl-ai/skore/pull/1838)

Thank You !

Issue #1761, opened 2025-05-25
`Metadata.reports` should allow to construct a `ComparisonReport`


glemaitre:
### What would you like to say?

Right now, the API of `Metadata.reports` is to return a list of `EstimatorReport`. We could extend the API such that a single `ComparisonReport` is provided.

We should introduce a parameter to select the behaviour. Maybe something like `return_as="list"/"dict'/"comparison"`?

glemaitre:
> Is your intention to provide a comparison report object? 

Yes.

> In that case, maybe comparator is better than comparison?

Yep it aligned more with the sense that it an object. We could be be sepecific and request the string "ComparisonReport" also. I don't have strong opinion right now.

sylvaincom:
What if, in the stored reports, there are classifiers and regressors (more generally, estimators applied to different datasets)? Then, I believe we should return a list of comparison reports.

PR #1791, opened 2025-06-03
feat(skore): Allow `Summary.reports` to return comparison report

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/1761.


skore/tests/unit/project/test_metadata.py
```diff
@@ -168,10 +168,34 @@ def test_reports_empty(self):
...
+
+        assert isinstance(comparison, ComparisonReport)
+        assert comparison.reports_ == [regression1, regression2]
+
+    def test_reports_exception(self, regression, binary_classification):
+        with raises(
+            RuntimeError,
+            match="Bad condition: it is not a valid `Metadata` object.",
+        ):
             Metadata([{"<column>": "<value>"}]).reports()
```
auguste-probabl: make it a new test

thomass-dev: Done in https://github.com/probabl-ai/skore/pull/1791/commits/08575168ccce6ce365fa40f47b0273f4a1c5d429.

auguste-probabl (approved):
Ok for me!

skore/src/skore/project/metadata.py
```diff
@@ -92,7 +100,21 @@ def reports(self, *, filter: bool = True) -> list[EstimatorReport]:
         if filter and (querystr := self._query_string_selection()):
             self = self.query(querystr)
 
-        return list(map(self.project.reports.get, self.index.get_level_values("id")))
+        reports = map(self.project.reports.get, self.index.get_level_values("id"))
```
glemaitre: Since you always need a list for the processing later.

```suggestion
        reports = [
            self.project.reports.get(idx) for idx in self.index.get_level_values("id")
        ]
```

skore/src/skore/project/metadata.py
```diff
@@ -92,7 +100,21 @@ def reports(self, *, filter: bool = True) -> list[EstimatorReport]:
...
+
+            if len(datasets) > 1:
+                raise RuntimeError(
+                    f"Bad condition: the comparison mode is only applicable when "
+                    f"reports have the same dataset.\n"
+                    f"Found '{datasets}'.\n"
+                    f"Please query the dataframe or use the widget to make your "
+                    f"selection."
+                )
+            return ComparisonReport(list(reports))
```
glemaitre: The comparison report will already fail on this so I would prefer to let the `ComparisonReport` raise the error. Potentially, the `ComparisonReport` can raise additional error when making the check. If really the error is not easy to get then, we can do a reraise.

```suggestion
            try:
                return ComparisonReport(reports)
            except ValueError as exc:
                raise RuntimeError(
                    f"Bad condition: the comparison mode is only applicable when "
                    f"reports have the same dataset.\n"
                    f"Found '{datasets}'.\n"
                    f"Please query the dataframe or use the widget to make your "
                    f"selection."
                ) from exc
```

glemaitre: Also not sure that `RuntimeError` is actually the right type here.

skore/src/skore/project/metadata.py
```diff
@@ -82,6 +88,8 @@ def reports(self, *, filter: bool = True) -> list[EstimatorReport]:
         filter : bool, optional
```
glemaitre:
```suggestion
        filter : bool, default=True
```

skore/src/skore/project/metadata.py
```diff
@@ -82,6 +88,8 @@ def reports(self, *, filter: bool = True) -> list[EstimatorReport]:
...
             Filter the reports to return with the user query string selection, default
             True.
+        return_as : Literal["list", "comparison"], optional
```
glemaitre:
```suggestion
        return_as : Literal["list", "comparison"], default="list"
```

skore/tests/unit/project/test_metadata.py
```diff
@@ -1,20 +1,22 @@
...
 from sklearn.model_selection import train_test_split
-from skore import EstimatorReport
 from skore.project.metadata import Metadata
+from skore.sklearn import ComparisonReport, EstimatorReport
 
 
 @fixture
-def regression():
-    X, y = make_regression(random_state=42)
+def regression(random_state=42):
```
glemaitre: It is not clear to me where you are using a different `random_state` in the code below?

glemaitre (commented):
It looks good. I would probably amend an example where we get the report to use this feature directly.

skore/src/skore/project/metadata.py
thomass-dev: It is not needed in the case where we reach the exception.

skore/src/skore/project/metadata.py
thomass-dev: > The comparison report will already fail on this

Initially, I had let the constructor raise its exception, but I thought that the error was not explicit enough in this configuration. I thought it was important to say `Please query the dataframe or use the widget to make your selection.`. What do you think?

> `if "ML usecase" in str(exc):`

Honestly, i think this is pretty bad. If you want me to follow this way, i'd rather specialize the error raised by `ComparisonReport` than base the behavior on the content of the exception.

skore/tests/unit/project/test_metadata.py
thomass-dev: I can revert this change. It was needed during my iteration, but it's no longer necessary.
I found the changes useful, so I let them.

skore/src/skore/project/metadata.py
thomass-dev: Both are allowed in the numpydoc convention, but i can change 😄 .

skore/src/skore/project/metadata.py
thomass-dev: Both are allowed in the numpydoc convention, but i can change 😄 .

glemaitre: Arff, yes. Here, I'm using what we do in scikit-learn (https://scikit-learn.org/dev/developers/contributing.html#guidelines-for-writing-docstrings) but feel free to ignore if is not consistent with the rest.

skore/tests/unit/project/test_metadata.py
glemaitre: Yep, let's keep the diff minimum and revisit when we need it.

skore/src/skore/project/metadata.py
glemaitre: > I thought it was important to say Please query the dataframe or use the widget to make your selection.. What do you think?

So let's reraise the error to have the full traceback in this case.

glemaitre: > Honestly, i think this is pretty bad.

Looking at the underlying error raised, we could avoid the `if` statement because any `ValueError` that can be triggered by the `ComparisonReport` from the interaction with the metadata would be related to not have the same dataset (either task or either different `y_test`). So a global reraise would provide the underlying error and the error at the metadata level.

skore/src/skore/project/metadata.py
glemaitre: Not if I delegate everything to the `ComparisonReport` :). To me, it is nice to have the fine-grain error of the `ComparisonReport` and the global reraised one.

thomass-dev: Yes, i will go this way!

skore/src/skore/project/metadata.py
thomass-dev: Fine, i can go this way with only a try/catch ValueError.

skore/tests/unit/project/test_metadata.py
thomass-dev: Fixed in 29805422157ce4917b613105fd568cb77290d83f.

skore/src/skore/project/metadata.py
thomass-dev: Just a remark that i have already made in the past: 

`ValueError` is a generic exception than can be raised by underlying functions of `ComparisonReport`.
In a next iteration we should specialize our explicitly raised exceptions.

thomass-dev: Fixed in 574e54d2fc77cc92e63ad81640cbee482665ff52.

glemaitre (approved):

glemaitre:
LGTM. Thanks @thomass-dev

Issue #1798, opened 2025-06-05
Add a helper entry point to disconnect from hub


rouk1:
### Is your feature request related to a problem? Please describe.

As it can be a bit cumbersome to find where your token file is stored, it could be nice to have an entrypoint that helps user to logout from the hub.

### Describe the solution you'd like

A `skore-hub-logout` entrypoint that deletes user's token.

PR #1799, opened 2025-06-05
feat(skore-hub-project): Add `skore-hub-logout` entry-point to remove cached token

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/1798.

### DIFF
--- a/skore-hub-project/tests/unit/authentication/test_logout.py
+++ b/skore-hub-project/tests/unit/authentication/test_logout.py
@@ -1,0 +1,14 @@
+from skore_hub_project.authentication.logout import logout
+from skore_hub_project.authentication.token import Token
+
+
+def test_logout(nowstr):
+    token = Token("A", "B", nowstr)
+
+    assert token.valid
+    assert token.filepath.exists()
+
+    logout()
+
+    assert token.valid
+    assert not token.filepath.exists()
--- a/skore-hub-project/pyproject.toml
+++ b/skore-hub-project/pyproject.toml
@@ -36,6 +36,7 @@ test-lts-cpu = ["skore-hub-project[test-base]", "polars-lts-cpu"]
 
 [project.scripts]
 skore-hub-login = "skore_hub_project.authentication.login:login"
+skore-hub-logout = "skore_hub_project.authentication.logout:logout"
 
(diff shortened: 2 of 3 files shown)


skore-hub-project/tests/unit/authentication/test_logout.py
```diff
@@ -0,0 +1,14 @@
...
+
+def test_logout(self, nowstr):
+    token = Token("A", "B", nowstr)
+
+    assert token.valid
+    assert token.filepath.exists()
+
+    logout()
+
+    assert token.valid
```
rouk1, line 13: Isn't it a bit surprising that a token that is supposed to be deleted is still valid ?

rouk1 (commented):
Thanks for the quick implementation. It is useful.

skore-hub-project/tests/unit/authentication/test_logout.py
thomass-dev: IMO no. On the contrary, I think it's great.
At runtime, i don't think we want to crash if a third party deleted the cached token.
The token should remain in RAM (as currently).

Do you want me to change the behavior, so that each time the token is called, we checked for the cached file?

thomass-dev: Imagine this workflow : 

User A, Terminal 1 -> skore-hub-login
User A, Terminal 1 -> run long process
User A, Terminal 2  -> skore-hub-logout

I don't think terminal 2 should have an impact on terminal 1, until the long process finished.

rouk1: TBH I dont know and indeed as the file is stored un tmp I agree with your point, but when explicitly logout it is really unexpected to still be authenticated because the token leaves in RAM.

I said that before but to me tmp is not a nice place to store auth tokens. It should be in app data hence never deleted by a third party that is not a human.

thomass-dev: Token is cached only for user convenience between two runs of its scripts.
After a laptop restart, we should enforce the user to re-login.
This is why the token is cached in `/tmp`.

rouk1: Okay !

Issue #1795, opened 2025-06-05
Revisit the API of the `skore.Metadata`


glemaitre:
### What would you like to say?

Right now, to access some report, we need to write:

```python
project.reports.metadata().reports()
```

Having twice `reports` is weird. I still feel that the API discussed in https://github.com/probabl-ai/skore/issues/1269 was nice even more with the current limitation of the `put`/`get`.

```python
report = project.metatada().reports()
```

thomass-dev:
@glemaitre so we can "just" remove the first namespace for now?

auguste-probabl:
Even just `metadata` is a strange name IMO: it's a very technical name that suggests the mechanic is for advanced users, which is not the case AFAIK. `project.reports()` or something like that would be preferable.

thomass-dev:
Maybe we can use `summary` as for `report_metrics` in https://github.com/probabl-ai/skore/issues/1794.

PR #1801, opened 2025-06-05
feat(skore): Change the API `project.reports.metadata` to `project.summary`

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/1795.


glemaitre:
It came to my mind that a verb would be better suited than a noun since it is a function. So, `project.summarize()` would be better.

So maybe we can slightly think a bit more about the naming to know if `summarize` is really the intended action that we perform to get the metadata. I think it is not bad I'll just take a couple of minutes to think of alternative.

thomass-dev:
> It came to my mind that a verb would be better suited than a noun since it is a function.

You're true! Totally agree. `def summarize(self) -> Summary` looks good.
Waiting for your return before making the change.

glemaitre:
In short, I have a feeling that `summarize` would trigger a computation to be able to do the summary while `stocktake` and `Stock` is just a listing.

Basically, since @auguste-probabl is a native speaker, I would take the one that he finds the most appropriate and we move on with it. I am already happy with both suggestions compared to the current state.

sphinx/reference/project.rst
```diff
@@ -9,54 +9,40 @@ Skore project
...
    :template: class_methods_no_index.rst
 
-   project.metadata.Metadata.reports
+   project.summary.Summary.reports
 
 .. note::
-   This class :class:`~skore.project.metadata.Metadata` is not meant to be used
-   directly. Instead, use the accessor :meth:`Project.reports.metadata`.
+   This class :class:`~skore.project.summary.Summary` is not meant to be used
+   directly. Instead, use the accessor :meth:`Project.summarize`.
```
glemaitre: Actually small nitpick.

```suggestion
   directly. Instead, use the method :meth:`Project.summarize`.
```

thomass-dev: Well spotted, fix it.

thomass-dev:
Ready to merge @glemaitre (do not update with main; use the merge queue 😄 ).

PR #1803, opened 2025-06-05
chore(skore): Extend test coverage

MarieSacksick:


skore/tests/unit/sklearn/estimator/test_estimator.py
```diff
@@ -917,26 +922,26 @@ def test_estimator_report_report_metrics_indicator_favorability(
...
+    "scoring, scoring_names, scoring_kwargs",
     [
-        ("accuracy", None),
-        ("neg_log_loss", None),
-        (accuracy_score, {"response_method": "predict"}),
-        (get_scorer("accuracy"), None),
+        ("accuracy", "accuracy", None),
+        ("neg_log_loss", "neg_log_loss", None),
+        (accuracy_score, "accuracy", {"response_method": "predict"}),
+        (get_scorer("accuracy"), "accuracy", None),
```
glemaitre: Could you add a case where, the `scoring_name` is "toto" (or whatever else). Basically, having a really different name make sure that we did not silently used the name of the scorer internally. Since that they are aligned with the metric here, it is not obvious when reading the test.

MarieSacksick: Updated and precised!

glemaitre (approved):
LGTM. Thanks @MarieSacksick

Issue #811, opened 2024-11-25
FEAT: enhance skrub's TableReport


sylvaincom:
### Is your feature request related to a problem? Please describe.

Following #806, we can improve the TableReport from skrub and have something more thorough in skore.

### Describe the solution you'd like

Have something like:
```python
skore.analyze_table(df, project="my_project", prefix="/my_data")
```
that separately puts in the skore UI (in the skore UI better than output of notebook because it clutters with output graphs and code cells):
- skrub Table Report
    - they choose not to use javascript because they want minimal dependencies, but we can do something nicer
- seaborn pairplot
- what y-data profiling does: percentage of missing values
- in `/my_data`, there is table analysis, correlations, column 1, etc
- compare different datasets

rouk1:
> * he skrub team chose not to use javascript because they want minimal dependencies, but we can do something nicer

FYI Skrub table report actually has [1000+ lines of js](https://github.com/skrub-data/skrub/blob/main/skrub/_reporting/_data/templates/report.js) ; )

sylvaincom:
Could be a step of #834

sylvaincom:
Got this from a user interview:

Make it possible to customize the report table as the data varies depending on the industry, whether it pertains to insurance or travel, the analysis approach differs; in some domains, the average is not relevant for example

glemaitre:
Now, that we have the different report, I think a first step is to compute the `TableReport` via a new accessor. We should make some tests to have a good API there but we could have:

My thoughts is to create a display. Internally, we might want to call the summarize function and not the `TableReport` itself.

Discussing with @Vincent-Maladiere, we thought that the API could be something like:

```python
display = report.data.analyze(data_souce="train"/"test"/"all")
display  # Show the HTML representation of the `TableReport`
display.plot(kind="distribution", feature=...)
```

Vincent-Maladiere:
Below are some additional feature ideas that could be included with the table report display.

- Plot multiple single distributions, which we already have in the TableReport:
	```python
	display.plot(kind="distribution_1d", cols=...)
	```
- Plot pairwise distributions, which is currently not supported in TableReport, between features and the target:
	```python
	display.plot(kind="distribution_2d", cols=...)
	```
- Plotting associations heatmap:
	```python
	display.plot(kind="pearson", cols=...)
	display.plot(kind="cramer", cols=...)
	```
	This would be close to what users would get with a regular `sns.heatmap(X.corr())`, but with our metrics.

[...]

Issue #1757, opened 2025-05-25
api(skore): `ConfusionMatrix` is not following the Display API


glemaitre:
### What would you like to say?

The class `ConfusionMatrix` is not following the Display API:

- It should inherit from the style and help mixin
- It should not inherit from the protocol `Display`
- It should not expose a `from_predictions` method
- It should be added in the API reference

glemaitre:
> if I understand it right, instead of the Display protocol, would you prefer adding base classes like StyleDisplayMixin and HelpDisplayMixin?

Yes indeed.

> And instead of using the from_predictions, we create a ConfusionMatrixDisplay instance directly using the confusion matrix from sklearn? [Ref](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.ConfusionMatrixDisplay.html)

Nop because the display in scikit-learn will not work with the `CrossValidationReport` and `ComparisonReport`. So we need to have a similar design as the `RocCurveDisplay` in term of a private class method to build the display. I would delay a bit this refactoring because @auguste-probabl is almost done refactoring the displays in #1669 #1701 #1709 to accommodate all type of reports. Once those merged, it will be much easier to go forward with a straightforward API.

GaetandeCast:
To align with the API of the other displays, I believe we need to change `ConfusionMatrix` in the following way:
- initialize with a method `_compute_data_for_display()` instead of `from_predictions()`
- adapt `_MetricsAccessor.confusion_matrix()` to use `_compute_data_for_display()`
- add `ConfusionMatrixDisplay` to the API reference in the visualization Displays
- add `_plot_single_estimator`, `_plot_cross_validated_estimator`, `_plot_comparison_estimator` and `_plot_comparison_cross_validation` to accommodate all types of report.

From what I can see, all other Display classes currently also inherit from `DisplayMixin` so I assume this should not be changed for `ConfusionMatrixDisplay`.

@glemaitre

PR #1805, opened 2025-06-05
feat(skore): Add a data accessor

Vincent-Maladiere:
Fixes https://github.com/probabl-ai/skore/issues/811

I played around with the API. This is a first simple POC for the data accessor using TableReport machinery.

```python

X, y = load_breast_cancer(return_X_y=True, as_frame=True)
X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=0)

rf = RandomForestClassifier(random_state=0)
rf_report = EstimatorReport(
    rf, X_train=X_train, X_test=X_test, y_train=y_train, y_test=y_test, pos_label=1
)
rf_report.data.analyze()
```

~However, I can't get `rf_report.help()` to display the data accessor. `rf_report.data.help()` looks good, though.~

Ping @glemaitre


skore/src/skore/sklearn/_estimator/data_accessor.py
```diff
@@ -0,0 +1,117 @@
...
+
+from skore.externals._pandas_accessors import DirNamesMixin
+from skore.sklearn._base import _BaseAccessor
+from skore.sklearn._estimator.report import EstimatorReport
+from skore.sklearn._plot import TableReportDisplay
+
+pandas_version = parse_version(parse_version(pd.__version__).base_version)
+
+
+# TODO: remove when skrub 0.7.0 is released
```
glemaitre: I'm thinking that we should do a `_skrub_compat` file for those types of thing. It will come back time to time.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,49 @@
...
+        str :
+            The HTML snippet.
+        """
+        return to_html(
+            self.summary,
+            standalone=False,
+            column_filters=self.column_filters,
+        )
+
+    def _repr_mimebundle_(self, include=None, exclude=None):
```
glemaitre: Let's do a mixin for those like in scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_repr_html/base.py#L118-L152

We are going to reuse it many time I think for the different display.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,49 @@
...
+
+from skore.sklearn._plot.base import Display
+
+
+class TableReportDisplay(Display):
+    def __init__(self, summary, column_filters=None):
+        self.summary = summary
+        self.column_filters = column_filters
+
+    @classmethod
```
glemaitre: Let's keep the factory private. For consistency, let's call it `_compute_data_for_display`

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,49 @@
+from skrub._reporting._html import to_html
+from skrub._reporting._summarize import summarize_dataframe
+
+from skore.sklearn._plot.base import Display
+
+
+class TableReportDisplay(Display):
```
glemaitre: Actually, you don't inherit from a Python `Protocol` ;). It is only useful to do something like `isinstance(table_report, Display)` without forcing inheritance and just checking the availability of methods.

So here, you need to do:

```suggestion
class TableReportDisplay(StyleDisplayMixin, HelpDisplayMixin):
```

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,49 @@
...
+    @classmethod
+    def from_frame(cls, df, with_plots=True, title=None):
+        summary = summarize_dataframe(
+            df,
+            with_plots=with_plots,
+            title=title,
+        )
+        return cls(summary)
+
+    def plot(self, kind=None):
```
glemaitre: To allow the `set_style` method as well:

```suggestion
    @StyleDisplayMixin.style_plot
    def plot(self, kind=None):
```

glemaitre (commented):
I think it is going in the right direction. Just a couple of comments regarding code organization.

skore/src/skore/sklearn/_plot/data/table_report.py
Vincent-Maladiere: Ok, got it, but note that `ConfusionMatrixDisplay` inherits from `Display`

glemaitre: Yep I open some time ago this issue: https://github.com/probabl-ai/skore/issues/1757

Vincent-Maladiere:
This is getting quite large, I'm stopping there on the feature side and will switch gears to the documentation (and tests).

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,130 @@
...
+            c_col,
+        )
+
+    @StyleDisplayMixin.style_plot
+    def plot_pearson(self):
+        return _plotting.plot_pearson(self.df)
+
+    @StyleDisplayMixin.style_plot
+    def plot_cramer(self):
+        return _plotting.plot_cramer(self.df)
```
glemaitre: I would prefer a single `plot` method as in `pandas`.

Up to now, we have a single occurrence for the `PredictionErrorDisplay` that has a `kind` parameter to switch the type of plot. It works nicely because we don't have some specific parameters for one kind or another. Here, I assume it will be different.

```python
display.plot(kind="distribution")
```

will have `x`, `y`, or `c` parameters but they are not useful for others. So it makes the documentation hard to follow and parameters difficult to discover. I really like the part of pandas that do:

```python
display.plot.distribution(x=...)
```

We are never going to go one level deeper but I find that in terms of documentation, discoverability, it makes just the experience nicer.

@Vincent-Maladiere WDYT?

glemaitre:
I'll give a detail review on the code itself. In terms of feature, I think that it is enough right now. Any other thing should be done in subsequent PRs.

skore/src/skore/sklearn/_plot/data/table_report.py
Vincent-Maladiere: I agree that consistency is of prime importance.

That said, my hunch is that `plot_something` can be discovered via `display.help()` and autocompletion, which show all plot options nicely. It's also shorter to write, and that's what I ended up preferring when writing it over and over for testing.

I'm slightly less in favour of `display.plot.distribution`, which complicates things a little bit, unless we foresee other features for the display. It makes sense to have this plot accessor in pandas DataFrame, but I have less certainty for the `display` class.

So overall, my preferences would be:
1. `display.plot_something(x=...)`
2. `display.plot(kind=...)`
3. `display.plot.distribution(x=...)`

Vincent-Maladiere:
I added an example but I guess you'll have to run it yourself since there is no doc build preview?

glemaitre:
The documentation preview is available here: https://github.com/probabl-ai/skore/pull/1805#issuecomment-2945159885.

It is updated at each push (when the sphinx CI succeed).

Vincent-Maladiere:
Almost ready to review, missing tests

skore/src/skore/sklearn/_plot/data/table_report.py
Vincent-Maladiere: As discussed, I went for option 2

skore/src/skore/sklearn/_plot/data/_plotting.py
```diff
@@ -0,0 +1,419 @@
+import itertools
+
+import numpy as np
+import pandas as pd
+import seaborn as sns
```
glemaitre: `skrub` and `seaborn` will be two new dependencies. I think that we should include them in `pyproject.toml` and then we need to update the lock file for the CI with `bash pip-compile.sh skore --upgrade`

skore/src/skore/sklearn/_plot/data/_plotting.py
```diff
@@ -0,0 +1,419 @@
...
+
+from skore.skrub._skrub_compat import is_in
+
+_RED = "#dd0000"
+_BLUE = "#4878d0"
+_ORANGE = "#ee854a"
+_TEXT_COLOR_PLACEHOLDER = "#123456"
+
+
+def _despine(ax):
```
glemaitre: We have already a package for small utils like this one in `skore/sklearn/_plot/utils.py`. I think that all the small tweak that can be applied to a matplotlib axis could be centralized there. In the current module, we could keep the core functions that make plotting given the data.

skore/src/skore/sklearn/_plot/data/_plotting.py
```diff
@@ -0,0 +1,419 @@
...
+import seaborn as sns
+from matplotlib import pyplot as plt
+from skrub import _column_associations
+from skrub import _dataframe as sbd
+from skrub._reporting import _utils
+
+from skore.skrub._skrub_compat import is_in
+
+_RED = "#dd0000"
+_BLUE = "#4878d0"
```
glemaitre: It is not used. I'm wondering if we should not just use `tab:blue`, `tab:orange`, and `tab:red` to be inline with all other plots.

skore/src/skore/sklearn/_plot/data/_plotting.py
```diff
@@ -0,0 +1,419 @@
...
+    ax.set_yticklabels(list(map(str, values)))
+    if title is not None:
+        ax.set_title(title, fontsize=16)
+
+    _adjust_fig_size(fig, ax, 7.0, 0.4 * len(values))
+    return fig
+
+
+def scatter(x_col, y_col, c_col):
+    fig, ax = plt.subplots(dpi=150)
```
glemaitre: I don't think that we should be creating any plots here. Everything should be created in the `.plot` method of the display, and we just passing the parameters.

skore/src/skore/skrub/_skrub_compat.py
```diff
@@ -0,0 +1,73 @@
+import pandas as pd
+from sklearn.utils.fixes import parse_version
+from skrub import _dataframe as sbd
+from skrub import _join_utils
+from skrub._dataframe._common import dispatch
+
+pandas_version = parse_version(parse_version(pd.__version__).base_version)
+
+
+# TODO: remove when skrub 0.7.0 is released
```
glemaitre: The idea is to not care about the removal and to have something like:

```python
skrub_version = parse_version(parse_version(skrub.__version__).base_version)

if skrub_version < parse_version("1.7"):
    # all the code that we have in the file
    ...
else:
    from skrub import ...  # noqa: F401
```

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,196 @@
...
+    def plot(
+        self,
+        x_col=None,
+        y_col=None,
+        c_col=None,
+        kind="dist",
+        top_k_categ=20,
+        n_subsample=None,
+        subsample_strategy="head",
+        random_state=None,
```
glemaitre: In terms of API, those should go in the `analyze` method but not in the `plot` because they modify the computation. The `plot` method is restrained only at modifying the visual of the visualization. `PredictionErrorDisplay` have the same behaviour. It means that one should interact as:

```python
report.metrics.analyze(subsample=1_000, random_state=10)
```

glemaitre: In addition, I would rename `n_subsample` to `subsample` and `random_state` to `seed` (we don't accept anything else that integer for caching purpose).

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,196 @@
...
+            title=title,
+        )
+        return cls(summary, df)
+
+    @StyleDisplayMixin.style_plot
+    def plot(
+        self,
+        x_col=None,
+        y_col=None,
+        c_col=None,
```
glemaitre: I would stick to the pandas/seaborn API and name `x`, `y`, `hue`

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,196 @@
...
+
+        y_col : str, default=None
+            The name of the column to use for the y-axis of the plot. Only used when
+            ``kind='dist'``.
+
+        c_col : str, default=None
+            The name of the column to use for the color or hue axis of the plot. Only
+            used when ``kind='dist'``.
+
+        top_k_categ : int, default=20
```
glemaitre: I would rename `top_k_categories` to have the full name.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,196 @@
...
+        summary = summarize_dataframe(
+            df,
+            with_plots=with_plots,
+            title=title,
+        )
+        return cls(summary, df)
+
+    @StyleDisplayMixin.style_plot
+    def plot(
+        self,
```
glemaitre: Keywords only are good :)

```suggestion
        self,
        *,
```

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,196 @@
...
+              distribution. The random seed is controlled by ``random_state``.
+
+        random_state : int, default=None
+            The random seed to use when randomly subsampling. It only has an effect when
+            ``n_subsample`` is not Noen and ``subsample_strategy='random'``.
+
+        Return
+        ------
+        fig : matplotlib.figure.Figure
+            The drawn figure.
```
glemaitre: The API of plot does not return a figure. A user can access via the attribute `axis_` and `figure_` once `plot` has been called once.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,196 @@
...
+
+        c_col : str, default=None
+            The name of the column to use for the color or hue axis of the plot. Only
+            used when ``kind='dist'``.
+
+        top_k_categ : int, default=20
+            For categorical columns, the number of most frequent elements to display.
+            Only used when ``kind='dist'``.
+
+        kind : {'dist', 'pearson', 'cramer'}, default='dist'
```
glemaitre: I'm wondering if we need Pearson and Cramer or if because Cramer cover all cases, then it is far enough. If it is the case, we could rename the two options with name that represents what we display: a marginal distribution or a pairwise distribution. So maybe renaming with `kind="marginal"` and `kind="pairwise"`.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,196 @@
...
+    @classmethod
+    def _compute_data_for_display(cls, df, with_plots=True, title=None):
+        summary = summarize_dataframe(
+            df,
+            with_plots=with_plots,
+            title=title,
+        )
+        return cls(summary, df)
+
+    @StyleDisplayMixin.style_plot
```
glemaitre: So @Vincent-Maladiere, you can have a look at the `StyleDisplayMixin` class but in short it will allow to add a `set_style` method by taking `whatever_kwargs` parameters. So in short, in `plot`, we could create a `scatter_kwargs` or any other `_kwargs` method which allow to tweak the plots. I did not yet go into details in the `_plotting` module but it might give you some perspective of changes that can be done.

glemaitre (commented):
Here is a first round of review regarding the higher level API on the `analyze` and `plot`. We can make a back and forth on this one and then I will go in details for the other part.

skore/src/skore/skrub/_skrub_compat.py
Vincent-Maladiere: right, makes more sense indeed

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+        return self._heatmap(
+            cramer_v_table,
+            title="Cramer's V",
+            heatmap_kwargs=heatmap_kwargs,
+        )
+
+    def frame(self):
+        return self.summary
+
+    def html_snippet(self):
```
glemaitre: I would prefer that for th moment we remain conservative in terms of API in the first step and to keep this function private.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+        str :
+            The HTML snippet.
+        """
+        return to_html(
+            self.summary,
+            standalone=False,
+            column_filters=self.column_filters,
+        )
+
+    def html(self):
```
glemaitre: Let's not implement this function for the moment. I would like that we tackle all display at once and expose a consistent API.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+            column_filters=self.column_filters,
+        )
+
+    def _html_repr(self, include=None, exclude=None):
+        return self.html_snippet()
+
+    def __repr__(self):
+        return f"<{self.__class__.__name__}: use .open() to display>"
+
+    def open(self):
```
glemaitre: We can therefore also remove this function for the moment.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+        y_col=y_col,
+        c_col=c_col,
+    )
+    for k, v in params.items():
+        if v is not None:
+            raise ValueError(f"When {kind=!r}, this function takes no {k!r} argument.")
+
+
+class TableReportDisplay(StyleDisplayMixin, HelpDisplayMixin, ReprHTMLMixin):
+    """Display a report of a dataset as an HTML table.
```
glemaitre:
```suggestion
    """Display reporting information about a given dataset.
```

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+
+    This display summarizes the dataset and provides a way to visualize
+    the distribution of its columns.
+
+    Parameters
+    ----------
+    summary : pandas DataFrame
```
glemaitre:
```suggestion
    summary : DataFrame
```

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+        The summary of the dataset, as returned by ``summarize_dataframe``.
+
+    df : pandas or polars DataFrame
```
glemaitre: I'm thinking that we should call this one dataset. Also, even if we support `polars` as a side-effect, let's not document it for the moment since we are not consistent. I think that in other location we only either write `DataFrame` or `pd.DataFrame`. We should probably improve the consistency here. You can use `DataFrame`.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+
+    _default_heatmap_kwargs: dict[str, Any] | None = None
+    _default_boxplot_kwargs: dict[str, Any] | None = None
+    _default_scatterplot_kwargs: dict[str, Any] | None = None
+    _default_stripplot_kwargs: dict[str, Any] | None = None
+
+    def __init__(self, summary, df, column_filters=None):
+        self.summary = summary
+        self.column_filters = column_filters
+        self._df = weakref.ref(df)
```
glemaitre: I have second thought right now and I'm thinking that `self.dataset = dataset` would be enough for the moment. We should also revisit later the issue. But right now, there is nothing in `skore` communication with the `hub` that will lead to keep a reference.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+
+        y : str, default=None
+            The name of the column to use for the y-axis of the plot. Only used when
+            ``kind='dist'``.
+
+        hue : str, default=None
+            The name of the column to use for the color or hue axis of the plot. Only
+            used when ``kind='dist'``.
+
+        kind : {'dist', 'cramer'}, default='dist'
```
glemaitre: I would prefer "dist" and "corr" or "distribution" and "correlation" (I always have a preference for fully qualified name).
We can explain that the correlation is based on Cramer's in the docstring below as what you did.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+            The kind of plot drawn.
+
+            - If ``'dist'``, plot a distribution parametrized by ``x_col``, ``y_col``
```
glemaitre: You might want to correct x_col y_col and c_col

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+            the distribution 2D plot, when both ``x_col`` and ``y_col`` are numeric.
+
+        stripplot_kwargs: dict, default=None
+            Keyword arguments to be passed to seaborn's ``stripplot`` for rendering
+            the distribution 2D plot, when either ``x_col`` or ``y_col`` is numeric, and
+            the other is categorical. This plot is drawn on top of the boxplot.
+
+        boxplot_kwargs: dict, default=None
+            Keyword arguments to be passed to seaborn's ``boxplot`` for rendering
+            the distribution 2D plot, when either ``x_col`` or ``y_col`` is numeric, and
```
glemaitre: same

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+            For categorical columns, the number of most frequent elements to display.
+            Only used when ``kind='dist'``.
+
+        scatterplot_kwargs: dict, default=None
+            Keyword arguments to be passed to seaborn's ``scatterplot`` for rendering
+            the distribution 2D plot, when both ``x_col`` and ``y_col`` are numeric.
+
+        stripplot_kwargs: dict, default=None
+            Keyword arguments to be passed to seaborn's ``stripplot`` for rendering
+            the distribution 2D plot, when either ``x_col`` or ``y_col`` is numeric, and
```
glemaitre: same

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+            - If ``'cramer'``, plot Cramer's V correlation among all columns. This
+              option doesn't take any ``x_col``, ``y_col`` or ``c_col`` argument.
+
+        top_k_categories : int, default=20
+            For categorical columns, the number of most frequent elements to display.
+            Only used when ``kind='dist'``.
+
+        scatterplot_kwargs: dict, default=None
+            Keyword arguments to be passed to seaborn's ``scatterplot`` for rendering
+            the distribution 2D plot, when both ``x_col`` and ``y_col`` are numeric.
```
glemaitre: same here about x_col and y_col

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,672 @@
...
+
+        boxplot_kwargs: dict, default=None
+            Keyword arguments to be passed to seaborn's ``boxplot`` for rendering
+            the distribution 2D plot, when either ``x_col`` or ``y_col`` is numeric, and
+            the other is categorical. This plot is drawn below the stripplot.
+
+        heatmap_kwargs: dict, default=None
+            Keyword arguments to be passed to seaborn's ``heatmap`` for rendering
+            the Cramer's V correlation matrix, when ``kind='cramer'`` or when
+            ``kind='dist'`` and both ``x_col`` and ``y_col`` are categorical.
```
glemaitre: same

MarieSacksick:
Hello @Vincent-Maladiere, can you please sign your commits? (cf https://docs.skore.probabl.ai/0.9/contributing.html#signing-commits)

glemaitre (commented):
Really partial review. I'll continue tomorrow.

skore/src/skore/sklearn/_plot/data/table_report.py
Vincent-Maladiere: It's actually a dict, my bad

Vincent-Maladiere:
Ready for another round of review @glemaitre

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,656 @@
...
+            column_filters=self.column_filters,
+        )
+
+    def _html_repr(self, include=None, exclude=None):
+        return self._html_snippet()
+
+    def __repr__(self):
+        return f"<{self.__class__.__name__}: use .open() to display>"
+
+    def _json(self):
```
Vincent-Maladiere: @rouk1 WDYT?

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,657 @@
...
+        *,
+        x: str | None = None,
+        y: str | None = None,
+        hue: str | None = None,
+        kind: str = "dist",
+        top_k_categories: int = 20,
+        scatterplot_kwargs: dict[str, Any] | None = None,
+        stripplot_kwargs: dict[str, Any] | None = None,
+        boxplot_kwargs: dict[str, Any] | None = None,
+        heatmap_kwargs: dict[str, Any] | None = None,
```
glemaitre: I would like to have the option `despine` with a default value at True.

My expectation is that the `despine` will apply for the 1d distribution:

- for histogram, we should remove the top and right spines
- for the value count plot, we should remove, top, right, and left spines (and also the left ticks)

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,657 @@
...
+            for (p, c) in zip(percent, counts, strict=False)
+        ]
+        small_percent = [
+            p if c <= counts[-1] / 2 else ""
+            for (p, c) in zip(percent, counts, strict=False)
+        ]
+
+        # those are written on top of the orange bars so we write them in black
+        self.ax_.bar_label(
+            rects, large_percent, padding=-30, color="black", fontsize=14
```
glemaitre: Instead to set specifically the size here, could we modify one of the matplotlib style such that we can globally set it if we want. What I mean is to modify the dict in `skore/sklearn/_plot/style.py` and set this property and to not have to set it here.

skore/src/skore/sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,657 @@
...
+        )
+        # those are written on top of the background so we write them in foreground
+        # color
+        self.ax_.bar_label(
+            rects, small_percent, padding=5, color=_TEXT_COLOR_PLACEHOLDER, fontsize=14
+        )
+
+        self.ax_.set_yticks(self.ax_.get_yticks())
```
glemaitre: I would like to improve a small thing regarding the ticks:

- I want to be sure that the axis always start and end with a tick corresponding to a value (usually a rounded one).
- I would like to make sure that we don't show too many ticks: in short, I think that above 6 ticks, we have too much info on the axis.

Do you think that we can easily tweak those on the figure?

glemaitre: I would have the same remark for the histogram.

skore/src/skore/sklearn/_estimator/data_accessor.py
```diff
@@ -0,0 +1,116 @@
...
+        )
+
+
+class _DataAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
+    def __init__(self, parent: EstimatorReport) -> None:
+        super().__init__(parent)
+
+    def analyze(
+        self,
+        source_dataset: str = "all",
```
auguste-probabl: Should be `data_source` for consistency

examples/model_evaluation/plot_dataset_display.py
```diff
@@ -0,0 +1,97 @@
...
+data = fetch_employee_salaries()
+X_train, X_test, y_train, y_test = train_test_split(data.X, data.y, random_state=0)
+
+pipeline = skrub.tabular_learner(HistGradientBoostingRegressor())
+pipeline
+
+# %%
+# We bring the dataset and pipeline into the report, and use ``.data.analyze()`` to get
+# our insights. ``dataset="all"`` means analyzing both train and test, and ``with_y=True``
+# include the target in the analysis.
```
sylvaincom:
```suggestion
# our insights. ``dataset="all"`` means analyzing both train and test sets, and ``with_y=True``
# includes the target in the analysis.
```

examples/model_evaluation/plot_dataset_display.py
```diff
@@ -0,0 +1,97 @@
...
+    X_train=X_train,
+    X_test=X_test,
+    y_train=y_train,
+    y_test=y_test,
+)
+display = report.data.analyze(source_dataset="all", with_y=True)
+display
+
+# %%
+# As usual, we can easy glance at the options using ``.help``:
```
sylvaincom:
```suggestion
# As usual, we can easily glance at the options using ``.help``:
```

examples/model_evaluation/plot_dataset_display.py
```diff
@@ -0,0 +1,97 @@
...
+# %%
+# The dataset is somewhat balanced, with a clear majority of males.
+# Next, we colorize this distribution using the salary to predict, in the column
+# ``current_annual_salary``.
+
+display.plot(x="gender", hue="current_annual_salary")
+
+# %%
+# Interestingly, we see that the median (the black vertical bar) is slightly higher for
+# males. We can also see there is an outlier at $300,000.
```
sylvaincom:
```suggestion
# Interestingly, we observe that the median (the black vertical bar) is slightly higher for
# males. We can also observe that there is an outlier at $300,000.
```

examples/model_evaluation/plot_dataset_display.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# Let's now add a third dimension to this plot by visualizing the hired year as the
+# y-axis (which becomes the x-axis since the plot is horizontal):
+display.plot(x="gender", y="year_first_hired", hue="current_annual_salary")
+
+# %%
+# The year has replaced the salary as the x-axis, and the salary is still represented by
+# the color. This plot is getting a bit hard to read due to the large number of data
+# points, we can subsample it slightly to see a pattern emerges:
```
sylvaincom:
```suggestion
# the color. As this plot is getting a bit hard to read due to the large number of data
# points, we can slightly subsample it to see that a pattern emerges:
```

examples/model_evaluation/plot_dataset_display.py
```diff
@@ -0,0 +1,97 @@
...
+report.data.analyze(source_dataset="all", with_y=True, subsample=1000).plot(
+    x="gender",
+    y="year_first_hired",
+    hue="current_annual_salary",
+)
+# %%
+# As expected, newcomers (at the right of the plot) are paid significantly less than
+# more senior employees (at the left of the plot). Let's now switch gears and observe
+# the correlation among our columns:
```
sylvaincom:
```suggestion
# the correlation among our columns.
```

examples/model_evaluation/plot_dataset_display.py
```diff
@@ -0,0 +1,97 @@
...
+#
+# Since Pearson's correlation is only defined between numerical columns and our
+# dataset contains mostly categorical columns, we're missing associations between
+# most of the columns.
+#
+# To get a broader view of our columns correlations, we can use another metric instead,
```
sylvaincom:
```suggestion
# To get a broader view of our columns' correlations, we can use another metric instead,
```

examples/model_evaluation/plot_dataset_display.py
```diff
@@ -0,0 +1,97 @@
...
+# the `Cramer's V correlation <https://en.wikipedia.org/wiki/Cram%C3%A9r%27s_V>`_,
+# whose interpretation is close to the Pearson's correlation:
```
sylvaincom:
```suggestion
# whose interpretation is close to the Pearson's correlation.
```

sylvaincom (commented):
Some minor suggestions to the example, mainly typos and grammar

skore/src/skore/sklearn/_plot/data/table_report.py
Vincent-Maladiere: Yeah so I think the easiest way forward would be to ditch all this skrub plotting and use seaborn. This way, we could control the parameters with more consistency. We can add some logic for robustness later. WDYT?

Vincent-Maladiere:
@glemaitre I removed all skrub custom plots and replaced them with seaborn: the logic is way simpler, and we control the style.

Vincent-Maladiere:
No strong opinion here, I also like being able to read an entire row without having to mentally concatenate a row + a column

glemaitre:
I'll make a new review. I'm bit stuck in the triage on the scikit-learn side.

sylvaincom:
> No strong opinion here, I also like being able to read an entire row without having to mentally concatenate a row + a column

Oh yeah, maybe reading from left to right is easier ; so we could fill only the upper right triangle

sylvaincom:
It would be nice to have the name of the feature (label) in the legend, for example in the [given example](https://skoredoc.k.probabl.dev/1805/auto_examples/model_evaluation/plot_dataset_display.html#sphx-glr-auto-examples-model-evaluation-plot-dataset-display-py) it would be nice to have `current_annual_salary` as the legend title:

Capture d’écran 2025-07-09 à 11 12 18

Vincent-Maladiere:
> It would be nice to have the name of the feature (label) in the legend, for example in the [given example](https://skoredoc.k.probabl.dev/1805/auto_examples/model_evaluation/plot_dataset_display.html#sphx-glr-auto-examples-model-evaluation-plot-dataset-display-py) it would be nice to have current_annual_salary as the legend title:

Fixed!

examples/use_cases/plot_employee_salaries.py
```diff
@@ -111,14 +113,48 @@
...
+# We also have access to some additional information regarding the dataset used for
+# training and testing of the model similar to what we have seen in the previous
+# section. For instance, let's check some information about the training dataset.
+train_data_display = hgbt_split_1.data.analyze(data_source="train")
+train_data_display
+
+# %%
+#
+# The display obtained allows to get a quick overview with the same HTML-based view
+# than the :class:`skore.TableReport` we have seen earlier. In addition, you can access
```
sylvaincom:
```suggestion
# than the :class:`skrub.TableReport` we have seen earlier. In addition, you can access
```

skore/src/skore/_sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,697 @@
...
+        return f"<{self.__class__.__name__}(...)>"
+
+    def _json(self) -> str:
+        """Serialize the data of this report in JSON format.
+
+        It is the serialization chosen to be sent to skore-hub.
+        """
+        to_remove = ["dataframe"]
+        data = {k: v for k, v in self.summary.items() if k not in to_remove}
+        return json.dumps(data, cls=JSONEncoder)
```
MarieSacksick: Why do we keep the sample_table here? It is dropped [in skrub](https://github.com/skrub-data/skrub/blob/main/skrub/_reporting/_table_report.py#L283). It triggers an error when there is timestamp data inside the table.   
Currently, we have a test that makes this function fails.

glemaitre: I need to check but I think that `sample_table`, is the preview table shown for the HTML representation and thus the information that Matthieu will need to show. So having them precomputed could be handy but I need to check more in details.

skore/src/skore/_sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,697 @@
...
+            raise ValueError(f"Invalid kind: {kind!r}")
+
+    def _html_repr(self) -> str:
+        """Show the HTML representation of the report."""
+        return to_html(self.summary, standalone=False, column_filters=None)
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__}(...)>"
+
+    def _json(self) -> str:
```
glemaitre: Could we name this function `_to_json`

skore/src/skore/_sklearn/_plot/data/table_report.py
glemaitre: So `sample_data` contains the data required to easily build the HTML snippet I think. I think that it could be nice to have @rouk1 in the loop to know what was required in the first version to rebuild the `TableReport` on the front-end.

We could drop it if not required. Otherwise the alternative is to solve the bug and make sure that `Timestamp` value get converted into a string when serializing (in isoformat for instance https://pandas.pydata.org/docs/reference/api/pandas.Timestamp.isoformat.html)

glemaitre: Be aware that this function is completely private for the moment and I would say that https://github.com/probabl-ai/skore-hub/issues/424 will defined the most adequate behaviour (it is the point 1 of the checklist). So we might merge without the `_to_json` or make it a TODO with minimal testing.

skore/tests/unit/sklearn/plot/table_report/test_estimator.py
```diff
@@ -0,0 +1,414 @@
...
+    with pytest.raises(ValueError, match=msg):
+        display.plot(x="gender", hue="division", top_k_categories=30)
+
+    display.plot(y="year_first_hired", hue="current_annual_salary")
+    assert display.ax_.get_xlabel() == "current_annual_salary"
+    assert display.ax_.get_ylabel() == "year_first_hired"
+    assert display.ax_.legend_.get_title().get_text() == "current_annual_salary"
+
+
+def test_1D_plot_with_duration_data(display):
```
glemaitre:
```suggestion
def test_1D_plot_with_duration_data(pyplot, display):
```

It is quite important to add the `pyplot` pytest fixture because it will be in charge of closing figure once the test is run. It avoids weird behaviour where figure are shown to developer when running the test suite.

So in general, if a test contains a `.plot` call, you need to add this `pyplot` in the signature.

skore/src/skore/_sklearn/_plot/data/table_report.py
```diff
@@ -0,0 +1,697 @@
...
+        if kind == "dataset":
+            return self.summary["dataframe"]
+        elif kind == "top-associations":
+            return pd.DataFrame(self.summary["top_associations"])
+        else:
+            raise ValueError(f"Invalid kind: {kind!r}")
+
+    def _html_repr(self) -> str:
+        """Show the HTML representation of the report."""
+        return to_html(self.summary, standalone=False, column_filters=None)
```
glemaitre, line 685: I think that it remains only this line to be covered. We could do similarly to https://github.com/skrub-data/skrub/blob/main/skrub/_reporting/tests/test_table_report.py#L52-L64 to check the output of this function.

glemaitre:
> Is it worth saying a few words about this data accessor in the user guide?

Yes, I think that we need to. We should rework slightly this page: https://docs.skore.probabl.ai/0.9/user_guide/reporters.html

Basically, I think that we need to add a section about "Data inspection" before the "Model evaluation" section. But somehow, the section about displays that are introduced in the "Model evaluation" seems odd then. I think that we need to move this introduction, being a bit more general above all to explain the Display API. Maybe, it could be OK to merge this PR to unlock the SPEC, and make a subsequent PR on the documentation with this reworking.

skore/src/skore/_sklearn/_plot/data/table_report.py
MarieSacksick: to continue this discussion, I opened a discussion in skore hub: https://github.com/probabl-ai/skore-hub/discussions/543

rouk1:
🎉

PR #1813, opened 2025-06-07
api(skore): Rename metrics.report_metrics to metrics.summarize

### LINKED ISSUE #1794: Rename `report_metrics` to `summarize`

It has been bugging me for some time to have to right something like:

```python
report.metrics.report_metrics(...)
```

The fact that we are repeating the same terms twice is weird. Speaking with @Vincent-Maladiere yesterday, an alternative name came to mind:

```python
report.metrics.summarize(...)
```

Which is really the role of this function. It would mean as well that we would create a `SummaryMetricsDisplay`.

glemaitre:
closes #1794 

Rename the `.metrics.report_metrics()` method to `.metrics.summarize()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,8 +93,8 @@ cv_report.help()
 
 ```python
 # Display the report metrics that was computed for you:
-df_cv_report_metrics = cv_report.metrics.report_metrics()
-df_cv_report_metrics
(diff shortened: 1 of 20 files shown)


MarieSacksick (approved):
LGTM!

PR #1816, opened 2025-06-09
feat: Add a plot to ComparisonReport summarize display

### LINKED ISSUE #2129: feat(skore): Add pairplot to show metric trade-off in `ComparisonReport`

In `ComparisonReport`, it is quite common to represent the trade-off between two metrics. Usually, we will be interested in the Pareto front in this plot.

Such example is:

Image

Since it is usually related to metrics, it looks like this should be a plot available under the `MetricSummaryDisplay`.

MarieSacksick:
The plot of this display will be a pair plot to visualize one metric against another.

This is part of the narrative "I have several models, how can I choose the best one?". 

Closes #2129


MarieSacksick:
tests to come; ready for a first high-level review.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
     This class should not be instantiated directly.
     """
 
-    def __init__(self, summarize_data):
+    # should be removed once transformed into a utils
```
glemaitre: Instead to redefine, let's create a variable called `verbose_metric_names` in `__init__`. It will be a dict mapping a name of a metric function to its verbose name.

This variable can be created from the metrics accessor by filtering the metric and metric name that were requested to be computed. So we should end-up with a dictionary string -> string.

It tells me that we have to test the following plot for all supported `scoring` cases (string, scorer, callable, etc.).

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+            {"report_a": report_a, "report_b": report_b}
+        )
+        >>> display = comparison_report.metrics.summarize()
+        >>> display.plot(x="accuracy", y="roc_auc")
         """
-        return self.summarize_data
+        if self.report_type in (
+            ["estimator", "cross-validation", "comparison-cross-validation"]
+        ):
+            raise NotImplementedError("To come soon!")
```
glemaitre: For "estimator", I'm pretty sure that we will never have any comparison. So you might want to dissociate the error message depending of the type of report.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+            self._plot_comparison_estimator(x, y)
 
-    @StyleDisplayMixin.style_plot
-    def plot(self):
-        raise NotImplementedError
+    def _plot_comparison_estimator(self, x, y):
+        self.figure_, self.ax_ = plt.subplots()
+
+        x_label = self._SCORE_OR_LOSS_INFO.get(x, {}).get("name", x)
+        y_label = self._SCORE_OR_LOSS_INFO.get(y, {}).get("name", y)
```
glemaitre: Since we are going to pass `verbose_metric_names`, we will have surely a match. So we just need to find verbose name in the dictionary.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+
+        # Check that the metrics are in the report
```
glemaitre: This check should be done at the start of `plot` because it is common to all plot.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+        # If the metric is not in the report, help the user by suggesting
+        # supported metrics
+        reverse_score_info = {
+            value["name"]: key for key, value in self._SCORE_OR_LOSS_INFO.items()
+        }
+        index = self.summarize_data.index
+        if isinstance(index, pd.MultiIndex):
```
glemaitre: Looking at this, I'm thinking that it would be beneficial to change the format of the dataframe similarly to what @auguste-probabl did in other PR meaning that we should not have any multi-index. Then, I have to see more but I think that the `.frame` should provide a bit of flexibility on the way to the dataframe should be structured.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+
+        self.ax_.scatter(x=x_data, y=y_data)
+        self.ax_.set_title(title)
+        self.ax_.set_xlabel(x_label_text)
+        self.ax_.set_ylabel(y_label_text)
+
+        # Add labels to the points with a small offset
+        text = self.summarize_data.columns.tolist()
+        for label, x_coord, y_coord in zip(text, x_data, y_data, strict=False):
+            self.ax_.annotate(
```
glemaitre: Looking at the rendering, we will want to have the option to put those as a legend outside of the plot instead of an annotation. It might a safer choice: one can avoid to show the legend and then tweak to have the annotation. Removing the annotation should be much more work.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+        if y == "fit_time" and self.data_source != "train":
+            y_label_text = y_label + " on train set"
+        else:
+            y_label_text = y_label
+
+        title = f"{x_label} vs {y_label}"
+        if self.data_source is not None:
+            title += f" on {self.data_source} set"
+
+        self.ax_.scatter(x=x_data, y=y_data)
```
glemaitre: I think that by default we should use a set of marker and colors to plot each individual data point.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+                xytext=(10, 0),
+                bbox=dict(
+                    boxstyle="round,pad=0.3",
+                    edgecolor="gray",
+                    facecolor="white",
+                    alpha=0.7,
+                ),
+            )
+
+        plt.tight_layout()
```
glemaitre: You will not need this line.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -9,19 +13,166 @@ class MetricsSummaryDisplay(HelpDisplayMixin, StyleDisplayMixin):
...
+            y_label_text = y_label
+
+        title = f"{x_label} vs {y_label}"
+        if self.data_source is not None:
+            title += f" on {self.data_source} set"
+
+        self.ax_.scatter(x=x_data, y=y_data)
+        self.ax_.set_title(title)
+        self.ax_.set_xlabel(x_label_text)
+        self.ax_.set_ylabel(y_label_text)
```
glemaitre: A more compact way to `set` stuff in matplotlib.

```suggestion
        self.ax_.set(title=title, xlabel=x_label_text, y_label=y_label_text)
```

glemaitre (commented):
It is a quick review. A couple of additional thoughts:

- we will want to have some logspace sometimes. In scikit-learn, we came with an heuristic with the time for the validation curve: https://github.com/scikit-learn/scikit-learn/blob/d4d4af8c4/sklearn/model_selection/_plot.py#L111-L117 I think that we can retake this implementation
- we need to put some thoughts regarding score vs. loss.
- I think that we need to tackle the following issue first: https://github.com/probabl-ai/skore/issues/1837

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
MarieSacksick: I don't really get why we would put this code into the MetricsAccessor, as it's only used in the display, and even only used in the plot of the display.  
Do you have something else in mind?

MarieSacksick: (however, this variable `verbose_metric_names` could be created in plot, and not only in `_plot_comparison_estimator`.)

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
MarieSacksick: Not exactly: for some plots, we will not have a y axis.  
I'd argue that we should do refactoring when we have the new code, and not before, as we don't know exactly what we will have in it.

MarieSacksick:
> we will want to have some logspace sometimes. In scikit-learn, we came with an heuristic with the time for the validation curve: https://github.com/scikit-learn/scikit-learn/blob/d4d4af8c4/sklearn/model_selection/_plot.py#L111-L117 I think that we can retake this implementation  

About this, it looks like an improvement to add this option imo. It could be a good first issue. Since we have a workshop soon, what keeping it for this workshop?  

> I think that we need to tackle the following issue first: https://github.com/probabl-ai/skore/issues/1837

If you think that this other issue can be done soon, let's wait indeed! Otherwise, I don't really like keeping PR open for a while, we tend to forget the code and the context.  

> we need to put some thoughts regarding score vs. loss.  

Do you have the pareto front in mind?  
As the first feedback, to me it's an improvement, not light this one, and I prefer to do step by step and releasing this first part when we are ready.

skore/src/skore/sklearn/_plot/metrics/metrics_summary_display.py
glemaitre: I think that it will depend if we succeed to remove any relationship between this dict and metrics accessor. I think the help is also using some part of it. I would prefer to have it in a single location to deduplicate it.

Ideally, it would be cool to have it only in the display.

glemaitre:
> About this, it looks like an improvement to add this option imo. It could be a good first issue. Since we have a workshop soon, what keeping it for this workshop?

It is 5 lines of code already tested somewhere else. I would not bother since I already know that we want it.

> Do you have the pareto front in mind?
As the first feedback, to me it's an improvement, not light this one, and I prefer to do step by step and releasing this first part when we are ready.

Not really, it was more on the direction of the loss/score. But I think we can just align the implementation with the display on the hub.

In terms of implementation, I don't like a partial implementation.

MarieSacksick:
@glemaitre, can you be more precise about your expectations about the following sentence please?

> Not really, it was more on the direction of the loss/score. But I think we can just align the implementation with the display on the hub.

So far, the only difference with the hub is the display of the pareto front and the colored aspect on the "worse" side of the front.

image

MarieSacksick:
Ready for a new round of review @glemaitre :) ! (still in draft because still waiting for your PR first).

PR #1817, opened 2025-06-09
test: Add data_source parameter tests for ROC curve and ComparisonReport

### LINKED ISSUE #1747: enh: Add tests for data_source parameter and the display of `ComparisonReport`

At least for ROC curve, it seems that we don't check the possibility to pass `data_source` while it is supported in `ComparisonReport`.

We should improve it.

sobiya-22:
Closes #1747

Hi @glemaitre , @MarieSacksick 
Added unit tests for:
- data_source parameter functionality in ROC curve
- ComparisonReport display verification


glemaitre:
@sobiya-22 Thanks for the PR but it is not really what we would like as a test.

To give an example, in `EstimatorReport`, we test the behaviour of `data_source` with the following test:

https://github.com/probabl-ai/skore/blob/main/skore/tests/unit/sklearn/plot/roc_curve/test_estimator.py#L113-L158

Here, I would expect to modify the files `skore/tests/unit/sklearn/plot/roc_curve/test_comparison_estimator.py` to add similar test for the case that we have a `ComparisonReport` built with `EstimatorReport`s to check that the `data_source` parameter has the expected behaviour.

We can limit the scope of this PR to this specific behaviour but I think that we need to extend the same test to the `test_comparison_cross_validation.py` file and most probably for the other type of visualization (PR curve and prediction-error).

sobiya-22:
Hi @glemaitre ,
Thank you for the feedback! I've now implemented the `data_source` parameter tests for ROC curve visualization in `test_comparison_estimator.py` and `test_comparison_cross_validation.py`

skore/tests/unit/sklearn/plot/roc_curve/test_comparison_estimator.py - Added test_data_source_binary_classification() and test_data_source_multiclass_classification()
skore/tests/unit/sklearn/plot/roc_curve/test_comparison_cross_validation.py - Added corresponding tests for CrossValidationReport scenarios

Both test files now include comprehensive coverage of the `data_source` parameter behavior ("train", "test", and "X_y" options) following the same pattern as the EstimatorReport tests you referenced.

Remaining task: 
PR curve visualization tests (precision_recall curve)
Prediction-error visualization tests

Before I proceed with implementing similar `data_source` tests for **PR curve** and **prediction-error visualizations**, could you please review the current ROC curve implementation to confirm I'm on the right track? I want to ensure the approach and test structure are correct before extending to the remaining visualization types.

Please let me know if this matches your expectations, and I'll proceed with the PR curve and prediction-error tests accordingly.
Thanks!

glemaitre:
> Before I proceed with implementing similar data_source tests for PR curve and prediction-error visualizations, could you please review the current ROC curve implementation to confirm I'm on the right track? 

Indeed, it should be implemented in some independent PRs to keep the scope focused. I'll make a review.

skore/tests/unit/sklearn/plot/roc_curve/test_comparison_cross_validation.py
```diff
@@ -232,3 +232,84 @@ def test_multiclass_classification_kwargs(pyplot, multiclass_classification_repo
...
+    estimator_2 = LogisticRegression(C=10)
+    
+    report = ComparisonReport(
+        reports={
+            "estimator_1": CrossValidationReport(estimator_1, X, y),
+            "estimator_2": CrossValidationReport(estimator_2, X, y),
+        }
+    )
+    
+    # Test with custom data source using X_y
```
glemaitre: Please remove all the comments for which the code is self explicit.

sobiya-22:
Hi @glemaitre,
I've now removed **self-explanatory comments** as suggested.
Please let me know if anything else needs to be addressed. 
Once approved, I’ll proceed with the separate PRs for the **PR curve and prediction-error visualizations** as you recommended.

sobiya-22:
Hi @glemaitre ,  Shall I proceed with separate PR's for `PR curve and prediction-error visualizations`?

skore/tests/unit/sklearn/plot/roc_curve/test_comparison_cross_validation.py
```diff
@@ -232,3 +232,79 @@ def test_multiclass_classification_kwargs(pyplot, multiclass_classification_repo
     display.plot(despine=False)
     assert display.ax_[0].spines["top"].get_visible()
     assert display.ax_[0].spines["right"].get_visible()
+
+
+def test_data_source_binary_classification(pyplot, binary_classification_data_no_split):
```
MarieSacksick, line 237: Can you check that the results when changing the data_source do change please?  
For instance, you can use a subsample of X and y when giving the data source `X_y` so you don't have to create a new one, and then check that the dataframes `display.frame()` outputed are different with `assert not` and the [equals](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.equals.html) function.

skore/tests/unit/sklearn/plot/roc_curve/test_comparison_cross_validation.py
```diff
@@ -232,3 +232,79 @@ def test_multiclass_classification_kwargs(pyplot, multiclass_classification_repo
...
+    n_reports = len(report.reports_)
+    n_splits = report.reports_[0]._cv_splitter.n_splits
+    expected_auc_entries = n_reports * n_splits
+
+    assert len(display.roc_auc) == expected_auc_entries
+    auc_values = display.roc_auc["roc_auc"].values
+    assert all(0 <= auc <= 1 for auc in auc_values)
+
+
+def test_data_source_multiclass_classification(
```
MarieSacksick, line 273: Same as above, can you check that the outputs are different please?

skore/tests/unit/sklearn/plot/roc_curve/test_comparison_estimator.py
```diff
@@ -141,6 +141,92 @@ def test_multiclass_classification(pyplot, multiclass_classification_data):
     assert display.ax_.get_title() == "ROC Curve"
 
 
+def test_data_source_binary_classification(pyplot, binary_classification_data):
```
MarieSacksick, line 144: same as in previous file: checking that the outputs are different.

MarieSacksick (changes requested):
Almost good to me :)!  
To be sure that the data_source is really used in the plot, and not just always the train or the test, can you check that there are some changes, other than in the parameter data_source please?

lisaleemcb:
Hi @sobiya-22 ! Is it okay if I finish up this issue for you? Although you're super close, so perhaps it would be easier for you.

glemaitre:
@lisaleemcb Looking at how the tests changed since, I think that we should be decorating either some common tests or already written test to try the different combination of the `data_source` without the need to rewrite entirely new tests.

Issue #1796, opened 2025-06-05
Remove Python 3.9 from supported versions


auguste-probabl:
### What would you like to say?

Follow-up to https://github.com/probabl-ai/skore/issues/1658.

Use the opportunity to update the code specific to Python 3.9; for example, some types can be made more readable by writing `X | Y` rather than `Union[X, Y]`. We should figure out if there are other such changes we could make.

thomass-dev:
Also, don't forget to remove the relative `pyproject` section 

```yml
[tool.ruff.lint.per-file-ignores]
"src/*" = ["UP007"]              
```

PR #1819, opened 2025-06-10
feat(skore): Define `importlib-metadata` explicitly as a `python<3.10` dependency instead of using it as a transitive dependency

thomass-dev:


glemaitre:
Do we want to drop python 3.9 also in the `main` branch instead?

thomass-dev:
> Do we want to drop python 3.9 also in the `main` branch instead?

Yes, I made this report before we decided on this issue. I take the https://github.com/probabl-ai/skore/issues/1796.

thomass-dev:
Closed in favor of https://github.com/probabl-ai/skore/pull/1827.

PR #1823, opened 2025-06-10
feat(train_test_split): Allow positional arguments with `as_dict=True`

### LINKED ISSUE #1544: Passing all datasets by keyword makes it annoying to use `as_dict=True`

nkapila6:
Fixes #1544.

- Allows the user to specify X and y positional arguments. 
- Warn users that single/double positional arguments default to 'X'/'y' names when as_dict=True.
- Updated docstrings.
- Added new and fixed old tests.

### DIFF
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -155,13 +155,13 @@ def train_test_split(
     keys = []
 
     if X is not None:
-        keys.append("X")
         new_arrays.append(X)
+        keys.append("X")
     if y is not None:
-        keys.append("y")
         new_arrays.append(y)
+        keys.append("y")
 
-    if as_dict:
+    if as_dict and arrays:
         # case when X or y is implicit as first 2 args for as_dict=True
         if len(arrays) == 1:
             warnings.warn(
@@ -183,7 +183,11 @@ def train_test_split(
                 "\nEx: train_test_split(X=X, y=y, z=z, sw=sample_weight, as_dict=True)"
             )
 
-    if keyword_arrays:  # case when keyword arg is specified but as_dict isn't used
+    if keyword_arrays:
+        if X is None and y is None:
+            arrays = tuple(
+                keyword_arrays.values()
+            )  # if X and y is not passed but other variables
         new_arrays.extend(keyword_arrays.values())
         keys.extend(keyword_arrays.keys())
 
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -138,12 +138,13 @@ def train_test_split(
 
     new_arrays = list(arrays)
     keys = []
+
     if X is not None:
+        keys.append("X")
         new_arrays.append(X)
-        keys += ["X"]
     if y is not None:
+        keys.append("y")
         new_arrays.append(y)
-        keys += ["y"]
 
     if as_dict:
         if len(arrays) == 1:
@@ -157,13 +158,9 @@ def train_test_split(
                 "Example: train_test_split(X, y, z=z, sw=sample_weight, as_dict=True)"
             )
 
-    if keyword_arrays:
-        if X is None and y is None:
-            arrays = tuple(
-                keyword_arrays.values()
-            )  # if X and y is not passed but other variables
-        keys += list(keyword_arrays.keys())
-        new_arrays += list(keyword_arrays.values())
+    if keyword_arrays:  # case when keyword arg is specified but as_dict isn't used
+        new_arrays.extend(keyword_arrays.values())
+        keys.extend(keyword_arrays.keys())
 
     output = sklearn.model_selection.train_test_split(
         *new_arrays,
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -176,8 +176,10 @@ def train_test_split(
             keys.extend(["X", "y"])
         elif len(arrays) > 2:
             raise ValueError(
-                "With as_dict=True, more than two arguments must be keyword arguments."
-                "\nEx: train_test_split(X=X, y=y, z=z, sw=sample_weight, as_dict=True)"
+                "With as_dict=True, expected no more than two positional arguments "
+                "(which will be interpreted as X and y). "
+                "The remaining arrays must be passed by keyword, "
+                "e.g. train_test_split(X, y, z=z, sw=sample_weights, as_dict=True)."
             )
 
     if keyword_arrays:
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -159,6 +159,12 @@ def train_test_split(
         keys.append("y")
 
     if as_dict and arrays:
+        # if X is passed as both keyword and pos args
+        if len(arrays) <= 2 and (X is not None or y is not None):
+            raise ValueError(
+                "With as_dict=True, X/y cannot be passed by both position and keyword."
+            )
+
         # if X or y are passed by position
         if len(arrays) == 1:
             warnings.warn(
@@ -174,6 +180,7 @@ def train_test_split(
                 stacklevel=2,
             )
             keys.extend(["X", "y"])
+
         elif len(arrays) > 2:
             raise ValueError(
                 "With as_dict=True, expected no more than two positional arguments "
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -145,11 +145,17 @@ def train_test_split(
         new_arrays.append(y)
         keys += ["y"]
 
-    if as_dict and arrays:
-        raise ValueError(
-            "When as_dict=True, arrays must be passed as keyword arguments.\n"
-            "Example: train_test_split(X=X, y=y, sw=sample_weight, as_dict=True)"
-        )
+    if as_dict:
+        if len(arrays) == 1:
+            keys += ["X"]
+        elif len(arrays) == 2:
+            keys.extend(["X", "y"])
+        elif len(arrays) > 2:
+            raise ValueError(
+                ""
+                "With as_dict=True, arguments must be passed as keyword arguments.\n"
+                "Example: train_test_split(X, y, z=z, sw=sample_weight, as_dict=True)"
+            )
 
     if keyword_arrays:
         if X is None and y is None:
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -162,12 +162,12 @@ def train_test_split(
         # if X or y is passed both by keyword and by position
         if X is not None:
             raise ValueError(
-                f"With as_dict=True, expected X to be passed either "
+                "With as_dict=True, expected X to be passed either "
                 "by position or keyword, not both."
             )
         if y is not None:
             raise ValueError(
-                f"With as_dict=True, expected y to be passed either "
+                "With as_dict=True, expected y to be passed either "
                 "by position or keyword, not both."
             )
 
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -172,7 +172,7 @@ def train_test_split(
             keys.append("X")
         elif len(arrays) == 2:
             warnings.warn(
-                "With as_dict=True, two arguments defaults to 'X' & 'y'."
+                "With as_dict=True, first 2 positional arguments defaults to 'X' & 'y'."
                 "\nSpecify as keyword argument to rename.",
                 stacklevel=2,
             )
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -135,16 +135,13 @@ def train_test_split(
      'y_train': ..., 'y_test': ...,
      'sample_weights_train': ..., 'sample_weights_test': ...}
 
-    >>> # With as_dict is True, first 2 arguments are implicitly treated as X and y.
-    >>> split_dict = train_test_split(X, as_dict=True)
-    >>> split_dict
+    >>> # With as_dict is True, the first 2 arguments are implicitly treated as X and y
+    >>> train_test_split(X, as_dict=True)
     {'X_train': ..., 'X_test': ...}
-    >>> split_dict = train_test_split(X, y, as_dict=True)
-    >>> split_dict
+    >>> train_test_split(X, y, as_dict=True)
     {'X_train': ..., 'X_test': ...,
      'y_train': ..., 'y_test': ...}
-    >>> split_dict = train_test_split(X, y, sample_weights=sample_weights, as_dict=True)
-    >>> split_dict
+    >>> train_test_split(X, y, sample_weights=sample_weights, as_dict=True)
     {'X_train': ..., 'X_test': ...,
      'y_train': ..., 'y_test': ...,
      'sample_weights_train': ..., 'sample_weights_test': ...}
@@ -162,7 +159,7 @@ def train_test_split(
         keys.append("y")
 
     if as_dict and arrays:
-        # case when X or y is implicit as first 2 args for as_dict=True
+        # if X or y are passed by position
         if len(arrays) == 1:
             warnings.warn(
                 "With as_dict=True, single argument defaults to 'X'."
@@ -177,7 +174,7 @@ def train_test_split(
                 stacklevel=2,
             )
             keys.extend(["X", "y"])
-        elif len(arrays) > 2:  # case when as_dict=True and given # arrays > 3
+        elif len(arrays) > 2:
             raise ValueError(
                 "With as_dict=True, more than two arguments must be keyword arguments."
                 "\nEx: train_test_split(X=X, y=y, z=z, sw=sample_weight, as_dict=True)"
--- a/skore/src/skore/sklearn/train_test_split/train_test_split.py
+++ b/skore/src/skore/sklearn/train_test_split/train_test_split.py
@@ -160,16 +160,14 @@ def train_test_split(
 
     if as_dict and arrays:
         # if X or y is passed both by keyword and by position
-        err = []
         if X is not None:
-            err.append("X")
-        if y is not None:
-            err.append("y")
-
-        if err:
-            args = " and ".join(err)
             raise ValueError(
-                f"With as_dict=True, expected {args} to be passed either "
+                f"With as_dict=True, expected X to be passed either "
+                "by position or keyword, not both."
+            )
+        if y is not None:
+            raise ValueError(
+                f"With as_dict=True, expected y to be passed either "
                 "by position or keyword, not both."
             )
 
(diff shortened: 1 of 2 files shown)


skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -138,26 +153,43 @@ class labels.
...
+            warnings.warn(
+                "With as_dict=True, first 2 positional arguments defaults to 'X' & 'y'."
+                "\nSpecify as keyword argument to rename.",
+                stacklevel=2,
+            )
+            keys.extend(["X", "y"])
+        elif len(arrays) > 2:  # case when as_dict=True and given # arrays > 3
+            raise ValueError(
+                "With as_dict=True, more than two arguments must be keyword arguments."
+                "\nEx: train_test_split(X=X, y=y, z=z, sw=sample_weight, as_dict=True)"
```
nkapila6: I suppose I could add another example in the ValueError but concerned it would be too long? Already have it in the doc string though.

Ex: train_test_split(X,y,z=z,sw=sample_weight,as_dict=True)

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -73,7 +73,8 @@ class labels.
     as_dict : bool, default is False
         If True, returns a Dictionary with keys values ``X_train``, ``X_test``,
         ``y_train``, and ``y_test`` instead of a List. Requires data to be
```
auguste-probabl, line 75:
```suggestion
        If True, returns a dictionary with keys values ``X_train``, ``X_test``,
        ``y_train``, and ``y_test`` instead of a list. Requires data to be
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -130,6 +131,20 @@ class labels.
     >>> split_dict = train_test_split(
     ...     X=X, y=y, sample_weights=sample_weights, as_dict=True, random_state=0)
     >>> split_dict
+    {'X_train': ..., 'X_test': ...,
+     'y_train': ..., 'y_test': ...,
+     'sample_weights_train': ..., 'sample_weights_test': ...}
+
+    >>> # With as_dict is True, first 2 arguments are implicitly treated as X and y.
```
auguste-probabl:
```suggestion
    >>> # With as_dict is True, the first 2 arguments are implicitly treated as X and y
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -130,6 +131,20 @@ class labels.
...
+    >>> split_dict = train_test_split(X, as_dict=True)
+    >>> split_dict
```
auguste-probabl:
```suggestion
    >>> train_test_split(X, as_dict=True)
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -130,6 +131,20 @@ class labels.
...
+    {'X_train': ..., 'X_test': ...}
+    >>> split_dict = train_test_split(X, y, as_dict=True)
+    >>> split_dict
```
auguste-probabl:
```suggestion
    >>> train_test_split(X, y, as_dict=True)
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -130,6 +131,20 @@ class labels.
...
+    {'X_train': ..., 'X_test': ...,
+     'y_train': ..., 'y_test': ...}
+    >>> split_dict = train_test_split(X, y, sample_weights=sample_weights, as_dict=True)
+    >>> split_dict
```
auguste-probabl:
```suggestion
    >>> train_test_split(X, y, sample_weights=sample_weights, as_dict=True)
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -138,26 +153,43 @@ class labels.
...
         new_arrays.append(y)
-        keys += ["y"]
+        keys.append("y")
 
     if as_dict and arrays:
-        raise ValueError(
-            "When as_dict=True, arrays must be passed as keyword arguments.\n"
-            "Example: train_test_split(X=X, y=y, sw=sample_weight, as_dict=True)"
-        )
+        # case when X or y is implicit as first 2 args for as_dict=True
```
auguste-probabl:
```suggestion
        # if X or y are passed by position
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -138,26 +153,43 @@ class labels.
...
+            )
+            keys.append("X")
+        elif len(arrays) == 2:
+            warnings.warn(
+                "With as_dict=True, first 2 positional arguments defaults to 'X' & 'y'."
+                "\nSpecify as keyword argument to rename.",
+                stacklevel=2,
+            )
+            keys.extend(["X", "y"])
+        elif len(arrays) > 2:  # case when as_dict=True and given # arrays > 3
```
auguste-probabl:
```suggestion
        elif len(arrays) > 2:
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,78 @@ def test_train_test_split_kwargs():
...
+
+    X = [[1]] * 20
+    y = [0] * 10 + [1] * 10
+
+    output = train_test_split(X, y, as_dict=True)
+    keys = list(output.keys())
+    # assert that keys have X and y
+    assert keys[0].startswith("X") and keys[2].startswith("y")
+    # assert that output is a dict
+    assert type(output) is dict
```
auguste-probabl: put this first, use `isinstance` and remove the redundant comment

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,78 @@ def test_train_test_split_kwargs():
...
+def test_train_test_split_two_posargs():
+    """Passing two positional argument with as_dict=True should return as dict."""
+
+    X = [[1]] * 20
+    y = [0] * 10 + [1] * 10
+
+    output = train_test_split(X, y, as_dict=True)
+    keys = list(output.keys())
+    # assert that keys have X and y
+    assert keys[0].startswith("X") and keys[2].startswith("y")
```
auguste-probabl:
```suggestion
    assert set(output.keys()) == {"X_train", "X_test", "y_train", "y_test"}
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,78 @@ def test_train_test_split_kwargs():
...
+    z = [0] * 10 + [1] * 10
+
+    output = train_test_split(X, y, z=z, as_dict=True)
+    keys = list(output.keys())
+    # assert that keys have X, y, z
+    assert (
+        keys[0].startswith("X") and keys[2].startswith("y") and keys[4].startswith("z")
+    )
+    # assert that output is a dict
+    assert type(output) is dict
```
auguste-probabl: see above

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,78 @@ def test_train_test_split_kwargs():
...
+    )
 
-    with pytest.raises(
-        ValueError,
-        match="When as_dict=True, arrays must be passed as keyword arguments",
-    ):
-        train_test_split(X, y, random_state=0, as_dict=True)
+    with pytest.raises(ValueError) as err:
+        train_test_split(X, y, z, random_state=0, as_dict=True)
+    assert str(err.value) == err_msg
```
auguste-probabl: use `raises`'s `match` argument
```suggestion
    with pytest.raises(ValueError, match=err_msg):
        train_test_split(X, y, z, as_dict=True)
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,78 @@ def test_train_test_split_kwargs():
...
 
 
 def test_train_test_split_check_dict():
-    """If `return_dict` is True then the result is a dict."""
+    """If `as_dict` is True then the result is a dict."""
     X = [[1]] * 20
     y = [0] * 10 + [1] * 10
     output = train_test_split(X=X, y=y, random_state=0, as_dict=True)
     assert type(output) is dict
```
auguste-probabl: that test seems redundant now

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -138,26 +153,43 @@ class labels.
...
+                "\nSpecify as keyword argument to rename.",
+                stacklevel=2,
+            )
+            keys.append("X")
+        elif len(arrays) == 2:
+            warnings.warn(
+                "With as_dict=True, first 2 positional arguments defaults to 'X' & 'y'."
+                "\nSpecify as keyword argument to rename.",
+                stacklevel=2,
+            )
```
auguste-probabl: i don't think the warnings are worth it. if anything this function needs [fewer warnings](https://github.com/probabl-ai/skore/discussions/1800)

skore/src/skore/sklearn/train_test_split/train_test_split.py
auguste-probabl:
```suggestion
                "With as_dict=True, expected no more than two positional arguments "
                "(which will be interpreted as X and y). "
                "The remaining arrays must be passed by keyword, "
                "e.g. train_test_split(X, y, z=z, sw=sample_weights, as_dict=True)."
```

auguste-probabl: see my other comment

auguste-probabl (changes requested):
Looking great. I have a few suggestions

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: I agree with you but just concerned if user does not follow our implicit assumptions of pos args.

Based on your last comment in the linked discussion, it could be integrated into the Panel UX on line [243](https://github.com/probabl-ai/skore/blob/c9fe7fc0454905ae6612a91314f623850470942a/skore/src/skore/sklearn/train_test_split/train_test_split.py#L243)? I could then remove these warnings for a future issue.

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: Added my thoughts!

auguste-probabl:
feel free to resolve comments yourself

skore/src/skore/sklearn/train_test_split/train_test_split.py
auguste-probabl: IMO the behavior is not surprising enough to warrant pestering the user with a warning, and if the user is in fact surprised then the warning would not change that (they would open an issue). Not to mention the behavior is clearly documented

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,88 @@ def test_train_test_split_kwargs():
...
+def test_train_test_split_single_posargs():
+    """Passing single positional argument with as_dict=True should return as dict."""
+
+    X = [[1]] * 20
+
+    output = train_test_split(X, as_dict=True)
+    # assert that first key contains X
+    assert next(iter(output)).startswith("X")
+    # assert that output is a dict
+    assert type(output) is dict
```
auguste-probabl:
```suggestion
    assert isinstance(output, dict)
    assert set(output.keys()) == {"X_train", "X_test"}
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
nkapila6: Hmm, you have a point, it is indeed documented in the docstrings. I'll remove the warnings in a bit.

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,88 @@ def test_train_test_split_kwargs():
...
+    X = [[1]] * 20
+    y = [0] * 10 + [1] * 10
+    z = [0] * 10 + [1] * 10
+
+    output = train_test_split(X, y, z=z, as_dict=True)
+    keys = set(output.keys())
+    # assert that keys have X, y, z
+    assert keys == {"X_train", "X_test", "y_train", "y_test", "z_train", "z_test"}
+    # assert that output is a dict
+    assert isinstance(output, dict)
```
auguste-probabl, line 242: remove comments, move `isinstance` before defining `keys`

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -138,26 +150,52 @@ class labels.
...
         new_arrays.append(y)
-        keys += ["y"]
+        keys.append("y")
 
     if as_dict and arrays:
-        raise ValueError(
-            "When as_dict=True, arrays must be passed as keyword arguments.\n"
-            "Example: train_test_split(X=X, y=y, sw=sample_weight, as_dict=True)"
-        )
+        # if X is passed as both keyword and pos args
```
auguste-probabl:
```suggestion
        # if X or y is passed both by keyword and by position
```

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -138,26 +150,52 @@ class labels.
...
+        if len(arrays) <= 2 and (X is not None or y is not None):
+            raise ValueError(
+                "With as_dict=True, X/y cannot be passed by both position and keyword."
```
auguste-probabl:
```suggestion
                "With as_dict=True, expected X and y to be passed either by position or by keyword. "
                "Got both."
```

auguste-probabl: Actually that's not clear enough. I think the best we can do here is to test separately for `X` and for `y`

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,88 @@ def test_train_test_split_kwargs():
...
-        ValueError,
-        match="When as_dict=True, arrays must be passed as keyword arguments",
-    ):
-        train_test_split(X, y, random_state=0, as_dict=True)
+    output = train_test_split(X, y, as_dict=True)
+    keys = set(output.keys())
+    # assert that keys have X and y
+    assert keys == {"X_train", "X_test", "y_train", "y_test"}
+    # assert that output is a dict
+    assert isinstance(output, dict)
```
auguste-probabl:
```suggestion
    assert isinstance(output, dict)
    assert set(output.keys()) == {"X_train", "X_test", "y_train", "y_test"}
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
auguste-probabl:
```suggestion
    assert isinstance(output, dict)
    assert set(output.keys()) == {"X_train", "X_test", "y_train", "y_test", "z_train", "z_test"}
```

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -182,30 +182,88 @@ def test_train_test_split_kwargs():
...
+    err_msg = (
+        "With as_dict=True, expected no more than two positional arguments "
+        "(which will be interpreted as X and y). "
+        "The remaining arrays must be passed by keyword, "
+        "e.g. train_test_split(X, y, z=z, sw=sample_weights, as_dict=True)."
+    )
+    import re
+
+    with pytest.raises(ValueError, match=re.escape(err_msg)):
+        train_test_split(X, y, z, as_dict=True)
```
auguste-probabl, line 263: I think just using `r"..."` is enough

auguste-probabl:
```suggestion
    err_msg = (
        r"With as_dict=True, expected no more than two positional arguments "
        r"(which will be interpreted as X and y). "
        r"The remaining arrays must be passed by keyword, "
        r"e.g. train_test_split(X, y, z=z, sw=sample_weights, as_dict=True)."
    )

    with pytest.raises(ValueError, match=err_msg):
        train_test_split(X, y, z, as_dict=True)
```

auguste-probabl: or maybe not. just resolve if it doesn't work

auguste-probabl (changes requested):
Almost there

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
nkapila6: It works fine with `r` but I need to manually escape regex meta and `re.escape()` was cleaner 😓

skore/src/skore/sklearn/train_test_split/train_test_split.py
```diff
@@ -138,26 +150,49 @@ class labels.
...
+            err.append("X")
+        if y is not None:
+            err.append("y")
+
+        if err:
+            args = " and ".join(err)
+            raise ValueError(
+                f"With as_dict=True, expected {args} to be passed either "
+                "by position or keyword, not both."
+            )
```
auguste-probabl: This is better than before, but I imagined something simpler:
```suggestion
        if X is not None:
            raise ValueError(
                f"With as_dict=True, expected X to be passed either "
                "by position or keyword, not both."
            )
        if y is not None:
            raise ValueError(
                f"With as_dict=True, expected y to be passed either "
                "by position or keyword, not both."
            )
```

nkapila6: 😅 I did that for the case: X, y, X=X, y=y

auguste-probabl (changes requested):
Just one last thing and I think we're good

skore/src/skore/sklearn/train_test_split/train_test_split.py
auguste-probabl: It's unnecessary IMO: the user will first hit the X error and then either figure out that it'll be the same for y, or get the error for y.

nkapila6: Alr, fair. I guess I was overthinking the edge cases to make it dynamic hahah

auguste-probabl: no worries, i just want to avoid over-complicating

auguste-probabl (dismissed):
Awesome work @nkapila6, thanks for the reactivity!!

nkapila6:
Thank you @auguste-probabl 😄

auguste-probabl (approved):

PR #1826, opened 2025-06-11
chore(refacto): Make the same variable common

MarieSacksick:
I have to use this variable in another PR (https://github.com/probabl-ai/skore/pull/1816), and I find it not clean to have to copy/paste the same information in different classes.


MarieSacksick:
@glemaitre in a previous comment (https://github.com/probabl-ai/skore/pull/1520/files#r2108991825) you seemed to say that using a utils might not be a good idea?

skore/src/skore/sklearn/_estimator/metrics_accessor.py
```diff
@@ -39,22 +40,6 @@ class _MetricsAccessor(_BaseAccessor["EstimatorReport"], DirNamesMixin):
...
-        "precision": {"name": "Precision", "icon": "(↗︎)"},
-        "recall": {"name": "Recall", "icon": "(↗︎)"},
-        "brier_score": {"name": "Brier score", "icon": "(↘︎)"},
-        "roc_auc": {"name": "ROC AUC", "icon": "(↗︎)"},
-        "log_loss": {"name": "Log loss", "icon": "(↘︎)"},
-        "r2": {"name": "R²", "icon": "(↗︎)"},
-        "rmse": {"name": "RMSE", "icon": "(↘︎)"},
-        "custom_metric": {"name": "Custom metric", "icon": ""},
-        "summarize": {"name": "Metrics summary", "icon": ""},
-        "confusion_matrix": {"name": "Confusion Matrix", "icon": ""},
```
auguste-probabl, line 55: `confusion_matrix` should remain here somehow right?

MarieSacksick: thanks, more attentive than I, I hadn't seen that difference!

glemaitre:
> in a previous comment (https://github.com/probabl-ai/skore/pull/1520/files#r2108991825) you seemed to say that using a utils might not be a good idea?

I did not like that we need to store a global dictionary in the `utils` while it was already defined in the classes. I don't really have an issue with `utils` in general :)

MarieSacksick:
> I did not like that we need to store a global dictionary in the utils while it was already defined in the classes. I don't really have an issue with utils in general :) 

I guess so :laughing:   
The question is more: is it worth it? To me, it's enough lines of code and information shared between classes to be worth centralizing it somewhere.

skore/src/skore/sklearn/utils.py
```diff
@@ -0,0 +1,21 @@
+"""Utility variable for Skore and Scikit-learn integration."""
+
+_SCORE_OR_LOSS_INFO: dict[str, dict[str, str]] = {
```
glemaitre: So indeed, I don't like it this way :)

So here, I would prefer to create in `skore/sklearn/_base.py` a class named `_BaseMetricsAccessor` from which each individual `_MetricsAccessor` will inherit.


The class should be something like

```python
class _BaseMetricsAccessor:
    # no need anymore for a global dict and we can use lower case
    _score_or_loss_info: dict[str, dict[str, str]] = {
        ...
    }
```

And then for the `_MetricAccessor` from estimator we can inherit and modify specifically for the need of estimator (because we did not yet do the job for the other one):

```python
class _MetricsAccessor(
    _BaseAccessor["EstimatorReport"], _BaseMetricsAccessor, DirNamesMixin
):
    _score_or_loss_info: dict[str, dict[str, str]] = {
        **_BaseMetricsAccessor._score_or_loss_info,
        "confusion_matrix": {"name": "Confusion Matrix", "icon": ""},
    }
```

glemaitre:
> The question is more: is it worth it? To me, it's enough lines of code and information shared between classes to be worth centralizing it somewhere.

With the `_BaseMetricsAccessor`, a next step would be to investigate if other part of the code specific to the metrics accessor should go there. Let's say this refactor is a first step.

skore/src/skore/sklearn/utils.py
MarieSacksick: ok I'll change!  

Question to learn: why do you prefer a parent class rather than a util? Intuitively, I would have used a parent class more for functions, and utils for variables.

glemaitre: Here, the `_score_or_loss_info` is really an attribute closely related to the metrics accessor. A utililty function/variable is something that I would share across many objects/functions and that will not have a tight scope. For instance, if this `_score_or_loss_info` starts to be used in different accessors then I would reconsider moving it in a more global utility module.

skore/src/skore/sklearn/_base.py
```diff
@@ -429,3 +429,74 @@ def _get_cached_response_values(
         (cache_key, predictions, False),
         (predict_time_cache_key, predict_time(), False),
     ]
+
+
+class _BaseMetricsAccessor:
+    _SCORE_OR_LOSS_INFO: dict[str, dict[str, str]] = {
```
glemaitre: Since it is not anymore a global variable, could you use the same name but in lower case.

PR #1839, opened 2025-06-15
refactor(skore): Modify dataframe format of MetricsSummaryDisplay

### LINKED ISSUE #1837: api(skore): Revisit the output of the `.frame()` from `.summarize()`

We recently added the `.summarize` method that create a `MetricsSummaryDisplay` to align the API of the object. We should now revisit a bit the API of the `summarize` function and the `frame` method and specifically the data structure stored in the display.

With the experience gather with the other display, we should store a "long" dataframe in the display because it can accommodate all the different combination of reports. The `frame` method will have a job to transform this "long" dataframe to a "wide" dataframe whenever possible depending on the case.

In terms of API, we should move the parameter `indicator_favorability` , `flat_index`, `aggregate` as parameter into the `frame` method because they just modify the structure of the dataframe. We can switch `flat_index` to `True` to make it easy to index the dataframe.

glemaitre:
closes #1837 

This PR intends to:

- transform the internal format of the dataframe from wide to long when calling `.summarize`
- make that `.frame` is converting from long to wide format when necessary
- add the necessary API to `frame` to provide flexibility


auguste-probabl:
Superseded by https://github.com/probabl-ai/skore/pull/2322 ? Please close if yes

glemaitre:
Closing because it is outdated.

Issue #1624, opened 2025-04-30
docs: Indicate that `indicator_favorability` appears only when a scorer is given, or a string referencing a scikit-learn scorer.


sylvaincom:
### Describe the (alleged) bug

With an adding a metric in the report metrics, `indicator_favorability=True` does not display anything for that new metric:

<img src="https://github.com/user-attachments/assets/df676caa-0f56-43e8-b845-74106b559ffb" width=650>

### Steps/Code to Reproduce

The code is in the following PR: https://github.com/probabl-ai/skore/pull/1623
This PR updates the `plot_employee_salaries` example

### Expected Behavior

`indicator_favorability` should be displayed

### Actual Behavior

`indicator_favorability` is not displayed

### Environment

```shell
System:
    python: 3.12.4 | packaged by Anaconda, Inc. | (main, Jun 18 2024, 10:07:17) [Clang 14.0.6 ]
executable: [/Users/sylvaincom/Documents/GitHub/skore/.venv/bin/python](https://file+.vscode-resource.vscode-cdn.net/Users/sylvaincom/Documents/GitHub/skore/.venv/bin/python)
   machine: macOS-15.4.1-arm64-arm-64bit

Python dependencies:
        skore: 0.0.0+unknown
          pip: 24.0
    diskcache: 5.6.3
       joblib: 1.4.2
   matplotlib: 3.10.1
        numpy: 2.2.5
       pandas: 2.2.3
         rich: 14.0.0
 scikit-learn: 1.6.1
        skops: 0.11.0
polars-lts-cpu; (platform_machine: None
polars; (platform_machine: None
```

auguste-probabl:
This is to be expected: you cannot expect skore to know that your user-defined function is lower/greater-is-better. Maybe this could be passed as an extra argument somehow...

Schefflera-Arboricola:
I'd like to work on this issue please.

sylvaincom:
> This is not a bug: you cannot expect skore to know that your user-defined function is lower/greater-is-better. Maybe this information could be passed as an extra argument somehow...

True. Some suggestions:
- If the metric is from scikit-learn such as MAE, we should know whether up or down. We could hardcode a dictionary somewhere.
- If the metric is custom made, the dev should put it as an extra argument indeed. There is very important to share with its colleagues and stakeholders.

sylvaincom:
@Schefflera-Arboricola 
> I'd like to work on this issue please.

[...]

PR #1866, opened 2025-06-26
docs: Add docs to explain link between scorer and favorability

MarieSacksick:
Closes https://github.com/probabl-ai/skore/issues/1624.

### DIFF
--- a/skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
+++ b/skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
@@ -95,9 +95,9 @@ class _MetricsAccessor(
               and will be forwarded with `scoring_kwargs`.
             - if the callable API is too restrictive (e.g. need to pass
               same parameter name with different values), you can use scikit-learn
-              scorers as provided by :func:`sklearn.metrics.make_scorer`. With its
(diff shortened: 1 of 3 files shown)


skore/src/skore/sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -95,7 +95,9 @@ def summarize(
               and will be forwarded with `scoring_kwargs`.
             - if the callable API is too restrictive (e.g. need to pass
               same parameter name with different values), you can use scikit-learn
-              scorers as provided by :func:`sklearn.metrics.make_scorer`.
+              scorers as provided by :func:`sklearn.metrics.make_scorer`. With its
+              `greater_is_better` parameter, the favorability of the metric will be
+              displayed.
```
auguste-probabl:
```suggestion
              scorers as provided by :func:`sklearn.metrics.make_scorer`. In this case,
              the metric favorability will only be displayed if it is given explicitly via
              `make_scorer`'s `greater_is_better` parameter.
```

MarieSacksick:
I also added indication when using the callable type of argument.

auguste-probabl (approved):
Nice!

PR #1759, opened 2025-05-25
refactor: Refactor confusion matrix

waridrox:
Following up on #1757

CC: @glemaitre

Issue #1562, opened 2025-04-18
bug: The "make html" command fails when run in Git Bash on Windows


paulinebanye:
### **Bug Description**
When attempting to build the HTML documentation using the _`make html`_ command, the process fails due to the cache.db file being in use. 

```python
(venv) C:\Users\Pauline PC\Documents\codes\ESOC\skore\sphinx>make html
Running Sphinx v8.1.3
loading translations [en]... done
Using Sphinx-Gallery to convert rst text blocks to markdown for .ipynb files.
[autosummary] generating autosummary for: auto_examples\getting_started\plot_quick_start.rst, contributing.rst, index.rst, install.rst, reference\api\skore.Compa_predictions.rst, reference\api\skore.ComparisonReport.clear_cache.rst, reference\api\skore.ComparisonReport.get_predictions.rst, reference\api\skore.ComparisonReference\api\skore.ComparisonReport.metrics.accuracy.rst, reference\api\skore.ComparisonReport.metrics.brier_score.rst, ..., reference\api\skore.set_config.rst, re.train_test_split.rst, reference\index.rst, reference\project.rst, reference\report\comparison_report.rst, reference\report\cross_validation_report.rst, refereys.rst, reference\report\estimator_report.rst, reference\report\index.rst, reference\utils.rst
loading intersphinx inventory 'python' from https://docs.python.org/3/objects.inv ...
loading intersphinx inventory 'numpy' from https://numpy.org/doc/stable/objects.inv ...
loading intersphinx inventory 'scipy' from https://docs.scipy.org/doc/scipy/objects.inv ...
loading intersphinx inventory 'matplotlib' from https://matplotlib.org/stable/objects.inv ...
loading intersphinx inventory 'sklearn' from https://scikit-learn.org/stable/objects.inv ...
loading intersphinx inventory 'skrub' from https://skrub-data.org/stable/objects.inv ...
loading intersphinx inventory 'pandas' from http://pandas.pydata.org/pandas-docs/stable/objects.inv ...
loading intersphinx inventory 'polars' from https://docs.pola.rs/py-polars/html/objects.inv ...
loading intersphinx inventory 'seaborn' from http://seaborn.pydata.org/objects.inv ...
generating gallery...
generating gallery for auto_examples... 
WARNING: g gallery for auto_examples\getting_started... [ 50%] plot_quick_start.py
..\examples\getting_started\plot_quick_start.py unexpectedly failed to execute correctly:

    Traceback (most recent call last):
[...]

PR #1868, opened 2025-06-26
chore: Add pixi

### LINKED ISSUE #1659: repo: Add pixi

The onboarding experience is made more difficult by having to give detailed explanations about virtual environments and e.g. conda. The new contributor needs to manually create an environment, activate it, install dependencies... these steps are error-prone and prevent the contributor from quickly and reliably doing their work, e.g. testing their changes.

`pixi` solves this problem by defining Makefile-like directives, called [tasks](https://pixi.sh/latest/workspace/advanced_tasks/), but with the important difference that pixi takes care of installing the environment prior to running any task.
By switching to it we can eliminate many future issues and reduce our docs size.

This issue is about using `pixi` for contributors, but since it also locks package dependencies a case could be made that if it works out we should also use it in CI. See #1331 for an attempt at doing this.

auguste-probabl:
Add `pixi.toml`, including tasks commonly run when working on skore. Since this accomplishes the same job as Makefiles, but additionally ensuring that the environment is always up-to-date, Makefiles have been removed.

Closes #1659
Addresses various issues where contributors forget the re-run `make install-skore` or some other command, such as https://github.com/probabl-ai/skore/pull/1759#issuecomment-2993563871 (not single-ing anyone out of course, this happens to everyone)


auguste-probabl:
I still need to test whether every command works out-of-the-box on Windows.

thomass-dev:
You are forcing user to use `pixi` by removing `/Makefile` and `sphinx/Makefile`.
IMO, as proposed in the issue, `pixi` should be an alternative, not the only way to do.

pixi.lock
```diff
@@ -0,0 +1,5031 @@
+version: 6
```
thomass-dev, line 1: Can't we let the user generate this file on-the-fly?

auguste-probabl:
> You are forcing user to use `pixi` by removing `/Makefile` and `sphinx/Makefile`. IMO, as discussed in the issue, `pixi` should be an alternative, not the only way to do.

Right now we are "forcing" users to use `make`, so in this sense not much is changing (except `make` is not present by default on Windows); also, note that the dependencies of each package remain in each project's `pyproject.toml`, so users can install individual package dependencies, run tests, etc. without using `pixi` (or `make`).

pixi.lock
auguste-probabl: The point of locking is that the environment is consistent across everyone's machine: for example, if you checkout a Git commit with a different `pixi.lock`, then running `pixi run ...` will pick up the new environment (i.e. you don't have to remember to `make install`)

thomass-dev: Otherwise you have to add all the machinery to check that it's up to date in each PR.

auguste-probabl:
Confirmed that running tests, linting, and building docs all work with the `pixi run` commands (apart from https://github.com/probabl-ai/skore/issues/1562#issuecomment-3012126301 which is not specific to pixi).

auguste-probabl:
I'm pretty sure the failure in https://github.com/probabl-ai/skore/actions/runs/15923718608/job/44916157923?pr=1868:
```
Handler <function update_config at 0x7f8a40a504a0> for event 'builder-inited' threw an exception (exception: Invalid URL 'https:///versions.json': No host supplied)
```
is due to `SPHINX_DOMAIN` being set to the empty string.

tuscland:
Should we rather force users to use `pixi` or `make`?
Right now, I would prefer to maintain configuration for a single build tool.

tuscland:
@glemaitre, we would need you to take the decision on wether pixi should be adopted, and if it complements or replaces the Makefile-based workflow.

auguste-probabl:
Closing due to lack of interest ATM

PR #1870, opened 2025-06-27
docs: Dot not fail on `examples/` when cleanup cant be achieved

thomass-dev:
Since we only support `python>=3.10`, use `ignore_cleanup_errors` parameter.

Source: https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory.cleanup


auguste-probabl (approved):
Tested manually; it seems to work! Thanks

Issue #1507, opened 2025-03-31
docs: be clearer about the obligation to use the same test set in comparison report


MarieSacksick:
### Which part of the documentation needs improvement?

It is [mentionned](https://skore.probabl.ai/0.7/auto_examples/getting_started/plot_skore_getting_started.html#comparing-estimators-reports) that the comparison report enables comparing on the same test set.

### Describe the problem found in the documentation

Does skore check that the two test sets are indeed identical ? maybe the datasets are stored in the EstimatorReport and their hashes are compared ? maybe not ? it would be nice to have the guide tell me about it, and showcase what happens if I try to use ComparisonReport if the datasets are in fact different.

### Suggested improvement

- [ ] explain how we do to ensure that the two test sets are identical
- [ ] show what happens otherwise

auguste-probabl:
See #1621

PR #1871, opened 2025-06-27
docs: Give more information about comparison report

MarieSacksick:
Closes https://github.com/probabl-ai/skore/issues/1507.  

One of the task of the issue is "show what happens otherwise" (ie when y_target is not the same), but I don't think anymore that it's super relevant.

### DIFF
--- a/skore/src/skore/sklearn/_comparison/report.py
+++ b/skore/src/skore/sklearn/_comparison/report.py
@@ -42,6 +42,7 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
     reports : list of reports or dict
         Reports to compare. If a dict, keys will be used to label the estimators;
         if a list, the labels are computed from the estimator class names.
+        Expects at least two reports to compare, with he same test target.
 
     n_jobs : int, default=None
         Number of jobs to run in parallel. Training the estimators and computing
--- a/skore/src/skore/sklearn/_comparison/report.py
+++ b/skore/src/skore/sklearn/_comparison/report.py
@@ -42,7 +42,7 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
     reports : list of reports or dict
         Reports to compare. If a dict, keys will be used to label the estimators;
         if a list, the labels are computed from the estimator class names.
-        Expects at least two reports to compare, with he same test target.
+        Expects at least two reports to compare, with the same test target.
 
     n_jobs : int, default=None
         Number of jobs to run in parallel. Training the estimators and computing
--- a/sphinx/user_guide/reporters.rst
+++ b/sphinx/user_guide/reporters.rst
@@ -133,7 +133,7 @@ performance of the different models.
 In order for the comparison to make sense, the reports must all have the same test target.
 However, they may have different training data or target; this might be the case when comparing a new model with the current production model, for example.
 They may also have different testing data ($X_\text{test}$), which means the compared model pipelines do not necessarily need to be the same.
-This comparison of testing target is done thanks to hashing of the table. Therefore, if
+The comparison of test targets is done by computing a hash of the arrays. Therefore, if
 the y_test are functionally equal, but have different data types, they will be
 considered as different.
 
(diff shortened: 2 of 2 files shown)


skore/src/skore/sklearn/_comparison/report.py
```diff
@@ -42,6 +42,7 @@ class ComparisonReport(_BaseReport, DirNamesMixin):
     reports : list of reports or dict
         Reports to compare. If a dict, keys will be used to label the estimators;
         if a list, the labels are computed from the estimator class names.
+        Expects at least two reports to compare, with he same test target.
```
auguste-probabl:
```suggestion
        Expects at least two reports to compare, with the same test target.
```

sphinx/user_guide/reporters.rst
```diff
@@ -123,15 +123,20 @@ metrics available in `skore` for cross-validation.
...
 :class:`ComparisonReport` takes a list (or a dictionary) of :class:`EstimatorReport` or
 :class:`CrossValidationReport` instances. It then provides methods to compare the
 performance of the different models.
 
+The reports to be compared must share the same test target, for the comparison to make
+sense. The reports may not share the same training data or target, therefore the
+:class:`ComparisonReport` can be used to diagnose the performance of a new model,
+compared to a historical one in production for instance. They may also not share the
+same testing X, to remove the constraints of having consistent pipeline input.
+This comparison of testing target is done thanks to hashing of the table.
```
auguste-probabl: "The reports may not share the same training data" is ambiguous, and could be interpreted as "The reports *must not* share the same training data".
I would make it a more direct statement like "the reports may have different training data".

```suggestion
In order for the comparison to make sense, the reports must all have the same test target.
However, they may have different training data or target; this might be the case when comparing a new model with the current production model, for example.
They may also have different testing data ($X_\text{test}$), which means the compared model pipelines do not necessarily need to be the same.
This comparison of testing target is done thanks to hashing of the table.
```

auguste-probabl: Do you mean that this is the method we'll use to check if the `y_test` are all the same? If so, this seems like an implementation detail; either remove it or add some detail on why that's important (e.g. I guess, that if the `y_test` are functionally equal, but have different data types, then the check will fail)

auguste-probabl: This is a comment on the last sentence only. sorry for the mis-attribution

MarieSacksick: You're right, we don't really need to know how the check is done actually.  
I don't know how hashing works exactly, "y_test are functionally equal, but have different data types, then the check will fail" --> was this just an example, or is it true?  For instance if I have one y with 1 and 0, and the other with 1.0 and 0.0 (on the same lines), does it output the same hash?

auguste-probabl: This is true, the two arrays in your example don't have the same hash

auguste-probabl: See https://github.com/probabl-ai/skore/blob/main/skore/src/skore/sklearn/_comparison/report.py#L174-L178 for the impl

MarieSacksick: I'm adding your precision then.

sphinx/user_guide/reporters.rst
```diff
@@ -123,15 +123,20 @@ metrics available in `skore` for cross-validation.
...
+:class:`ComparisonReport`.
 
 :class:`ComparisonReport` takes a list (or a dictionary) of :class:`EstimatorReport` or
 :class:`CrossValidationReport` instances. It then provides methods to compare the
 performance of the different models.
 
+In order for the comparison to make sense, the reports must all have the same test target.
+However, they may have different training data or target; this might be the case when comparing a new model with the current production model, for example.
+They may also have different testing data ($X_\text{test}$), which means the compared model pipelines do not necessarily need to be the same.
+This comparison of testing target is done thanks to hashing of the table. Therefore, if
```
auguste-probabl:
```suggestion
The comparison of test targets is done by computing a hash of the arrays. Therefore, if
```

MarieSacksick: I'm not sure about arrays / table, because it can be a multi-dim in case of multi-output. Even though by "excès de langage", indeed, we always talk about arrays.

auguste-probabl: You mean that they might be pandas sometimes? Otherwise, a numpy array can be multi-dimensional so I don't really see the issue.

MarieSacksick: true! let's go.

auguste-probabl (approved):
Nice!

PR #1888, opened 2025-07-02
feat: Show permutation at different stages of a pipeline

### LINKED ISSUE #1398: enh: Show permutation at different stage of a pipeline
Currently, we compute the permutation importance for the feature at the entrance of the the pipeline. However, one might be interested to know the importance at the entrance of the predictive model after some feature engineering.

In pseudo-code, in scikit-learn it would look like:

```python
pipeline = ...
X_engineered, y_engineered = pipeline[:-1].fit_transform(X, y)
permutation_importance(pipeline[-1], X_engineered, y_engineered)
```

I therefore think that we could have a parameter to control this behaviour. Another thought is that I think that we have only 2 location where we want to have this feature importance: start of the pipeline and start of the predictive model (last stage). I think that intermediate steps would be YAGNI for the moment.

waridrox:
Closes #1398 

Added support for calculating permutation importance at different stages of a pipeline. One can choose to compute feature importance either at the start of the pipeline or at the end.

Added a new `stage` parameter to `permutation()` and `_feature_permutation()` methods with values:
  - `start` (default): Computes importance using raw input features.
  - `end`: Computes importance using engineered features after pipeline transformations.

- [x] Added documentation and tests for the implementation.

CC: @MarieSacksick @glemaitre

### DIFF
--- a/examples/use_cases/plot_feature_importance.py
+++ b/examples/use_cases/plot_feature_importance.py
@@ -1072,6 +1072,55 @@ plot_permutation_train_test(tree_report)
 # and ``AveOccup``.
 
 # %%
+# Computing permutation importance at different stages of a pipeline
+# =================================================================
+#
+# When working with pipelines that contain feature engineering steps, we might be interested
+# in measuring feature importance at different stages:
+#
+# - At the beginning of the pipeline (raw features)
+# - Before the final estimator (engineered features)
+#
+# The `stage` parameter in the `permutation` method allows us to control where the feature
+# importance is calculated.
+
+# %%
+# Here's a pipeline with feature engineering (standardization) and a Ridge regression:
+
+# %%
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+
+pipeline_model = make_pipeline(StandardScaler(), Ridge())
+pipeline_model.fit(X_train, y_train)
+
+pipeline_report = EstimatorReport(
+    pipeline_model,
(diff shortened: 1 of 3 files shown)


examples/use_cases/plot_feature_importance.py
```diff
@@ -1071,6 +1071,55 @@ def plot_permutation_train_test(est_report):
...
+pipeline_report.feature_importance.permutation(seed=0, stage="start")
+
+# %%
+# And also at the end of feature engineering, right before the Ridge regression:
+
+# %%
+pipeline_report.feature_importance.permutation(seed=0, stage="end")
+
+# %%
+# This capability is particularly useful we want to understand:
```
auguste-probabl:
```suggestion
# This capability is particularly useful when we want to understand:
```

skore/tests/unit/sklearn/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -314,3 +314,51 @@ def test_not_fitted(regression_data):
     error_msg = "This LinearRegression instance is not fitted yet"
     with pytest.raises(NotFittedError, match=error_msg):
         report.feature_importance.permutation()
+
+
+def test_stage_parameter():
```
auguste-probabl: I'd rather have one test for start and one for end. Use `pytest.mark.parametrize` if it helps save space

auguste-probabl:
Superseded by #1988

Issue #1808, opened 2025-06-06
skore-remote: Switch storage


MarieSacksick:
### Is your feature request related to a problem? Please describe.

We are going to switch how we do storage on skore-hub. We need to adapt how we send and fetch artefacts with skore-lib-infra.  

### Describe the solution you'd like

TBD.

thomass-dev:
Today we discussed with @tuscland and @retromechs that the choice of the hashing method is going to be an important issue, and must be explicitly called/referenced in the code.

- https://docs.python.org/3/library/hashlib.html#hashlib.sha3_256
- https://github.com/oconnor663/blake3-py

In addition, it should be interesting to prepend the checksum with its method of compute, such as `sha3-256:<checksum>`.
It will help the `skore-hub-project` on future retro-compatibility needs.

thomass-dev:
Quoting @tuscland : 

Here is a note from Claude worth checking out (and verify!):

>     SHA3-256 has limited parallel implementations compared to SHA-2. The Keccak sponge construction allows some parallelism, but it's not as naturally parallel as algorithms like Blake3. There are SIMD-optimized implementations (AVX2, NEON) that provide modest speedups.
> 
>     Hardware support:
> 
>         Intel/AMD laptops: No dedicated SHA-3 instructions in consumer CPUs yet
>         ARM: Some newer ARM cores include SHA-3 acceleration (e.g., Cortex-A78AE, Neoverse)
>         Servers: Limited - mainly in specialized crypto accelerators or newer ARM server chips
> 
>     Software support:
> 
>         Excellent across all platforms (OpenSSL, libsodium, native crypto libraries)
>         Performance is generally 2-4x slower than SHA-256 on x86 without hardware acceleration
> 
>     Practical considerations:
>     If you need maximum performance on current laptop/server hardware, SHA-256 or Blake3 are better choices. Blake3 offers excellent parallelism and is often faster than both SHA-2 and SHA-3 on modern CPUs.
>     SHA3-256 is primarily chosen for its different security foundation (sponge vs Merkle-Damgård) rather than performance advantages.
> 

https://claude.ai/public/artifacts/38aec2b4-9c7a-4477-a7c5-620eb883b368

thomass-dev:
To confirm what I said IRL to @retromechs, this issue will result to two visions:

- short-term, without breaking the API and with as little effort as possible,
- long-term, the most scalable solution.

[...]

PR #1891, opened 2025-07-02
feat(skore-hub-project)!: Send pickled report to S3

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/1878.

This PR changes the way reports are sent to the hub and retrieved from it, with as few changes as possible.

A complete refactor will be done in a future iteration: we don't need the "item" notion anymore, as long as the only supported objects are `skore` reports. It can simplify a lot the code of `skore-hub-project`.


thomass-dev:
@retromechs please take a look for the validity of the workflow (it works well on dev).

skore/src/skore/sklearn/_base.py
```diff
@@ -120,6 +120,7 @@ class _BaseReport(_HelpMixin):
     _X_test: ArrayLike | None
     _y_train: ArrayLike | None
     _y_test: ArrayLike | None
+    _cache: dict[tuple[Any, ...], Any]
```
thomass-dev: @glemaitre No contraindication ?

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -30,6 +37,37 @@ class Metadata(TypedDict):  # noqa: D101
...
+    report._cache = {}
+
+    with io.BytesIO() as stream:
+        try:
+            joblib.dump(report, stream)
+        finally:
+            report._cache = cache
+
+        pickle = stream.getvalue()
+        hasher = Blake3(max_threads=(1 if len(pickle) < 10**6 else Blake3.AUTO))
```
rouk1, line 66: I do not get why you compare with $`1^6`$ (which is ... 1).

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -135,18 +173,52 @@ def put(self, key: str, report: EstimatorReport):
...
+                url=f"projects/{self.tenant}/{self.name}/artefacts",
+                json=[{"checksum": checksum, "content_type": "<default>"}],
+            )
+
+        url = next(iter(response.json()), None)
 
+        if url:
+            # upload pickled report
+            with io.BytesIO(pickle) as stream, Client() as client:
+                client.put(
```
retromechs, line 197: Here, you can use a client that doesn't need authentication. This way, you avoid accidentally sharing the Skore Hub authentication token with the S3 storage. Although this isn't a big security risk right now because we manage the S3 storage, it's still better to be safe.

Extra:  you can check if the upload was successful by looking at the response status before you call the `/complete` endpoint.

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -135,18 +173,52 @@ def put(self, key: str, report: EstimatorReport):
...
 
-        item = item_module.object_to_item(report)
+        # pickle report
+        pickle, checksum = dumps(report)
+
+        # ask for upload url
+        with AuthenticatedClient(raises=True) as client:
+            response = client.post(
+                url=f"projects/{self.tenant}/{self.name}/artefacts",
+                json=[{"checksum": checksum, "content_type": "<default>"}],
```
retromechs: could you use a more specific term like `"report-pickle"` here instead of `"<default>"`? This way, we'll have a bit more context about the artefact at the S3 level.

retromechs (changes requested):
Nice work!
Only minor changes

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: > Here, you can use a client that doesn't need authentication.

This is the case: i used `Client`, not `AuthenticatedClient`.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: Yes no worries, this is just a temporary placeholder (i forgot to ping you about it, sorry).
What is your preference?

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: > Extra: you can check if the upload was successful by looking at the response status before you call the /complete endpoint.

The client automatically makes a `response.raise_for_status()`, so with a response other than 2**, it will quit.

skore-hub-project/src/skore_hub_project/project/project.py
retromechs: I don't have a particular preference, as long as it allows us to easily identify an artefact!

thomass-dev: Let's go for `estimator-report-pickle` for now?

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -155,20 +227,33 @@ def reports(self):
...
-                item_class = getattr(item_module, item_class_name)
-                item_parameters = report["raw"]["parameters"]
-                item = item_class(**item_parameters)
-                return item.__raw__
+            metadata = response.json()
+            checksum = metadata["raw"]["checksum"]
 
+            # ask for read url
             with AuthenticatedClient(raises=True) as client:
                 response = client.get(
```
rouk1, line 247: As a file may be large you should [stream](https://www.python-httpx.org/quickstart/#streaming-responses) the response instead of laoding it in RAM.

skore-hub-project/src/skore_hub_project/project/project.py
retromechs: seems good to me, let's go with this wording for now :+1:

skore-hub-project/tests/unit/project/test_project.py
```diff
@@ -68,20 +69,207 @@ def test_put_exception(self):
...
+                    "category": "model",
+                    "attributes": {},
+                    "parameters": {},
+                    "representation": {"media_type": "text/html", "value": None},
+                },
+            ],
+            "parameters": {"checksum": checksum},
+            "key": "<key>",
+            "run_id": "<run_id>",
+        }
```
rouk1: Why not just checking that th checksum is here ? Comparing the whole structure seems outside of the scope os this particular test.

skore-hub-project/pyproject.toml
```diff
@@ -4,6 +4,7 @@ dynamic = ["license", "readme", "version"]
 requires-python = ">=3.10"
 maintainers = [{ name = "skore developers", email = "skore@signal.probabl.ai" }]
 dependencies = [
+  "blake3",
```
rouk1, line 7: This choice needs to be documented, if we change the hashing algorythme it will be a breaking change. Would you consider writing add an ADR for this ?

rouk1 (changes requested):

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: This was fixed during your review by https://github.com/probabl-ai/skore/pull/1891/commits/934dd79e39a10274f3ba898907aae4dc593ba7b0.

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -30,6 +37,37 @@ class Metadata(TypedDict):  # noqa: D101
...
+        try:
+            joblib.dump(report, stream)
+        finally:
+            report._cache = cache
+
+        pickle = stream.getvalue()
+        hasher = Blake3(max_threads=(1 if len(pickle) < 10**6 else Blake3.AUTO))
+        checksum = hasher.update(pickle).digest()
+
+    return pickle, bytes_to_b64_str(checksum)
```
thomass-dev:
```suggestion
    return pickle, f"blake3-{bytes_to_b64_str(checksum)}"
```

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: Sorry, i don't get it.

I need the entire bytes array to unpickle the report, so in state it have to be in RAM.
I don't see the benefit of streaming, because i should still leave each part in ram and reconstruct the complete bytes array before unpickling it.

Otherwise, you suggest i save the bytes array in a temporary file?

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: Fixed in 1853f960d31beb65fb484e80d3de789141b13fd9.

skore-hub-project/src/skore_hub_project/project/project.py
rouk1: Maybe this is a nitpick indeed, but httpx will mount the response body in RAM then you'll unpickle it. Indeed you need it at the end. But the way it is for now will spike at 2 times the needed amout of RAM.

A temp file could be an elegant way to avoid this. Ignore this comment if the threshold to switch to chunk is tool low to be considered harmful.

thomass-dev: I will change to stream + temp file.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: Fixed in a04cb7db2b7fd4c6ac777360ed876850c4dce110.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: Fixed in 92f5371eb4e13ab63d565164565ecd3b86d11573.

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -250,15 +251,21 @@ def get(id: str) -> EstimatorReport:
...
-            # download pickled report
-            with Client() as client:
-                response = client.get(url=url)
-
-            pickle = response.content
-
-            # unpickle report
-            with io.BytesIO(pickle) as stream:
-                return joblib.load(stream)
+            # Download pickled report before unpickling it.
```
rouk1, line 254: 😻

skore-hub-project/tests/unit/project/test_project.py
thomass-dev: Fixed in d862f0c1f0efd4deb1fc48cd64d8f74c89cbf1a3.

skore-hub-project/pyproject.toml
thomass-dev: > if we change the hashing algorithm it will be a breaking change

There is no real breaking change:
- during the "put" phase, artifact will be sent to the S3 with a hash that has never been met (indeed there is a chance to duplicate data in state but we can had some strategies to avoid it),
- during the "get" phase, there is no impact: the hash is immutable in the hub table and will remain unchanged whenever the hashing algorithm changes or not.  

So changing the hashing algorithm can only cause data duplication if not done intelligently.

Quoting https://github.com/probabl-ai/skore/issues/1808#issuecomment-2976750564, one solution is to maintain both hashing algorithms during the transition: during the "put" phase, we can first search for an existing artifact using the older hashing algorithm, before searching with the new one.

> 
> In addition, it should be interesting to prepend the checksum with its method of compute, such as `sha3-256:<checksum>`.
> It will help the `skore-hub-project` on future retro-compatibility needs.

rouk1: Okay !

retromechs (approved):

PR #1904, opened 2025-07-05
chore: Turn skore.utils as a private module

### LINKED ISSUE #1858: Turn skore.utils as a private module
For full description, see parent issue.   
We want to turn non user-facing module private.   

Transform `skore.utils` to `skore._utils`.

smazoyer:
fix #1858


thomass-dev (approved):
Thanks for your contribution!

Issue #1910, opened 2025-07-08
feat(skore-hub-project): Send a heavy report in several chunks


thomass-dev:
Following https://github.com/probabl-ai/skore/issues/1878.

PR #1911, opened 2025-07-08
feat(skore-hub-project): Send a heavy report in several chunks

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/1910.

---

Tested both on linux and windows.
Tested across scripts, IPython and notebooks.

Also, i tested other back-ends to parallelize IO:
- `asyncio`, issues with notebooks,
- `asyncio` x `threading` (use new event loop in a dedicated thread), works, but i decided to use only `threading` for simplicity,
- `anyio`, same issues as `asyncio`,
- `trio`, lower performance that `threading`.

https://github.com/user-attachments/assets/a20169a9-405f-4350-bff1-d9eb7d940e1e

### DIFF
--- a/skore-hub-project/src/skore_hub_project/project/artefact.py
+++ b/skore-hub-project/src/skore_hub_project/project/artefact.py
@@ -123,7 +123,7 @@ def upload_chunk(
             url=url,
             content=file.read(length),
             headers={"Content-Type": "application/octet-stream"},
-            timeout=None,
+            timeout=30,
         )
 
         return response.headers["etag"]
--- a/skore-hub-project/src/skore_hub_project/project/artefact.py
+++ b/skore-hub-project/src/skore_hub_project/project/artefact.py
@@ -1,3 +1,5 @@
+"""Function definition of the artefact ``upload``."""
+
 from __future__ import annotations
 
 from concurrent.futures import ThreadPoolExecutor, as_completed
@@ -38,7 +40,7 @@ Progress = partial(
 
 
 class Serializer:
-    """Serialize an object using ``joblib``, on disk to avoid RAM overhead."""
+    """Serialize an object using ``joblib``, on disk to reduce RAM footprint."""
 
     def __init__(self, o: Any):
         with self.filepath.open("wb") as file:
@@ -58,7 +60,18 @@ class Serializer:
 
     @cached_property
     def checksum(self) -> str:
-        """The checksum of the serialized object."""
+        """
+        The checksum of the serialized object.
+
+        Notes
+        -----
+        Depending on the size of the serialized object, the checksum can be computed on
+        one or more threads:
+
+            Note that this can be slower for inputs shorter than ~1 MB
+
+        https://github.com/oconnor663/blake3-py
+        """
         hasher = Blake3(max_threads=(1 if self.size < 1e6 else Blake3.AUTO))
         checksum = hasher.update_mmap(self.filepath).digest()
 
@@ -77,7 +90,32 @@ def upload_chunk(
     offset: int,
     length: int,
 ) -> str:
-    """"""
+    """
+    Upload a chunk of the serialized object to the artefacts storage.
+
+    Parameters
+    ----------
+    filepath : ``Path``
+        The path of the file containing the serialized object.
+    client : ``httpx.Client``
+        The client used to upload the chunk to the artefacts storage.
+    url : str
+        The url used to upload the chunk to the artefacts storage.
+    offset : int
+        The start of the chunk in the file containing the serialized object.
+    length: int
+        The length of the chunk in the file containing the serialized object.
+
+    Returns
+    -------
+    etag : str
+        The ETag assigned by the artefacts storage to the chunk, used to acknowledge the
+        upload.
+
+    Notes
+    -----
+    This function is in charge of reading its own chunk to reduce RAM footprint.
+    """
     with filepath.open("rb") as file:
         file.seek(offset)
 
@@ -92,7 +130,31 @@ def upload_chunk(
 
 
 def upload(project: Project, o: Any, type: str, chunk_size: int) -> str:
-    """"""
+    """
+    Upload an object to the artefacts storage.
+
+    Parameters
+    ----------
+    project : ``Project``
+        The project where to upload the object.
+    o : Any
+        The object to upload.
+    type : str
+        The type to associate to object in the artefacts storage.
+    chunk_size : int
+        The size of the object not to be exceeded before to be uploaded in chunks.
+        The size of the chunks to upload.
+
+    Returns
+    -------
+    checksum : str
+        The checksum of the object after upload to the artefacts storage, based on its
+        ``joblib`` serialization.
+
+    Notes
+    -----
+    An object that was already uploaded in its whole will be ignored.
+    """
     with (
         Serializer(o) as serializer,
         AuthenticatedClient(raises=True) as authenticated_client,
@@ -115,11 +177,11 @@ def upload(project: Project, o: Any, type: str, chunk_size: int) -> str:
         if urls := response.json():
             task_to_chunk_id = {}
 
-            # Upload each chunk of the pickled report to the artefacts storage, using
+            # Upload each chunk of the serialized object to the artefacts storage, using
             # disk temporary file.
             #
-            # Each task is in charge of reading its own file chunk at runtime, to avoid
-            # RAM overhead.
+            # Each task is in charge of reading its own file chunk at runtime, to reduce
+            # RAM footprint.
             #
             # Use `threading` over `asyncio` to ensure compatibility with Jupyter
             # notebooks, where the event loop is already running.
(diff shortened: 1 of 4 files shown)


thomass-dev:
@rouk1 @retromechs you can start your review, even if the PR is still in draft. I want to do a manual test on Windows platform.

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -38,35 +44,48 @@ class Metadata(TypedDict):  # noqa: D101
...
-        hasher = Blake3(max_threads=(1 if len(pickle) < 10**6 else Blake3.AUTO))
-        checksum = hasher.update(pickle).digest()
+        size = tmpfile.tell()
+        hasher = Blake3(max_threads=(1 if size < 1e6 else Blake3.AUTO))
+        checksum = hasher.update_mmap(tmpfile.name).digest()
+
+        tmpfile.seek(0)
 
-    return pickle, f"blake3-{bytes_to_b64_str(checksum)}"
+        yield tmpfile, f"blake3-{bytes_to_b64_str(checksum)}", size
```
thomass-dev, line 88: Ensure that is working well on Windows platform.

thomass-dev:
Ready to review @rouk1 @retromechs , thanks.

skore-hub-project/src/skore_hub_project/project/artefact.py
```diff
@@ -0,0 +1,232 @@
...
+    This function is in charge of reading its own chunk to reduce RAM footprint.
+    """
+    with filepath.open("rb") as file:
+        file.seek(offset)
+
+        response = client.put(
+            url=url,
+            content=file.read(length),
+            headers={"Content-Type": "application/octet-stream"},
+            timeout=None,
```
rouk1: You should always have a timeout. Infintie timeout leads to unresponsive programs when internet connexion goes down.

skore-hub-project/src/skore_hub_project/project/artefact.py
```diff
@@ -0,0 +1,232 @@
...
+            json=[
+                {
+                    "checksum": serializer.checksum,
+                    "content_type": type,
+                    "chunk_number": ceil(serializer.size / chunk_size),
+                }
+            ],
+        )
+
+        # If no url, it has been already uploaded.
```
rouk1:
```suggestion
        # If server does not respond an url,
        # then according to it's checksum this chunk 
        # already exists in the storage.
```

skore-hub-project/src/skore_hub_project/project/artefact.py
```diff
@@ -0,0 +1,232 @@
...
+        file.seek(offset)
+
+        response = client.put(
+            url=url,
+            content=file.read(length),
+            headers={"Content-Type": "application/octet-stream"},
+            timeout=None,
+        )
+
+        return response.headers["etag"]
```
rouk1, line 129: Maybe you could implement a retry strategy here ?
If connexion goes down for some reason you may want to retry the upload.

I reccoment to implement this using an exponential backoff algorithm.

skore-hub-project/src/skore_hub_project/project/artefact.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def upload(project: Project, o: Any, type: str, chunk_size: int) -> str:
```
rouk1: ~I think that we should have two functions `chunk_upload` & `atomic_upload` the upload function shall remain and dispatch to the new functions according to a size threshold of the artefact.~
`chunk_size` should be hidden in this module and not known by `Projects` (it could be a const or an env var).
Threshold should be a const I think that @retromechs recommend ~100MB.

rouk1 (changes requested):
Implementation is really elegant ghood job on this !
I think you removed the possibility to do an atomic uplaod for small artefacts though.

skore-hub-project/src/skore_hub_project/project/artefact.py
thomass-dev: Actually, the `chunk_size` is the threshold that you mention, and is defined to ~10MB.
Thus, any object larger than ~10MB is divided into ~10MB chunks.

But if you prefer, we can differentiate the `chunk_size` of the threshold, and define these two values as constant.

thomass-dev: > I think that we should have two functions chunk_upload & atomic_upload the upload function shall remain and dispatch to the new functions according to a size threshold of the artefact.

Sorry, i don't have the point 😅 .
Currently, objects smaller than 10MB are sent as they are in one chunk.
Otherwise, if the rule is to chunk any object that is larger than a threshold, i don't see the use of an explicit "atomic" upload function.

skore-hub-project/src/skore_hub_project/project/artefact.py
thomass-dev: 30 minutes looks like a good tradeoff. Is it good for you?

skore-hub-project/src/skore_hub_project/project/artefact.py
thomass-dev: I will implement a retry strategy on connection error and connection timeout raised by the client.
Is that sufficient or i have to treat some server status codes?

rouk1: You may also want to retry if you receive some status codes:

- 408 Request Timeout
- 425 Too Early
- 429 Too Many Requests
- 502 Bad Gateway
- 503 Service Unavailable
- 504 Gateway Timeout

skore-hub-project/src/skore_hub_project/project/artefact.py
rouk1: Ok as discussed IRL, I did not understood that API to get URL was with a default of 1 chunk.

skore-hub-project/src/skore_hub_project/project/artefact.py
rouk1: This is way to big ! Httpx raises timeout on network inactivity not on the total duration of a request.
You should use very small timeouts (max 30 seconds).

thomass-dev: Fixed in 49c117fd246002366f6b1758c6675d33132272b1.

skore-hub-project/src/skore_hub_project/project/artefact.py
```diff
@@ -0,0 +1,234 @@
...
+    def checksum(self) -> str:
+        """
+        The checksum of the serialized object.
+
+        Notes
+        -----
+        Depending on the size of the serialized object, the checksum can be computed on
+        one or more threads:
+
+            Note that this can be slower for inputs shorter than ~1 MB
```
auguste-probabl, line 71:
```suggestion
        Depending on the size of the serialized object, the checksum can be computed on
        one or more threads. Note that this can be slower for inputs shorter than ~1 MB.
```

thomass-dev: The extra indentation is here because it is a citation from the source.

skore-hub-project/src/skore_hub_project/project/artefact.py
thomass-dev: Postponed to another PR https://github.com/probabl-ai/skore/pull/1931.

skore-hub-project/src/skore_hub_project/project/artefact.py
```diff
@@ -0,0 +1,234 @@
...
+                    "chunk_number": ceil(serializer.size / CHUNK_SIZE),
+                }
+            ],
+        )
+
+        # If no url, it has been already uploaded.
+        if urls := response.json():
+            task_to_chunk_id = {}
+
+            # Upload each chunk of the serialized object to the artefacts storage, using
```
rouk1, line 183:
```suggestion
            # Upload each chunk of the serialized object to the artefacts storage, 
            # using a disk temporary file.
```

rouk1 (dismissed):
Appart some comment which I do not find very understandable this is good to me LGTM !

skore-hub-project/src/skore_hub_project/project/artefact.py
thomass-dev: The line is <89chars, why this change? ^^

skore-hub-project/src/skore_hub_project/project/artefact.py
thomass-dev: Fixed in 483ad143b80aab6a0bd08e5626f7be47e0cd0cf4.

skore-hub-project/src/skore_hub_project/project/artefact.py
thomass-dev: Fixed in 483ad143b80aab6a0bd08e5626f7be47e0cd0cf4.

rouk1 (approved):

PR #1912, opened 2025-07-09
fix(skore): Force the upgrade of python dependencies on `make install-skore`

thomass-dev:
@sylvaincom had troubles caused by the non-automatic update of `skore-local-project`.


sylvaincom (approved):
Thanks!

PR #1924, opened 2025-07-17
feat(skore): Add infrastructure to support different plotting backends for Displays

### LINKED ISSUE #1079: feat(skore): Add plotly as plotting backend for `Display.plot`

divakaivan:
Closes #1079

### DIFF
--- a/skore/src/skore/_config.py
+++ b/skore/src/skore/_config.py
@@ -1,13 +1,19 @@
 """Global configuration state and functions for management."""
 
+from __future__ import annotations
+
 import threading
 import time
 from collections.abc import Generator
 from contextlib import contextmanager
-from typing import Any
+from typing import TYPE_CHECKING, Any
+
+if TYPE_CHECKING:
+    from skore._sklearn.types import PlotBackend
 
 _global_config: dict[str, Any] = {
     "show_progress": True,
+    "plot_backend": "matplotlib",
 }
 _threadlocal = threading.local()
 
@@ -48,7 +54,9 @@ def get_config() -> dict[str, Any]:
 
 
 def set_config(
+    *,
     show_progress: bool | None = None,
+    plot_backend: PlotBackend | None = None,
 ) -> None:
     """Set skore configuration.
 
@@ -61,6 +69,13 @@ def set_config(
     show_progress : bool, default=None
         If True, show progress bars. Otherwise, do not show them.
 
+    plot_backend : {"matplotlib", "plotly"}, default=None
+        The plotting backend to be used.
+
+        - `"matplotlib"`: Use Matplotlib for plotting
+        - `"plotly"`: Use Plotly for plotting
+        - `None`: Plotting backend is unchanged
+
     See Also
     --------
     config_context : Context manager for skore configuration.
@@ -70,18 +85,21 @@ def set_config(
     --------
     >>> # xdoctest: +SKIP
     >>> from skore import set_config
-    >>> set_config(show_progress=False)
+    >>> set_config(show_progress=False, plot_backend="plotly")
     """
     local_config = _get_threadlocal_config()
 
     if show_progress is not None:
         local_config["show_progress"] = show_progress
+    if plot_backend is not None:
+        local_config["plot_backend"] = plot_backend
 
 
 @contextmanager
 def config_context(
     *,
     show_progress: bool | None = None,
+    plot_backend: PlotBackend | None = None,
 ) -> Generator[None, None, None]:
     """Context manager for skore configuration.
 
@@ -93,6 +111,13 @@ def config_context(
     ----------
     show_progress : bool, default=None
         If True, show progress bars. Otherwise, do not show them.
+
+    plot_backend : {"matplotlib", "plotly"}, default=None
+        The plotting backend to be used.
+
+        - `"matplotlib"`: Use Matplotlib for plotting
+        - `"plotly"`: Use Plotly for plotting
+        - `None`: Plotting backend is unchanged
 
     Yields
     ------
@@ -115,7 +140,7 @@ def config_context(
     >>> from sklearn.datasets import make_classification
     >>> from sklearn.linear_model import LogisticRegression
     >>> from skore import CrossValidationReport
-    >>> with skore.config_context(show_progress=False):
+    >>> with skore.config_context(show_progress=False, plot_backend="matplotlib"):
     ...     X, y = make_classification(random_state=42)
     ...     estimator = LogisticRegression()
     ...     report = CrossValidationReport(estimator, X=X, y=y, cv_splitter=2)
@@ -123,6 +148,7 @@ def config_context(
     old_config = get_config()
     set_config(
         show_progress=show_progress,
+        plot_backend=plot_backend,
     )
 
     try:
--- a/skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
+++ b/skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
@@ -15,6 +15,7 @@ from skore._sklearn._plot.style import StyleDisplayMixin
 from skore._sklearn._plot.utils import (
     LINESTYLE,
     HelpDisplayMixin,
+    PlotBackendMixin,
     _ClassifierCurveDisplayMixin,
     _despine_matplotlib_axis,
     _validate_style_kwargs,
@@ -63,7 +64,7 @@ def _add_chance_level(
 
 
 class RocCurveDisplay(
-    StyleDisplayMixin, HelpDisplayMixin, _ClassifierCurveDisplayMixin
+    StyleDisplayMixin, HelpDisplayMixin, _ClassifierCurveDisplayMixin, PlotBackendMixin
 ):
     """ROC Curve visualization.
 
@@ -707,7 +708,7 @@ class RocCurveDisplay(
         return self.ax_, lines, info_pos_label
 
     @StyleDisplayMixin.style_plot
-    def plot(
+    def _plot_matplotlib(
         self,
         *,
         estimator_name: str | None = None,
--- a/skore/src/skore/_sklearn/_plot/utils.py
+++ b/skore/src/skore/_sklearn/_plot/utils.py
@@ -16,7 +16,14 @@ from sklearn.utils.validation import (
     check_consistent_length,
 )
 
-from skore._sklearn.types import MLTask, PositiveLabel, ReportType, YPlotData
+from skore._config import get_config
+from skore._sklearn.types import (
+    MLTask,
+    PlotBackend,
+    PositiveLabel,
+    ReportType,
+    YPlotData,
+)
 
 LINESTYLE = [
     ("solid", "solid"),
@@ -324,6 +331,27 @@ def _adjust_fig_size(
     width = _get_adjusted_fig_size(fig, ax, "width", target_width)
     height = _get_adjusted_fig_size(fig, ax, "height", target_height)
     fig.set_size_inches((width, height))
+
+
+class PlotBackendMixin:
+    """Mixin class for Displays to dispatch plotting to the configured backend."""
+
+    def plot(self, **kwargs):
+        plot_backend = get_config()["plot_backend"]
+        if plot_backend == "matplotlib":
+            return self._plot_matplotlib(**kwargs)
+        elif plot_backend == "plotly":
+            return self._plot_plotly(**kwargs)
+        else:
+            raise NotImplementedError(
+                f"Plotting backend {plot_backend} not available. "
+                f"Available options are {PlotBackend.__args__}."
+            )
+
+    def _plot_plotly(self, **kwargs):
+        raise NotImplementedError(
+            "Plotting with plotly is not supported for this Display."
+        )
 
 
 def _despine_matplotlib_axis(
--- a/skore/src/skore/_sklearn/_plot/base.py
+++ b/skore/src/skore/_sklearn/_plot/base.py
@@ -8,7 +8,7 @@ class Display(Protocol):
     """Protocol specifying the common API for all `skore` displays."""
 
     def plot(self, **kwargs: Any) -> None:
-        """Display a matplotlib figure containing the information of the display."""
+        """Display a figure containing the information of the display."""
 
     def set_style(self, **kwargs: Any) -> None:
         """Set the style of the display."""
--- a/skore/src/skore/_sklearn/_plot/data/table_report.py
+++ b/skore/src/skore/_sklearn/_plot/data/table_report.py
@@ -20,6 +20,7 @@ from skrub._reporting._utils import (
 from skore._sklearn._plot.style import StyleDisplayMixin
 from skore._sklearn._plot.utils import (
     HelpDisplayMixin,
+    PlotBackendMixin,
     _adjust_fig_size,
     _rotate_ticklabels,
     _validate_style_kwargs,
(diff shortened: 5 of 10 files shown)


divakaivan:
`make test` output:  `926 passed, 5 skipped, 909 warnings`

skore/src/skore/_config.py
```diff
@@ -61,6 +63,14 @@ def set_config(
     show_progress : bool, default=None
         If True, show progress bars. Otherwise, do not show them.
 
+    plot_backend : str, default=None
+        The default plotting backend to be used.
+
+        - `"default"`: Default backend for the type of plot
+        - `"matplotlib"`: Use Matplotlib for plotting
+        - `"plotly"`: Use Plotly for plotting
+        - `None`: Plotting backend is unchanged
```
thomass-dev:
```python
# skore/src/skore/_sklearn/_plot/utils.py
class PlotBackendMixin:
    def plot(self, **kwargs):
        plot_backend = get_config()["plot_backend"]
        if plot_backend in ["default", "matplotlib"]:
            return self._plot_matplotlib(**kwargs)
[...]            
```

As i can see in the above snippet, `default` and `matplotlib` are exchangeable.
The `default` cannot be finely defined for each type of display.
So i don't se the point of `default` compared to using `matplotlib`.

IMO, `default` should be removed and replaced by `matplotlib`.

```suggestion
    plot_backend : str | None, default=None
        The plotting backend to be used.

        - `"matplotlib"`: Use Matplotlib for plotting
        - `"plotly"`: Use Plotly for plotting
        - `None`: Plotting backend is unchanged
```

skore/src/skore/_config.py
```diff
@@ -49,6 +50,7 @@ def get_config() -> dict[str, Any]:
 
 def set_config(
     show_progress: bool | None = None,
+    plot_backend: str | None = None,
```
thomass-dev:
```suggestion
    *,
    show_progress: bool | None = None,
    plot_backend: str | None = None,
```

skore/src/skore/_config.py
```diff
@@ -70,18 +80,21 @@ def set_config(
...
     >>> # xdoctest: +SKIP
     >>> from skore import set_config
-    >>> set_config(show_progress=False)
+    >>> set_config(show_progress=False, plot_backend="plotly")
     """
     local_config = _get_threadlocal_config()
 
     if show_progress is not None:
         local_config["show_progress"] = show_progress
+    if plot_backend is not None:
```
thomass-dev, line 94: Nitpick: as `None` is not a valid value for `plot_backend` in this context, we should use ellipsis as a sentinel.

skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
```diff
@@ -837,6 +838,13 @@ def plot(
             if despine:
                 _despine_matplotlib_axis(self.ax_)
 
+    @StyleDisplayMixin.style_plot
+    def _plot_plotly(self, **kwargs):
+        raise NotImplementedError(
+            "Plots with plotly is not implemented. "
+            "Please use one of ['default', 'matplotlib']."
+        )
```
thomass-dev: Why is this necessary (identical for each metric), when mixin defines this same default behavior?

thomass-dev (changes requested):
Nice! The overall structure LGTM, just few thoughts.

skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
divakaivan: Yes. I added the Mixin class after and forgot to remove ~ Will remove

skore/src/skore/_config.py
divakaivan: At the moment default and matplotlib do the same thing. I think the idea was that when `plot_backend` is untouched/set to `default`, some displays will use matplotlib and some plotly (whichever we choose to be best for a particular display). Ultimately ~ `default` will allow to use both type of backends for different displays. Would something like this be better, **if** we want this behaviour of using both type of plots:

For example:

```python
### set an attribute in each class for its default plot backend
class ConfusionMatrixDisplay(PlotBackendMixin, Display):
    def __init__(self, ...):
        self._default_plot_backend = self._plot_matplotlib


class PlotBackendMixin:
    if plot_backend == "default":
        return self._default_plot_backend(**kwargs)
    elif plot_backend == "matplotlib":
        return self._plot_matplotlib(**kwargs)
    elif plot_backend == "plotly":
        return self._plot_plotly(**kwargs)
    else: ...
```

skore/src/skore/_config.py
divakaivan: 1. I can get the mypy to be happy if I set 
```python
def set_config(..., plot_backend: str | Any = ...:
    if plot_backend is not ...:
        local_config["plot_backend"] = plot_backend
```
But this would be using Any

2. `from types import EllipsisType` ([added in Python 3.10](https://stackoverflow.com/questions/66636262/get-python-ellipsis-type-without-having-to-define-it-as-type))
```python
def set_config(..., plot_backend: str | EllipsisType = ...):
    if plot_backend is not ...:
        local_config["plot_backend"] = plot_backend
```

3.  only accepting `str`:
```python
def set_config(..., plot_backend: str = "default"):
    if plot_backend != "default":
        local_config["plot_backend"] = plot_backend
```

skore/src/skore/_config.py
auguste-probabl, line 1: Define a new type `PlotBackend = Literal["default", "matplotlib", "plotly"]`

skore/src/skore/_config.py
auguste-probabl: I also think that `default` is superfluous as it feels indistinguishable from `None`

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -242,6 +243,29 @@ def _validate_from_predictions_params(
...
+    def plot(self, **kwargs):
+        plot_backend = get_config()["plot_backend"]
+        if plot_backend in ["default", "matplotlib"]:
+            return self._plot_matplotlib(**kwargs)
+        elif plot_backend == "plotly":
+            return self._plot_plotly(**kwargs)
+        else:
+            raise NotImplementedError(
+                f"Plotting backend {plot_backend} not available. "
+                "Available options are ['default', 'matplotlib', 'plotly']."
```
auguste-probabl: don't hard-code this; define the correct values in the same place as the new `PlotBackend` type

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -242,6 +243,29 @@ def _validate_from_predictions_params(
...
+            )
+
+    def _plot_matplotlib(self, **kwargs):
+        raise NotImplementedError("_plot_matplotlib must be implemented by subclass.")
```
auguste-probabl:
```suggestion
        raise NotImplementedError("Plotting with matplotlib is not supported.")
```

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -242,6 +243,29 @@ def _validate_from_predictions_params(
...
+
+    def _plot_plotly(self, **kwargs):
+        raise NotImplementedError(
+            "Plots with plotly is not implemented. "
```
auguste-probabl:
```suggestion
            "Plotting with plotly is not supported. "
```

skore/src/skore/_sklearn/_plot/utils.py
divakaivan: I created a `PlotBackend` as suggested, and used `PlotBackend.__args__` to list the options. Found it in [this stackoverflow thread](https://stackoverflow.com/questions/48572831/how-to-access-the-type-arguments-of-typing-generic/50101934#50101934)

skore/src/skore/_config.py
```diff
@@ -48,7 +54,9 @@ def get_config() -> dict[str, Any]:
 
 
 def set_config(
+    *,
```
auguste-probabl, line 57: nice find!

skore/src/skore/_config.py
```diff
@@ -61,6 +69,13 @@ def set_config(
     show_progress : bool, default=None
         If True, show progress bars. Otherwise, do not show them.
 
+    plot_backend : str, default="matplotlib"
```
auguste-probabl:
```suggestion
    plot_backend : {"matplotlib", "plotly"}, default="matplotlib"
```

skore/src/skore/_config.py
```diff
@@ -61,6 +69,13 @@ def set_config(
...
+        The default plotting backend to be used.
```
auguste-probabl:
```suggestion
        The plotting backend to be used.
```

skore/src/skore/_config.py
auguste-probabl: I don't see what could go wrong here; just in case, if we do something I'd prefer using `skore._sklearn.types._DEFAULT` rather than ellipsis.

skore/src/skore/_config.py
```diff
@@ -94,6 +112,13 @@ def config_context(
     show_progress : bool, default=None
         If True, show progress bars. Otherwise, do not show them.
 
+    plot_backend : str, default="matplotlib"
```
auguste-probabl: See above

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -242,6 +249,28 @@ def _validate_from_predictions_params(
...
+        elif plot_backend == "plotly":
+            return self._plot_plotly(**kwargs)
+        else:
+            raise NotImplementedError(
+                f"Plotting backend {plot_backend} not available. "
+                f"Available options are {PlotBackend.__args__}."
+            )
+
+    def _plot_matplotlib(self, **kwargs):
+        raise NotImplementedError("Plotting with matplotlib is not supported.")
```
auguste-probabl: Since matplotlib is the preferred implementation for now, I think this base case shouldn't exist -- if _plot_matplotlib is not implemented then something has gone terribly wrong.

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -242,6 +249,28 @@ def _validate_from_predictions_params(
...
+
+    def _plot_plotly(self, **kwargs):
+        raise NotImplementedError(
+            "Plotting with plotly is not supported. Please use one of ['matplotlib']."
```
auguste-probabl:
```suggestion
            "Plotting with plotly is not supported for this Display."
```
just to be clear that it might be supported for other displays. Eventually I guess we'd need a catalogue of which Displays support which plotting backends, and this error message would refer to that.

skore/src/skore/_config.py
```diff
@@ -94,6 +112,13 @@ def config_context(
     show_progress : bool, default=None
         If True, show progress bars. Otherwise, do not show them.
 
+    plot_backend : {"matplotlib", "plotly"}, default="matplotlib"
+        The default plotting backend to be used.
```
auguste-probabl: here also
```suggestion
        The plotting backend to be used.
```
I would like something like DocStrands to synchronize the docstrings...

divakaivan: that's nice. recently published - [DocStrands repo](https://github.com/multimeric/DocStrands)

skore/src/skore/_config.py
```diff
@@ -61,6 +69,13 @@ def set_config(
     show_progress : bool, default=None
         If True, show progress bars. Otherwise, do not show them.
 
+    plot_backend : {"matplotlib", "plotly"}, default="matplotlib"
```
thomass-dev:
```suggestion
    plot_backend : {"matplotlib", "plotly"}, default=None
```

I don't think it is the right place to mention that the default backend is `matplotlib`.

skore/src/skore/_config.py
```diff
@@ -94,6 +112,13 @@ def config_context(
...
```
thomass-dev:
```suggestion
    plot_backend : {"matplotlib", "plotly"}, default=None
```

I don't think it is the right place to mention that the default backend is matplotlib.

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -242,6 +249,27 @@ def _validate_from_predictions_params(
...
+        else:
+            raise NotImplementedError(
+                f"Plotting backend {plot_backend} not available. "
+                f"Available options are {PlotBackend.__args__}."
+            )
+
+    def _plot_plotly(self, **kwargs):
+        raise NotImplementedError(
+            "Plotting with plotly is not supported for this Display."
+        )
```
thomass-dev, line 354: Just my 2cts @auguste-probabl @divakaivan : after re-reading the PR, i find it quite strange to have both the `Display` protocol and the `PlotBackendMixin` mixin, with almost the same purpose.

What about merging the two in a new abstract class `Display`?

divakaivan: I can see them being merged. Auguste mentioned something about a possible restructuring of the Display later on. I think he’s looking at this from more of a bird’s-eye view than I am so curiouis of his PoV ~

auguste-probabl: I'd like to keep the two separate. The two classes don't serve the same purpose: Display is an interface while PlotBackendMixin is an implementation.

auguste-probabl:
@divakaivan Can you clarify the commit message? This PR doesn't actually make it possible to plot via plotly unfortunately.

skore/src/skore/_sklearn/_plot/utils.py
thomass-dev: I disagree. 

As long as there is only one `plot` implementation (what i see), the interface `Display` and the implementation `PlotBackendMixin` have the same purpose: defining the behavior of the display object. Having them in one place improves readability. IMO, there is no problem to have function definition in an interface. Something like: 

```python
from abc import ABC

class Display(ABC):  # DisplayMixin
    def plot(): ...
    def plot_plotly(): raise NotImplementedError
    def plot_matplotlib(): raise NotImplementedError    
    
    @abstractmethod
    def set_style(self, **kwargs): ...
    
    @abstractmethod
    def frame(self, **kwargs): ...
```

It's not blocking, just an architectural question. Feel free to resolve the conversation @auguste-probabl.

thomass-dev (approved):

glemaitre:
Just catching up here.

Regarding the `"default"` case, it is exactly what @divakaivan mentioned: we want a display to define its default. So why do we need `None`: in case of nested configuration where we need to detect if we set `"default"` that semantically means choose the default of the display or `None` that semantically means "do not change anything".

From what I recall, without the two options, you are not going to properly detect thing in a nested case.

To be explicit:

```python
with config_context(plot_backend=None):  
    with config_context(plot_backend="plotly"):
        print(get_config()["plot_backend"])
```

There is no way that I revert it to the `"default"` backend, while what I wanted is:


```python
with config_context(plot_backend="default"):  
    with config_context(plot_backend="plotly"):
        print(get_config()["plot_backend"])
```

glemaitre:
I open https://github.com/probabl-ai/skore/issues/1951 to avoid continuing the discussion in a merged PR ;)

Issue #1451, opened 2025-03-21
docs: Add a technical example showing scikit-learn API compatible estimators that work with skore


sylvaincom:
*Note*: following https://github.com/probabl-ai/skore/pull/1814#issuecomment-2961525382 in June 2025, this issue has broadened its scope from just deep learning to any estimator compatible with the scikit-learn API

### Which part of the documentation needs improvement?

Examples gallery

### Describe the problem found in the documentation

No deep learning methods, no foundations models, etc are mentioned in our examples, mainly traditional scikit-learn models, while skore can deal with basically any estimator that is compatible with the scikit-learn API.

### Suggested improvement

Add an example in the technical section

Its goal is to explain that you can pass to skore any model that has a predict method; and use skorch, keras (see https://github.com/keras-team/keras/pull/20599), tabpfn, etc, a manual wrapped model and explain a bit the internal details to show when you will get some guidance with skore (as the choice of metrics)

### Additional context

Could also do a wrapper for a LLM, as done in https://medium.com/capgemini-invent-lab/quantifying-llms-uncertainty-with-conformal-predictions-567870e63e00

MarieSacksick:
@warith-harchaoui, @emmatysinger, would you be willing to take this one? Should we plan a discussion about it?

MarieSacksick:
I heard about it, but the output is not clear to me; should this be done, and if yes, by whom?

nkapila6:
Hi @sylvaincom, ref our discussion at https://github.com/probabl-ai/skore/pull/1557

I don't have access to the minutes of meeting but looking at your code [here](https://github.com/probabl-ai/skore/blob/1a5bc1f2cb1cd3ac1caea838fa9269ea3647e724/examples/use_cases/plot_resnet.py), it seems like you're reporting top-k scores for k=1,5,10 using `skore` on a pre-trained ResNet.

Would you like me to implement the same using skorch?

sylvaincom:
Hi @nkapila6, thanks for following this up. No need to have access to the minutes, we were just discussing potential deep learning use cases. Sure, you can try something with skorch. Beware of the memory usage and processing time: the dataset must be not too large but still interesting (not a synthetic dataset) and the example is actually executed on our documentation. Let me know if you need any more information. You can also directly try something on the same datasets used in the examples of skorch, but applying skore on it.

nkapila6:
Hey @sylvaincom

How does this look? 
[...]

PR #1814, opened 2025-06-07
docs(usecases): deep learning example with skorch.

nkapila6:
Following up on #1451.

This PR adds a deep learning example with skorch. I would have loved to showcase a CNN + MNIST example but I understand these are being executed during the build and hence, did not go that route.

You can see the notebook output here: https://github.com/nkapila6/skorch-skore-example/blob/main/skore.ipynb

I added some feature pre-processing and dropping steps with explanations and the building of the model using PyTorch+Skorch for Skore to utilize.

Let me know your thoughts! @sylvaincom

PR #1927, opened 2025-07-21
docs: Add a technical example showing scikit-learn API compatible estimators that work with skore

sylvaincom:
Fix https://github.com/probabl-ai/skore/issues/1451

Supersedes https://github.com/probabl-ai/skore/pull/1814 that is too long but archived for safe-keeping

### DIFF
--- a/examples/technical_details/plot_sklearn_api.py
+++ b/examples/technical_details/plot_sklearn_api.py
@@ -129,7 +129,7 @@ class CustomClassifier(ClassifierMixin, BaseEstimator):
 # .. note::
 #
 #   The estimator above does not have a `predict_proba` method, therefore
-#   we cannot display the ROC as above.
+#   we cannot display its ROC curve as done previously.
 
 # %%
 # We can now use this model with skore:
--- a/examples/technical_details/plot_sklearn_api.py
+++ b/examples/technical_details/plot_sklearn_api.py
@@ -15,30 +15,28 @@ been fitted.
 
 .. note::
 
-  When computing the ROC AUC or ROC curve
-  for a classification task, the estimator must have a ``predict_proba`` method.
+  When computing the ROC AUC or ROC curve for a classification task, the estimator must
+  have a ``predict_proba`` method.
 
 In this example, we showcase a gradient boosting model
 (`XGBoost <https://github.com/dmlc/xgboost>`_) and a custom estimator.
 
-.. note::
+Note that this example is not exhaustive; many other scikit-learn compatible models can
+be used with skore:
 
-    This example is not exhaustive; many other scikit-learn compatible models can be
-    used with skore:
+-   More gradient boosting libraries like
+    `LightGBM <https://github.com/microsoft/LightGBM>`_, and
+    `CatBoost <https://github.com/catboost/catboost>`_,
 
-    -   More gradient boosting libraries like
-        `LightGBM <https://github.com/microsoft/LightGBM>`_, and
-        `CatBoost <https://github.com/catboost/catboost>`_,
+-   Deep learning frameworks such as `Keras <https://github.com/keras-team/keras>`_
+    and `skorch <https://github.com/skorch-dev/skorch>`_ (a wrapper for
+    `PyTorch <https://github.com/pytorch/pytorch>`_).
 
-    -   Deep learning frameworks such as `Keras <https://github.com/keras-team/keras>`_
-        and `skorch <https://github.com/skorch-dev/skorch>`_ (a wrapper for
-        `PyTorch <https://github.com/pytorch/pytorch>`_).
+-   Tabular foundation models such as
+    `TabICL <https://github.com/soda-inria/tabicl>`_ and
+    `TabPFN <https://github.com/PriorLabs/TabPFN>`_,
 
-    -   Tabular foundation models such as
-        `TabICL <https://github.com/soda-inria/tabicl>`_ and
-        `TabPFN <https://github.com/PriorLabs/TabPFN>`_,
-
-    -   etc.
+-   etc.
 
 """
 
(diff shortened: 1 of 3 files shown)


examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+Using skore with scikit-learn compatible estimators
+=====================================================================================
+
+This example shows how to use skore with scikit-learn compatible estimators.
+For these models, skore provides a unified interface for reporting and evaluation.
+Any model that can be used with the scikit-learn API can be used with skore.
+Skore's :class:`~skore.EstimatorReport` can be used to report on any estimator
+that has a ``fit`` and ``predict`` method.
+In fact, skore only requires the ``predict`` method if the estimator has already
+been fitted.
```
MarieSacksick:
```suggestion
This example shows how to use skore with scikit-learn compatible estimators.

Any model that can be used with the scikit-learn API can be used with skore.
Skore's :class:`~skore.EstimatorReport` can be used to report on any estimator
that has a ``fit`` and ``predict`` method.
In fact, skore only requires the ``predict`` method if the estimator has already
been fitted.
```
To simplify, otherwise it's both redundant and misleading: "for these models", sounds like not for the others.

examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+
+.. note::
+
+  When using an :class:`~skore.EstimatorReport` to compute the ROC AUC and ROC curve
+  for a classification task, the estimator must have a ``predict_proba`` method.
```
MarieSacksick:
```suggestion
  When computing the ROC AUC or ROC curve
  for a classification task, the estimator must have a ``predict_proba`` method.
```
because it will be the case for comparison and crossvalidation reports too.

examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+
+In this example, we showcase a gradient boosting model
+(`XGBoost <https://github.com/dmlc/xgboost>`_) and a custom estimator.
+
+.. note::
```
MarieSacksick: I don't think the highlight with note is required here. It attracts a lot the attention of the user, while it's just a list.

examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+from sklearn.datasets import make_classification
+
+X, y = make_classification(
+    n_samples=1_000,
+    n_features=5,
+    n_classes=2,
+    class_sep=0.3,
+    n_clusters_per_class=1,
+    random_state=42,
+)
```
MarieSacksick: As in the previous examples, let's keep it simple here: you can remove at the very least the `n_classes` argument, and I would argue everything except the random_state: we want to show the interface, at no moment the actual performance.

examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+print(f"{X.shape = }")
+
+# %%
+# We split our data:
+
+# %%
+from skore import train_test_split
+
+split_data = train_test_split(X, y, shuffle=True, random_state=42, as_dict=True)
```
MarieSacksick: why is shuffle explicitely True, since it's the default, plus it triggers a warning?  Let's either remove it, or pass it to False.

examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+#
+# For this binary classification task, we consider a gradient-boosted decision trees
+# model from a library external to scikit-learn.
+# One of the most popular is `XGBoost <https://github.com/dmlc/xgboost>`_.
+
+# %%
+from skore import EstimatorReport
+from xgboost import XGBClassifier
+
+xgb = XGBClassifier(n_estimators=50, max_depth=3, learning_rate=0.1, random_state=42)
```
MarieSacksick, line 76: when hovering `XGBClassifier`, the docs redirects to https://scikit-learn.org/stable/modules/generated/sklearn.base.ClassifierMixin.html#sklearn.base.ClassifierMixin; it should redirect to https://xgboost.readthedocs.io/en/latest/python/python_api.html#xgboost.XGBClassifier instead, if anything.

image

examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+        check_is_fitted(self)
+        X = validate_data(self, X, reset=False)
+        closest = np.argmin(euclidean_distances(X, self.X_), axis=1)
+        return self.y_[closest]
+
+
+# %%
+# .. note::
+#
+#   The estimator above does not have a `predict_proba` method.
```
MarieSacksick:
```suggestion
# .. note::
#
#   The estimator above does not have a `predict_proba` method, therefore 
#   we cannot display the ROC as above.
```

examples/technical_details/plot_sklearn_api.py
```diff
@@ -0,0 +1,170 @@
...
+# %%
+custom_report = EstimatorReport(CustomClassifier(), pos_label=1, **split_data)
+custom_report.metrics.precision()
+
+# %%
+# Conclusion
+# ==========
+#
+# This example demonstrates how skore can be used with scikit-learn compatible estimators.
+# Any estimator that follows the scikit-learn API can be integrated with skore for evaluation and reporting.
```
MarieSacksick:
```suggestion
# This example demonstrates how skore can be used with scikit-learn compatible estimators.
```

The second sentence is redundant.

MarieSacksick (changes requested):

examples/technical_details/plot_sklearn_api.py
sylvaincom: Ok, done

examples/technical_details/plot_sklearn_api.py
sylvaincom: Ok, done

examples/technical_details/plot_sklearn_api.py
sylvaincom: Ok, done

examples/technical_details/plot_sklearn_api.py
sylvaincom: Ok, done

MarieSacksick (approved):
Thanks, good for me :)!

PR #1931, opened 2025-07-24
feat(skore-hub-project): Add a retry strategy to client based on exponential backoff algorithm

thomass-dev:
> Maybe you could implement a retry strategy here ?
> If connexion goes down for some reason you may want to retry the upload.
> 
> I reccoment to implement this using an exponential backoff algorithm.

_Originally posted by @rouk1 in https://github.com/probabl-ai/skore/pull/1911#discussion_r2224599273_

---

Mainly based on [urllib3.util.Retry](https://urllib3.readthedocs.io/en/stable/reference/urllib3.util.html#urllib3.util.Retry) and inspired by [httpx-retries](https://github.com/will-ockmore/httpx-retries).

### DIFF
--- a/skore-hub-project/src/skore_hub_project/client/api.py
+++ b/skore-hub-project/src/skore_hub_project/client/api.py
@@ -1,19 +1,9 @@
 """API used to exchange with ``skore hub``."""
 
 from datetime import datetime
-from functools import partial
-from os import environ
 from time import sleep
-from urllib.parse import urljoin
 
-import httpx
-
-URI = environ.get("SKORE_HUB_URI", "https://api.skore.probabl.ai")
-Client = partial(
-    httpx.Client,
-    follow_redirects=True,
-    event_hooks={"response": [httpx.Response.raise_for_status]},
-)
+from httpx import HTTPError, TimeoutException
 
 
 def get_oauth_device_login(success_uri: str | None = None):
@@ -39,10 +29,12 @@ def get_oauth_device_login(success_uri: str | None = None):
         - user_code: str
             The user code that needs to be entered on the authorization page
     """
-    url = urljoin(URI, "identity/oauth/device/login")
+    from skore_hub_project.client.client import HUBClient
+
+    url = "identity/oauth/device/login"
     params = {"success_uri": success_uri} if success_uri is not None else {}
 
-    with Client() as client:
+    with HUBClient(authenticated=False) as client:
         response = client.get(url, params=params).json()
 
         return (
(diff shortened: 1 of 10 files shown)


thomass-dev:
@rouk1 ready to review!

skore-hub-project/src/skore_hub_project/client/api.py
```diff
@@ -39,10 +29,12 @@ def get_oauth_device_login(success_uri: str | None = None):
         - user_code: str
             The user code that needs to be entered on the authorization page
     """
-    url = urljoin(URI, "identity/oauth/device/login")
+    from skore_hub_project.client.client import HUBClient
```
rouk1, line 32: Why do you import this in each function instead of a top level import ?

skore-hub-project/src/skore_hub_project/client/client.py
```diff
@@ -3,34 +3,160 @@
...
         5: "Server error",
     }
 
-    def __init__(self, *, raises=False):
-        super().__init__(follow_redirects=True, timeout=300)
+    def __init__(
+        self,
+        *,
+        raises: bool = True,
+        retry: bool = True,
```
rouk1: That is personal test btu when naming boolean I like to let the user knows that it is a boolean value.
- `raises` could be `is_raising`
- `retry` could be `is_retrying`

This more readable to me especially when writing conditions.

skore-hub-project/src/skore_hub_project/client/client.py
```diff
@@ -3,34 +3,160 @@
...
+                    or (response.status_code not in self.RETRYABLE_STATUS_CODES)
+                ):
+                    break
+
+            timeout = self.retry_backoff_factor * (2**retries)
+            retries += 1
+
+            sleep(min(timeout, self.retry_backoff_max))
+
+        if self.raises and not response.is_success:
```
rouk1: Why would you want not to raise ?

rouk1 (dismissed):
Nice work, a few small questions but it is overall good to me.

skore-hub-project/src/skore_hub_project/client/client.py
thomass-dev: If you want to treat the status code separately, without the need to try/catch (its the default behaviour of `httpx.Client`). This behavior is left untouched in this PR.

skore-hub-project/src/skore_hub_project/client/api.py
thomass-dev: To avoid circular imports: api uses client, but client uses token that uses api.

skore-hub-project/src/skore_hub_project/client/client.py
thomass-dev: On the contrary, I prefer it to be an action verb... a matter of preference.

skore-hub-project/src/skore_hub_project/client/client.py
thomass-dev: But actually, `raises=False` is never used for now. Do you want me to delete this feature?

thomass-dev: Let's remove it, the logic will be simpler.

thomass-dev: Changed in c2535f054d7cdc2afca2b5a61676aa83b3ea0470.

skore-hub-project/src/skore_hub_project/client/client.py
rouk1: Then retries if raises : )

PR #1947, opened 2025-07-30
docs: Rework the README file such that it does need minimum revision

### LINKED ISSUE #1762: Rework the README file such that it does need minimum revision

Right now, the README file required update at each release. However, the only goal of the README in the GitHub project is to redirect to the right documentation entry point that should almost never change, e.g. getting started, how to contribute, how to install, social media relay, etc.

I think that it would be a real improvement to change the structure of the README to not have one more document to be in sync.

sylvaincom:
Fix #1762 

See the resulting README [here](https://github.com/probabl-ai/skore/tree/1762-rework-the-readme-file-such-that-it-does-need-minimum-revision?tab=readme-ov-file#own-your-data-science)

Co-authored-by: tuscland <cam@probabl.ai>

Done:
- [x] remove the part about the skore project and the put
- [x] some nitpicks
- [x] insist on _why_ rather than _what_ (no listing of the features)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ When it comes to data science, you have excellent tools at your disposal: `panda
 
 **But here's the challenge**: Your experience is key to choosing the right building blocks and methodologies. You often spend significant time navigating documentation, writing boilerplate code for common evaluations, and struggling to maintain clear project structure.
 
-**Skore is the cherry on top** that transforms your data science pipeline into structured, meaningful artifacts. It reduces the time you spend on documentation navigation, eliminates boilerplate code, and guides you toward the right methodological information to answer your questions.
+**Skore is the conductor** that transforms your data science pipeline into structured, meaningful artifacts. It reduces the time you spend on documentation navigation, eliminates boilerplate code, and guides you toward the right methodological information to answer your questions.
 
 ### What Skore does for you:
 


MarieSacksick (changes requested):
This issue is not (only) about removing the code. It's about a whole revamp of the readme.  
The goal is to present the purpose of skore, instead of presenting a list of features. It requires some product synthesis: what are the pain points that we want to remove from the shoulders of our users? How do we provide this value? Once you have the answer to these two questions, the last step will be to write down the couple of lines of code that deliver the first elements of value.

sylvaincom:
From what I recall from a live discussion with @glemaitre and @auguste-probabl, the issue was mostly about the code snippets shared in the README: as no automatic tests are ran to check if that code is still up to date, it's quite time-consuming to check it by hand after each release.

Up to now, following discussions with @tuscland, we have a list of features to present Skore Lib. Same as the list of features could change one day, a synthetic description of skore's purpose would change as well if our product takes a slightly different direction.

MarieSacksick:
That is actually the bet: the product reached a maturity where it will not change a lot. Improvement can always be done certainly, but it should have reached something stable.  
We want to change from a list of "what" to the "why".

sylvaincom:
Ok I'll try to draft something on the "why" based on what Guillaume has done in the user guide

README.md
```diff
@@ -55,9 +56,9 @@ We recommend using a [virtual environment (venv)](https://docs.python.org/3/tuto
 
 Then, you can install skore by using `pip`:
 ```bash
-# If you just use skore locally
+# If you plan to use skore only locally
```
tuscland: I would use capital S to refer to the product.

> If you plan to use Skore locally

README.md
```diff
@@ -55,9 +56,9 @@ We recommend using a [virtual environment (venv)](https://docs.python.org/3/tuto
...
 pip install -U skore
-# If you wish to also interact with the skore hub
+# If you wish to also interact with skore hub
```
tuscland: And here as well

> If you wish to interact with Skore Hub as well

README.md
```diff
@@ -71,81 +72,43 @@ conda install conda-forge::skore
...
+# See what insights are available
 cv_report.help()
-```
 
-```python
-# Display the report metrics that was computed for you:
-metrics_summary = cv_report.metrics.summarize()
-metrics_summary
-```
+# Access the metrics summary
```
tuscland: Metrics summary is an insight example amongst others, so I would suggest:

> # Example: Access the metrics summary

README.md
```diff
@@ -71,81 +72,43 @@ conda install conda-forge::skore
...
+metrics_summary = cv_report.metrics.summarize().frame()
 
-```python
-# Display the ROC curve that was generated for you:
+# Get the ROC curve
```
tuscland: Same for the ROC curve:

> # Example: Get the ROC curve

sylvaincom:
@MarieSacksick could you please have a look? You had requested changes

README.md
```diff
@@ -23,27 +23,28 @@ Elevate ML Development with Built-in Recommended Practices \
...
-- **Skore Hub**: the collaborative layer where teams connect, learn more on our [product page](https://probabl.ai/skore).
+When it comes to data science, you have excellent tools at your disposal: `pandas` and `polars` for data exploration, `skrub` for stateful transformations, and `scikit-learn` for model training and evaluation. These libraries are designed to be generic and accommodate a wide range of use cases.
 
-⭐ Support us with a star and spread the word - it means a lot! ⭐
+**But here's the challenge**: Your experience is key to choosing the right building blocks and methodologies. You often spend significant time navigating documentation, writing boilerplate code for common evaluations, and struggling to maintain clear project structure.
 
-### Key features of Skore Lib
+**Skore is the cherry on top** that transforms your data science pipeline into structured, meaningful artifacts. It reduces the time you spend on documentation navigation, eliminates boilerplate code, and guides you toward the right methodological information to answer your questions.
```
MarieSacksick: "skore is the cherry on the top" does not convey the idea that it's a must have.  
Can you change it to something more impactful? Like "skore is the glue you need between all these tools" for instance.

MarieSacksick (changes requested):
it's much better, thank you!!  
I just have one request.

README.md
tuscland: I suggest "Skore is the conductor […]"

tuscland: The conductor metaphor works really well because:
- A conductor coordinates multiple instruments (tools) to create harmony
- It implies active guidance and direction, not just passive connection
- It suggests expertise and intelligence in knowing when and how to bring different elements together
- It positions Skore as essential: an orchestra without a conductor can function, but not nearly as effectively

sylvaincom: I applied "Skore is the conductor […]"

PR #1949, opened 2025-07-30
chore: Correct test to be compliant with several skrub versions

### LINKED ISSUE #1954: tests: Two tests failing on a fresh setup of the repo

I did a fresh pull and setup of the repo, and found these two tests failing. 

Result of `make test` after a fresh pull and setup is below.

<details>
<summary><h3>The tests in question</h3></summary>
<br>

```python
# tests/unit/sklearn/plot/table_report/test_estimator.py

def test_json_dump(display):
    json_dict = json.loads(display._to_json())
    assert list(json_dict.keys()) == [
        "dataframe_module",
        "n_rows",
        "n_columns",
        "columns",
        "dataframe_is_empty",
        "plots_skipped",
        "associations_skipped",
        "n_constant_columns",
        "top_associations",
    ]

# ======== AND ========

def test_table_report_display_constructor(display):
    """Check the value that are stored in the display constructor."""
    assert isinstance(display, Display)

    assert hasattr(display, "summary")
    assert isinstance(display.summary, dict)
    assert list(display.summary.keys()) == [
        "dataframe",
        "dataframe_module",
        "n_rows",
        "n_columns",
        "columns",
        "dataframe_is_empty",
        "plots_skipped",
        "associations_skipped",
        "sample_table",
        "n_constant_columns",
        "top_associations",
    ]

```

</details>

```
FAILED tests/unit/sklearn/plot/table_report/test_estimator.py::test_json_dump - AssertionError: assert ['dataframe_module', 'n_rows', 'n_columns', 'columns', 'dataframe_is_empty', 'plots_skipped', 'associations_skipped', 'n_constant_columns', 'top_associations'] == ['dataframe_module', 'n_rows', 'n_columns', 'columns', 'dataframe_is_empty', 'plots_skipped', 'n_constant_columns', 'top_associations']

  At index 6 diff: 'associations_skipped' != 'n_constant_columns'
  Left contains one more item: 'top_associations'

  Full diff:
    [
        'dataframe_module',
        'n_rows',
        'n_columns',
        'columns',
        'dataframe_is_empty',
        'plots_skipped',
  +     'associations_skipped',
        'n_constant_columns',
        'top_associations',
    ]
FAILED tests/unit/sklearn/plot/table_report/test_estimator.py::test_table_report_display_constructor - AssertionError: assert ['dataframe', 'dataframe_module', 'n_rows', 'n_columns', 'columns', 'dataframe_is_empty', 'plots_skipped', 'associations_skipped', 'sample_table', 'n_constant_columns', 'top_associations'] == ['dataframe', 'dataframe_module', 'n_rows', 'n_columns', 'columns', 'dataframe_is_empty', 'plots_skipped', 'sample_table', 'n_constant_columns', 'top_associations']

  At index 7 diff: 'associations_skipped' != 'sample_table'
  Left contains one more item: 'top_associations'

  Full diff:
    [
        'dataframe',
        'dataframe_module',
        'n_rows',
        'n_columns',
        'columns',
        'dataframe_is_empty',
        'plots_skipped',
  +     'associations_skipped',
        'sample_table',
        'n_constant_columns',
        'top_associations',
    ]
```

 Does "associations_skipped" need to be added? Maybe related to #1805  ?

MarieSacksick:
With current assertions, we are actually testing the `summarize_dataframe` function of skrub (cf [code](https://github.com/skrub-data/skrub/blob/693bff1afb3b87832c242f858be605cfbc87c47b/skrub/_reporting/_summarize.py#L12)), which changes over versions.  
For instance, they added with_associations parameter to the function, and the same key to the output: therefore these tests are ok with skrub 0.5.4, but fail with 0.6.0, while it doesn't affect skore.  

Fixes #1954

### DIFF
--- a/skore/tests/unit/sklearn/plot/table_report/test_estimator.py
+++ b/skore/tests/unit/sklearn/plot/table_report/test_estimator.py
@@ -379,7 +379,7 @@ def test_corr_plot(pyplot, estimator_report):
 
 def test_json_dump(display):
     json_dict = json.loads(display._to_json())
-    assert isinstance(json_dict, dict)
+    assert set(json_dict.keys()).issubset(display.summary.keys())
 
 
 def test_repr(display):
--- a/skore/tests/unit/sklearn/plot/table_report/test_estimator.py
+++ b/skore/tests/unit/sklearn/plot/table_report/test_estimator.py
@@ -379,7 +379,7 @@ def test_corr_plot(pyplot, estimator_report):
 
 def test_json_dump(display):
     json_dict = json.loads(display._to_json())
-    assert set(json_dict.keys()).issubset(display.summary.keys())
+    assert isinstance(json_dict, dict)
 
 
 def test_repr(display):
--- a/skore/tests/unit/sklearn/plot/table_report/test_estimator.py
+++ b/skore/tests/unit/sklearn/plot/table_report/test_estimator.py
@@ -188,18 +188,6 @@ def test_table_report_display_constructor(display):
 
     assert hasattr(display, "summary")
     assert isinstance(display.summary, dict)
-    assert list(display.summary.keys()) == [
-        "dataframe",
-        "dataframe_module",
-        "n_rows",
-        "n_columns",
-        "columns",
-        "dataframe_is_empty",
-        "plots_skipped",
-        "sample_table",
-        "n_constant_columns",
-        "top_associations",
-    ]
 
 
 def test_table_report_display_frame_error(display):
@@ -391,16 +379,7 @@ def test_corr_plot(pyplot, estimator_report):
 
 def test_json_dump(display):
     json_dict = json.loads(display._to_json())
-    assert list(json_dict.keys()) == [
-        "dataframe_module",
-        "n_rows",
-        "n_columns",
-        "columns",
-        "dataframe_is_empty",
-        "plots_skipped",
-        "n_constant_columns",
-        "top_associations",
-    ]
+    assert isinstance(json_dict, dict)
 
 
 def test_repr(display):
--- a/ci/requirements/skore/python-3.10/scikit-learn-1.4/test-requirements.txt
+++ b/ci/requirements/skore/python-3.10/scikit-learn-1.4/test-requirements.txt
@@ -93,7 +93,6 @@ packaging==25.0
     #   matplotlib
     #   plotly
     #   pytest
-    #   skrub
 pandas==2.3.1
     # via
     #   skore (skore/pyproject.toml)
@@ -129,13 +128,17 @@ pure-eval==0.2.3
     # via stack-data
 pyarrow==20.0.0
     # via skore (skore/pyproject.toml)
+pydot==4.0.1
+    # via skrub
 pygments==2.19.2
     # via
     #   ipython
     #   pytest
     #   rich
 pyparsing==3.2.3
-    # via matplotlib
+    # via
+    #   matplotlib
+    #   pydot
 pytest==8.4.1
     # via
     #   skore (skore/pyproject.toml)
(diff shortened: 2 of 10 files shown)


skore/tests/unit/sklearn/plot/table_report/test_estimator.py
```diff
@@ -391,16 +379,7 @@ def test_corr_plot(pyplot, estimator_report):
...
-        "dataframe_module",
-        "n_rows",
-        "n_columns",
-        "columns",
-        "dataframe_is_empty",
-        "plots_skipped",
-        "n_constant_columns",
-        "top_associations",
-    ]
+    assert isinstance(json_dict, dict)
```
thomass-dev, line 382: I think the test should be something like:

```python
def test_json_dump(display):
    assert json.loads(display._to_json()) == {k: v for k, v in display.summary.items() if k not in ("dataframe", "sample_table")}
```

thomass-dev:
You can take this opportunity to update `skrub` in the lockfiles, using 

```python
bash ci/pip-compile.sh skore --upgrade-package skrub
```

skore/tests/unit/sklearn/plot/table_report/test_estimator.py
MarieSacksick: if feels really weird to test by more or less writing down the function itself...  
What are we trying to test here, is it the JSONEncoder? Because it comes from skrub, so it's not really ours to test, is it?

thomass-dev: > if feels really weird to test by more or less writing down the function itself...

Unittest 🤷

> What are we trying to test here, is it the JSONEncoder?

No, we test that we serialize what we want to serialize, not the encoder itself.
But i haven't in mind that the encoder is custom, and the output can slightly differ from the original object.
So maybe my test will failed.

In any case, IMO the current test isn't really good.

MarieSacksick: indeed, checking the values fail. I proposed something else, a bit more exhaustive than current version.

glemaitre: > In any case, IMO the current test isn't really good.

I agree. The idea is to full proof it once we know what is expected on the hub side (the `_to_json` is only expose for that). Once we know the exact set of information that we need to send, then we can be enforcing it in the test and then being able to detect any regression that would break the hub.

Right now, we are not really sure how `_to_json` is used :)

MarieSacksick: Ok, so the test is not complete, but we don't know how to complete it.  
What should we do then? Put this PR back to draft until hub implement it, or merge it to be compliant with several versions of skrub, and improve it later once we know how the function is actually used?

thomass-dev: > Right now, we are not really sure how _to_json is used :)

Currently the implementation of `_to_json` is striclty equivalent to `skrub.TableReport.json`, isn't ?
This is what we want for now: sending the `TableReport` in a JSON format that `skrub` can managed.
For now, if we can just rely on `skrub.TableReport.json`, let's do it. 

(as a short-term safeguard, I will also send the HTML representation of the `TableReport`)

MarieSacksick: If I understand correctly, you propose to drop the function _to_json altogether?

MarieSacksick: (additional note: we now have an issue about this: https://github.com/probabl-ai/skore/issues/1954)

thomass-dev: For now, you can just simplify as much as possible the test.
We will improve the `_to_json` function and its test after its integration in the hub.

```python
def test_json_dump(display):
    display._to_json()
```

cc @glemaitre

MarieSacksick: ok, I resimplified the test - keeping the assertion that the json dump can be reloaded at least.

thomass-dev (approved):
Let's move forward. LGTM in state for now.

PR #1956, opened 2025-07-31
feat(skore): Save the indices of the splits in `CrossValidationReport`

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/1955.

PR #1958, opened 2025-08-01
chore(skore)!: Rename `cv_splitter` to `splitter` in cross-validation

thomass-dev:
> [!CAUTION]
> Must be merged after https://github.com/probabl-ai/skore/issues/1955.

> Can i rename `cv_splitter` in `splitter` 👼🏻 ?
> It looks obvious that splitter in the `CrossValidationReport` is the splitter of the cross-validation.

_Originally posted by @thomass-dev in https://github.com/probabl-ai/skore/pull/1956#discussion_r2245737102_

### DIFF
--- a/examples/getting_started/plot_skore_getting_started.py
+++ b/examples/getting_started/plot_skore_getting_started.py
@@ -127,7 +127,7 @@ plt.tight_layout()
 # %%
 from skore import CrossValidationReport
 
-cv_report = CrossValidationReport(rf, X, y, cv_splitter=5)
+cv_report = CrossValidationReport(rf, X, y, splitter=5)
 
 # %%
 # We display the cross-validation report helper:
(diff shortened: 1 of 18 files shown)


MarieSacksick:
Waiting for a review from @glemaitre to double-check this API change, but to me it would be fine!  
Historically, in scikit-learn, this parameter is called "cv" (cf function [cross_validate](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.cross_validate.html)), hence the reason why we extended the param name, to make it clear but keeps some roots with what people are used to. But seeing your proposition, I find it fine to change.

glemaitre (approved):
+1 with the change.

I see that it makes more sense that `splitter` is related to the "cross-validation" of the report. I can think of a similar renaming from `scoring` (that means higher is always better) to `metric` (that can be both side). But this is for another PR.

PR #1965, opened 2025-08-04
feat: Add the data accessor to CrossValidationReport

### LINKED ISSUE #1964: feat(data): Extend data accessor to the CrossValidationReport

# Goal

Just as we have the data accessor in the EstimatorReport, we would like to have the same accessor in the CrossValidationReport. 

# TODO
 
- [ ] Add this new functionnality

MarieSacksick:
Closes #1964   

On the way, I split the tests of the data accessor for estimator report between things specifics and common.


skore/tests/unit/displays/table_report/test_common.py
```diff
@@ -0,0 +1,174 @@
+import json
```
glemaitre: The test need to be moved into a new file:

```
skore/tests/unit/displays/table_report/test_cross_validation.py
```

glemaitre: Oh I see that you did it below. So I did not see the next file which is exactly this. I think it makes sense to have the private helpers in this specific file.

skore/tests/unit/displays/table_report/test_estimator.py
```diff
@@ -370,28 +259,6 @@ def test_hue_plots_2d(pyplot, estimator_report):
...
-    assert isinstance(json_dict, dict)
-
-
-def test_repr(display):
-    """Check the string representation of the `TableReportDisplay`."""
-    repr = display.__repr__()
-    assert repr == "<TableReportDisplay(...)>"
-
-
 def test_html_repr(estimator_report):
```
glemaitre: I think that we can have the same test in the `test_common.py`. We only need to create the report (estimator or cross-validation) with a pytest parametrization.

skore/tests/unit/sklearn/test_find_estimators.py
```diff
@@ -19,7 +19,7 @@ def test_find_estimators():
 
 
 def test_find_estimators_with_skrub():
-    pip = skrub.tabular_learner("classification")
+    pip = skrub.tabular_pipeline("classification")
```
glemaitre: I think that we should make an effort to support skrub version 0.5 and 0.6 for the moment. The one way to handle this deprecation would be to add a `tabular_pipeline` in the `skore/_externals/_skrub_compat.py` where in this file it will call `tabular_learner` if the version is `<0.6` and `tabular_pipeline` otherwise.

Then, we will never need to change code internally.

MarieSacksick: Is that important to be compatible with both versions in the tests? As it's not in the usage of skore?

skore/tests/unit/displays/table_report/test_estimator.py
MarieSacksick: Indeed! Should I really to a parametrization? After all, we are testing the display itself here, the fact that it comes from an estimator or a cv report shouldn't change anything, sould it?

skore/tests/unit/sklearn/test_find_estimators.py
glemaitre: Second solution is to skip the test based on the version.

However, if we start to have many places that use it and that we don't run the tests, it become problematic. I don't think it is actually a problem here because I don't even think that we use this `find_estimators` functionality. We should probably remove it.

auguste-probabl: Deleted `find_estimators`, and added a `tabular_pipeline` export to `_skrub_compat.py`

skore/tests/unit/displays/table_report/test_estimator.py
auguste-probabl: test_html_repr seems just specific enough to EstimatorReports that a parametrize would be awkward (lots of `if-else`). If you want we can add a similar test to `test_cross_validation.py` but IMO it's not worth it.

skore/tests/unit/displays/table_report/test_common.py
```diff
@@ -0,0 +1,159 @@
...
+from skore._sklearn._plot.data.table_report import (
+    _compute_contingency_table,
+    _resize_categorical_axis,
+    _truncate_top_k_categories,
+)
+from skrub.datasets import fetch_employee_salaries
+
+
+@pytest.fixture
+def estimator_report():
```
glemaitre: I find it weird that in `test_common.py`, we don't test both the `EstimatorReport` and `CrossValidationReport`. Basically, in this file I expect to have private routines that are used by both classes or user code that would either class and provide the display on which we want to make some checks.

Concretely, whenever this fixture is used, I would expect a parametrization that would use as well a display generated via a `CrossValidationReport`. In short, it looks like there is three functions impacted: the `corr` method call, the `repr` and the `to_json`.

skore/src/skore/__init__.py
```diff
@@ -18,7 +18,6 @@
     PredictionErrorDisplay,
     RocCurveDisplay,
     TableReportDisplay,
-    find_estimators,
```
glemaitre: I would prefer to have separate PR for this matter. When `git` blaming it will be much easier to understand the history of the code.

glemaitre: I'll provide a quick PR such that only a small rebasing is required.

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -35,14 +35,21 @@ def _generate_estimator_report(
     train_indices: ArrayLike,
     test_indices: ArrayLike,
 ) -> EstimatorReport | KeyboardInterrupt | Exception:
+    if y is None:
```
glemaitre: Regarding this fix, I also want to isolate it in its own PR because it is a bug fix and it will be easier to understand the history.

skore/tests/unit/displays/table_report/test_cross_validation.py
```diff
@@ -0,0 +1,135 @@
...
+        {
+            "dataframe",
+            "dataframe_module",
+            "n_rows",
+            "n_columns",
+            "columns",
+            "dataframe_is_empty",
+            "plots_skipped",
+            # skrub>=0.6
+            # "associations_skipped",
```
glemaitre: Instead to add the key here, I would prefer to have a comment on above the assert mentioning that we check the necessary keys require by `skore` to generate the display.

skore/tests/unit/displays/table_report/test_cross_validation.py
```diff
@@ -0,0 +1,135 @@
...
+        ),
+    ],
+)
+@pytest.mark.parametrize(
+    "y",
+    [
+        np.ones((100, 1)),
+        np.ones(100),
+        # pd.Series(np.ones(100)),
+        # pd.DataFrame(np.ones((100, 1)), columns=["Target"]),
```
glemaitre: I think that they are legit case and we need to test them as well. If there is a bug (I suspect that there is one because I stumbled in https://github.com/probabl-ai/skore/issues/1990), then we need to make a PR to fix it first. So here, I think that we should pass a `Series` that does not have a `name` as it is now, and another where we fix the `name`.

skore/tests/unit/displays/table_report/test_cross_validation.py
```diff
@@ -0,0 +1,135 @@
...
+    ],
+)
+def test_retrieve_data_as_frame(X, y):
```
glemaitre: It is a bit weird for me to call this function `retrieve_data_as_frame` while we don't have an assertion in this direction.
I would suggest to make some assertion the retrieved dataframe store in `display.summary["dataframe"]` that should be concatenation of the `X` and `y` that we have in the parametrization. I think that the test does not need to check up to the data in the dataframe but at least the shape and the column names.

glemaitre: Also it would be nice to have a small docstring explaining what we try to to check for the function.

skore/tests/unit/displays/table_report/test_cross_validation.py
```diff
@@ -0,0 +1,135 @@
...
+            # skrub>=0.6
+            # "associations_skipped",
+            "sample_table",
+            "n_constant_columns",
+            "top_associations",
+        }
+    )
+
+
+def test_table_report_display_frame(cross_validation_report):
```
glemaitre, line 46:
```suggestion
def test_table_report_display_frame(cross_validation_report):
    """Check that we return the expected kind of data when calling `.frame`."""
```

glemaitre (commented):
It looks really. We still need the following:

- Similarly to `skore/tests/unit/reports/estimator/data/test_accessor.py`, we need to create `skore/tests/unit/reports/cross-validation/data/test_accessor.py` with the same type of tests.
- We miss the documentation in the rst file `cross_validation_report.rst` (we have the same issue for the `estimator_report.rst` and I'll make a quick PR to solve this problem).
- In terms of documentation, we should scan the user guide to see where the information make sense to be added and we should use it somewhere in the example. I recall that it was complex when using the `EstimatorReport` because our example used the `CrossValidationReport` so it should be quite easy to find the example where the data analysis will fit.
- As commented I'll make a couple of PR that will isolate some fix or refactoring that are not related to the PR itself to make the history cleaner when it comes to the future us using `git blame`.

skore/src/skore/_sklearn/_cross_validation/report.py
glemaitre: Uhm weird because it looks like it was merged already in a separate PR: https://github.com/probabl-ai/skore/pull/1969

most probably this would disappear after up merging `main` in this PR ;)

skore/tests/unit/displays/table_report/test_cross_validation.py
MarieSacksick: not clear to me what you mean here - would you like to check that all the required elements are passed to the display to be created? As it's already created... it means it received everything needed, doesn't it?
Now that we started to parametrize, actually, this test could go into the common, as we only check that the TableReportDisplay has been created properly.

skore/tests/unit/displays/table_report/test_cross_validation.py
auguste-probabl: See #2013

MarieSacksick: Actually it may have been another error. I fixed it, but also did a ticket on skrub side to be sure: https://github.com/skrub-data/skrub/issues/1593.

PR #1977, opened 2025-08-11
fix: EstimatorReport thinks skorch model is fitted but it is not

### LINKED ISSUE #1966: bug(skore): EstimatorReport thinks skorch model is fitted but it is not
## Context

<details>
  <summary>show</summary>

I tried using a skorch model in a EstimatorReport

```python
# env google colab: pip install -U skore skorch
import numpy as np 
from torch import nn
from skorch import NeuralNetClassifier

X, y = make_classification(n_samples=1000, n_features=10, n_informative=4, n_classes=2)
X = X.astype(np.float32)

class MyModule(nn.Module):
    def __init__(self, num_units=10, nonlin=nn.ReLU()):
        super().__init__()

        self.dense0 = nn.Linear(10, num_units)
        self.nonlin = nonlin
        self.dropout = nn.Dropout(0.5)
        self.dense1 = nn.Linear(num_units, num_units)
        self.output = nn.Linear(num_units, 2)
        self.softmax = nn.Softmax(dim=-1)

    def forward(self, X, **kwargs):
        X = self.nonlin(self.dense0(X))
        X = self.dropout(X)
        X = self.nonlin(self.dense1(X))
        X = self.softmax(self.output(X))
        return X

net = NeuralNetClassifier(
    MyModule,
    max_epochs=10,
    lr=0.1,
    # Shuffle training data on each epoch
    iterator_train__shuffle=True,
)

report = EstimatorReport(
    net, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
)
```

The above code runs fine. However

```python
report.metrics.summarize() # or accuracy, roc, etc
```

Produces:

```
AttributeError                            Traceback (most recent call last)
[/usr/local/lib/python3.11/dist-packages/skorch/classifier.py](https://localhost:8080/#) in classes_(self)
    102         try:
--> 103             return self.classes_inferred_
    104         except AttributeError as exc:

AttributeError: 'NeuralNetClassifier' object has no attribute 'classes_inferred_'

The above exception was the direct cause of the following exception:

AttributeError                            Traceback (most recent call last)
6 frames
[/usr/local/lib/python3.11/dist-packages/skorch/classifier.py](https://localhost:8080/#) in classes_(self)
    113                 f"initializing {self.__class__.__name__}"
    114             )
--> 115             raise AttributeError(msg) from exc
    116 
    117     # pylint: disable=signature-differs

AttributeError: NeuralNetClassifier could not infer the classes from y; this error probably occurred because the net was trained without y and some function tried to access the '.classes_' attribute; a possible solution is to provide the 'classes' argument when initializing NeuralNetClassifier
```

</details>

## Root cause

<details>
  <summary>show</summary>

After some digging I think the EstimatorReport thinks the skorch net is fit but actually it's not.

In EstimatorReport ([line 152](https://github.com/probabl-ai/skore/blob/e073e200e59fcf735ef3c9f6ffe7e5c8f3ee092a/skore/src/skore/_sklearn/_estimator/report.py#L152)), we check if the estimator is fitted

```python
class EstimatorReport
    ...
    def __init__(self, ...):
        ...
        if fit == "auto":
            try:
                # An error is not raised when not fitted skorch model is passed
                check_is_fitted(estimator) # <---- HERE. 
                self._estimator = self._copy_estimator(estimator)
            except NotFittedError:
                self._estimator, fit_time = self._fit_estimator(
                    estimator, X_train, y_train
                )
        ...
```

This `check_is_fitted` comes from [sklearn](https://github.com/scikit-learn/scikit-learn/blob/c5497b7f7eacfaff061cf68e09bcd48aa93d4d6b/sklearn/utils/validation.py#L1753) and calls [`_is_fitted`](https://github.com/scikit-learn/scikit-learn/blob/c5497b7f7eacfaff061cf68e09bcd48aa93d4d6b/sklearn/utils/validation.py#L1664) 

```python
def check_is_fitted(...):
    ...
    if not _is_fitted(estimator, attributes, all_or_any):
        raise NotFittedError(msg % {"name": type(estimator).__name__})
```

`_is_fitted` goes to check this condition:

```python

def _is_fitted(...):
    ...
    fitted_attrs = [
        v for v in vars(estimator) if v.endswith("_") and not v.startswith("__")
    ]
    return len(fitted_attrs) > 0
```

- If True then the model is fitted
- If False then the model is not fitted. 

*Said differently, when not fitted the above check returns an empty list which in turn raises an error from `check_is_fitted`* 

This check works fine for sklearn models. Example:

```python

unfit_ridge = Ridge()

from sklearn.utils.validation import check_is_fitted

check_is_fitted(unfit_ridge)  # raises an error
```

Under, the call to `_is_fitted` returns an empty list, which raises an error.

The error I got at the top of the issue suggested the model is not fit. And indeed it was not. Because in the example code at the top, before I pass `net` (the skorch model) to the `EstimatorReport`, if I fit it (`net.fit(X_train, y_train)`), then `report.metrics.summarize()` works ✅ (as net is fit) (alongside other `report.metrics.<function>`). 

</details>

## Issue: EstimatorReport's check_is_fitted returns True when it should be False

<details>
  <summary>show</summary>

The EstimatorReport init method calls `check_is_fitted`, however here the output on a **not fitted** **skorch model** is not an empty list: 

```python
# check_is_fitted calls _is_fitted which checks if the list below is empty
# reminder: if the list is empty -> model is not fit
not_fitted_net_attrs = [v for v in vars(unfit_net) if v.endswith("_") and not v.startswith("__")]
# ['history_', 'initialized_', 'virtual_params_'] 
```

So this *does not* raise any errors:

```python
unfit_net = NeuralNetClassifier(
    MyModule,
    max_epochs=10,
    lr=0.1,
    # Shuffle training data on each epoch
    iterator_train__shuffle=True,
)

check_is_fitted(unfit_net) # doesnt raise anything but it should
```

The list is not empty, so `check_is_fitted` returns True and the code up to but not including `report.metrics.summarize()` runs fine. 

</details>

## Suggestion

<details>
  <summary>show</summary>

To not check if the model is a skorch model and avoid adding skorch as a dependency, we could wrap the `check_is_fitted` in skore utils

```python
def skore_check_is_fitted(estimator):
    if not all(attr in vars(estimator) for attr in ['history_', 'initialized_', 'virtual_params_']):
        check_is_fitted(estimator) # non-skorch cases
    else: 
        check_is_fitted(estimator, attributes=[
            'init_context_',
            'callbacks_',
            'prefixes_',
            'cuda_dependent_attributes_',
            'module_',
            'criterion_',
            'optimizer_',
            'classes_inferred_'
        ])
```

The attributes `['history_', 'initialized_', 'virtual_params_']` exist on the skorch model when its not fitted, but dont exist on other sklearn models when not fitted. Will update if I find something better

- for CVReport the model is fit in splits so there seems to be no issue
- for ComparisonReport - this report cares that the underlying EstReport/CVReport has fit the model so the fix for EstReport will cascade to here as well

Example:

```python
net = NeuralNetClassifier(
    MyModule,
    max_epochs=10,
    lr=0.1,
    # Shuffle training data on each epoch
    iterator_train__shuffle=True,
)

check_is_fitted(net, attributes=['init_context_', 'callbacks_', 'prefixes_', 'cuda_dependent_attributes_', 'module_', 'criterion_', 'optimizer_', 'classes_inferred_'])
```

Results in:

```
NotFittedError                            Traceback (most recent call last)
/tmp/ipython-input-1381609148.py in <cell line: 0>()
      9 )
     10 
---> 11 check_is_fitted(net, attributes=[
     12     'init_context_',
     13     'callbacks_',

/usr/local/lib/python3.11/dist-packages/sklearn/utils/validation.py in check_is_fitted(estimator, attributes, msg, all_or_any)
   1755 
   1756     if not _is_fitted(estimator, attributes, all_or_any):
-> 1757         raise NotFittedError(msg % {"name": type(estimator).__name__})
   1758 
   1759 

NotFittedError: This NeuralNetClassifier instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator.
```

So if we use a skorch model in an EstReport, it is fitted and then the EstReport functionality is available. 

</details>

---

## [Google colab with code](https://colab.research.google.com/drive/1gKkpQe6Ran0rZDwCXI8-ua_SNFfhb0de?usp=sharing)

divakaivan:
Fixes #1966 

Please check the description of the above issue for exact details. 
I did not want to include skorch as a dependency so this fix does not require it.

### DIFF
--- a/skore/src/skore/_utils/_fixes.py
+++ b/skore/src/skore/_utils/_fixes.py
@@ -28,6 +28,9 @@ def skore_check_is_fitted(estimator):
     estimator has these three attributes to determine if the estimator is a skorch
     model. If so, extra attributes are passed to sklearn's `check_is_fitted`
     to avoid silently passing not fitted skorch models to `EstimatorReport`s.
+
+    TODO: Remove this when the issue is solved skorch-side
+          (https://github.com/skorch-dev/skorch/issues/1118)
     """
     if not all(
         attr in vars(estimator)
--- a/skore/src/skore/_utils/_fixes.py
+++ b/skore/src/skore/_utils/_fixes.py
@@ -1,6 +1,7 @@
 from typing import Any
 
 import joblib
+from sklearn.utils.validation import check_is_fitted
 
 from skore._externals._sklearn_compat import parse_version
 
@@ -16,3 +17,34 @@ def _validate_joblib_parallel_params(**kwargs: Any) -> dict[str, Any]:
     if joblib_version < parse_version("1.4"):
         kwargs.pop("return_as", None)
     return kwargs
+
+
+def skore_check_is_fitted(estimator):
+    """More thorough version of sklearn's `check_is_fitted` covering skorch models.
+
+    Skorch models that are not fitted pass through the sklearn `check_is_fitted`
+    when they are not fitted. Skorch models have three attributes that exist on not
+    fitted models which do not exist on other models. This function checks if the
+    estimator has these three attributes to determine if the estimator is a skorch
+    model. If so, extra attributes are passed to sklearn's `check_is_fitted`
+    to avoid silently passing not fitted skorch models to `EstimatorReport`s.
+    """
+    if not all(
+        attr in vars(estimator)
+        for attr in ["history_", "initialized_", "virtual_params_"]
+    ):
+        check_is_fitted(estimator)
+    else:  # skorch models
+        check_is_fitted(
+            estimator,
+            attributes=[
+                "init_context_",
+                "callbacks_",
+                "prefixes_",
+                "cuda_dependent_attributes_",
+                "module_",
+                "criterion_",
+                "optimizer_",
+                "classes_inferred_",
+            ],
+        )
--- a/skore/src/skore/_sklearn/_estimator/report.py
+++ b/skore/src/skore/_sklearn/_estimator/report.py
@@ -9,7 +9,6 @@ from numpy.typing import ArrayLike
 from sklearn.base import BaseEstimator, clone
 from sklearn.exceptions import NotFittedError
 from sklearn.pipeline import Pipeline
-from sklearn.utils.validation import check_is_fitted
 
 from skore._externals._pandas_accessors import DirNamesMixin
 from skore._externals._sklearn_compat import is_clusterer
@@ -17,6 +16,7 @@ from skore._sklearn._base import _BaseReport, _get_cached_response_values
 from skore._sklearn.find_ml_task import _find_ml_task
 from skore._sklearn.types import _DEFAULT, PositiveLabel
 from skore._utils._fixes import _validate_joblib_parallel_params
+from skore._utils._fixes import skore_check_is_fitted as check_is_fitted
 from skore._utils._measure_time import MeasureTime
 from skore._utils._parallel import Parallel, delayed
 from skore._utils._progress_bar import progress_decorator
--- a/skore/tests/unit/reports/cross_validation/test_report.py
+++ b/skore/tests/unit/reports/cross_validation/test_report.py
@@ -11,9 +11,9 @@ from sklearn.exceptions import NotFittedError
 from sklearn.linear_model import LinearRegression
 from sklearn.metrics import accuracy_score
 from sklearn.pipeline import Pipeline
-from sklearn.utils.validation import check_is_fitted
 from skore import CrossValidationReport, EstimatorReport
 from skore._sklearn._cross_validation.report import _generate_estimator_report
+from skore._utils._fixes import skore_check_is_fitted as check_is_fitted
 from skore._utils._testing import MockEstimator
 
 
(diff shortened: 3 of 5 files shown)


skore/src/skore/_utils/_fixes.py
auguste-probabl, line 1: Was an issue also opened with the skorch folks? It seems like they should know that they don't adhere to the sklearn protocol

divakaivan: No issue raised with `skorch`. Could you share the protocol you are referring to? I am not aware of those :/ 

Edit: https://scikit-learn.org/stable/developers/develop.html#estimated-attributes (found this)

divakaivan: It seems skorch needs to look at the [developer API for check_is_fitted](https://scikit-learn.org/stable/auto_examples/developing_estimators/sklearn_is_fitted.html#sphx-glr-auto-examples-developing-estimators-sklearn-is-fitted-py), which suggests they should implement a custom `__sklearn_is_fitted__` method on their estimator. If that’s the case, I should close or draft this PR? and raise an issue with them.

auguste-probabl: It was `__sklearn_is_fitted__` I was thinking of, yes! This PR is still useful, but we should first open an issue with skorch and add a comment in our code like "TODO: Remove this when the issue is solved skorch-side (`<issue link>`)"

divakaivan: [Fixed on skorch's side](https://github.com/skorch-dev/skorch/pull/1119#pullrequestreview-3123893736)

closing this PR

PR #1979, opened 2025-08-12
fix(skore): Prototype accessors directly in `EstimatorReport` to make mypy happy

thomass-dev:
Fixing `error: "EstimatorReport" has no attribute "data"  [attr-defined])` issue.


skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -98,7 +104,10 @@ class EstimatorReport(_BaseReport, DirNamesMixin):
         "feature_importance": {"name": "feature_importance"},
         "data": {"name": "data"},
     }
-    metrics: "_MetricsAccessor"
+
+    metrics: _MetricsAccessor
+    feature_importance: _FeatureImportanceAccessor
+    data: _DataAccessor
```
thomass-dev, line 110: My 2cts: as this is required to satisfy `mypy`, the accessors should be declared in the same place instead of in the `__init__.py` file.

thomass-dev:
@MarieSacksick can we plan a `0.10.1` with this fix (and the latest ones)?

glemaitre (approved):
I would postpone the accessors. I don't have a strong opinions where to declare the accessors. I think the first time I wrote it it was in this file. It seems also clean when moving it in the `__init__`. The fact that we need to declare the type is indeed one argument to move it here.

PR #1961, opened 2025-08-01
feat: Allow access to sub-estimators via a dict instead of a list in ComparisonReport

HimanshuSharmaKUL:
Now ComparisonReport exposes sub-estimators through the reports_list_ attribute, and the combination of report names and reports themselves is exposed through reports_ (a dict). [#1577](https://github.com/probabl-ai/skore/issues/1577)

Instead of writing like 
`for name, report in zip(comparison_report.report_names_, comparison_report.reports_):`
Users can simply write:
`for name, report in comparison_report.reports_.items():`

Following tasks done:
- [ ] The earlier reports_ attribute renamed to reports_list_,
- [ ] reports_ now a dict.
- [ ] removed report_names_ 

p.s. This is my first open-source contribution. I hope it is correct and meets the criteria of open course contribution.

PR #1986, opened 2025-08-27
feat(ComparisonReport)!: Make `reports_` a dict instead of a list

### LINKED ISSUE #1577: feat(ComparisonReport): Give access to sub-estimators through a dict rather than lists

Currently, `ComparisonReport` exposes sub-estimators through the `reports_` attribute, and the report names through `report_names`.

It might turn out that this leads to users constantly writing something like
```python
for name, report in zip(comparison_report.report_names_, comparison_report.reports_):
    ...
```

In that case, it would make more sense to make `reports_` a dict:
```python
for name, report in comparison_report.reports_.items():
    ...
```

But we'd have to see with usage.

Originates from https://github.com/probabl-ai/skore/pull/1512/files/815b1d2df8fc06e15d9133b4a9667ae72d56f527

@sylvaincom @MarieSacksick do you have any thoughts on this?

auguste-probabl:
Remove `report_names_`.

Closes #1577

Supersedes #1961

Co-authored-by: Himanshu Sharma <himanshuxstats@gmail.com>

### DIFF
--- a/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
@@ -242,7 +242,7 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
 
             generator = parallel(
                 joblib.delayed(getattr(report.metrics, report_metric_name))(**kwargs)
-                for report in self._parent.reports_
+                for report in self._parent.reports_.values()
             )
             individual_results = []
             for result in generator:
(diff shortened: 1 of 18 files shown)


auguste-probabl:
Notice that we often have to write weird statements like `list(report.reports_.values())[0]`. Personally I don't think one phrasing is better than the other, so I would just not touch it.

auguste-probabl:
The amount of tests failing with the error `rich.errors.LiveError: Only one live display may be active at once` is also concerning; this indicates a thread safety issue with the progress bar logic. I don't think it makes sense to fix the remaining tests until this is addressed.

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
```diff
@@ -256,15 +256,15 @@ def _compute_metric_scores(
             if self._parent._reports_type == "EstimatorReport":
                 results = _combine_estimator_results(
                     individual_results,
-                    estimator_names=self._parent.report_names_,
+                    estimator_names=self._parent.reports_.keys(),
```
thomass-dev, line 259: If `reports` is a dict, you don't need to specified `reports.keys()` to iterate over its keys.

auguste-probabl: I agree that it works, but it's less explicit and thus more bug-prone.

thomass-dev (approved):

PR #1988, opened 2025-08-28
feat(EstimatorReport): Show permutation at different stages of a pipeline

### LINKED ISSUE #1398: enh: Show permutation at different stage of a pipeline

auguste-probabl:
Added support for calculating permutation importance at different stages of a pipeline.
One can choose to compute feature importance either at the start of the pipeline or at the end.

Closes #1398

Supersedes #1888

Co-authored-by: waridrox <mohdwarid4@gmail.com>

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
@@ -612,14 +612,8 @@ class _FeatureImportanceAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
             score = self._parent._cache[cache_key]
         else:
             if not isinstance(self._parent.estimator_, Pipeline) or at_step == 0:
-                estimator = self._parent.estimator_
+                feature_engineering, estimator = None, self._parent.estimator_
                 X_transformed = X_
-
-                feature_names = _get_feature_names(
-                    estimator,
-                    X=X_transformed,
-                    transformer=None,
-                )
 
             else:
                 pipeline = self._parent.estimator_
@@ -639,14 +633,17 @@ class _FeatureImportanceAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
                             "number of steps in the Pipeline, which is "
                             f"{len(pipeline.steps)}; got {at_step}"
                         )
-                    feature_eng, estimator = pipeline[:at_step], pipeline[at_step:]
-                    X_transformed = feature_eng.transform(X_)
+                    feature_engineering, estimator = (
+                        pipeline[:at_step],
+                        pipeline[at_step:],
+                    )
+                    X_transformed = feature_engineering.transform(X_)
 
-                feature_names = _get_feature_names(
-                    estimator,
-                    X=X_transformed,
-                    transformer=feature_eng,
-                )
+            feature_names = _get_feature_names(
+                estimator,
+                X=X_transformed,
+                transformer=feature_engineering,
+            )
 
             if issparse(X_transformed):
                 X_transformed = X_transformed.todense()


skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -386,6 +389,16 @@ def permutation(
             Whether to flatten the multi-index columns. Flat index will always be lower
             case, do not include spaces and remove the hash symbol to ease indexing.
 
+        stage : {"start", "end"}, default="start"
```
glemaitre: In the terminology of the scikit-learn `Pipeline`, we would call this a step. I think that I would call this parameter `at_step`:

```python
report.feature_importance.permutation(at_step="...")
```

I know that originally I stated that intermediate stage would be YAGNI but I'm thinking that we could end-up with a more natural API if we support it. In short, `at_step` can accept both the name of a step as a string or its position. With some reworking we would have:

```python
model = make_pipeline(
    StandardScaler(),
    PCA(),
    LogisticRegression(),
)
model.named_steps
```

```
{'standardscaler': StandardScaler(),
 'pca': PCA(),
 'logisticregression': LogisticRegression()}
```

```python
at_step = "pca"
index_at_step = list(model.named_steps.keys()).index(at_step)
X_tran = pipeline[:index_at_step].fit_transform(X)
permutation_importance(pipeline[-1], X_engineered, y)
```

and if `at_step` is an integer then we it will already be equivalent to `index_at_step`. And by default, we can force `at_step=0` to be equivalent to `start` and `end` would be equivalent to `-1`.

glemaitre:
Since we still don't have the permutation importance across the different reporters, there is not documentation to change in the user guide.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: Done

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -386,6 +389,22 @@ def permutation(
             Whether to flatten the multi-index columns. Flat index will always be lower
             case, do not include spaces and remove the hash symbol to ease indexing.
 
+        at_step : int or str, default=0
+            If the estimator is a :class:`~sklearn.pipeline.Pipeline`, at which step of
+            the pipeline the importance is computed. If `n`, then the features that
```
jeremiedbb, line 420: Would it be possible/not to painful to have `permutation` not be the same method when the estimator is a Pipeline and when it's not. That way you only get the `at_step` param for the Pipelines and don't get an unused param for other estimators.

auguste-probabl: I think it would be quite painful. For example, I don't know how we would get this distinction across to readers in our docs.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -29,6 +29,8 @@
 
 DataSource = Literal["test", "train", "X_y"]
 
+PipelineStep = int | str
```
glemaitre: Since we have only 2 occurrences of the type and that they are only 2 built-in types, then I would avoid to define it.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -501,6 +549,7 @@ def _feature_permutation(
                 self._parent._hash,
                 "permutation_importance",
                 data_source,
+                at_step,
```
glemaitre: I'm thinking that we could always translate `at_step` into a positive index before entering in the if/else and use it for the cache key. In this way the cache key is a single type and we would hit the cache if the name corresponding to the index as well (it is micro optimization.)

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +581,31 @@ def _feature_permutation(
...
+                X_transformed = X_
+            else:
+                pipeline = self._parent.estimator_
+                if isinstance(at_step, str):
+                    at_step = list(pipeline.named_steps.keys()).index(at_step)
+                if at_step == 0:
+                    estimator = pipeline
+                    X_transformed = X_
+                elif isinstance(at_step, int):
+                    if at_step >= len(pipeline.steps):
```
glemaitre: I think that we should also check when the index is negative and too large. Otherwise it will raise an `IndexError`.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +581,31 @@ def _feature_permutation(
...
+                        raise ValueError(
+                            "at_step must be strictly smaller than the number of steps "
+                            f"in the Pipeline, which is {len(pipeline.steps)}; "
+                            f"got {at_step}"
+                        )
+                    feature_eng, estimator = pipeline[:at_step], pipeline[at_step:]
+                    X_transformed = feature_eng.transform(X_)
+                else:
+                    raise ValueError(f"at_step must be an integer; got {at_step!r}")
```
glemaitre: It could also be a string.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -314,3 +315,76 @@ def test_not_fitted(regression_data):
...
+
+    result = pipeline_report.feature_importance.permutation(seed=42, at_step=at_step)
+
+    assert isinstance(result.index, pd.MultiIndex)
+    assert result.index.nlevels == 2
+    assert result.index.names == ["Metric", "Feature"]
+    assert result.shape[0] > 0
+
+
+def test_at_step_parameter_too_large(pipeline_report):
```
glemaitre: You need to add when the integer is negatively too large.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -314,3 +315,76 @@ def test_not_fitted(regression_data):
...
+    """
+    result = pipeline_report.feature_importance.permutation(seed=42, at_step="pca")
+
+    assert isinstance(result.index, pd.MultiIndex)
+    assert result.index.nlevels == 2
+    assert result.index.names == ["Metric", "Feature"]
+    assert result.shape[0] > 0
+
+
+def test_at_step_parameter_pipeline_str_wrong(pipeline_report):
```
glemaitre: I think that we can parametrize for the int/str at the same time: we need to provide `at_step`, the error message and the type of error. The call to the `permutation` error will be the same.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: That requires adding a bit too much logic in the cache key computation, for little gain.

auguste-probabl:
> amend the example called `plot_feature_importance.py` where we can demonstate the feature. We have a `Ridge` model with some feature importance and we can show that we can compute with index 0 and -1.

~TODO: I found a bug while doing this, at a certain step the permutation computation fails because it doesn't like sparse matrices.~ Fixed.

Now I need to figure out how to showcase the new feature in the example, hopefully in a way that doesn't disrupt the flow.

examples/use_cases/plot_feature_importance.py
```diff
@@ -1062,19 +1074,51 @@ def plot_permutation_train_test(est_report):
 # pipeline (with regards to the original features):
 
 # %%
-plot_permutation_train_test(selectk_ridge_report)
+plot_permutation_train_test(**compute_permutation_importances(selectk_ridge_report))
+
+# %%
+# Since this estimator is a Pipeline, it can be also be interesting to look at the *engineered*
+# features rather than the raw input features; this can be done by setting the ``at_step`` parameter:
```
glemaitre:
```suggestion
# Since this estimator resorts on complex feature engineering, it is interesting to look
# at the impact of the *engineered* features rather than the original input features.
# For instance, it is interesting to check if feature with low level of importance are
# not present in the engineered features.
```

examples/use_cases/plot_feature_importance.py
```diff
@@ -991,38 +991,39 @@ def add_y_true_pred(model_report, split):
...
+def plot_permutation_train_test(*, train_importance, test_importance):
     _, ax = plt.subplots(figsize=(8, 6))
 
     train_color = "blue"
     test_color = "orange"
 
-    est_report.feature_importance.permutation(data_source="train", seed=0).T.boxplot(
+    features = list(train_importance.reset_index()["Feature"])
+
+    train_importance.T.boxplot(
```
glemaitre: Let's use `.plot.box` to be able to pass `whis` parameter to control the outlier

```suggestion
    train_importance.T.box.plot(
```

examples/use_cases/plot_feature_importance.py
```diff
@@ -991,38 +991,39 @@ def add_y_true_pred(model_report, split):
...
         ax=ax,
         vert=False,
         widths=0.35,
         patch_artist=True,
         boxprops=dict(facecolor=train_color, alpha=0.7),
         medianprops=dict(color="black"),
-        positions=[x + 0.4 for x in range(len(est_report.X_train.columns))],
+        positions=[x + 0.4 for x in range(len(features))],
```
glemaitre:
```suggestion
        positions=[x + 0.4 for x in range(len(features))],
        whis=100,
```

examples/use_cases/plot_feature_importance.py
```diff
@@ -991,38 +991,39 @@ def add_y_true_pred(model_report, split):
...
-    est_report.feature_importance.permutation(data_source="test", seed=0).T.boxplot(
+    test_importance.T.boxplot(
         ax=ax,
         vert=False,
         widths=0.35,
         patch_artist=True,
         boxprops=dict(facecolor=test_color, alpha=0.7),
         medianprops=dict(color="black"),
-        positions=range(len(est_report.X_test.columns)),
+        positions=range(len(features)),
```
glemaitre:
```suggestion
        positions=range(len(features)),
        whis=100,
```

examples/use_cases/plot_feature_importance.py
```diff
@@ -991,38 +991,39 @@ def add_y_true_pred(model_report, split):
...
         vert=False,
         widths=0.35,
         patch_artist=True,
         boxprops=dict(facecolor=train_color, alpha=0.7),
         medianprops=dict(color="black"),
-        positions=[x + 0.4 for x in range(len(est_report.X_train.columns))],
+        positions=[x + 0.4 for x in range(len(features))],
     )
-    est_report.feature_importance.permutation(data_source="test", seed=0).T.boxplot(
+    test_importance.T.boxplot(
```
glemaitre:
```suggestion
    test_importance.T.box.plot(
```

examples/use_cases/plot_feature_importance.py
```diff
@@ -991,38 +991,39 @@ def add_y_true_pred(model_report, split):
...
+# Since the permutation importance involves permuting values of a feature at random, by default it is computed
+# several times, each time with different permutations of the feature.
+# For this reason, if `seed` is not passed, skore does not cache the permutation importance results.
 
 # %%
+# Now, we plot the permutation feature importance on the train and test sets using boxplots:
 
 
-def plot_permutation_train_test(est_report):
+def plot_permutation_train_test(*, train_importance, test_importance):
```
glemaitre: I think that we can actually simplify further by using seaborn:

```python
def plot_permutation_train_test(importances):
    _, ax = plt.subplots(figsize=(8, 6))

    # create a long format required by seaborn
    importances = importances.stack().reset_index()
    importances.columns = ["Dataset", "Feature", "Repeats", "Importance"]

    sns.boxplot(
        data=importances,
        x="Importance",
        y="Feature",
        hue="Dataset",
        orient="h",
        whis=100,
        order=importances["Feature"].unique()[::-1],
        ax=ax,
    )
    ax.set_xlabel("Decrease of R2 score")
    ax.set_title("Permutation feature importance (Train vs Test)")

    plt.tight_layout()
    plt.show()


# %%


def compute_permutation_importances(report, at_step=0):
    train_importance = report.feature_importance.permutation(
        data_source="train", seed=0, at_step=at_step
    )
    test_importance = report.feature_importance.permutation(
        data_source="test", seed=0, at_step=at_step
    )

    return pd.concat(
        {"train": train_importance, "test": test_importance},
        axis=0,
        names=["dataset"],
    ).droplevel(level=1, axis=0)
```

examples/use_cases/plot_feature_importance.py
```diff
@@ -1062,19 +1074,51 @@ def plot_permutation_train_test(est_report):
...
+# Take only the 15 most important train features
+best_15_features = (
+    train_importance.aggregate("mean", axis=1).sort_values(key=abs).tail(15).index
+)
+train_importance = train_importance.loc[best_15_features]
+test_importance = test_importance.loc[best_15_features]
+
+plot_permutation_train_test(
+    train_importance=train_importance, test_importance=test_importance
+)
```
glemaitre: If we consider that `compute_permutation_importances` return a single dataframe with the train-test in the index, then we need to slightly change the code here:

```python
importances = compute_permutation_importances(
    selectk_ridge_report, at_step=-1
)

# rename some features for clarity
importances = importances.reset_index()
importances["Feature"] = (
    importances["Feature"]
    .str.replace("remainder__", "")
    .str.replace("kmeans__", "geospatial__")
)
importances = importances.set_index(["Dataset", "Feature"])

# Take only the 15 most important train features
best_15_features = (
    importances.loc[("test", slice(None))]
    .aggregate("mean", axis=1)
    .sort_values(key=abs)
    .tail(15)
    .index
)

importances = importances[importances.index.isin(best_15_features, level=1)]
```

examples/use_cases/plot_feature_importance.py
```diff
@@ -1062,19 +1074,51 @@ def plot_permutation_train_test(est_report):
...
 
 # %%
 # Hence, contrary to coefficients, although we have created many features in our
 # preprocessing, the interpretability is easier.
 # We notice that, due to our preprocessing using a clustering on the geospatial data,
 # these features are of great importance to our model.
+#
+# Also, Average Bedrooms and Average Rooms appear often in the plot, whereas they were not considered
+# as important when looking at the coefficients.
```
glemaitre:
```suggestion
#
# Also, Average Bedrooms and Average Rooms appear often in the plot, whereas they were
# not considered as important when looking at the coefficients. It means that once
# combined with other features, they become more relevant.
```

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +581,54 @@ def _feature_permutation(
...
+                if hasattr(estimator, "feature_names_in_"):
+                    feature_names = estimator.feature_names_in_
+                elif hasattr(estimator, "n_features_in_"):
+                    feature_names = [
+                        f"Feature #{i}" for i in range(estimator.n_features_in_)
+                    ]
+                else:
+                    feature_names = [
+                        f"Feature #{i}" for i in range(X_transformed.shape[1])
+                    ]
```
glemaitre: I'm thinking that if the estimator does not follow the scikit-learn convention, we should be able to still take the columns if we have a dataframe.

I think that we can also directly rely on the `else` without checking the `n_features_in_`. Usually the `feature_names_in_` is set at the same time than the `n_features_in_` if one follow the guideline. In practice, I think that you end-up with the same results. I will check in the tests are you covered those individual case, thought to see how likely my comment could be a valid corner case.

```suggestion
                if hasattr(estimator, "feature_names_in_"):
                    # infer from model if it follows scikit-learn API
                    feature_names = estimator.feature_names_in_
                elif hasattr(X_transformed, columns):
                    # infer the column from the dataframe if it is one
                    feature_names = X_transformed.columns.tolist()
                else:
                    feature_names = [
                        f"Feature #{i}" for i in range(X_transformed.shape[1])
                    ]
```

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +581,54 @@ def _feature_permutation(
...
+                            "number of steps in the Pipeline, which is "
+                            f"{len(pipeline.steps)}; got {at_step}"
+                        )
+                    feature_eng, estimator = pipeline[:at_step], pipeline[at_step:]
+                    X_transformed = feature_eng.transform(X_)
+
+                if hasattr(estimator, "feature_names_in_"):
+                    feature_names = estimator.feature_names_in_
+                elif hasattr(feature_eng, "get_feature_names_out"):
+                    feature_names = feature_eng.get_feature_names_out()
```
glemaitre: So maybe we can still get a chance if the estimator does not follow the API but return a dataframe.

```suggestion
                    feature_names = feature_eng.get_feature_names_out()
                elif hasattr(X_transformed, "columns"):
                    feature_names = X_transformed.columns.tolist()
```

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +581,54 @@ def _feature_permutation(
...
+                else:
+                    [f"Feature #{i}" for i in range(estimator.n_features_in_)]
+
+            if issparse(X_transformed):
+                X_transformed = X_transformed.todense()
```
glemaitre, line 649: I'm surprised. I need to check the test :)

glemaitre: OK I see. Scikit-learn does not support it. We could maybe implement a shuffling in the CSC case by permuting non-zero rows. However, it might not be valid statistically.

Here, I think that the only thing that we should do is to add a note in the docstring mentioning that we are going to densify sparse matrix.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -314,3 +315,99 @@ def test_not_fitted(regression_data):
...
+        """
+
+        result = pipeline_report.feature_importance.permutation(
+            seed=42, at_step=at_step
+        )
+
+        assert isinstance(result.index, pd.MultiIndex)
+        assert result.index.nlevels == 2
+        assert result.index.names == ["Metric", "Feature"]
+        assert result.shape[0] > 0
```
glemaitre, line 368: Since we have a scikit-learn pipeline, let's check the "Features" from the dataframe as well. It should corresponds to the `feature_names_in_` of the steps

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -314,3 +315,99 @@ def test_not_fitted(regression_data):
...
+    def test_str(self, pipeline_report):
+        """
+        Test the `at_step` string parameter for permutation importance with a pipeline.
+        """
+        result = pipeline_report.feature_importance.permutation(seed=42, at_step="pca")
+
+        assert isinstance(result.index, pd.MultiIndex)
+        assert result.index.nlevels == 2
+        assert result.index.names == ["Metric", "Feature"]
+        assert result.shape[0] > 0
```
glemaitre, line 379: same regarding the value of "Features"

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +581,54 @@ def _feature_permutation(
...
+                        raise ValueError(
+                            "at_step must be strictly smaller in magnitude than the "
+                            "number of steps in the Pipeline, which is "
+                            f"{len(pipeline.steps)}; got {at_step}"
+                        )
+                    feature_eng, estimator = pipeline[:at_step], pipeline[at_step:]
+                    X_transformed = feature_eng.transform(X_)
+
+                if hasattr(estimator, "feature_names_in_"):
+                    feature_names = estimator.feature_names_in_
```
glemaitre: From the coverage report, it seems that we don't cover the following two lines: 620, 624. Could you amend the test to check those as well.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -314,3 +315,99 @@ def test_not_fitted(regression_data):
     error_msg = "This LinearRegression instance is not fitted yet"
     with pytest.raises(NotFittedError, match=error_msg):
         report.feature_importance.permutation()
+
+
+class TestAtStep:
```
glemaitre, line 339: So here, we always test estimator that follow the `feature_names_in_` API. We should test a basic scaler and classifier that does not follow this API but still can be used in a pipeline.

So for instance, the following code could work reasonably in real life:

```python
import numpy as np
from sklearn.base import BaseEstimator, TransformerMixin, ClassifierMixin
from sklearn.pipeline import make_pipeline

X, y = load_iris(return_X_y=True, as_frame=True)


class Scaler(TransformerMixin, BaseEstimator):
    def fit(self, X, y=None):
        self.mean_ = X.mean(axis=0)
        self.std_ = X.std(axis=0)
        return self

    def transform(self, X, y=None):
        X_ = X.copy()
        return (X_ - self.mean_) / self.std_


class Classifier(ClassifierMixin, BaseEstimator):
    def fit(self, X, y):
        self.classes_ = np.unique(y)
        return self

    def predict(self, X):
        return np.ones_like(y) * self.classes_[0]


model = make_pipeline(Scaler(), Classifier())
model.fit(X, y).predict(X)
assert not hasattr(model, "feature_names_in_")
```

without enforcing the scikit-learn API regarding the feature names.

glemaitre: And I think that we should make sure that the `EstimatorReport` is working with both `numpy` array and `dataframe` then.

examples/use_cases/plot_feature_importance.py
auguste-probabl: reworded slightly.

examples/use_cases/plot_feature_importance.py
auguste-probabl: nice!

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: so essentially, don't use `n_features_in_`?

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: added note

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
auguste-probabl: You're right, using a non-sklearn Scaler does lead trigger a bug: Even if `Pipeline(steps=[('scaler', Scaler())])` does have a `get_feature_names_out` method, it defers to `Scaler`, which doesn't have such a method, so the whole thing fails. Should be fixed.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
auguste-probabl: in this case X is a numpy array so the Features column will always be `[f"Feature #{i}" for i in range(X_transformed.shape[1])]`. Point taken in https://github.com/probabl-ai/skore/pull/1988#discussion_r2460750280 that there should be tests for DataFrames.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
auguste-probabl: point about arrays vs. dataframes is acknowledged elsewhere; closing this

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
auguste-probabl: made it so every test also tries dataframes.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: so the aforementioned lines are now covered, but now https://github.com/probabl-ai/skore/blob/1dbbc94fe638a27f1776e960d5c86670332bfc38/skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py#L603
and https://github.com/probabl-ai/skore/blob/1dbbc94fe638a27f1776e960d5c86670332bfc38/skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py#L635
is not covered, even though we run every test both with arrays and with dataframes. i think these lines can be safely deleted.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
auguste-probabl: >  It should corresponds to the `feature_names_in_` of the steps

hm so I've found an example where the `feature_names_in_` of `estimator[at_step]` is not set, but `get_feature_names_out` of the previous step does work, and that's what we use for the output.
The test is
`tests/unit/reports/estimator/feature_importance/test_permutation_importance.py::TestAtStep::test_int[dataframe--1]`

So to me this seems that in order to test this properly we'd need to repeat the logic that is already in the code.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
glemaitre: I don't think so. It is just that the test `test_non_sklearn_pipeline` is not triggering those lines while it could.

For the second line, it is because you are using a `LinearRegression` that is a fully compatible scikit-learn classifier and then implement the `feature_names_in_`. If we would have a simple classifier that only inherit from `BaseEstimator` and `ClassifierMixin` but does not validate the input, then we would not have this attribute and the `get_feature_names_out` will as well fail. If I recall, it was what I propose in a previous comment to by implementing a kind of `DummyClassifier`.

For l.603, we need to have the same test than `test_non_sklearn_pipeline` but just passing the classifier or forcing `at_step=0`.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -339,3 +341,125 @@ def test_not_fitted(regression_data):
...
+        these names should appear in the output."""
+        X, y = data
+        pipeline = make_pipeline(SplineTransformer(), LinearRegression())
+        report = EstimatorReport(pipeline, X_train=X, X_test=X, y_train=y, y_test=y)
+
+        result = report.feature_importance.permutation(seed=42, at_step=-1)
+        last_step_feature_names = list(report.estimator_[0].get_feature_names_out())
+        assert list(result.index.levels[1]) == last_step_feature_names
+
+    def test_non_sklearn_pipeline(self, data):
```
glemaitre:
```suggestion
    @pytest.mark.parametrize("at_step", [0, -1])
    def test_non_sklearn_pipeline(self, data, at_step):
```

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
```diff
@@ -339,3 +341,125 @@ def test_not_fitted(regression_data):
...
+                self.mean_ = X.mean(axis=0)
+                self.std_ = X.std(axis=0)
+                return self
+
+            def transform(self, X, y=None):
+                X_ = X.copy()
+                return (X_ - self.mean_) / self.std_
+
+        X, y = data
+        pipeline = make_pipeline(Scaler(), LinearRegression())
```
glemaitre: So here, you need to replace `LinearRegression` with:

```python
        class DummyRegressor(RegressorMixin, BaseEstimator):
            def fit(self, X, y):
                self.mean_ = y.mean()
                return self

            def predict(self, X):
                return np.ones(X.shape[0]) * self.mean_
  ```

File: skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py [hunk 26]
```diff
@@ -339,3 +341,125 @@ def test_not_fitted(regression_data):
...
+        report = EstimatorReport(pipeline, X_train=X, y_train=y, X_test=X, y_test=y)
+
+        report.feature_importance.permutation(seed=42, at_step=-1)
```
glemaitre:
```suggestion
        report.feature_importance.permutation(seed=42, at_step=at_step)
```

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
glemaitre: Yep there is no need because they come together with `feature_names_in_`

examples/use_cases/plot_feature_importance.py
```diff
@@ -991,62 +991,58 @@ def add_y_true_pred(model_report, split):
...
 
 # %%
-# The permutation importance is often calculated several times, each time
-# with different permutations of the feature.
-# Hence, we can have measure its variance (or standard deviation).
-# Now, we plot the permutation feature importance on the train and test sets using boxplots:
-
+# Since the permutation importance involves permuting values of a feature at random, by default it is computed
+# several times, each time with different permutations of the feature.
+# For this reason, if `seed` is not passed, skore does not cache the permutation importance results.
```
glemaitre: Can you wrap those lines to 88 chars.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -62,6 +64,14 @@
 }
 
 
+def _function_call_succeeds(func: Callable) -> bool:
+    try:
+        func()
+        return True
+    except Exception:
```
glemaitre: Let's specialize to what the `get_feature_names_out` can do.

```suggestion
    except AttributeError:
```

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +593,58 @@ def _feature_permutation(
...
+                if hasattr(estimator, "feature_names_in_"):
+                    feature_names = estimator.feature_names_in_
+                elif hasattr(X_transformed, "columns"):
+                    # Infer the column from the dataframe if it is one
+                    feature_names = X_transformed.columns.tolist()
+                else:
+                    feature_names = [
+                        f"Feature #{i}"
+                        for i in range(X_transformed.shape[1])  # type: ignore
+                    ]
```
glemaitre: Since this code is almost the same as the one below starting at l.631, we could create a `get_feature_names(estimator, transformer=None)` where `transformer=None` when we have no pipeline or `at_step=0` and corresponds to `feature_eng` otherwise.

skore/tests/unit/reports/estimator/feature_importance/test_permutation_importance.py
glemaitre: It would also be nice to check the feature names that we get instead to only have a smoke test here.

auguste-probabl:
Now it's https://github.com/probabl-ai/skore/blob/612dee8ac2868ddbdb17db00c33a2b77d1df4579/skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py#L632 that is not covered

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -532,9 +611,49 @@ def _feature_permutation(
...
+                            f"{len(pipeline.steps)}; got {at_step}"
+                        )
+                    feature_eng, estimator = pipeline[:at_step], pipeline[at_step:]
+                    X_transformed = feature_eng.transform(X_)
+
+                feature_names = _get_feature_names(
+                    estimator,
+                    X=X_transformed,
+                    transformer=feature_eng,
+                )
```
glemaitre:
```suggestion
            if not isinstance(self._parent.estimator_, Pipeline) or at_step == 0:
                feature_eng, estimator = None, self._parent.estimator_
                X_transformed = X_

            else:
                pipeline = self._parent.estimator_
                if not isinstance(at_step, str | int):
                    raise ValueError(
                        f"at_step must be an integer or a string; got {at_step!r}"
                    )

                if isinstance(at_step, str):
                    # Make at_step an int and process it as usual
                    at_step = list(pipeline.named_steps.keys()).index(at_step)

                if isinstance(at_step, int):
                    if abs(at_step) >= len(pipeline.steps):
                        raise ValueError(
                            "at_step must be strictly smaller in magnitude than the "
                            "number of steps in the Pipeline, which is "
                            f"{len(pipeline.steps)}; got {at_step}"
                        )
                    feature_eng, estimator = pipeline[:at_step], pipeline[at_step:]
                    X_transformed = feature_eng.transform(X_)

            feature_names = _get_feature_names(
                estimator, X=X_transformed, transformer=feature_eng
            )
```

glemaitre (dismissed):
Just the last comment. Otherwise, LGTM.

.pre-commit-config.yaml
```diff
@@ -39,7 +39,7 @@ repos:
         name: mypy skore/
         files: ^skore/
         args: [--config-file=skore/pyproject.toml, skore]
-        additional_dependencies: [IPython, matplotlib, numpy, polars, rich]
+        additional_dependencies: [IPython, matplotlib, numpy, polars, rich, scipy-stubs]
```
thomass-dev, line 42: Please also add `scipy-stubs` to the `skore` test dependencies.

thomass-dev: Fix https://github.com/probabl-ai/skore/pull/2119.

PR #1998, opened 2025-09-01
fix(ci): Revert #1982 after `loky==3.5.6` fix

thomass-dev:
Revert the `fix(ci): Temporary pin 3.13.6 in actions/setup-python` https://github.com/probabl-ai/skore/pull/1982 temporary fix after https://github.com/joblib/loky/issues/459 (and https://github.com/joblib/joblib/pull/1740).


thomass-dev:
Thanks @lesteve and @tomMoral for your investigations.

PR #2000, opened 2025-09-02
feat(ComparisonReport): Delegate `available_if` to sub-reports

### LINKED ISSUE #1473: chore: ComparisonReport should delegate to underlying estimator the availability of metrics

Related to #1472 

`ComparisonReport` should also delegate to the underlying `EstimatorReport` is a method is available.
Here, we need to think a bit more because we should be more lenient than for `CrossValidationReport`: we could compute the score if it is possible for at least one estimator and set to `np.nan` for whenever it is not possible to compute the score.

auguste-probabl:
Replaces the checks in available_if with a check that at least one of the compared reports has the requested metric.

Closes #1473

### DIFF
--- a/skore/src/skore/_utils/_accessor.py
+++ b/skore/src/skore/_utils/_accessor.py
@@ -191,3 +191,19 @@ def _check_comparison_report_sub_estimators_have_coef() -> Callable:
         return all(_check_has_coef(e) for e in parent_estimators)
 
     return check
+
+
+########################################################################################
+# Accessor related to `ComparisonReport`
+########################################################################################
+
+
+def _check_any_sub_report_has_metric(metric: str) -> Callable[[Any], bool]:
+    """Check whether any sub-report of the ComparisonReport supports `metric`."""
+
+    def check(accessor: Any) -> bool:
+        return any(
+            hasattr(report.metrics, metric) for report in accessor._parent.reports_
+        )
+
+    return check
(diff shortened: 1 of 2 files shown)


skore/src/skore/_utils/_accessor.py
```diff
@@ -191,3 +191,19 @@ def check(accessor: Any) -> bool:
...
+# Accessor related to `ComparisonReport`
+########################################################################################
+
+
+def _check_any_sub_report_has_metric(metric: str) -> Callable[[Any], bool]:
+    """Check whether any sub-report of the ComparisonReport supports `metric`."""
+
+    def check(accessor: Any) -> bool:
+        return any(
+            hasattr(report.metrics, metric) for report in accessor._parent.reports_
```
thomass-dev, line 206: Nitpick: you can factorize your function to `check_any_sub_report_has_accessor` with the following :

```suggestion
            reduce(getattr, accessor.split("."), report)
            for report in accessor._parent.reports_
```

```
@available_if(_check_any_sub_report_has_accessor("metrics.rmse"))
```

thomass-dev (changes requested):
Do you have any example in mind where the previous implementation would cause problems?
If so, please add a dedicated test, otherwise LGTM.

auguste-probabl:
> Do you have any example in mind where the previous implementation would cause problems? If so, please add a dedicated test, otherwise LGTM.

I think this was the point of @guillaume that I split into #2001. That issue is proving hard to solve, so I could either add tests here and change them in #1985, or we could wait for #1985 (might be a while) and then merge this one.

thomass-dev (approved):

PR #2011, opened 2025-09-04
fix(skore-local-project): Go through storage in insertion order

thomass-dev:
Rather than in alphabetical key order.

> Caches may be iterated by either insertion order or sorted order. The default ordering uses insertion order. To iterate by sorted order, use iterkeys. The sort order is determined by the database which makes it valid only for str, bytes, int, and float data types. Other types of keys will be serialized which is likely to have a meaningless sorted order.
> 
> ```python
> >>> for key in 'cab':
> ...     cache[key] = None
> >>> list(cache)
> ['c', 'a', 'b']
> >>> list(cache.iterkeys())
> ['a', 'b', 'c']
> >>> cache.peekitem()
> ('b', None)
> >>> cache.peekitem(last=False)
> ('c', None)
> ```

https://grantjenks.com/docs/diskcache/tutorial.html

### DIFF
--- a/skore-local-project/src/skore_local_project/storage.py
+++ b/skore-local-project/src/skore_local_project/storage.py
@@ -75,7 +75,7 @@ class DiskCacheStorage:
 
     def __iter__(self) -> Iterator[str]:
         """
-        Yield the keys in the storage.
+        Yield the keys in the storage in insertion order.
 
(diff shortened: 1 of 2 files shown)


auguste-probabl (approved):
Nice catch

PR #2014, opened 2025-09-04
ci: Run tests serially

auguste-probabl:
Counter-intuitively, tests take longer to run in parallel.


thomass-dev:
How big is the difference?

auguste-probabl:
> How big is the difference?

At least 8 minutes before (we had issues where we reached the 10-minute timeout in the past); 2 minutes now.

thomass-dev (approved):
Let's try it!

Issue #1923, opened 2025-07-16
Add CrossValidationReports to Project.summarize()


MarieSacksick:
### Is your feature request related to a problem? Please describe.

Today, we have project.summarize(), which outputs an extended pandas.DataFrame, containing only EstimatorReports.
Tomorrow, we would like to store and retrieve from the project both EstimatorReports and CrossValidationReports. Therefore, what should project.summarize() outputs?
--> 

it's consistant with hub, where we have different tabs for Estimator and CrossValidation.
what is displayed makes sense.
in the parallel coordinates plot, the line for a given CrossValidationReport is the average of the metrics. For now, we do not want to display the std.

A crossvalidation report is made of several Estimator Reports. Those sub-Estimator Reports shouldn't be displayed not fetchable by project.summarize(report_type="Estimator").

### Describe the solution you'd like

As presented in https://github.com/probabl-ai/skore/issues/1892, I would like to have:  
`project.summarize(report_type="Estimator")`.  (or "CrossValidation")
report_type cannot be "all", because we cannot display both Estimator and CrossValidation reports in the same parallel plot, it wouldn't make sense.

### Describe alternatives you've considered, if relevant

To give access to all reports in `project.summarize()`, but it wouldn't make sense because these two objects are different by nature.

thomass-dev:
Decided jointly with @glemaitre:

Use a single dataframe with all columns of `EstimatorReport` and `CrossValidationReport`.
Add a column with the type of report, using `skore/_sklearn/types.py:ReportType`.

thomass-dev:
PRs that start to address this issue: 
- https://github.com/probabl-ai/skore/pull/2008,
- https://github.com/probabl-ai/skore/pull/2020.

PR #2019, opened 2025-09-05
feat(skore-hub-project)!: Get `CrossValidationReport`

thomass-dev:
Part of https://github.com/probabl-ai/skore-hub/issues/629 and https://github.com/probabl-ai/skore/issues/1944.

PR #2020, opened 2025-09-08
feat(skore-hub-project)!: Add `CrossValidationReport` metadata to `summarize`

thomass-dev:
Part of https://github.com/probabl-ai/skore/issues/1923.


thomass-dev:
Ready to review. Waiting https://github.com/probabl-ai/skore/pull/2019 to be merged.

skore-hub-project/tests/unit/project/test_project.py
```diff
@@ -274,32 +299,64 @@ def test_reports_metadata(self, nowstr, respx_mock):
 
         assert metadata == [
             {
-                "id": "<report_id_0>",
+                "id": "skore:report:estimator:<report_id_0>",
```
retromechs, line 302: I think you should add a dedicated "urn": "skore:report:estimator:<report_id_0>", field while keeping id as it is, since we use it in the hub response.
Alternatively, you could remove id and use urn in its place.

retromechs (changes requested):
It’s more of a nitpick than a strong request. Updating the id would introduce inconsistencies between the behavior in the hub and in the lib

skore-hub-project/tests/unit/project/test_project.py
thomass-dev: I understand your reasoning but here, `id` is more generic.

This `metadata` list is used by `skore-lib` at an higher level. In this list, `id` is the identifier of the report. It can be used to retrieve its pickled content regardless if it comes from local or hub project, while `urn` is hub specific.

retromechs: Sounds good, thanks!

Issue #1996, opened 2025-08-30
api(skore): Revisit the `PlotBackendMixin` class and API


glemaitre:
### What would you like to say?

https://github.com/probabl-ai/skore/pull/1924 introduced a new mixin to be able to switch from matplotlib backend to plotly backend.

One issue that I had as a user is that now, we don't get anymore the proper documentation because we get the generic documentation of the mixin that does not define the args and kwargs.

So we need to revisit the implementation to make sure that `help(display.plot)` shows a good documentation.

I think that before to address this issue, we should make a PR for https://github.com/probabl-ai/skore/issues/1957. Indeed, it would allow to move the problematic only on the `set_style` method since the `plot` method should not have any parameter related to plotting backend.

Once done, a question will be how to specialize `set_style` enough, showing the right documentation depending on the backend selected. A first thought would be to have a general signature `set_style(self, **kwargs)` which is far to be perfect but at least, the mixin could then inject the documentation of the specialized `_set_style_matplotlib` or `_set_style_plotly` that need to be properly document then.

glemaitre:
One potential solution is to change the mixin and add an `__init__` where the configuration would happen and dynamically wrap the `_set_style` that could be specialized. Something like:

```python
import functools
from typing import Any

class PlotBackendMixin:
    def __init__(self):
        plot_backend = get_config()["plot_backend"]

        if plot_backend == "matplotlib":
            self.set_style = functools.wraps(self._set_style_matplotlib)(
                lambda *args, **kwargs: self._set_style_matplotlib(*args, **kwargs)
            )
        else:
            self.set_style = functools.wraps(self._set_style_plotly)(
                lambda *args, **kwargs: self._set_style_plotly(*args, **kwargs)
            )

class A(PlotBackendMixin):

    def _set_style_matplotlib(self, color: str, size: int = 10):
        print(f"Matplotlib style: color={color}, size={size}")

    def _set_style_plotly(self, font: str, bold: bool = False):
        print(f"Plotly style: font={font}, bold={bold}")
```

If somehow, relying solely on the `set_style` for the backend, then we could imagine to wrap still the `plot` method.

glemaitre:
Actually the solution above cannot work because if change the backend after the creation of the instance of `A` then you are not going to call `__init__` again and it will not work. Need to think of a better solution :)

Edit:

So using a property instead of `__init__`.

```python
class PlotBackendMixin:
    @property
    def set_style(self) -> Callable:
[...]

PR #2027, opened 2025-09-10
fix(skore): Document  via explicit implementation

glemaitre:
Partially fixing https://github.com/probabl-ai/skore/issues/1996

This PR implement a specialized `plot` method for each display. It allows to get the proper documentation that was broken when using a `mixin` that overwrite it.

This approach is simpler than trying to overwrite the name function, signature, and documentation of the dispatch plotting function. It should be easier to maintain.

While implementing this PR, I also fix the following:

- group all the `mixin` in `base.py` and create a `DisplayMixin` since all specialized displays will inherit from all those mixins.
- I fixed slightly the `ConfusionMatrixDisplay` but it will need more works.


skore/src/skore/_sklearn/_plot/base.py
```diff
@@ -21,3 +40,250 @@ def frame(self, **kwargs: Any) -> pd.DataFrame:
...
+
+########################################################################################
+# Display mixin inheriting from the different mixins
+########################################################################################
+
+
+class DisplayMixin(HelpDisplayMixin, PlotBackendMixin, StyleDisplayMixin):
+    """Mixin inheriting help, plotting, and style functionality."""
+
+    pass
```
thomass-dev:
```suggestion
```

You don't need `pass` when a docstr is present.

skore/src/skore/_sklearn/_plot/base.py
```diff
@@ -21,3 +40,248 @@ def frame(self, **kwargs: Any) -> pd.DataFrame:
...
+        return self
+
+    @staticmethod
+    def style_plot(plot_func: Callable) -> Callable:
+        """Apply consistent style to skore displays.
+
+        This decorator:
+        1. Applies default style settings
+        2. Executes `plot_func`
+        3. Applies `tight_layout`
```
auguste-probabl, line 157: does it apply tight layout?

auguste-probabl:
I can confirm that this solves the issue, using the reproducer given [here](https://github.com/probabl-ai/skore/issues/1996#issuecomment-3243208776)

PR #2030, opened 2025-09-11
chore(skore-hub-project/client): Add logging

auguste-probabl:
Helps understand why skore-hub-project hangs, in case the server is down or the server URL is incorrect.

### DIFF
--- a/skore-hub-project/src/skore_hub_project/client/client.py
+++ b/skore-hub-project/src/skore_hub_project/client/client.py
@@ -27,8 +27,6 @@ from httpx._types import HeaderTypes
 
 from ..authentication import token as Token
 
-logging.basicConfig()
-
 logger = logging.getLogger(__name__)
 
 
--- a/skore-hub-project/src/skore_hub_project/__init__.py
+++ b/skore-hub-project/src/skore_hub_project/__init__.py
@@ -1,5 +1,6 @@
 """Package that provides APIs to communicate between ``skore`` and ``skore hub``."""
 
+import logging
 from base64 import b64decode, b64encode
 from contextlib import contextmanager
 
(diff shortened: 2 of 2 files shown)


skore-hub-project/src/skore_hub_project/client/client.py
```diff
@@ -106,7 +111,14 @@ def request(self, *args, **kwargs) -> Response:
             timeout = self.retry_backoff_factor * (2**retries)
             retries += 1
 
-            sleep(min(timeout, self.retry_backoff_max))
+            sleep_duration = min(timeout, self.retry_backoff_max)
+
+            logger.warn(
+                "Request failed to reach server. "
```
thomass-dev, line 115: I think you can even add the exception type in case of exception, and the response status otherwise.

skore-hub-project/src/skore_hub_project/client/client.py
```diff
@@ -106,7 +111,14 @@ def request(self, *args, **kwargs) -> Response:
             timeout = self.retry_backoff_factor * (2**retries)
             retries += 1
 
-            sleep(min(timeout, self.retry_backoff_max))
+            sleep_duration = min(timeout, self.retry_backoff_max)
+
+            logger.warn(
```
thomass-dev, line 114: My 2cts: maybe we can use `logger.debug` instead to not spam the user?

auguste-probabl: the warning shouldn't happen often (e.g. when a dev forgets to start the hub locally ^^'), so I think `warning` is good

skore-hub-project/src/skore_hub_project/client/client.py
auguste-probabl: I don't want to make the message too noisy, and I don't have access to `response` at this stage; could you suggest a way to do it?

PR #2032, opened 2025-09-12
refactor(skore/progress): Add comments and simplify

auguste-probabl:
In particular, make it clearer that `progress_decorator` is very specific to `skore.*Report`s


skore/src/skore/_utils/_progress_bar.py
```diff
@@ -67,15 +73,20 @@ def wrapper(*args: Any, **kwargs: Any) -> T:
                 progress.start()
                 created_progress = True
 
-            # assigning progress to child reports
+            # Make child reports share their parent's Progress instance
+            # so that there is only one Progress instance at any given point
             reports_to_cleanup: list[Any] = []
-            if hasattr(self_obj, "reports_"):
+            if isinstance(self_obj, ComparisonReport):
```
thomass-dev, line 79: This change is a prerequisite for https://github.com/probabl-ai/skore/pull/2033.

skore/src/skore/_utils/_progress_bar.py
```diff
@@ -67,15 +73,20 @@ def wrapper(*args: Any, **kwargs: Any) -> T:
                 progress.start()
                 created_progress = True
 
-            # assigning progress to child reports
+            # Make child reports share their parent's Progress instance
+            # so that there is only one Progress instance at any given point
             reports_to_cleanup: list[Any] = []
```
glemaitre, line 78: Thinking a bit more about the problem, I think that it would be cleaner to have a parameter in the decorator (i.e. `clean_child_reports`) that would trigger the cleaning. It means that we call in the `ComparatorReport`, the decorator with `clean_child_reports=True` that trigger this code without having to check the type of report. For the other report, then we need to pass `clean_child_reports=False`.

skore/tests/unit/utils/test_progress_bar.py
```diff
@@ -125,41 +125,3 @@ def run(self):
 
     # Verify progress bar was cleaned up
     assert task._progress_info is None
-
-
-def test_child_report_cleanup():
```
glemaitre, line 131: With my proposal then we can test the procedure because we are not tight anymore to a `ComparisonReport` instance specifically.

glemaitre (commented):
Sorry for the delay. I think to not have something tight to the `ComparisonReport` would make it better.

skore/src/skore/_utils/_progress_bar.py
auguste-probabl: The point of this PR is to reduce `progress_decorator`'s complexity (which we have been bit by many times now) by recognizing that it is only used within `*Report` objects, and making it less general.

In other words, what I mean is, the proposed parameter is really just another way of asking "is the report a `ComparisonReport`", and I'd rather just ask that question directly.

glemaitre: What I don't like here is the fact that we import `ComparisonReport`.

Could we instead create a `report._report_type` that is aligned with the other case, for instance when calling `_compute_data_for_display` that takes `report_type` (https://github.com/probabl-ai/skore/blob/006fa4b00ba2e12dfdff071c9e3b3e3d614d63a6/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py#L1187)

If we have this attribute for each type of report, we only need to propagate it instead of hard-coding it. It means that here we only need to have `if "comparison" in self_obj.report_type`.

But otherwise, I'm fine with the rest.

thomass-dev:
Please clean the history/diff 😅 .

auguste-probabl:
Obsoleted by #2357

PR #2043, opened 2025-09-18
skore: Follow regarding the implentation of a display for linear model coefficents

### LINKED ISSUE #1995: api(skore): Improve API related to coefficients feature importance

In https://github.com/probabl-ai/skore/pull/1948, we introduce a new display for coefficients of linear model. I did not get time to provide review and I have the following comments that I think should be addressed:

- The display should be documented in the API and user guide. I see that we modify the example which is great.
- The name `FeatureImportanceDisplay` is not right: I would expect this class to have a `kind` parameter to be able to select the type of feature importance I want to compute. However, in terms of API, we already do this job when accessing the method, e.g. `report.feature_importance.coefficients()`. So we are going to have a specialize display for each type of importance and I think it makes sense because the plotting will be different and one class to rule all of them will not be even more difficult to maintain.
- I see a reference to the report via the `_parent` argument. I think that we should avoid that. The display should be independent as much as possible. The only thing that we need is actually the type of report that could be a string and otherwise, all the other information could be passed as public attribute of the display.
- The current attribute `_parent` and `_coefficients` are private but we can expose them as for other display.
- In general, I think that we can store all the information in a single dataframe. There is only the case where we want to compare coefficients of two models that do not have shared features: (i) we should check for it, (ii) we can still plot the results but they should not be on the same plot but in two different plots. I see that we already kind of manage several plots but we should actually have a single figure with several axis. I recall that there is an issue discussion an API to specify the layout and I think that it would be handy to have this API, to properly plot here.
- In terms of rendering, I think that we should make the feature names on the y-axis by default. It allows to have a more compact view, having name that you can read with turning your head or rotating the label.
- We should also despine the axis and we don't really need an axis for the name but only the name aligned with the box or the coefficient)
- We should drop the underlying grid.

divakaivan:
Closes #1995 

(will add pics of plots)

### DIFF
--- a/skore/src/skore/_sklearn/_comparison/feature_importance_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/feature_importance_accessor.py
@@ -95,7 +95,12 @@ class _FeatureImportanceAccessor(_BaseAccessor["ComparisonReport"], DirNamesMixi
         else:
             raise TypeError(f"Unexpected report type: {self._parent._reports_type}")
 
-        return FeatureImportanceCoefficientsDisplay(self._parent, coef_frames)
+        return FeatureImportanceCoefficientsDisplay(
+            "comparison-estimator"
+            if self._parent._reports_type == "EstimatorReport"
+            else "comparison-cross-validation",
+            coef_frames,
+        )
 
     ####################################################################################
     # Methods related to the help tree
--- a/skore/src/skore/_sklearn/_comparison/feature_importance_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/feature_importance_accessor.py
@@ -43,6 +43,12 @@ class _FeatureImportanceAccessor(_BaseAccessor["ComparisonReport"], DirNamesMixi
         together.
         When some reports share the same features and others do not, those with the same
         features are plotted together.
+
+        Returns
+        -------
+        FeatureImportanceDisplay
+            The feature importance display containing model coefficients and
+            intercept.
         """
         similar_reports = defaultdict(list)
 
--- a/skore/src/skore/_sklearn/_comparison/feature_importance_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/feature_importance_accessor.py
@@ -10,8 +10,8 @@ from skore._externals._pandas_accessors import DirNamesMixin
 from skore._sklearn._base import _BaseAccessor
 from skore._sklearn._cross_validation import CrossValidationReport
 from skore._sklearn._estimator import EstimatorReport
-from skore._sklearn._plot.metrics.feature_importance_display import (
-    FeatureImportanceDisplay,
+from skore._sklearn._plot.metrics.feature_importance_coefficients_display import (
+    FeatureImportanceCoefficientsDisplay,
 )
 from skore._utils._accessor import _check_comparison_report_sub_estimators_have_coef
 
@@ -29,7 +29,7 @@ class _FeatureImportanceAccessor(_BaseAccessor["ComparisonReport"], DirNamesMixi
         super().__init__(parent)
 
     @available_if(_check_comparison_report_sub_estimators_have_coef())
-    def coefficients(self) -> FeatureImportanceDisplay:
+    def coefficients(self) -> FeatureImportanceCoefficientsDisplay:
         """Retrieve the coefficients for each report, including the intercepts.
 
         If the compared reports are `EstimatorReport`s, the coefficients from each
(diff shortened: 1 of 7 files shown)


skore/src/skore/_sklearn/_comparison/feature_importance_accessor.py
```diff
@@ -29,21 +29,27 @@ def __init__(self, parent: ComparisonReport) -> None:
...
         When some reports share the same features and others do not, those with the same
         features are plotted together.
-        """
+
+        Returns
+        -------
+        FeatureImportanceCoefficientsDisplay
+            The feature importance display containing model coefficients and
+            intercept.
+        """  # noqa: D301
```
divakaivan: [the backtick is required for after double-backticks](https://documatt.com/sphinx-errors/5-inline-literal-start-string-without-end/), and `# noqa: D301` because ruff thinks there is a regex because of the backslash.

sphinx/reference/report/comparison_report.rst
auguste-probabl, line 1: I can't find any reference to this documentation in the preview. Wonder what's going on

sphinx/user_guide/reporters.rst
```diff
@@ -84,6 +84,16 @@ Refer to the :ref:`displays` section for more details regarding the `skore` disp
...
 
+Model interpretability
+^^^^^^^^^^^^^^^^^^^^^^
+
+:obj:`EstimatorReport.feature_importance` is the entry point to interpret and explain a
+predictive model. This accessor provides methods that return a `skore` :class:`Display`
+object. As with other display objects, they expose three methods:
+(i) `plot` that visualizes the information.
+(ii) `set_style` that sets graphical options.
+(iii) `frame` that returns a `pandas.DataFrame` with the underlying data.
```
auguste-probabl: Same, I can't find any reference to this in the preview, to check if the rendering looks good

sphinx/reference/report/comparison_report.rst
divakaivan: [isn't this it](https://skoredoc.k.probabl.dev/2043/reference/api/skore.ComparisonReport.feature_importance.coefficients.html)

sphinx/user_guide/reporters.rst
divakaivan: [here](https://skoredoc.k.probabl.dev/2043/user_guide/reporters.html#model-interpretability) and will update it to look like bullet points (atm it doesn't)

skore/src/skore/_sklearn/_plot/metrics/feature_importance_coefficients_display.py
```diff
@@ -110,53 +107,71 @@ def plot(self, **kwargs) -> None:
...
-        self._coefficient_data.plot.bar(ax=self.ax_)
-        self.ax_.set_title(f"{self._parent.estimator_name_} Coefficients")
+        self._coefficient_data.plot.barh(ax=self.ax_)
+        self.ax_.set_title("Coefficients")
         self.ax_.legend(loc="best", bbox_to_anchor=(1, 1), borderaxespad=1)
+        self.ax_.grid(False)
+        self.ax_.spines["top"].set_visible(False)
+        self.ax_.spines["right"].set_visible(False)
+        self.ax_.spines["left"].set_visible(False)
+        self.ax_.tick_params(axis="y", length=0)
```
auguste-probabl: This can be factorized

sphinx/user_guide/reporters.rst
divakaivan: actually, with the (i), (ii), (iii) it doesn't look like bullet points but I followed the format from the [Model evaluation part](https://skoredoc.k.probabl.dev/2043/user_guide/reporters.html#model-evaluation)

skore/src/skore/_sklearn/_plot/metrics/feature_importance_coefficients_display.py
```diff
@@ -110,53 +107,71 @@ def plot(self, **kwargs) -> None:
...
+        if self._parent == "estimator":
             return self._plot_estimator_report()
-        elif isinstance(self._parent, CrossValidationReport):
+        elif self._parent == "cross-validation":
             return self._plot_cross_validation_report()
-        elif isinstance(self._parent, ComparisonReport):
-            return self._plot_comparison_report()
         else:
-            raise TypeError(f"Unrecognised report type: {self._parent}")
+            return self._plot_comparison_report()
```
auguste-probabl: I would rather have `_plot_comparison_report_estimator` and `_plot_comparison_report_cross_validation`

auguste-probabl (approved):

PR #2067, opened 2025-10-16
fix(skore-hub-project): URL encode tenant and project before using it

thomass-dev:
Partially addresses https://github.com/probabl-ai/skore-hub/issues/750.

### DIFF
--- a/skore-hub-project/src/skore_hub_project/project/project.py
+++ b/skore-hub-project/src/skore_hub_project/project/project.py
@@ -4,12 +4,12 @@ from __future__ import annotations
 
 import itertools
 import re
-from functools import wraps
+from functools import cached_property, wraps
 from operator import itemgetter
 from tempfile import TemporaryFile
 from types import SimpleNamespace
 from typing import TYPE_CHECKING
-from urllib.parse import quote_plus
+from urllib.parse import quote
 
 import joblib
 import orjson
@@ -47,7 +47,9 @@ def ensure_project_is_created(method):
     def wrapper(project: Project, *args, **kwargs):
         if not project.created:
             with HUBClient() as hub_client:
-                hub_client.post(f"projects/{project.tenant}/{project.name}")
+                hub_client.post(
+                    f"projects/{project.quoted_tenant}/{project.quoted_name}"
+                )
 
             project.created = True
 
@@ -114,8 +116,8 @@ class Project:
         """
         self.created = False
 
-        self.__tenant = quote_plus(tenant)
-        self.__name = quote_plus(name)
+        self.__tenant = tenant
+        self.__name = name
 
     @property
     def tenant(self) -> str:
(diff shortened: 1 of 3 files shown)


skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -113,8 +114,8 @@ def __init__(self, tenant: str, name: str):
         """
         self.created = False
 
-        self.__tenant = tenant
-        self.__name = name
+        self.__tenant = quote_plus(tenant)
```
tuscland: I don't recommend you store URL-encoded names, unless the variable name clearly states it.

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -309,7 +310,7 @@ def delete(tenant: str, name: str):
         """
         with HUBClient() as hub_client:
             try:
-                hub_client.delete(f"projects/{tenant}/{name}")
+                hub_client.delete(f"projects/{quote_plus(tenant)}/{quote_plus(name)}")
```
tuscland: We should use a URL builder/library, because I'm almost certain errors will be done with path components, or forgetting about encoding parameters.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: Why? I think the opposite: the quoted name is the identifier of the project in the hub, not the original name.

tuscland: What is the value of this Project class? Which layer does it belong to?

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: I usually used `urllib.parse.urljoin`, but my colleagues found it less explicit that a f-string.

tuscland: I don't know if a f-string normalizes two consecutive / characters.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: But i should add a comment on the transformation in the docstr.

thomass-dev: This `Project` class is an intermediate API that can be address by `skore.Project` by `name`, `put`, `get`, `summarize`. It is not directly used by end-user.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: This is covered by https://github.com/probabl-ai/skore/blob/main/skore-hub-project/src/skore_hub_project/client/client.py#L187.

skore-hub-project/src/skore_hub_project/project/project.py
tuscland: URL encoding is done at client level, or just before when building the URL. In my experience, I have never seen instance variables storing URL-encoded content because this is something that is part of the URL building process. There is no point of risking confusion storing content in an encoded form, and there is no performance to gain doing so.

thomass-dev: Ok. Let me change.

thomass-dev: Fixed in 76997b55fef870e690d4ea44639fe318a510c8c7.

PR #2072, opened 2025-10-17
chore(skore-local-project): Set `pandas` from transitive to explicit test dependency

thomass-dev:
…st dependency


thomass-dev:
@auguste-probabl `deps` is not allowed

image

PR #2085, opened 2025-10-27
chore: Remove unecessary labels and workflow

glemaitre:
I think that we discussed in the past that:

- we don't really need to assign someone to its own PR
- we don't really point out "user-reported" because it does not really help

Removing those two things can reduce a bit the noise on the repository.


glemaitre:
> I find quite useful to see who is responsible (the face of the profil) for the PR at a glance.

It is redundant with the "by xxx" that you have just under the PR title.

I would reserve this assign when the author is not the person opening the PR because it get my attention then when doing the review (I need to address someone else).

PR #2089, opened 2025-10-27
chore(all): PEP 639 compliance

DimitriPapadopoulos:
In all packages metadata, change `licence` (deprecated in [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-field)) to `licence-files`.

### DIFF
--- a/skore-hub-project/hatch/metadata.py
+++ b/skore-hub-project/hatch/metadata.py
@@ -34,10 +34,10 @@ class MetadataHook(MetadataHookInterface):
             }
         """
         # Copy LICENCE file from root to `sdist`
-        license_filepath = self.config["license"]["file"]
-        license_filepath = Path(self.root, license_filepath)
-        copy(license_filepath, self.root)
+        license_filepath = Path(self.root, self.config["license"]["file"])
         license_filename = license_filepath.name
+
+        copy(license_filepath, self.root)
 
         # Retrieve README from root files
         readme_filepath = self.config["readme"]["file"]
--- a/skore-local-project/hatch/metadata.py
+++ b/skore-local-project/hatch/metadata.py
@@ -34,10 +34,10 @@ class MetadataHook(MetadataHookInterface):
             }
         """
         # Copy LICENCE file from root to `sdist`
-        license_filepath = self.config["license"]["file"]
-        license_filepath = Path(self.root, license_filepath)
-        copy(license_filepath, self.root)
+        license_filepath = Path(self.root, self.config["license"]["file"])
         license_filename = license_filepath.name
+
+        copy(license_filepath, self.root)
 
         # Retrieve README from root files
         readme_filepath = self.config["readme"]["file"]
--- a/skore/hatch/metadata.py
+++ b/skore/hatch/metadata.py
@@ -34,10 +34,10 @@ class MetadataHook(MetadataHookInterface):
             }
         """
         # Copy LICENCE file from root to `sdist`
-        license_filepath = self.config["license"]["file"]
-        license_filepath = Path(self.root, license_filepath)
-        copy(license_filepath, self.root)
+        license_filepath = Path(self.root, self.config["license"]["file"])
         license_filename = license_filepath.name
+
+        copy(license_filepath, self.root)
 
         # Retrieve README from root files
         readme_filepath = self.config["readme"]["file"]


thomass-dev:
The main reason we did it this way using hatch trigger, was because the build badly integrates the shared files in the build artifacts (in this case, the shared `../LICENSE` and the shared `../README.md`).

We need to check if your change properly integrates the said files into the build artifacts.

thomass-dev:
> As far as i can understand, dynamic license (files) and readme are not needed, because the respective files are not generated during the build process.

The properties `licence` and `readme` are currently dynamic because they are generated by the hatch hooks (which is the prelude of the build process) by reading the dedicated files from the root of the repository.

thomass-dev:
With your PR, `hatch==1.15.1` and `hatchling==1.27.0`, i can confirm my suspicions:

```bash
$ python -m build --sdist && tar -tvf dist/skore*.tar.gz | less +G

[...]
-rw-r--r-- 0/0            1069 2020-02-02 01:00 skore-0.0.0+unknown/../LICENSE
-rw-r--r-- 0/0            5746 2020-02-02 01:00 skore-0.0.0+unknown/../README.md
```
```bash
$ python -m build --wheel && unzip -l dist/skore*.whl | less +G

[...]
1069  2020-02-02 00:00   skore-0.0.0+unknown.dist-info/licenses/../LICENSE
```

which will lead to errors when installing the sdist:

```
ERROR: Invalid member in the tar file [...]/skore/skore/dist/skore-0.0.0+unknown.tar.gz: '../LICENSE' would be extracted to '/tmp/LICENSE', which is outside the destination
```

Also for the wheel, i don't think the `METADATA` are valid either: `License-File: ../LICENSE`.

DimitriPapadopoulos:
> The properties `licence` and `readme` are currently dynamic because they are generated by the hatch hooks (which is the prelude of the build process) by reading the dedicated files from the root of the repository.

Why do you need to generate these properties in the the hatch hooks?

DimitriPapadopoulos:
I guess this could be a bug of the hatch implementation of PEP 639. Whatever the values of `readme` and `license-files`, the files should never end up outside of the sdist root directory or outside the wheel `licenses` directory.

I'll compare with setuptools and flint, and create a hatch issue if they behave differently. I'll check PEP 639 as well, although I doubt it gets into such details.

DimitriPapadopoulos:
From [Add `license-files` key](https://peps.python.org/pep-0639/#add-license-files-key) in [PEP 639](https://peps.python.org/pep-0639/#add-license-files-key):
> * Parent directory indicators (`..`) MUST NOT be used.

So I guess there is no way to point to files outside the directory containing `pyproject.toml`, hence the need for a hatch hook. OK, I'll keep the existing hooks and simply change `license` → `license-files`. How does this sound?

skore-hub-project/hatch/metadata.py
```diff
@@ -34,15 +35,16 @@ def update(self, metadata):
...
         # Retrieve LICENCE from root files
         license_filepath = self.config["license"]["file"]
-        license = Path(self.root, license_filepath).read_text(encoding="utf-8")
+        license_text = Path(self.root, license_filepath).read_text(encoding="utf-8")
 
         # Copy LICENCE file in `sdist`
-        with open(Path(self.root, "LICENSE"), "w") as f:
-            f.write(license)
+        license_filepath = basename(license_filepath)
+        with open(Path(self.root, license_filepath), "w") as f:
```
DimitriPapadopoulos: I figured it's better to reuse the name of the licence file in the root directory instead of hard-wiring it here.

skore/pyproject.toml
```diff
@@ -5,13 +5,13 @@ description = """\
 Skore's open-source Python library accelerates ML model development with automated evaluation reports, \
 smart methodological guidance, and comprehensive cross-validation analysis.\
 """
-dynamic = ["license", "readme", "version"]
+dynamic = ["license-files", "readme", "version"]
 requires-python = ">=3.10"
 maintainers = [{ name = "skore developers", email = "skore@signal.probabl.ai" }]
+license = "MIT"
```
DimitriPapadopoulos, line 11: Shouldn't this appear in all `pyproject.toml` files?

thomass-dev: Yes, you can do it.

skore-hub-project/hatch/metadata.py
```diff
@@ -34,15 +35,16 @@ def update(self, metadata):
...
+            f.write(license_text)
 
         # Retrieve README from root files
         readme_filepath = self.config["readme"]["file"]
-        readme = Path(self.root, readme_filepath).read_text(encoding="utf-8")
+        readme_text = Path(self.root, readme_filepath).read_text(encoding="utf-8")
```
thomass-dev: Please revert this change from all `pyproject.toml` files to limit the diff.

DimitriPapadopoulos: I can do that, but this change really made the existing code easier to understand.

thomass-dev: It may have helped in your first iteration, where you wrote a new README file.
I don't think it is necessary now. Thanks.

thomass-dev:
> Please use `cp` instead of read/open/write to feed `licence-files`.

Please commit this change without force-pushing. This will make the diff easier to review.

skore-hub-project/hatch/metadata.py
```diff
@@ -34,15 +35,16 @@ def update(self, metadata):
         """
         # Retrieve LICENCE from root files
         license_filepath = self.config["license"]["file"]
-        license = Path(self.root, license_filepath).read_text(encoding="utf-8")
+        license_text = Path(self.root, license_filepath).read_text(encoding="utf-8")
 
         # Copy LICENCE file in `sdist`
-        with open(Path(self.root, "LICENSE"), "w") as f:
-            f.write(license)
+        license_filepath = basename(license_filepath)
```
thomass-dev: Prefer to use `pathlib`:

```python
license_filepath = Path(self.root, self.config["license"]["file"])
license_basename = license_filepath.name
```

DimitriPapadopoulos: I actually used [`shutil.copy`](https://docs.python.org/3/library/shutil.html#shutil.copy) so that I don't have to extract the basename.

skore-hub-project/hatch/metadata.py
```diff
@@ -32,13 +34,9 @@ def update(self, metadata):
...
         """
-        # Retrieve LICENCE from root files
+        # Retrieve LICENCE from root files into `sdist`
         license_filepath = self.config["license"]["file"]
-        license = Path(self.root, license_filepath).read_text(encoding="utf-8")
-
-        # Copy LICENCE file in `sdist`
-        with open(Path(self.root, "LICENSE"), "w") as f:
-            f.write(license)
+        copy(license_filepath, self.root)
```
thomass-dev, line 40: `licence_filepath` should be relative to the root, not to the current `metadata.py` file.

```suggestion
        copy(Path(self.root, license_filepath), self.root)
```

skore-hub-project/hatch/metadata.py
DimitriPapadopoulos: On the other hand, this has become slightly irrelevant now that the code has been simplified.

skore-hub-project/hatch/metadata.py
```diff
@@ -1,6 +1,8 @@
 """Hatchling hooks used to dynamically update the metadata of the package."""
 
+from os.path import basename
```
thomass-dev: It is a leftover of your iterations. Please revert.

skore-local-project/hatch/metadata.py
```diff
@@ -32,13 +34,9 @@ def update(self, metadata):
                 'readme': {'file': '../README.md'},
             }
         """
-        # Retrieve LICENCE from root files
+        # Retrieve LICENCE from root files into `sdist`
```
thomass-dev:
```suggestion
        # Copy LICENCE file from root to `sdist`
```

DimitriPapadopoulos:
Finally, should the classifiers from [`skore/pyproject.toml`](https://github.com/probabl-ai/skore/blob/main/skore/pyproject.toml) be copied to other `pyproject.toml` files?

https://github.com/probabl-ai/skore/blob/2eefa988d44c80dbad754c62a53c3030bfec993e/skore/pyproject.toml#L11-L28

More generally, how about striving to make the `pyproject.toml` files identical, for example `tool.ruff.lint.select`?

thomass-dev:
This is not necessary, because the only visible/searchable package we want is `skore/`.

skore-local-project/hatch/metadata.py
```diff
@@ -51,6 +48,6 @@ def update(self, metadata):
             version = self.config["version-default"]
 
         # Update metadata
-        metadata["license"] = {"text": license, "content-type": "text/plain"}
+        metadata["license-files"] = [license_filepath]
```
thomass-dev: This is not valid, because it will refer to `["../LICENCE"]`.

thomass-dev: I think you will need something like (untested):

```python
license_filepath = Path(self.root, self.config["license"]["file"])
sdist_license_filepath = Path(self.root, license_filepath.name)

copy(licence_filepath, sdist_license_filepath)

[...]

metadata["license-files"] = [sdist_license_filepath]
```

DimitriPapadopoulos: Ah, right. So we do need `basename` after all.

DimitriPapadopoulos: Or use `.name` as suggested earlier...

skore-hub-project/hatch/metadata.py
```diff
@@ -32,13 +33,11 @@ def update(self, metadata):
...
+        # Copy LICENCE file from root to `sdist`
         license_filepath = self.config["license"]["file"]
-        license = Path(self.root, license_filepath).read_text(encoding="utf-8")
-
-        # Copy LICENCE file in `sdist`
-        with open(Path(self.root, "LICENSE"), "w") as f:
-            f.write(license)
+        license_filepath = Path(self.root, license_filepath)
+        copy(license_filepath, self.root)
+        license_filename = license_filepath.name
```
thomass-dev, line 39: I can't push on your PR, so can you please make a small changes:

```python
# Copy LICENCE file from root to `sdist`
license_filepath = Path(self.root, self.config["license"]["file"])
license_filename = license_filepath.name

copy(license_filepath, self.root)

# Retrieve README from root files
```

DimitriPapadopoulos:
> This is not necessary, because the only visible/searchable package we want is `skore/`.

Makes sense, I would nevertheless suggest:
* keeping `tool.ruff.lint.select` identical
* adding some links such as `project.urls.Homepage` and `project.urls.Repository` to all packages

skore-hub-project/hatch/metadata.py
DimitriPapadopoulos: Strange, I allow pushing on my PRS (the default) as far as I know. Any way, done.

PR #2090, opened 2025-10-27
style: Enforce ruff/flake8-comprehensions rules (C4)

DimitriPapadopoulos:
I wonder whether you would be interested in adding a few ruff rule sets.

For example, I find `C4` rules improve the code. Some of them, such as rule [`C408`](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/) might be discutable - I would be happy to ignore such rules.

### DIFF
--- a/skore-hub-project/pyproject.toml
+++ b/skore-hub-project/pyproject.toml
@@ -80,7 +80,7 @@ exclude_also = [
 src = ["skore_hub_project"]
 
 [tool.ruff.lint]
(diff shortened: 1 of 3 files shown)


thomass-dev (commented):
I can be happy about this change, since it is added and applied by ruff and therefore standardizes the code.
What do you think @auguste-probabl? I know that, under certain conditions, you prefer the dict constructor to `{}`.

PR #2091, opened 2025-10-27
style: Enforce `RUF010` and `RUF015` rules

DimitriPapadopoulos:
Enforce [RUF010](https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/) and [RUF015](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/) rules to standardize the code base.

### DIFF
--- a/sphinx/conf.py
+++ b/sphinx/conf.py
@@ -14,8 +14,8 @@ from sphinx_gallery.sorting import ExplicitOrder
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath("sphinxext"))
-from github_link import make_linkcode_resolve
-from matplotlib_skore_scraper import matplotlib_skore_scraper
+from github_link import make_linkcode_resolve  # noqa
+from matplotlib_skore_scraper import matplotlib_skore_scraper  # noqa
 
 project = "skore"
 copyright = "2024, Probabl"


thomass-dev:
I think you have spotted good points:
- `{self!r}` instead of `repr(self)`,
- `next(iter(...))` instead of `list(...)[0]`.

Please add the related rules to ruff to enforce it.

sphinx/conf.py
```diff
@@ -14,8 +14,8 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 sys.path.insert(0, os.path.abspath("sphinxext"))
-from github_link import make_linkcode_resolve  # noqa
-from matplotlib_skore_scraper import matplotlib_skore_scraper  # noqa
+from github_link import make_linkcode_resolve
+from matplotlib_skore_scraper import matplotlib_skore_scraper
```
thomass-dev: This directory is actually not covered by `.pre-commit`.
Please revert, otherwise i think we will have red IDE with some alerts on `non-top imports`.

DimitriPapadopoulos:
> Please add the related rules to ruff to enforce it.

Would you prefer I add all `RUF` rules and ignore some of them, or add only [`RUF010`](https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/) and [`RUF015`](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/) for now?

thomass-dev:
> > Please add the related rules to ruff to enforce it.
> 
> Would you prefer I add all `RUF` rules and ignore some of them, or add only [`RUF010`](https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/) and [`RUF015`](https://docs.astral.sh/ruff/rules/unnecessary-iterable-allocation-for-first-element/) for now?

Only the ones you mention for now.

sphinx/conf.py
DimitriPapadopoulos: I ran ruff manually instead of pre-commit.

By the way, which tool does your IDE use to lint and raise such alerts? I think the relevant ruff alert would be [`E402`](https://docs.astral.sh/ruff/rules/module-import-not-at-top-of-file/) from pycodestyle, but the IDE might run a different linter.

DimitriPapadopoulos:
By the way, `T` rules are not documented:
https://docs.astral.sh/ruff/rules/

I think you want [`T20`](https://docs.astral.sh/ruff/rules/#flake8-print-t20) rules here:
https://github.com/probabl-ai/skore/blob/2eefa988d44c80dbad754c62a53c3030bfec993e/skore/pyproject.toml#L149-L150

Unless you also want [`T10`](https://docs.astral.sh/ruff/rules/#flake8-debugger-t10) rules.

thomass-dev (approved):
Nice! I think we have enough rules for our code-style. Thanks.

thomass-dev:
> By the way, `T` rules are not documented: https://docs.astral.sh/ruff/rules/
> 
> I think you want [`T20`](https://docs.astral.sh/ruff/rules/#flake8-print-t20) rules here:
> 
> https://github.com/probabl-ai/skore/blob/2eefa988d44c80dbad754c62a53c3030bfec993e/skore/pyproject.toml#L149-L150
> 
> Unless you also want [`T10`](https://docs.astral.sh/ruff/rules/#flake8-debugger-t10) rules.

We want all rules prefixed by `T`.

DimitriPapadopoulos:
So that would be rule sets `flake8-print` and `flake8-debug`.

PR #2092, opened 2025-10-28
fix(skore): Cache all permutations importance

glemaitre:
Always put the permutation importance inside the cache such that we fetch from it when sending to the hub. If not in the cache, we should avoid the computation and just show a placeholder on the hub side.


thomass-dev (approved):

thomass-dev:
Partially addresses https://github.com/probabl-ai/skore-hub/issues/812.

PR #2097, opened 2025-10-29
chore(all): Enforce `PT003` and `PT014` rules

DimitriPapadopoulos:
* [PT003](https://docs.astral.sh/ruff/rules/pytest-extraneous-scope-function/#pytest-extraneous-scope-function-pt003) pytest-extraneous-scope-function
* [PT014](https://docs.astral.sh/ruff/rules/pytest-duplicate-parametrize-test-cases/#pytest-duplicate-parametrize-test-cases-pt014) pytest-duplicate-parametrize-test-cases

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -150,6 +150,8 @@ select = [
   "SIM",
   # flake8-print
   "T",
+  # flake8-pytest-style
+  "PT003", "PT014",
   # isort
   "I",
   # pydocstyle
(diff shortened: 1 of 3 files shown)


skore/pyproject.toml
```diff
@@ -150,6 +150,8 @@ select = [
   "SIM",
   # flake8-print
   "T",
+  # flake8-pytest-style
+  "PT003", "PT014",
```
thomass-dev:
```suggestion
  # flake8-pytest-style
  "PT003", "PT014",
```

thomass-dev (approved):
LGTM, thanks.

`PT014` is really a good rule, compare to `PT003` which is a bit overkill... I don't think we need more rules like this latest one.

PR #2101, opened 2025-10-29
chore(all): Enforce PERF and PYI rules

### LINKED ISSUE #2099: Some more ruff rules?

Again, I know you would rather avoid linter rules inflation, but I would like to suggest a few that do fix a couple things.

[**flake8-pyi (PYI)**](https://docs.astral.sh/ruff/rules/#flake8-pyi-pyi)

I suggest applying these rules only:
* [unnecessary-type-union (PYI055)](https://docs.astral.sh/ruff/rules/unnecessary-type-union/#unnecessary-type-union-pyi055)
  ```
  PYI055 Multiple `type` members in a union. Combine them into one.
  ```
* [generic-not-last-base-class (PYI059)](https://docs.astral.sh/ruff/rules/generic-not-last-base-class/#generic-not-last-base-class-pyi059)
  ```
  PYI059 `Generic[]` should always be the last base class
  ```

----------

[**Perflint (PERF)**](https://docs.astral.sh/ruff/rules/#perflint-perf)

I know these rules are about micro-optimizations and may be debatable in terms of readability/maintainability. Neverthless, I have become used to them over time so  readability is not a problem any more (for me):
* [manual-list-comprehension (PERF401)](https://docs.astral.sh/ruff/rules/manual-list-comprehension/#manual-list-comprehension-perf401)
  ```
  PERF401 Use a list comprehension to create a transformed list
  ```
* [manual-dict-comprehension (PERF403)](https://docs.astral.sh/ruff/rules/manual-dict-comprehension/#manual-dict-comprehension-perf403)
  ```
  PERF403 Use a dictionary comprehension instead of a for-loop
  ```

----------

[**pandas-vet (PD)**](https://docs.astral.sh/ruff/rules/#pandas-vet-pd)

The only issue is about consistency but makes sense:
* [pandas-use-of-dot-is-null (PD003)](https://docs.astral.sh/ruff/rules/pandas-use-of-dot-is-null/#pandas-use-of-dot-is-null-pd003)
  ```
  PD003 `.isna` is preferred to `.isnull`; functionality is equivalent
  ```

See also:
* [Pandas isna() and isnull(), what is the difference?](https://stackoverflow.com/questions/52086574/pandas-isna-and-isnull-what-is-the-difference#52628380)
* https://github.com/pandas-dev/pandas/pull/16972

-----

[**refurb (FURB)**](https://docs.astral.sh/ruff/rules/#refurb-furb)

One rule of interest:
* [if-expr-min-max (FURB136)](https://docs.astral.sh/ruff/rules/if-expr-min-max/#if-expr-min-max-furb136)
  ```
  FURB136 Replace `if` expression with `max(..., ...)`
  ```

DimitriPapadopoulos:
Fixes #2099.

### DIFF
--- a/skore-hub-project/pyproject.toml
+++ b/skore-hub-project/pyproject.toml
@@ -83,7 +83,22 @@ exclude_also = [
 ]
 
 [tool.ruff.lint]
-select = ["E", "F", "UP", "B", "C4", "SIM", "T", "PYI055", "PYI059", "PT003", "PT014", "I", "PERF", "D", "RUF010", "RUF015"]
+select = [
+    "B",
+    "C4",
+    "T10",
+    "T20",
+    "PYI055", "PYI059",
+    "PT003", "PT014",
+    "SIM",
+    "I",
+    "PERF",
+    "E",
+    "D",
+    "F",
+    "UP",
+    "RUF010", "RUF015",
+]
 ignore = ["PERF203"]
 
 [tool.ruff.lint.pydocstyle]
--- a/skore-local-project/pyproject.toml
+++ b/skore-local-project/pyproject.toml
@@ -74,7 +74,22 @@ exclude_also = [
 ]
 
 [tool.ruff.lint]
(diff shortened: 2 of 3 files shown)


thomass-dev (approved):
LGTM, thanks!

PR #2146, opened 2025-11-12
chore(skore): Rename `data_source` option from `"all"` to `"both"`

auguste-probabl:
In `metrics` and `data` accessors.


MarieSacksick (approved):
good for me, thanks!

Issue #1367, opened 2025-02-27
feat(skore): Automatically rescale coefficients of linear model during inspection


MarieSacksick:
### Is your feature request related to a problem? Please describe.

Due to the scaling done prior to linear models, we can not interpret the coefficient values with regards to the original unit of the feature, while it is interesting, in particular in regression.  
It would be nice to provide easily unscaled coefficients, without forgetting the intercerpt, so that the coefficients can interpreted using the original units.

### Describe the solution you'd like

Maybe what is done in https://github.com/probabl-ai/skore/blob/a577cad91663adfb61dc9b66e7fb220aac97339a/examples/model_evaluation/plot_feature_importance.py#L259-L265 (lines 259 to 265)

### Additional context

cf the example plot_feature_importance.py

glemaitre:
It could be tricky when you have a complex preprocessing, e.g. a step with `PolynomialFeatures` or `PCA`. So we would need to think thoroughly this feature to know in which case we can provide it and thus what should control the behaviour.

MarieSacksick:
Yes, at first I thought it could be a "good first issue" since the code was available in Sylvain's notebook. And then I realized it wouldn't be :).

PR #2152, opened 2025-11-12
fix(skore): Use display API for coefficients display

glemaitre:
Follow-up to #2043 
PRs needed to be merged before:
- #2181
- #2182

This PR intends to follow more strictly our `Display` API.

- expose a `_compute_data_for_display` in the `Display` class
- store a long format dataframe containing all the information required independently of the type of report
- use `.frame` to drop the unnecessary columns depending of the type of report
- leverage the `.frame` long format to directly use seaborn out of the box plotting API
- add the `subplots_by` API to show on different subplot

### DIFF
--- a/skore/src/skore/_sklearn/_plot/utils.py
+++ b/skore/src/skore/_sklearn/_plot/utils.py
@@ -253,7 +253,7 @@ def _despine_matplotlib_axis(
         The matplotlib axis to despine.
     axis_to_despine : tuple of str, default=("top", "right")
         The axes to despine.
-    remove_ticks : bool, default=True
+    remove_ticks : bool, default=False
         Whether to remove the ticks when adding "left" and "bottom" to the axis to
         despine.
     x_range : tuple of float, default=(0, 1)


skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,680 @@
...
+        if add_background_features:
+            for feature_idx in range(n_features):
+                if feature_idx % 2 == 0:
+                    ax.axhspan(
+                        feature_idx - 0.5,
+                        feature_idx + 0.5,
+                        color="lightgray",
+                        alpha=0.1,
+                        zorder=0,
+                    )
```
auguste-probabl:
```suggestion
            for feature_idx in range(0, n_features, 2):
                ax.axhspan(
                    feature_idx - 0.5,
                    feature_idx + 0.5,
                    color="lightgray",
                    alpha=0.1,
                    zorder=0,
                )
```

skore/src/skore/_sklearn/_plot/base.py
```diff
@@ -118,11 +135,18 @@ def _style_params(self) -> list[str]:
...
+        self, *, policy: Literal["override", "update"] = "override", **kwargs: Any
+    ):
         """Set the style parameters for the display.
 
         Parameters
         ----------
+        policy : {"override", "update"}, default="override"
+            Policy to use when setting the style parameters. If "override", the style
+            parameters will be set to the values provided. If "update", the style
+            parameters will be updated with the values provided.
```
auguste-probabl: If override, existing settings are set to the provided values. If update, existing settings are not changed; only settings that were previously unset are changed.

glemaitre: Yep I think that I'll take out those changes in another PR.

glemaitre: I opened the corresponding PR.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,686 @@
...
+                ax=self.ax_,
+                **plot_function_kwargs,
+            )
+            add_background_features = hue is not None and plot_function == sns.boxplot
+            self._decorate_matplotlib_axis(
+                ax=self.ax_,
+                add_background_features=add_background_features,
+                n_features=frame["feature"].nunique(),
+            )
+            self.ax_.set_title(f"{name}")
```
auguste-probabl:
```suggestion
            self.ax_.set_title(name)
```

auguste-probabl:
```python
from skore import EstimatorReport, train_test_split, ComparisonReport, CrossValidationReport
iris = load_iris(as_frame=True)
X, y = iris.data, iris.target
y = iris.target_names[y]
split_data = train_test_split(
    X=X, y=y, random_state=0, as_dict=True, shuffle=True
)
report_1 = CrossValidationReport(LogisticRegression(), X, y)
report_2 = CrossValidationReport(LogisticRegression(max_iter=500), X, y)
display = report.feature_importance.coefficients()
display.plot()
```
image

The variance is very low which makes it difficult to see the colors.

glemaitre:
> The variance is very low which makes it difficult to see the colors.

I'm not sure that we can do much. Here, the intercept value is dominant compared to the coefficients because we did not standardize the data. If you use a `StandardScaler` and the `LogisticRegression`, you will end-up with:

image

which is then manageable.

So I'm wondering if we should not revisit this issue when solving https://github.com/probabl-ai/skore/issues/1367 or with an option to whether or not to show the intercept.

Also I see that there is a small regression with the background. I think that the following is easier to read with the ordering of the legend and grouping.

image

auguste-probabl:
> I'm not sure that we can do much. 

I think switching back to a horizontal bar plot with error bars would be a good first step

> Also I see that there is a small regression with the background.

Agreed, the shading does help!

glemaitre:
OK, so I switch to seaborn to handle the columns as we discussed in https://github.com/probabl-ai/skore/issues/2190

I also implemented a mix of stripplot and boxplot for the problematic plot and allow to exclude the intercept if it is desired.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,630 @@
...
+        )
+
+    def _plot_matplotlib(
+        self,
+        *,
+        include_intercept: bool = True,
+        subplots_by: Literal["estimator", "label", "output"] | None = None,
+        barplot_kwargs: dict[str, Any] | None = None,
+        boxplot_kwargs: dict[str, Any] | None = None,
+        stripplot_kwargs: dict[str, Any] | None = None,
```
auguste-probabl, line 227:
```suggestion
        barplot_kwargs: dict[str, Any] = {},
        boxplot_kwargs: dict[str, Any] = {},
        stripplot_kwargs: dict[str, Any] = {},
```

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,630 @@
...
+                data=frame,
+                x="coefficients",
+                y="feature",
+                hue=hue,
+                col=col,
+                kind="strip",
+                dodge=True,
+                **stripplot_kwargs,
+            )
+            facet = facet.map_dataframe(
```
auguste-probabl:
```suggestion
            ).map_dataframe(
```

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,630 @@
...
+                x="coefficients",
+                y="feature",
+                hue=hue,
+                dodge=True,
+                **boxplot_kwargs,
+            )
+            add_background_features = hue is not None
+
+        self.figure_, self.ax_ = facet.figure, facet.axes.squeeze()
+        # self.figure_.suptitle(name)
```
auguste-probabl: forgot to remove this line?

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
glemaitre: We cannot do this one because a dictionary is mutable: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

auguste-probabl: Right.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,630 @@
...
+        stripplot_kwargs: dict[str, Any] | None = None,
+    ) -> None:
+        """Plot the coefficients for the different features.
+
+        Parameters
+        ----------
+        include_intercept : bool, default=True
+            Whether or not to include the intercept in the dataframe.
+
+        subplots_by : {"estimator", "label", "output"}, default=None
```
GaetandeCast: Let's choose between `subplots_by` and `subplot_by`. I prefer without the s (`subplot_by`) but its not a strong opinion.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,630 @@
...
+            The column to use for subplotting and dividing the coefficients into
+            subplots. If `None`, an automatic choice is made depending on the type of
```
GaetandeCast: IMO `None` and `"auto"` should be distinct choices for the `subplots_by`. There may be situations where we want to allow not subploting (`subplots_by=None`) but not have it be the default behaviour e.g. the default is `subplots_by="label"`.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,630 @@
...
+            facet = sns.catplot(
+                data=frame,
+                x="coefficients",
+                y="feature",
+                hue=hue,
+                col=col,
+                kind="bar",
+                **barplot_kwargs,
+            )
+            add_background_features = False
```
GaetandeCast: Why not have the shaded background in every case ? I think keeping consistent between reports is better.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
GaetandeCast: The suptitle is missing. Having it contain information on the data source and pos_label, when relevant, [like here](https://github.com/probabl-ai/skore/pull/2193/changes#diff-91d04a59777637c71fb2df63da34b82fa94a8b0a02790fa7acfbccf7095ee4d4R285-R287) would be nice I think.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
glemaitre: yep let's go with singular to be aligned with `group_by` for instance.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
glemaitre: Yes it is a good remark.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
glemaitre: It does not look nice (I tried it). The bars already give an effect where you can easily parse the information.

GaetandeCast: Indeed in binary classif/regression I agree it does not look good. I was trying in multiclass classif and there I think it makes sense, see:
image
image
So I suggest setting 
```add_background_features = hue is not None``` 
independently of the report type.

Comment (GaetandeCast) [reply skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py]: also the axis is a bit squished in this example :thinking:

Comment (glemaitre) [reply skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py]: I'll check if I can limit the number of ticks

File: skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py [hunk 10]
```diff
@@ -0,0 +1,645 @@
...
+            Keyword arguments to be passed to seaborn's :func:`seaborn.stripplot` for
+            rendering the coefficients with a :class:`~skore.ComparisonReport` of
+            :class:`~skore.CrossValidationReport`.
+        """
+        # help mypy to understand the following variable types
+        hue: str | None = None
+
+        # {"estimator", "label"} or {"estimator", "output"} or {"estimator"}
+        columns_to_groupby = self._get_columns_to_groupby(frame=frame)
+        if subplot_by not in ("auto", None) and subplot_by not in columns_to_groupby:
```
GaetandeCast, line 502: `subplot_by=None` should not be allowed in multiclass classification for ComparisonReports as it lead to unreadable plots:

Image

Image

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,645 @@
...
+            frame=frame,
+            report_type=report_type,
+            hue=hue,
+            col=col,
+            barplot_kwargs=barplot_kwargs,
+            boxplot_kwargs=boxplot_kwargs,
+            stripplot_kwargs=stripplot_kwargs,
+        )
+
+        self.figure_.suptitle(f"Coefficients of {estimator_name}")
```
GaetandeCast: In multiclass classif  `subplot_by` can be `"label"` but it is not reflected in the title, making it inconsistent with ComparisonReport.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,645 @@
...
+        """
+        # {"label"} or {"output"} or {}
+        columns_to_groupby = self._get_columns_to_groupby(frame=frame)
+        if subplot_by == "auto":
+            subplot_by = None
+
+        if subplot_by is not None and not len(columns_to_groupby):
+            raise ValueError(
+                "No columns to group by. `subplot_by` is expected to be None."
+            )
```
GaetandeCast, line 414:
```suggestion
        if subplot_by is not None and not len(columns_to_groupby):
            raise ValueError(
                "No columns to group by. `subplot_by` is expected to be None or auto."
            )
```

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,645 @@
...
+        elif subplot_by is not None and subplot_by not in columns_to_groupby:
+            raise ValueError(
+                f"Column {subplot_by} not found in the frame. It should be one "
+                f"of {', '.join(columns_to_groupby)}."
```
GaetandeCast:
```suggestion
                f"of {', '.join(columns_to_groupby + ["auto", "None"])}."
```

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -0,0 +1,645 @@
...
+        """
+        # help mypy to understand the following variable types
+        hue: str | None = None
+
+        # {"estimator", "label"} or {"estimator", "output"} or {"estimator"}
+        columns_to_groupby = self._get_columns_to_groupby(frame=frame)
+        if subplot_by not in ("auto", None) and subplot_by not in columns_to_groupby:
+            raise ValueError(
+                f"Column {subplot_by} not found in the frame. It should be one "
+                f"of {', '.join(columns_to_groupby)}."
```
GaetandeCast: This error message should mention "auto", and None when in binary classification.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
auguste-probabl: "auto" with quotes

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
glemaitre: Done and added test

examples/use_cases/plot_feature_importance.py
```diff
@@ -708,12 +712,13 @@ def add_y_true_pred(model_report, split):
...
-    title="Model weights",
-    xlabel="Coefficient",
-    ylabel="Feature",
+(
+    selectk_ridge_report.feature_importance.coefficients()
+    .frame()
+    .set_index("feature")
+    .sort_values(by="coefficients", key=abs, ascending=True)
+    .tail(15)
+    .plot.barh(title="Model weights", xlabel="Coefficient", ylabel="Feature")
```
GaetandeCast, line 721: Just a thought: it would be nice to have a parameter that allows us to only plot the top or bottom n features to avoid such code. We could promote the use of such a parameter in a warning when the number of features is large.
It could be the subject of a follow-up PR.

glemaitre: Yep, let's do in another PR but it is a good suggestion. It should not be the default because it depends on what type of analysis you are going to do with the magnitude.

glemaitre:
OK so I addressed all reviews.

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -250,19 +251,29 @@ def _despine_matplotlib_axis(
     ----------
     ax : matplotlib.axes.Axes
         The matplotlib axis to despine.
+    axis_to_despine : tuple of str, default=("top", "right")
+        The axes to despine.
+    remove_ticks : bool, default=True
```
GaetandeCast:
```suggestion
    remove_ticks : bool, default=False
```

Line 243: `remove_ticks: bool = False,`

glemaitre: Done

Issue #2153, opened 2025-11-13
Make cache thread-safe


auguste-probabl:
### Is your feature request related to a problem? Please describe.

We are currently delegating adding keys to the cache to the callers:

https://github.com/probabl-ai/skore/blob/9bae5d2ee9ce212add2fb6b35514e5020148912a/skore/src/skore/_sklearn/_base.py#L337-L341

This makes skore code more complex as callers have to know about this peculiarity of the cache.

### Describe the solution you'd like

The current code is a quick fix to preserve thread-safety. We should rework the caching mechanism to make it thread-safe, and move back all the adding keys logic into the cache module.

thomass-dev:
Before reworking the caching mechanism, we could simply share an [explicit lock](https://docs.python.org/3/library/threading.html#threading.Lock) wherever the cache is managed.

Funny fact: i asked @glemaitre this morning if it will be a problem to compute metrics in parallel (in https://github.com/probabl-ai/skore/pull/2114), and he answered me "the cache could be an issue".

Can we prioritize a quick fix with what i proposed if you are agree?

glemaitre:
@thomass-dev for your case, would it be enough to trigger the collection of the prediction using the `cache_predictions` and then parallelize the rest that should mainly read or write with different key in the dictionary?

glemaitre:
Oh I'm reading your message about the lock. Until that the lock is for writing in the dict then it would be fine. I think that accessing in "read-only" mode should not be locked.

glemaitre:
Let's add this issue to the meta issue about the cache for when we will redesign. In the meanwhile, let's find a workaround that works for @thomass-dev

thomass-dev:
>  Until that the lock is for writing in the dict then it would be fine. I think that accessing in "read-only" mode should not be locked.

Yes, i also think that locking operations that can change the state of the dictionary is sufficient.
IMO, the worst case in reading is: i try to access a resource using the cache-key `x` that is unavailable, so i compute it. In the meantime it is finally available. I will overwrite the value. I may have done the work twice.

I'm starting an implementation.

auguste-probabl:
I'd like to test whether this impacts performance on something like a big ComparisonReport of many CrossValidationReports.

auguste-probabl:
still relevant?

PR #2158, opened 2025-11-13
feat(skore): Make the cache thread-safe

thomass-dev:
Partially addresses https://github.com/probabl-ai/skore/issues/2153.

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -1648,10 +1648,18 @@ class _MetricsAccessor(
         data_source: DataSource,
         response_method: str | list[str] | tuple[str, ...],
         display_class: type[
-            RocCurveDisplay | PrecisionRecallCurveDisplay | PredictionErrorDisplay
+            RocCurveDisplay
+            | PrecisionRecallCurveDisplay
+            | PredictionErrorDisplay
+            | ConfusionMatrixDisplay
         ],
         display_kwargs: dict[str, Any],
-    ) -> RocCurveDisplay | PrecisionRecallCurveDisplay | PredictionErrorDisplay:
+    ) -> (
+        RocCurveDisplay
+        | PrecisionRecallCurveDisplay
+        | PredictionErrorDisplay
+        | ConfusionMatrixDisplay
+    ):
         """Get the display from the cache or compute it.
 
         Parameters
@@ -1676,7 +1684,7 @@ class _MetricsAccessor(
             The display class.
 
         display_kwargs : dict
-            The display kwargs used by `display_class._from_predictions`.
+            The display kwargs used by `display_class._compute_data_for_display`.
 
         Returns
         -------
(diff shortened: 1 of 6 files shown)


skore/src/skore/_utils/_cache.py
```diff
@@ -0,0 +1,26 @@
...
+
+class Cache(UserDict):
+    """Thread-safe cache based on ``dict``, with an explicit lock on write/delete."""
+
+    __setitem__ = method_with_explicit_lock(UserDict.__setitem__)
+    __delitem__ = method_with_explicit_lock(UserDict.__setitem__)
+    clear = method_with_explicit_lock(UserDict.clear)
+    pop = method_with_explicit_lock(UserDict.pop)
+    popitem = method_with_explicit_lock(UserDict.popitem)
+    update = method_with_explicit_lock(UserDict.update)
```
thomass-dev, line 27: Add a test ensuring methods are well wrapped.

thomass-dev: Fixed in [3c81980](https://github.com/probabl-ai/skore/pull/2158/commits/3c8198052715883baca8e5bb5f6c31b7f076e6f8).

skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -1129,7 +1129,7 @@ def _get_display(
         total_estimators = len(self._parent.estimator_reports_)
         progress.update(main_task, total=total_estimators)
 
-        if cache_key in self._parent._cache:
+        if cache_key and cache_key in self._parent._cache:
```
thomass-dev, line 1132: `cache_key` can be `None` [here](https://github.com/probabl-ai/skore/pull/2158/files#diff-1c71e143c03f8a23fe57afa65cdb3d1375e1e369e6972c9c70d2bb33e58dfae7L1116), and should be ignored.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -1699,7 +1699,7 @@ def _get_display(
                 cache_key_parts.append(data_source)
             cache_key = tuple(cache_key_parts)
 
-        if cache_key in self._parent._cache:
+        if cache_key and cache_key in self._parent._cache:
```
thomass-dev, line 1714: `cache_key` can be `None` [here](https://github.com/probabl-ai/skore/pull/2158/files#diff-cdd2327c0ba22f04adfecaeceb07b2149698137a132d3eb4ccdb2b32f06a9ae0R1692), and should be ignored.

auguste-probabl: If you look at the typing yes `cache_key` can be None, but in practice `cache_key` being `None` means that the result should not be cached (e.g. if the computation is not deterministic), therefore, if there are no bugs, then `None` will never be a cache key. We can make that more explicit, maybe by setting `cache_key` to a dummy value like `_DEFAULT` or `_DONT_CACHE` instead of `None`?

auguste-probabl:
As mentioned in #2153 I'd be interested to compare the performance of fitting e.g. a ComparisonReport of CrossValidationReports before and after this change. Other than that the design is clean.

PR #2165, opened 2025-11-18
fix(skore): Add tests and follow-up fixes for ConfusionMatrixDisplay

### LINKED ISSUE #2160: api(skore): follow-up fixes for #1759

#1759 was merged and already improved by a lot the API of the `ConfusionMatrixDisplay`. However, we need still to address those little glitches:

- `report_type` is not documented in the docstring of the class while being an attribute.
- we are passing `**kwargs` in plot. However, our API is to define a variable, for instance `imshow_kwargs` which is a Python dictionary. One needs to look at the `set_style` implementation to observe that we need to have such parameter to properly set the property.
- l.156/l.157 `self.include_values = include_values` and `self.values_format = values_format` in the `plot` method. It is weird because we are not supposed to set any attribute without ending `_`. The question would be also "do we need to have those attribute stored?"
- l.285 has a local `import pandas as pd`. It should be placed on the top of the file.
- in the frame, we have a `gettattr` to get `display_labels`. It is weird because it is for sure an attribute of the class.

As another future works, I think that `_compute_data_for_display` could store a long dataframe where each row contain the number of observation, the name of the true label, the name of the predicted label (and because of #2159, certainly the estimator name and the split). It is quite aligned with the other plot. The job of `.frame()` will therefore be to reshape this long frame into a real confusion matrix. Not sure yet what would be the output of the `ComparisonReport` however.

But the first batch of comment could already be addressed.

GaetandeCast:
Fixes #2160, improves on #1759.

cc @glemaitre

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,17 +1,20 @@
 version: 2
 updates:
-
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: "monthly"
+    cooldown:
+      default-days: 14
     commit-message:
       prefix: "chore(dependencies): GITHUB-ACTIONS"
-
   - package-ecosystem: "pip"
     directory: "/"
     schedule:
       interval: "weekly"
+    cooldown:
+      default-days: 14
+      semver-patch-days: 7
     commit-message:
       prefix: "chore(dependencies): PIP"
     groups:
--- a/skore-hub-project/src/skore_hub_project/artifact/media/feature_importance.py
+++ b/skore-hub-project/src/skore_hub_project/artifact/media/feature_importance.py
@@ -45,7 +45,7 @@ class Permutation(FeatureImportance[EstimatorReport], ABC):  # noqa: D101
     name: Literal["permutation"] = "permutation"
 
     def content_to_upload(self) -> bytes | None:  # noqa: D102
-        for key, obj in reversed(self.report._cache.items()):
+        for key, obj in reversed(list(self.report._cache.items())):
             if len(key) < 7:
                 continue
 
--- a/skore-hub-project/tests/conftest.py
+++ b/skore-hub-project/tests/conftest.py
@@ -1,4 +1,6 @@
 from datetime import datetime, timezone
+from functools import partial
+from importlib import reload
 from unittest.mock import Mock
 from urllib.parse import urljoin
 
@@ -141,11 +143,10 @@ def monkeypatch_tmpdir(monkeypatch, tmp_path):
 
     https://docs.python.org/3/library/tempfile.html#tempfile.gettempdir
     """
-    import importlib
     import tempfile
 
     monkeypatch.setenv("TMPDIR", str(tmp_path))
-    importlib.reload(tempfile)
+    reload(tempfile)
 
 
 @fixture
@@ -155,7 +156,10 @@ def monkeypatch_skrub(monkeypatch):
 
     https://github.com/skrub-data/skrub/blob/35f573ce586fe61ef2c72f4c0c4b188ebf2e664b/skrub/_reporting/_html.py#L153
     """
+    from skrub._dataframe import sample
+
     monkeypatch.setattr("secrets.token_hex", lambda: "<token>")
+    monkeypatch.setattr("skrub._dataframe.sample", partial(sample, seed=42))
 
 
 @fixture
@@ -210,6 +214,16 @@ def monkeypatch_sklearn_estimator_html_repr(monkeypatch):
     monkeypatch.setattr(importpath, lambda self: "<id>")
 
 
+@fixture
+def monkeypatch_rich(monkeypatch):
+    """Make `rich` silent."""
+    from rich.console import Console
+
+    monkeypatch.setattr("rich.console.Console", partial(Console, quiet=True))
+    monkeypatch.setattr("skore.console.quiet", True)
+    monkeypatch.setattr("skore_hub_project.console.quiet", True)
+
+
 @fixture(autouse=True)
 def setup(
     monkeypatch_tmpdir,
@@ -217,4 +231,5 @@ def setup(
     monkeypatch_skrub,
     monkeypatch_skore_hub_envars,
     monkeypatch_sklearn_estimator_html_repr,
+    monkeypatch_rich,
 ): ...
--- a/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
@@ -67,12 +67,14 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
 
         Parameters
         ----------
-        data_source : {"test", "train", "X_y"}, default="test"
+        data_source : {"test", "train", "X_y", "both"}, default="test"
             The data source to use.
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
             - "X_y" : use the provided `X` and `y` to compute the metric.
+            - "both" : use both the train and test sets to compute the metrics and
+              present them side-by-side.
 
         X : array-like of shape (n_samples, n_features), default=None
             New data on which to compute the metric. By default, we use the validation
@@ -254,6 +256,7 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
                     indicator_favorability=metric_kwargs.get(
                         "indicator_favorability", False
                     ),
+                    data_source=data_source,
                 )
             else:  # "CrossValidationReport"
                 results = _combine_cross_validation_results(
--- a/skore/src/skore/_sklearn/_comparison/utils.py
+++ b/skore/src/skore/_sklearn/_comparison/utils.py
@@ -1,14 +1,16 @@
 import copy
+from typing import Literal
 
 import pandas as pd
 
-from skore._sklearn.types import Aggregate
+from skore._sklearn.types import Aggregate, DataSource
 
 
 def _combine_estimator_results(
     individual_results: list[pd.DataFrame],
     estimator_names: list[str],
     indicator_favorability: bool,
+    data_source: DataSource | Literal["both"],
 ) -> pd.DataFrame:
     """Combine a list of dataframes provided by `EstimatorReport`s.
 
@@ -23,6 +25,9 @@ def _combine_estimator_results(
 
     indicator_favorability : bool
         Whether to keep the Favorability column.
+
+    data_source : {"test", "train", "X_y", "both"}
+        The data source.
 
     Examples
     --------
@@ -55,6 +60,7 @@ def _combine_estimator_results(
     ...     individual_results,
     ...     estimator_names,
     ...     indicator_favorability=False,
+    ...     data_source="test",
     ... )
     Estimator    LogisticRegression_1  LogisticRegression_2
     Metric
@@ -73,7 +79,17 @@ def _combine_estimator_results(
     else:
         favorability = None
 
-    results.columns = pd.Index(estimator_names, name="Estimator")
+    if data_source == "both":
+        results.columns = pd.Index(
+            [
+                x
+                for name in estimator_names
+                for x in [f"{name} (train)", f"{name} (test)"]
+            ],
+            name="Estimator",
+        )
+    else:
+        results.columns = pd.Index(estimator_names, name="Estimator")
 
     if favorability is not None:
         results["Favorability"] = favorability
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -1976,7 +1976,7 @@ class _MetricsAccessor(
     def prediction_error(
         self,
         *,
-        data_source: DataSource = "test",
+        data_source: DataSource | Literal["both"] = "test",
         X: ArrayLike | None = None,
         y: ArrayLike | None = None,
         subsample: float | int | None = 1_000,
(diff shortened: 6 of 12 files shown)


skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -28,18 +30,38 @@ class ConfusionMatrixDisplay(DisplayMixin):
...
+        ``n_classes - 1``.
+
+    normalize : {'true', 'pred', 'all'} or None
+        Normalizes confusion matrix over the true (rows), predicted (columns)
+        conditions or all the population. If None, confusion matrix will not be
+        normalized.
+
+    report_type : {"comparison-cross-validation", "comparison-estimator", \
+            "cross-validation", "estimator"}
+        The type of report.
```
glemaitre: The convention that we used up to now is the same than scikit-learn. We don't document again the attributes that are just affected in `__init__`, e.g. `confusion_matrix`. Usually, we only document the attribute that are created (the one finishing with an `_`).

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -96,6 +120,7 @@ def plot(
             values_format=values_format,
             cmap=cmap,
```
glemaitre: Let's remove the `cmap` parameter because it is actually a parameter that can be passed to `imshow_kwargs` directly. It is some old legacy in scikit-learn.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -96,6 +120,7 @@ def plot(
             values_format=values_format,
```
glemaitre: In terms of API, I think that we should move the `value_format` into a `text_kwargs`. In short, it would be handy to benefit from the `set_style` for this part as well.

```python
display.set_style(text_kwargs=(value_format=".5g"))
display.plot()  # is going to use the format without the need to specify it again.
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -282,9 +314,7 @@ def frame(self):
         frame : pandas.DataFrame
             The confusion matrix as a dataframe.
         """
-        import pandas as pd
-
         cm = self.confusion_matrix
-        display_labels = getattr(self, "display_labels", None)
+        display_labels = self.display_labels
 
         return pd.DataFrame(cm, index=display_labels, columns=display_labels)
```
glemaitre:
```suggestion
        return pd.DataFrame(
            self.confusion_matrix,
            index=self.display_labels,
            columns=self.display_labels,
        )
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
glemaitre: The `normalize` parameter should also be added to this dict. It controls the visual that we display and does not really require any processing. Right now, we don't have a way to modify it once we instantiated the class.

glemaitre: Oh I see that we call `sklearn_confusion_matrix` and pass the normalize. So let's never normalize the results and instead make that `plot` make the normalization. One thing we can do is to precompute in some attribute the some to be used for the different normalization.

glemaitre (commented):
First thought before looking at the test

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: fixed in 27b66bc3f25dcfd8b1a6c8632c3086fda11777c4

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: done in 27b66bc3f25dcfd8b1a6c8632c3086fda11777c4 for `cmap` and in 20ddf6becba50dc6b883c7b53509b0d1054b445b for `normalize`. I also added normalize in frame.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: done in 27b66bc3f25dcfd8b1a6c8632c3086fda11777c4

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: done in c5da6364650b4494554045c768c8d6a8c1ea6d2b

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -16,17 +18,16 @@ class ConfusionMatrixDisplay(DisplayMixin):
 
     Parameters
     ----------
-    confusion_matrix : ndarray of shape (n_classes, n_classes)
-        Confusion matrix.
+    confusion_matrix_data : pd.DataFrame
```
GaetandeCast: I'm not a big fan of this name tbh

glemaitre: Yep. maybe we could just keep `confusion_matrix` directly?

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -16,17 +18,16 @@ class ConfusionMatrixDisplay(DisplayMixin):
...
+        Confusion matrix data. Each row contains a confusion matrix.
 
-    display_labels : list of str, default=None
+    display_labels : list of str
```
glemaitre, line 25: Oh I did not realise but we don't really need this variable. `estimator.classes_` can directly be some labels with text and then we are going to reuse those strings.

glemaitre: Arff, you can disregard my previous comment. It is not an input from the user so fine to have them there.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -139,29 +146,47 @@ def _plot_single_estimator(
...
+        cm = self.confusion_matrix_data.values.reshape(n_classes, n_classes)
+        if normalize == "true":
+            cm = cm / cm.sum(axis=1, keepdims=True)
+        elif normalize == "pred":
+            cm = cm / cm.sum(axis=0, keepdims=True)
+        elif normalize == "all":
+            cm = cm / cm.sum()
+
+        imshow_kwargs_validated = _validate_style_kwargs(
+            {"cmap": "Blues"},
```
glemaitre: We can put this directly in the `_default_imshow_kwargs` when we create it.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -175,16 +200,28 @@ def _plot_single_estimator(
...
 
-        self.text_ = np.empty_like(cm, dtype=object)
-        if self.include_values:
-            fmt = self.values_format or (".2f" if self.normalize else "d")
+        if include_values:
+            text_kwargs_validated = _validate_style_kwargs(
+                {},
+                text_kwargs or self._default_text_kwargs or {},
+            )
+            values_format = text_kwargs_validated.pop("values_format", None)
```
glemaitre: We can as well add the `value_format` default in the `_default_text_kwargs` dictionary.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -265,26 +297,42 @@ def _compute_data_for_display(
...
-
-        return pd.DataFrame(cm, index=display_labels, columns=display_labels)
+        n_classes = len(self.display_labels)
+        cm = self.confusion_matrix_data.values.reshape(n_classes, n_classes)
+        if normalize == "true":
+            cm = cm / cm.sum(axis=1, keepdims=True)
+        elif normalize == "pred":
+            cm = cm / cm.sum(axis=0, keepdims=True)
+        elif normalize == "all":
+            cm = cm / cm.sum()
```
glemaitre: Could you create a small `staticmethod` that normalize the dataframe?

Like this we can share it between the `.plot` and `.frame`

glemaitre:
Now, I also realised that `report.metrics.confusion_matrix()` signature should be modify. We should remove `normalize` and `display_labels`. The former is only pass at `.plot` or `.frame` and the latter is not needed. Inside the function we can make that:

```
# we need to convert the classes to a list to make it hashable an used by the
# cache.
display_kwargs = {"display_labels": self._parent.estimator_.classes_.tolist()}
```

`classes_` is part of the scikit-learn API

glemaitre:
As discussed, we can simplify a lot the plotting by using `seaborn.heatmap` instead of managing the matplotlib element ourself.

So let's refactor to have this solution. However it would imply that we need to revisit the format of the dataframe to have the following columns: "actual", "predicted", "count", "normalized_by_all", "normalized_by_predicted", "normalized_by_actual".

Then, the `.plot` and `.frame` will need to filter the right columns. For the plot when then need to make `.pivot` as shown in the seaborn documentation.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: reverted to `confusion_matrix`

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: done

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: done

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: No longer needed since we store the normalized values now, and thus only compute them once.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -16,17 +17,17 @@ class ConfusionMatrixDisplay(DisplayMixin):
 
     Parameters
     ----------
-    confusion_matrix : ndarray of shape (n_classes, n_classes)
-        Confusion matrix.
+    confusion_matrix : pd.DataFrame
+        Confusion matrix data in long format with columns: "True label",
+        "Predicted label", "count", "normalized_by_true", "normalized_by_pred",
+        and "normalized_by_all". Each row represents one cell of the confusion matrix.
```
glemaitre, line 23: I would suggest to always use lower case and without space "true_label" and "predicted_label" when storing in the DataFrame.

At the moment of the plot, we should make the name pretty by replacing `_` by a space and capitalizing it. I don't think that we have a utility for the moment but something that I thought about in https://github.com/probabl-ai/skore/pull/1839 (still lagging behind work). For this PR, just do it locally, do not bother about the prettify function.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -249,42 +184,50 @@ def _compute_data_for_display(
         cm = sklearn_confusion_matrix(
             y_true=y_true_values,
             y_pred=y_pred_values,
-            normalize=normalize,
```
glemaitre, line 252: small details but could you write explicitly `normalize=None` to show that we force to not have any normalization at this stage.

skore/tests/unit/displays/confusion_matrix/test_estimator.py
```diff
@@ -77,40 +75,65 @@ def test_normalization(pyplot, forest_binary_classification_with_train_test):
...
-    custom_labels = ["Negative", "Positive"]
-    display = report.metrics.confusion_matrix(display_labels=custom_labels)
 
-    assert display.display_labels == custom_labels
-    frame = display.frame()
-    assert list(frame.index) == custom_labels
-    assert list(frame.columns) == custom_labels
+    display_true = report.metrics.confusion_matrix()
+    display_true.plot(normalize="true")
+    assert np.allclose(
```
glemaitre: Could you use `np.testing.assert_allclose` instead.

glemaitre: You can even have `np.testing.assert_allclose(arr, 1.0)` that would broadcast automatically.

skore/tests/unit/displays/confusion_matrix/test_estimator.py
```diff
@@ -77,40 +75,65 @@ def test_normalization(pyplot, forest_binary_classification_with_train_test):
...
+        display_pred.confusion_matrix.pivot(
+            index="True label", columns="Predicted label", values="normalized_by_pred"
+        ).sum(axis=0),
+        np.ones(2),
+    )
+    assert np.allclose(display_pred.frame(normalize="pred").sum(axis=0), np.ones(2))
+
+    display_all = report.metrics.confusion_matrix()
+    display_all.plot(normalize="all")
+    assert np.isclose(
```
glemaitre: If it is a single value, I prefer:

```python
assert whatever_func() == pytest.approx(1.0)
```

It shows that we have a scalar.

glemaitre (commented):
It really looks good. I only have tiny changes to make it good to be merged.

PR #2169, opened 2025-11-20
refactor: Let `_get_cached_response_values` save values to the cache

### LINKED ISSUE #2153: Make cache thread-safe

auguste-probabl:
This contributes to #2153 

When used in the context of `joblib.Parallel` with `n_jobs != 1` we still have to keep the saving logic in the caller because `joblib` parallelism is process-wise, not thread-wise.

### DIFF
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -335,11 +335,6 @@ def _get_cached_response_values(
 ) -> list[tuple[tuple[Any, ...], Any, bool]]:
     """Compute or load from local cache the response values.
 
-    Be aware that the predictions will be loaded from the cache if present, but they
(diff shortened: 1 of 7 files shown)


thomass-dev:
Should the branch be deleted @auguste-probabl ?

PR #2173, opened 2025-11-25
ci: Set dependabot cooldown to 14 days

thomass-dev:
> Dependency cooldowns are a free, easy, and incredibly effective way to mitigate the large majority of open source supply chain attacks. More individual projects should apply cooldowns (via tools like Dependabot and Renovate) to their dependencies, and packaging ecosystems should invest in first-class support for cooldowns directly in their package managers.

https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,17 +1,20 @@
 version: 2
 updates:
-
   - package-ecosystem: "github-actions"
     directory: "/"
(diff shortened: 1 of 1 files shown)


auguste-probabl (approved):
Thanks!

Issue #2161, opened 2025-11-15
api(skore): Extend the API of the `.frame` of display


glemaitre:
### What would you like to say?

Up to today, we converged towards storing a main dataframe in each display that is used by both `.plot()` and `.frame()`. To accommodate all type of reports, this dataframe required to be a long-format dataframe. This format is pretty handy for plotting because we can directly use the seaborn API in this case.

This long-format is not really readable when it comes to get it from `.frame` with an `EstimatorReport` and `CrossValidationReport`. One could get a much simpler dataframe using a wide format. It is a bit more complex when it comes to the `ComparisonReport`.

However, by extending the `.frame()` API, we could come with a compromise where a user could get the long or wide format, but we choose a sensible default. However, there is a need of an API discussion. A brain dump of possibilities:

- add a `format="long"/"wide"/"auto"` parameter where the "wide" option is disallowed with `ComparisonReport`
- more granular option where we would add parameters such as `by` or parameter used by `pivot_table` to return an hand crafted dataframe.

auguste-probabl:
My 2 cents: Unlike in the `.plot()` case where one cannot simply take a Matplotlib figure and change its style to fit their needs, in the `.frame()` case the user can use classic pandas operations to get the format they want. For this reason, I don't think this is high-priority.

glemaitre:
Just to provide a compelling example to try to convince a bit @auguste-probabl 

A confusion matrix to be used in seaborn would require three columns `true`, `predicted`, `value` with `n_classes x n_classes` rows. I think that a user would expect only `n_classes x n_classes` dataframe out of this operation.

So I still think that depending on the case, it might be kind of useful if we don't want to upset users :). But it can be done step by step and maybe not right away.

Premkumar-2004:
Hi. I am a new contributor here and would like to take up this issue. Thanks :).

Premkumar-2004:
my  Approach is to Add a format parameter ("long", "wide", or "auto", default "auto") to each display's .frame() method,Create a helper that converts long to wide using pivot_table(). Default to "wide" for EstimatorReport/CrossValidationReport and "long" for ComparisonReport. In .frame(), if format == "auto", use the default; if format == "wide", convert (raise an error for ComparisonReport); otherwise return long. This keeps backward compatibility and lets users choose the format.

GaetandeCast:
[...]

PR #2177, opened 2025-11-27
feat: Add a `threshold` parameter to the confusion_matrix display

### LINKED ISSUE #2112: enh(skore): Add a `decision_threshold` parameter to the `confusion_matrix` display

This issue is related to #2111. The entries of the confusion matrix depends on the decision threshold used to threshold the output of the `model.predict_proba`.

Showing a `confusion_matrix.frame(decision_threshold=0.1)` or `confusion_matrix.plot(decision_threshold=0.1)` would be really handy.

In terms of design, we have 2 choices:

- pre-computed a bunch of confusion matrix for different threshold ahead of time and when calling `.plot` or `.frame`, we pick-up the matrix with the decision threshold the closest to the one requested
- or only store the predictions and when calling `.plot` and `.frame`, we should threshold and then compute the matrix.

Given the current limitation on the hub side regarding the compute, the pre-computing could be the best solution at the short term.

GaetandeCast:
Closes #2112.

With this PR, for binary classification, calling `report.metrics.confusion_matrix(threshold=True)` will compute and store the confusion matrices for all thresholds of the decision function of the classifier. They can then be plotted with `.plot(threshold_value=x)` and accessed via `.frame()`.

This makes use of the new scikit-learn function `confusion_matrix_at_thresholds`, available in 1.8 and back-ported for earlier versions.

The storage structure extends what we converged towards in #2165 : a long format dataframe with one cell of one matrix per row. Columns are the raw count number, all three possible normalized values, the threshold value, and the true and predicted labels.

The default threshold value is `0.5`. We could add an `"auto"` option to select the "best" threshold if we find a satisfactory universal metric to define what "best" means (balanced accuracy may not be desirable for instance, as argued [here](https://github.com/probabl-ai/skore/issues/2112#issuecomment-3542916321)).

### DIFF
--- a/skore/tests/unit/displays/confusion_matrix/test_estimator.py
+++ b/skore/tests/unit/displays/confusion_matrix/test_estimator.py
@@ -2,6 +2,8 @@ import matplotlib as mpl
 import numpy as np
 import pandas as pd
 import pytest
+from sklearn.linear_model import LogisticRegression
+from sklearn.svm import SVC
 
 from skore import EstimatorReport
 from skore._sklearn._plot import ConfusionMatrixDisplay
@@ -377,29 +379,6 @@ def test_plot_with_threshold(pyplot, forest_binary_classification_with_train_tes
     assert "threshold" in display.ax_.get_title().lower()
 
 
-def test_plot_with_default_threshold(
-    pyplot, forest_binary_classification_with_train_test
-):
-    """Check that the default threshold (0.5) is used when not specified."""
-    estimator, X_train, X_test, y_train, y_test = (
-        forest_binary_classification_with_train_test
-    )
-    report = EstimatorReport(
-        estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
-    )
-    display = report.metrics.confusion_matrix()
-    display.plot()
-
-    expected_title = "Confusion Matrix\nDecision threshold: 0.50"
-    assert display.ax_.get_title() == expected_title
-
-    default_vals = display.ax_.collections[0].get_array().flatten()
-    display.plot(threshold_value=0.5)
-    np.testing.assert_allclose(
-        default_vals, display.ax_.collections[0].get_array().flatten()
-    )
-
-
 def test_frame_with_threshold(forest_binary_classification_with_train_test):
     """Check that we can get a frame at a specific threshold."""
     estimator, X_train, X_test, y_train, y_test = (
@@ -416,13 +395,15 @@ def test_frame_with_threshold(forest_binary_classification_with_train_test):
     assert frame.shape == (n_classes * n_classes, 4)
 
 
-def test_frame_default_threshold_predict_proba(
-    forest_binary_classification_with_train_test,
+@pytest.mark.parametrize(
+    "estimator, expected_default_threshold",
+    [(SVC(probability=False), 0), (LogisticRegression(), 0.5)],
+)
+def test_frame_default_threshold(
+    binary_classification_train_test_split, estimator, expected_default_threshold
 ):
-    """Check that frame() uses default threshold (0.5) for predict_proba response."""
-    estimator, X_train, X_test, y_train, y_test = (
-        forest_binary_classification_with_train_test
-    )
+    """Check that frame() uses default threshold."""
+    X_train, X_test, y_train, y_test = binary_classification_train_test_split
     report = EstimatorReport(
         estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
     )
@@ -433,32 +414,9 @@ def test_frame_default_threshold_predict_proba(
     n_classes = len(display.display_labels)
     assert frame.shape == (n_classes * n_classes, 4)
     assert frame["threshold"].nunique() == 1
-    closest_threshold = display.thresholds[np.argmin(abs(display.thresholds - 0.5))]
-    assert frame["threshold"].unique() == closest_threshold
-
-
-def test_frame_default_threshold_decision_function(
-    svc_binary_classification_with_train_test,
-):
-    """Check that frame() uses default threshold (0) for decision_function response."""
-    estimator, X_train, X_test, y_train, y_test = (
-        svc_binary_classification_with_train_test
-    )
-    report = EstimatorReport(
-        estimator.set_params(probability=False),
-        X_train=X_train,
-        y_train=y_train,
-        X_test=X_test,
-        y_test=y_test,
-    )
-    display = report.metrics.confusion_matrix()
-
-    frame = display.frame(threshold_value=None)
-    assert isinstance(frame, pd.DataFrame)
-    n_classes = len(display.display_labels)
-    assert frame.shape == (n_classes * n_classes, 4)
-    assert frame["threshold"].nunique() == 1
-    closest_threshold = display.thresholds[np.argmin(abs(display.thresholds - 0.0))]
+    closest_threshold = display.thresholds[
+        np.argmin(abs(display.thresholds - expected_default_threshold))
+    ]
     assert frame["threshold"].unique() == closest_threshold
 
 
@@ -584,3 +542,17 @@ def test_threshold_values_are_unique(forest_binary_classification_with_train_tes
     display = report.metrics.confusion_matrix()
 
     assert len(display.thresholds) == len(np.unique(display.thresholds))
+
+
+def test_threshold_greater_than_max(forest_binary_classification_with_train_test):
+    """Check that a threshold greater than the maximum threshold is set to the maximum
+    threshold."""
+    estimator, X_train, X_test, y_train, y_test = (
+        forest_binary_classification_with_train_test
+    )
+    report = EstimatorReport(
+        estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
+    )
+    display = report.metrics.confusion_matrix()
+    frame = display.frame(threshold_value=1.1)
+    assert frame["threshold"].unique() == display.thresholds[-1]


skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -2082,6 +2084,19 @@ def confusion_matrix(
             New target on which to compute the metric. By default, we use the target
             provided when creating the report.
 
+        threshold : bool, default=False
```
glemaitre: Do we need this parameter?

If it is not so expensive to be called (I assume not more expensive than a ROC or PR curve) then I would always compute the thresholds. Instead, I would have `threshold_value` in both `.plot()` and `.frame()` with a default value to filter and get a single confusion matrix.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -2097,16 +2112,40 @@ def confusion_matrix(
...
+            "threshold": threshold,
+        }
+
+        response_method: str | list[str] | tuple[str, ...]
+        if threshold:
+            if self._parent._ml_task == "binary-classification":
+                response_method = ("predict_proba", "decision_function")
+            else:
+                raise ValueError(
+                    "Threshold support can only be set for binary classification."
```
glemaitre: It means that we are going to raise the error a bit later when calling the `.frame` or `.plot`

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -184,38 +255,96 @@ def _compute_data_for_display(
         """
         y_true_values = y_true[0].y
         y_pred_values = y_pred[0].y
+        cms = []
+        if isinstance(pos_label, str):
+            neg_label = next(label for label in display_labels if label != pos_label)
+            display_labels = [neg_label, pos_label]
+
+        if threshold:
```
glemaitre: So here, we need to pass the `ml-task` of the report (we can store it in the constructor as well), and if binary classification we alway compute `confusion_matrix_at_thresholds`.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -184,38 +255,96 @@ def _compute_data_for_display(
...
+            tns, fps, fns, tps, thresholds = confusion_matrix_at_thresholds(
+                y_true=y_true_values,
+                y_score=y_pred_values,
+                pos_label=pos_label,
+            )
+            for tn, fp, fn, tp in zip(tns, fps, fns, tps, strict=True):
+                cms.append(np.array([[tn, fp], [fn, tp]]).astype(int))
```
glemaitre: I think it will be more efficient to concatenate all columns and then just reshape (`reshape` is only modifying metadata) instead of looping other the values:


```suggestion
            cms = np.column_stack([tns, fps, fns, tps]).reshape(-1, 2, 2).astype(int)
```

Looping of `cms` will loop other the `n_thresholds` confusion matrix.

Then, I think that we don't need `cms` as well. It should only be a numpy array.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -184,38 +255,96 @@ def _compute_data_for_display(
...
+        else:
+            cms.append(
+                sklearn_confusion_matrix(
+                    y_true=y_true_values,
+                    y_pred=y_pred_values,
+                    normalize=None,  # we will normalize later
+                    labels=display_labels,
+                )
+            )
```
glemaitre:
```suggestion
            cms = sklearn_confusion_matrix(
                y_true=y_true_values,
                y_pred=y_pred_values,
                normalize=None,  # we will normalize later
                labels=display_labels,
            )[np.newaxis, ...]
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -184,38 +255,96 @@ def _compute_data_for_display(
...
+            cm_pred = np.divide(
+                cm,
+                cm.sum(axis=0, keepdims=True),
+                where=cm.sum(axis=0, keepdims=True) != 0,
+            )
+            cm_all = np.divide(
+                cm,
+                cm.sum(),
+                where=cm.sum() != 0,
+            )
```
glemaitre: Since that we have numpy array with shape `(n_thresholds,)`, then we should be able to benefit from numpy vectorization without looping other the confusion matrix:

It should look like:

```python
        row_sums = cms.sum(axis=2, keepdims=True)
        cm_true = np.divide(
            cms_array,
            row_sums,
            where=row_sums != 0,
        )

        col_sums = cms_array.sum(axis=1, keepdims=True)
        cm_pred = np.divide(
            cms_array,
            col_sums,
            where=col_sums != 0,
        )

        total_sums = cms_array.sum(axis=(1, 2), keepdims=True)
        cm_all = np.divide(
            cms_array,
            total_sums,
            where=total_sums != 0,
        )
```

It makes me think that since we compute the normalization ourself now, we should have the smallest example on which we could check that the normalized computation are working properly (or that it is aligned with the scikit-learn implementation).,

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -184,38 +255,96 @@ def _compute_data_for_display(
...
+                confusion_matrix_records.append(
+                    {
+                        "true_label": true_label,
+                        "predicted_label": pred_label,
+                        "count": count,
+                        "normalized_by_true": normalized_true,
+                        "normalized_by_pred": normalized_pred,
+                        "normalized_by_all": normalized_all,
+                        "threshold": threshold_value,
+                    }
```
glemaitre: Then, I think that we should be also able to loop other and create a list of records instead of using directly a dict of numpy arrays.

It should look like:

```python
        n_thresholds = len(thresholds)
        n_classes = len(display_labels)
        n_cells = n_classes * n_classes

        true_labels = np.tile(np.repeat(display_labels, n_classes), n_thresholds)
        pred_labels = np.tile(np.tile(display_labels, n_classes), n_thresholds)
        threshold_values = np.repeat(thresholds, n_cells)

        counts = cms.reshape(-1)
        normalized_true_values = cm_true.reshape(-1)
        normalized_pred_values = cm_pred.reshape(-1)
        normalized_all_values = cm_all.reshape(-1)

        confusion_matrix_dict = {
            "true_label": true_labels,
            "predicted_label": pred_labels,
            "count": counts,
            "normalized_by_true": normalized_true_values,
            "normalized_by_pred": normalized_pred_values,
            "normalized_by_all": normalized_all_values,
            "threshold": threshold_values,
        }
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -184,38 +255,96 @@ def _compute_data_for_display(
         """
         y_true_values = y_true[0].y
         y_pred_values = y_pred[0].y
+        cms = []
+        if isinstance(pos_label, str):
```
glemaitre: Let's also store the `pos_label`. It could be handy to show in the matrix as an extra information.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -29,8 +31,15 @@ class ConfusionMatrixDisplay(DisplayMixin):
             "cross-validation", "estimator"}
         The type of report.
 
+    threshold : bool, default=False
```
glemaitre: I would not store this information.

glemaitre: Instead, we will have a column of `np.nan` when no threhsold are available in the `confusion_matrix` dataframe.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -44,10 +53,16 @@ def __init__(
...
+        threshold: bool = False,
     ):
         self.confusion_matrix = confusion_matrix
         self.display_labels = display_labels
         self.report_type = report_type
+        self.threshold = threshold
+        if threshold:
+            self.thresholds_ = self.confusion_matrix["threshold"].unique()[::-1]
+        else:
+            self.thresholds_ = None
```
glemaitre: If we need this information then we need to compute it in the `_compute_data_for_display` and just pass it as an attribute.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -118,29 +143,65 @@ def _plot_single_estimator(
...
+            if not self.threshold:
+                raise ValueError(
+                    "Enable threshold support by passing `threshold=True` to "
+                    "`report.metrics.confusion_matrix()` before calling `plot()` with "
+                    "`threshold_value`. This is only applicable for binary "
+                    "classification."
+                )
+            # Find the existing threshold that is closest to the given threshold_value
+            threshold_value = self.thresholds_[
+                np.argmin(np.abs(self.thresholds_ - threshold_value))
```
glemaitre: I think that the most effective algorithm here is the `numpy.searchsorted`: https://numpy.org/doc/stable/reference/generated/numpy.searchsorted.html

Then using it, you don't really need a comment.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -118,29 +143,65 @@ def _plot_single_estimator(
...
+            The decision threshold to use when applicable.
+            If None and thresholds are available, plots the confusion matrix at the
+            default threshold (0.5).
+
         heatmap_kwargs : dict, default=None
             Additional keyword arguments to be passed to seaborn's `sns.heatmap`.
-
         """
+        if threshold_value is not None:
+            if not self.threshold:
```
glemaitre: You need to adapt a bi this code because we don't need `threshold` anymore but have

```python
if threshold_value is not None and ml_task != "binary-classification":
```

Here it directly read as what is the limitation actually.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -118,29 +143,65 @@ def _plot_single_estimator(
...
+                    "`threshold_value`. This is only applicable for binary "
+                    "classification."
+                )
+            # Find the existing threshold that is closest to the given threshold_value
+            threshold_value = self.thresholds_[
+                np.argmin(np.abs(self.thresholds_ - threshold_value))
+            ]
+            cm = self.confusion_matrix[
+                self.confusion_matrix["threshold"] == threshold_value
+            ]
```
glemaitre: just a taste of style but I love `df.query` because it avoid so many masking:

```python
cm = self.confusion_matrix.query("threshold == @threshold_value")
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -118,29 +143,65 @@ def _plot_single_estimator(
...
+        else:
+            if self.threshold:
+                # Use the default threshold (0.5) if no threshold_value is provided
+                threshold_value = self.thresholds_[
+                    np.argmin(np.abs(self.thresholds_ - 0.5))
+                ]
+                cm = self.confusion_matrix[
+                    self.confusion_matrix["threshold"] == threshold_value
```
glemaitre: same remarks than above.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
glemaitre: Here, I would argue that we should be calling `self.frame(normalize=normalize, threshold_value=threshold_value)` to get easier confusion matrix to be displayed.

It should avoid to make twice the same query.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -225,14 +354,54 @@ def frame(self, normalize: Literal["true", "pred", "all"] | None = None):
...
+            cm = self.confusion_matrix.copy()
+            if threshold_value is not None:
+                # Find the existing threshold closest to given threshold_value
+                threshold_value = self.thresholds_[
+                    np.argmin(np.abs(self.thresholds_ - threshold_value))
+                ]
+                cm = cm[cm["threshold"] == threshold_value]
+            return cm.pivot(
+                index="true_label", columns="predicted_label", values=normalize_by
+            ).reindex(index=self.display_labels, columns=self.display_labels)
```
glemaitre: For the moment, let's only focus querying the confusion matrix and return the long format matrix.

Let's then think about an API design for https://github.com/probabl-ai/skore/issues/2161 that would help us to return a more "human"-readable matrix.

I think that for the confusion matrix, it would be best to implement the `CrossValidationReport` and `ComparisonReport` support and then make a decision of what to do because we will have at hand all the cases.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: >It makes me think that since we compute the normalization ourself now, we should have the smallest example on which we could check that the normalized computation are working properly (or that it is aligned with the scikit-learn implementation).

I'm not sure what you mean by that.

GaetandeCast:
Hi @glemaitre, thanks for the review. I have have implemented the requested changes. I will wait until we are satisfied with the API to update the tests and the doc example. So until then, the CI will be red.

GaetandeCast:
@glemaitre I did changes suggested orally:
- add an asterisk on the positive label and a legend explaining it
- the decision threshold is now on a new line in the title
- mention "Decision threshold: 0.5" in the title when no threshold was explicitly chosen 
- describe the usage and utility of the `threshold` in the docstring of `plot()` and `frame()`.

Here is how the `display.plot()` looks like now:
image

Generated with:
```
cm_display = report.metrics.confusion_matrix(pos_label="disallowed")
cm_display.plot(threshold_value=0.3)
plt.show()
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -44,10 +58,16 @@ def __init__(
...
     ):
         self.confusion_matrix = confusion_matrix
         self.display_labels = display_labels
         self.report_type = report_type
+        self.thresholds_ = thresholds
+        self.ml_task = ml_task
+        self.pos_label = pos_label
 
     _default_heatmap_kwargs: dict = {
         "cmap": "Blues",
```
GaetandeCast, line 82: We could change `"cbar"` to `False` in `_default_heatmap_kwargs` as I'm not sure it adds any real value in a confusion matrix.

glemaitre: yep we can do that.

glemaitre:
We have a test failing.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -5,11 +5,13 @@
 import numpy as np
 import pandas as pd
 import seaborn as sns
+from numpy._typing import NDArray
```
glemaitre: I think that you can import from a public module instead: https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -71,6 +97,11 @@ def plot(
             conditions or all the population. If None, the confusion matrix will not be
             normalized.
 
+        threshold_value : float or None, default=None
+            The decision threshold to use when applicable.
+            If None and thresholds are available, plots the confusion matrix at the
+            default threshold (0.5).
```
glemaitre: Actually it is 0.5 if the estimator expose a `predict_proba` but it is 0 if `decision_function` is used.

Indeed, it could be worth to use `SVC` classifier that only expose `decision_function` in the test to ensure the behaviour.

glemaitre:
Let me have a pass on the test now.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -29,8 +34,23 @@ class ConfusionMatrixDisplay(DisplayMixin):
...
+        The class considered as the positive class when displaying the confusion
+        matrix.
+
+    response_method : str
+        The estimator's method that was used to get the predictions. The possible
+        values are: "predict", "predict_proba", and "decision_function".
+
     Attributes
     ----------
+    thresholds_ : array-like of shape (n_thresholds,)
```
glemaitre: It is actually an `ndarray`. Array-like is less constrained (it could be a list or tuple for instance).

```suggestion
    thresholds_ : ndarray of shape (n_thresholds,)
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -185,54 +290,121 @@ def _compute_data_for_display(
...
         disp = cls(
             confusion_matrix=confusion_matrix,
             display_labels=display_labels,
             report_type=report_type,
+            ml_task=ml_task,
+            pos_label=pos_label,
+            response_method=_check_response_method(
+                estimators[0], response_method
+            ).__name__,
+            thresholds=np.unique(thresholds),
```
glemaitre, line 358: What is the reason for calling `np.unique`? Can we duplicate?

Also, `np.unique` sorts the values. Since we do a `searchsorted` it probably makes sense but then we should document in the constructor that it is in an increasing order.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -44,25 +64,39 @@ def __init__(
...
         report_type: ReportType,
+        ml_task: MLTask,
+        thresholds: NDArray,
+        pos_label: PositiveLabel,
+        response_method: str,
     ):
         self.confusion_matrix = confusion_matrix
         self.display_labels = display_labels
         self.report_type = report_type
+        self.thresholds_ = thresholds
```
glemaitre: This one looks weird. I would expect:


```suggestion
        self.thresholds = thresholds
```

If in `plot` or `frame`, we are changing the thresholds, then it makes sense to create also an attribute `thresholds_` then.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -44,25 +64,39 @@ def __init__(
...
+        threshold_value: float | None = None,
         heatmap_kwargs: dict | None = None,
     ):
-        """Plot visualization.
+        """Plot the confusion matrix.
+
+        In binary classification, the confusion matrix can be displayed at various
+        decision thresholds. This is useful for understanding how the model's
+        predictions change as the decision threshold varies. If no threshold is
+        provided, the confusion matrix is displayed at the default threshold (0.5).
```
glemaitre: The "default" threshold need to be update as well for the decision function case.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -185,54 +290,121 @@ def _compute_data_for_display(
...
+    def frame(
+        self,
+        threshold_value: float | None = None,
+    ):
+        """Return the confusion matrix as a long format dataframe.
+
+        In binary classification, the confusion matrix can be returned at various
+        decision thresholds. This is useful for understanding how the model's
+        predictions change as the decision threshold varies. If no threshold is
+        provided, the default threshold (0.5) is used.
```
glemaitre: need an update here as well.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -185,54 +290,121 @@ def _compute_data_for_display(
...
+
+        The matrix is returned as a long format dataframe where each line represents one
+        cell of the matrix. The columns are "true_label", "predicted_label", "count",
+        "normalized_by_true", "normalized_by_pred", "normalized_by_all" and "threshold".
```
glemaitre: Oh, let's align the API of plot and frame: add a `normalize` parameter. If people wants to access everything they can always request the attribute `.confusion_matrix`.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -118,30 +162,72 @@ def _plot_single_estimator(
...
         )
-
         normalize_by = "normalized_by_" + normalize if normalize else "count"
+
         sns.heatmap(
-            self.confusion_matrix.pivot(
-                index="true_label", columns="predicted_label", values=normalize_by
-            ),
+            self.frame(threshold_value=threshold_value)
+            .pivot(index="true_label", columns="predicted_label", values=normalize_by)
```
glemaitre: Allowing `self.frame(normalize=normalize)` could simplify a bit here.

skore/tests/unit/displays/confusion_matrix/test_estimator.py
```diff
@@ -285,30 +275,315 @@ def test_frame_structure(forest_binary_classification_with_train_test):
...
+    report = EstimatorReport(
+        estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
+    )
+    display = report.metrics.confusion_matrix()
+
+    display.plot(threshold_value=0.3)
+    assert "threshold" in display.ax_.get_title().lower()
+
+
+def test_plot_with_default_threshold(
```
glemaitre: Is this test redundant with the one below checking for `predict_proba`?

skore/tests/unit/displays/confusion_matrix/test_estimator.py
```diff
@@ -285,30 +275,315 @@ def test_frame_structure(forest_binary_classification_with_train_test):
...
+    )
+    display = report.metrics.confusion_matrix()
+
+    frame = display.frame(threshold_value=None)
+    assert isinstance(frame, pd.DataFrame)
+    n_classes = len(display.display_labels)
+    assert frame.shape == (n_classes * n_classes, 7)
+    assert frame["threshold"].nunique() == 1
+    closest_threshold = display.thresholds_[np.argmin(abs(display.thresholds_ - 0.0))]
+    assert frame["threshold"].unique() == closest_threshold
```
glemaitre: Here, I would parametrize the test with the estimator (only use the data from the fixture declared in `conftest.py`:


```python
@pytest.mark.parametrize(
    "estimator, expected_default_threshold",
    [(SVC(), 0), (LogisticRegression(), 0.5)]
)
def test_...(estimator, expected_default_threshold):
    ...
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -185,54 +290,121 @@ def _compute_data_for_display(
...
+            )
+        if threshold_value is None:
+            if self.ml_task == "binary-classification":
+                threshold_value = 0.5 if self.response_method == "predict_proba" else 0
+            else:
+                return self.confusion_matrix
+
+        index_right = np.searchsorted(self.thresholds_, threshold_value)
+        if index_right == len(self.thresholds_):
+            index_right = index_right - 1
```
glemaitre, line 416: This line is not covered. It might be worth to create such case.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: There can technically be duplicate threshold values e.g. for a decision tree with no split or due to finite float precision. I have documented the increasing order in the constructor's docstring.

skore/tests/unit/displays/confusion_matrix/test_estimator.py
GaetandeCast: yeah it is since `plot()` calls `frame()`, I've removed it.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
glemaitre: 👍

glemaitre:
LGTM. Merging and bypassing the failure since it is not related.

PR #2181, opened 2025-11-28
feat(skore): Allow to override or update style using

glemaitre:
Small improvement factor out from #2152

This enhancement either allow to update or override the style parameter of a `Display`.

### DIFF
--- a/skore/src/skore/_sklearn/_plot/base.py
+++ b/skore/src/skore/_sklearn/_plot/base.py
@@ -2,7 +2,7 @@ import inspect
 from collections.abc import Callable
 from functools import wraps
 from io import StringIO
-from typing import Any, Protocol, runtime_checkable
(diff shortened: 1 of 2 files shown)


skore/src/skore/_sklearn/_plot/base.py
```diff
@@ -118,11 +118,19 @@ def _style_params(self) -> list[str]:
...
+    ):
         """Set the style parameters for the display.
 
         Parameters
         ----------
+        policy : Literal["override", "update"], default="override"
+            Policy to use when setting the style parameters.
+            If "override", existing settings are set to the provided values.
+            If "update", existing settings are not changed; only settings that were
+            previously unset are changed.
```
thomass-dev, line 132: Nitpick:

> only settings that were previously unset are changed.

This is confusing/not true, as you will update an existing setting (`current_value`) with additional values (`param_value`).

PR #2189, opened 2025-12-02
chore(skore)!: Rename `scoring` to `metric`

### LINKED ISSUE #2106: api(skore): Rename the `scoring` and `scoring_kwargs` to `metrics` and `metrics_kwargs` in `skore`

In several places in `skore`, we chose to stick to the scikit-learn API naming by choosing `scoring` and `scoring_kwargs`.

However, scores carry the semantic "higher is better" and is particularly useful when trying to optimize them, e.g. `GridSearchCV` always maximizes scores.

`skore` however is never doing any optimization and instead only report "numbers". Therefore, it is completely OK to report scores or losses and to not stick to the same semantic.

In this case, renaming those occurrences to `metrics` would be better because they can represent a value that is both "higher is better" or "lower is better". Accordingly, the `scoring_kwargs` could be rename to `metric_kwargs`.

Currently `scoring` and `scoring_kwargs` is exposed into the `report.metrics.summary` and the `report.feature_importance.permutation` methods.

MarieSacksick:
Closes #2106  

Breaking changes since changes in the API.

Plenty of changes to make the naming consistant everywhere, from the API to the variable names and test function names.

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -211,11 +211,11 @@ class _MetricsAccessor(
         metric_names = None
         if isinstance(metric, dict):
             metric_names = list(metric.keys())
-            metric_list = list(metric.values())
+            metrics = list(metric.values())
         elif metric is not None and not isinstance(metric, list):
-            metric_list = [metric]
+            metrics = [metric]
         elif isinstance(metric, list):
-            metric_list = metric
+            metrics = metric
 
         if data_source == "X_y":
             # optimization of the hash computation to avoid recomputing it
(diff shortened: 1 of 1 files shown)


skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -8,7 +8,7 @@
 import numpy as np
 import pandas as pd
 from numpy.typing import ArrayLike, NDArray
-from sklearn import metrics
+from sklearn import metrics as sklearn_metrics
```
MarieSacksick, line 11: not mandatory, but I think it's more explicit

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
auguste-probabl: Rename this to `MetricName` and `Scoring` to `Metric`?

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -226,63 +228,66 @@ class is set to the one provided when creating the report. If `None`,
...
+        if metric_names is None:
+            metric_names = [None] * len(metric_list)  # type: ignore
 
         scores = []
         favorability_indicator = []
-        for metric_name, metric in zip(scoring_names, scoring, strict=False):
-            if isinstance(metric, str) and not (
-                (metric.startswith("_") and metric[1:] in self._score_or_loss_info)
-                or metric in self._score_or_loss_info
+        for metric_name, metric_item in zip(metric_names, metric_list, strict=False):
```
auguste-probabl: Not sure that the rename to `metric_item` is clearer

MarieSacksick: If I don't rename, then I'm overwriting the parameter metric, which doesn't look used below in the function, but it could be misleading if we want to add stuff later. Wdyt?

auguste-probabl: Ah true. Then maybe just `metric_`?

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -226,63 +228,66 @@ class is set to the one provided when creating the report. If `None`,
         else:
             data_source_hash = None
 
-        if scoring is None:
+        if metric is None:
             # Equivalent to _get_scorers_to_add
             if self._parent._ml_task == "binary-classification":
-                scoring = ["_precision", "_recall", "_roc_auc"]
+                metric_list = ["_precision", "_recall", "_roc_auc"]
```
auguste-probabl: `metrics`

PR #2194, opened 2025-12-05
feat(skore): Give Project a `ml_task` attribute

### LINKED ISSUE #2179: feat(skore): limit a project to a single `ML task`
Closes https://github.com/probabl-ai/skore-hub/issues/852.

As a user: 
- in an **empty** project, i can add a report with any ML task (`report.ml_task`)
- in a **non-empty** project, i can only add a report with the same ML task as the previously added report
    - an exception is raised when this condition is not met

---

In technical terms, in  the `skore.Project` class:
- during the project initialization, request the project summary to retrieve the list of `ML task`
    - raise an exception if the list is not uniform
- every time i `put` a report, i compare its `ML task` to the one retrieved previously
    - raise an exception if different

auguste-probabl:
The attribute is set as soon as a first report is `put` in the Project. Once set, if a new report is `put` with a different ML task, the operation fails.

Closes #2179

### DIFF
--- a/skore/src/skore/project/project.py
+++ b/skore/src/skore/project/project.py
@@ -197,6 +197,16 @@ class Project:
         self.__name = name
         self.__project = plugin(**(kwargs | parameters))
 
+        ml_tasks = {report["ml_task"] for report in self.__project.summarize()}
+
+        if len(ml_tasks) > 1:
+            raise RuntimeError(
+                "Expected every report in the Project to have the same ML task. "
+                f"Got ML tasks {ml_tasks}."
+            )
+
+        self.ml_task: MLTask | None = ml_tasks.pop() if ml_tasks else None
+
     @property
     def mode(self):
         """The mode of the project."""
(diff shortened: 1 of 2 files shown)


skore/src/skore/project/project.py
```diff
@@ -7,6 +7,7 @@
 from typing import Any, Literal
```
thomass-dev, line 8: This is good for a freshly created project, but you miss to treat the use-case where the project already exists on the hub/local storage.

> - during the project initialization, request the project summary to retrieve the list of ML task
>     - raise an exception if the list is not uniform

---

Directly during the `Project.__init__`, something like (untested):

```python
[...]

# retrieve previous `ML task`
ml_tasks = self.__project.summarize()["ml_task"].unique()

if len(ml_tasks) > 1:
    raise RuntimeError

self.ml_task = ml_tasks[0] if ml_tasks else None
```

PR #2178, opened 2025-11-27
feat(skore-hub-project): Thread the compute of report metrics

thomass-dev:
Partially addresses https://github.com/probabl-ai/skore/issues/2140, part of the `reduce put time` task.

---

<ins>Thread the compute of the metrics of a report</ins>, and:

- Add a skinned progress-bar tracking the compute of the metrics of the report,
- Merge all metric tests into one parametrized test.
- Disambiguate the `metric.value` property: when `metric.value` was called, the computing was performed implicitly, now, `metric.compute()` must be called before explicitly.

PR #2205, opened 2025-12-09
fix(skore/cache): Explicitly lock `__iter__`

thomass-dev:
Fix `RuntimeError: dictionary changed size during iteration` exception when the cache is read and updated in two concurrent threads.


thomass-dev:
Needed for https://github.com/probabl-ai/skore/pull/2178.

Issue #2203, opened 2025-12-09
Question - API not natural in custom metrics


MarieSacksick:
### What would you like to say?

hello!

I'm using skore for a presentation I'm preparing, and I tried to use custom metric, and found some behavior I find not natural at all.  
To give you some context, here is my code (it's an extract of the notebook, hopefully I didn't forget any required piece and this runs): 

```python
# %%
from skore import train_test_split
from sklearn.datasets import make_regression
X, y = make_regression(random_state=42)
X_train, X_test, y_train, y_test = train_test_split(X, y)

# %%
from skore import EstimatorReport
from sklearn.linear_model import LinearRegression

lr_est_rep = EstimatorReport(
                        LinearRegression(),
                        X_train=X_train,
                        y_train=y_train,
                        X_test=X_test,
                        y_test=y_test
                    )

# %%
def business_loss(y_true, y_pred):
    ### if under the market: 100% loss for me
    if y_true > y_pred:
        return y_true - y_pred
    ### If I'm above the market, I will waste some time to sell.
    ### Each month costs me 2k, and every month, I lower the price by 10k
    else: 
        return 2*(y_pred-y_true)/10

# %%
from sklearn.metrics import make_scorer
business_score = make_scorer(business_loss, greater_is_better=False, response_method="predict")
lr_est_rep.metrics.custom_metric(metric_function=business_score)
```

This last cell results in:  

```
TypeError                                 Traceback (most recent call last)
Cell In[3], [line 3](vscode-notebook-cell:?execution_count=3&line=3)
      1 from sklearn.metrics import make_scorer
      2 business_score = make_scorer(business_loss, greater_is_better=False, response_method="predict")
----> [3](vscode-notebook-cell:?execution_count=3&line=3) lr_est_rep.metrics.custom_metric(metric_function=business_score)

TypeError: _MetricsAccessor.custom_metric() missing 1 required positional argument: 'response_method'
```

This is not a bug, as it's the behavior documented in https://docs.skore.probabl.ai/0.8/reference/api/skore.ComparisonReport.metrics.custom_metric.html.  
But it's unexpected, for two reasons:  
1. the response_method is already given in the business score
2. I shouldn't even need to precise a response_method, since there is a default in business_score.  

Is there a reason why we want to force our users to fill the parameter, and not have a default? As you've understood through this ticket, I don't like it :)

auguste-probabl:
Indeed, `custom_metric` does not know about sklearn Scorers yet. It should!

[...]

PR #2207, opened 2025-12-09
feat: Add default to response method used in custom metric

MarieSacksick:
Partially fixes https://github.com/probabl-ai/skore/issues/2203.  

Does not fix the fact that we cannot pass a scorer done by make_scorer.


MarieSacksick:
Closing it as it's not the solution we want. We don't want to always have a default to avoid computing a metric that should be using the proba, and feed it with the label (most of the time it will not break, but output something false).  
However, we do want to avoid to the user the repetition, when the response method is already defined in the scorer.

Issue #2217, opened 2025-12-12
chore: Pin dependencies for sphinx-build job


auguste-probabl:
### What would you like to say?

The sphinx-build job is currently failing with no apparent change in code; see for example https://github.com/probabl-ai/skore/actions/runs/20168036995/job/57896226747 and https://github.com/probabl-ai/skore/actions/runs/20109676866/job/57703301488.

It would be good to bisect `main` to see if this is indeed due to our code changes, but in the meantime we need to future-proof things by pinning dependencies for this job, ideally using something like `uv` or `pixi`. Besides, doing this means the dependencies are cached which will save CI time.

thomass-dev:
Thanks for the report, it's been a while since we talked about it!
I'm looking how to integrate in the current lockfiles generated using uv, and how to update the GH workflow.

thomass-dev:
Thanks for your report. We've talked about it several times.
I'm looking to introduce the sphinx lockfile inside our current lockfiles generated by uv.

PR #2222, opened 2025-12-16
ci: Add sphinx lockfile

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/2217.

In this first iteration:
- i modified the `ci/pip-compile.sh` script to generate a lock-file including the `sphinx` extra dependencies
- i used the generated file to setup the `sphinx` workflow.

In the next iteration, i will improve the `sphinx` workflow by using cache mechanism.

### DIFF
--- a/.github/actions/sphinx/build/action.yml
+++ b/.github/actions/sphinx/build/action.yml
@@ -11,11 +11,11 @@ inputs:
 runs:
   using: composite
   steps:
-    - working-directory: skore
-      shell: bash
+    - shell: bash
       run: |
         python -m pip install --upgrade pip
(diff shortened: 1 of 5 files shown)


.github/actions/sphinx/build/action.yml
auguste-probabl, line 1: can you remind me why this is in a different file than `sphinx.yml`?

thomass-dev: To be reused in several workflows (reusable action).
As it is now only used in `sphinx.yml`, i will merge it directly into the workflow in the next iteration (that i have already started).

I want to keep this PR as small as possible, to fix the issue.

Issue #2206, opened 2025-12-09
Bug when trying to compute a custom metric on a comparison report of CV


MarieSacksick:
### Describe the bug

When I try to do 
```python
comp_report.metrics.summarize(
    scoring = business_loss, response_method = "predict"
).frame()
```

or without response_method, I have errors :/ 

### Steps/Code to Reproduce

```python
# %%
import sklearn

X, y = sklearn.datasets.fetch_california_housing(return_X_y=True, as_frame=True)

# %%
from skore import train_test_split

X_train, X_test, y_train, y_test = train_test_split(X, y)

# %%
def business_loss(y_true_list, y_pred_list):
    loss = 0
    for y_true_, y_pred_ in zip(y_true_list, y_pred_list):
        ### if under the market: 100% loss for me
        if y_true_ > y_pred_:
            loss = loss + float(y_true_ - y_pred_)
        ### If I'm above the market, I will waste some time to sell.
        ### Each month costs me 2k, and every month, I lower the price by 10k
        else:
            loss = loss + float(2 * (y_pred_ - y_true_) / 10)

    return loss

# %%
from skore import CrossValidationReport
from sklearn.linear_model import LinearRegression
from sklearn.ensemble import HistGradientBoostingRegressor

hist_cv_rep = CrossValidationReport(HistGradientBoostingRegressor(), X=X, y=y)

lr_cv_rep = CrossValidationReport(LinearRegression(), X=X, y=y)

# %%
from skore import ComparisonReport

comp_report = ComparisonReport(
    reports={"histogram": hist_cv_rep, "linear_reg": lr_cv_rep}
)

comp_report.metrics.summarize(scoring=business_loss, response_method="predict").frame()
```

### Expected Behavior

No error is thrown.

### Actual Behavior

**With response_method**

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[5], line 1
----> 1 comp_report.metrics.summarize(
      2     scoring = business_loss, response_method = "predict"
      3 ).frame()

TypeError: _MetricsAccessor.summarize() got an unexpected keyword argument 'response_method'
```

**Without response_method**

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 1
----> 1 comp_report.metrics.summarize(
      2     scoring = business_loss
      3 ).frame()

[...]

PR #2247, opened 2026-01-07
fix: Clarify error when using response_method in ComparisonReport.metrics.summarize

Adrien-1997:
Clarifies `response_method` handling in `ComparisonReport.metrics.summarize`.

Passing `response_method=` previously raised a generic "unexpected keyword argument"
TypeError, which is confusing given that `ComparisonReport.metrics.custom_metric`
and sklearn scorers (`make_scorer`) are the intended APIs for controlling the
prediction method (`predict`, `predict_proba`, etc.).

This PR:
- adds `response_method` to the `summarize` signature to raise a clear, actionable
  error message guiding users to `custom_metric(..., response_method=...)` or
  `make_scorer(..., response_method=...)`.
- adds a regression test covering both `predict` and `predict_proba`.

Refs #2206

### DIFF
--- a/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
@@ -156,9 +156,11 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
         """
         if response_method is not None:
             raise TypeError(
-                "`response_method` is not supported in `ComparisonReport.metrics.summarize`. "
+                "`response_method` is not supported in "
+                "`ComparisonReport.metrics.summarize`. "
                 "To compute a custom metric with an explicit response method, use "
-                "`ComparisonReport.metrics.custom_metric(metric_function=..., response_method=...)`, "
+                "`ComparisonReport.metrics.custom_metric("
+                "metric_function=..., response_method=...)`, "
                 "or pass a scikit-learn scorer created with "
                 "`sklearn.metrics.make_scorer(..., response_method=...)` via `metric=`."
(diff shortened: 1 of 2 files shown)


GaetandeCast:
Hey @Adrien-1997, thanks for the PR! 
You have two elements failing in the CI: the title of the PR is missing an upper case at the beginning and there is a linting issue. If you haven't already, consider using pre-commit hooks to automatically run the linter on commit. You can find [more details here](https://docs.skore.probabl.ai/0.11/contributing.html).

adrien-morel:
Thanks for the feedback!
PR title and linting issues are now fixed, and I’ve set up pre-commit hooks as well.

GaetandeCast:
So I'm not a huge fan of introducing a parameter just to raise an error when it is used, I find it misleading. Maybe use a `NotImplementedError` to indicate that it will be properly implemented later ? although I'm not even sure we are planning to do so.
WDYT @auguste-probabl ? You're more knowledgeable than me on this subject.

adrien-morel:
I was mainly trying to improve the current behavior, which already raises confusing errors.
Given the current separation between built-in metrics, `custom_metric`, and scikit-learn scorers, I agree that adding a parameter only to error out can be misleading.
Happy to adjust or drop this approach if it’s not aligned with the intended direction.

auguste-probabl:
Thanks for the PR!

I think this is going in the right direction, and in fact there isn't much more work needed to actually implement the right behaviour. I'm working on it.

auguste-probabl:
https://github.com/probabl-ai/skore/pull/2276 is now open so I'll go ahead and close this. Thanks again!

PR #2193, opened 2025-12-05
feat(skore): Refactor PR curve display and add `subplot_by` option

GaetandeCast:
Towards #1445 and #2190: adds an option `subplot_by` to the `PrecisionRecallCurveDisplay.plot()` and switches to seaborn API for plotting.

Here are how the plots look with the default value for `subplot_by`, in multi-class classification:
## `EstimatorReport`:
image

## `CrossValidationReport`
image

## `ComparisonReport`
image

## `ComparisonReport` of `CrossValidationReports`

image

PR #2273, opened 2026-01-09
fix(skore): Allow subplot_by=None for PRC in multiclass

GaetandeCast:
Follow up to #2193.

Allow `subplot_by=None` for in multiclass classification for `EstimatorReport` and `CrossValidationReport`, in the `plot` method of the `PrecisionRecallCurveDisplay`. 

The default is changed `None` for `EstimatorReport` and `CrossValidationReport`and `"estimator"` for `ComparisonReport`.

Here is how it looks like:

image
image

cc @glemaitre

### DIFF
--- a/skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
+++ b/skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
@@ -134,8 +134,9 @@ class PrecisionRecallCurveDisplay(_ClassifierCurveDisplayMixin, DisplayMixin):
         ----------
         subplot_by : {"auto", "label", "estimator"} or None, default="auto"
             Column to use for creating subplots. Options:
-            - "auto": None for EstimatorReport and Cross-Validation Report,
-                "label" for ComparisonReport
+
+            - "auto": None for EstimatorReport and Cross-Validation Report, \
+              "label" for ComparisonReport
             - "label": one subplot per class (multiclass only)
             - "estimator": one subplot per estimator (comparison only)
             - None: no subplots (Not available for comparison in multiclass)


skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -439,9 +439,6 @@ def _format_legend_label(
         if style_column_name == "data_source":
             style_value = cast(str, style_value)
             style_value = style_value.title() + " set"
-        if hue_column_name == "data_source":
-            hue_value = cast(str, hue_value)
-            hue_value = hue_value.title() + " set"
```
GaetandeCast, line 442: The `data_source` will never be used as hue on the plot so we can remove this code. (Improves coverage)

skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -137,7 +137,7 @@ def plot(
             - "auto": "label" for multiclass, None for binary
```
glemaitre: I'm thinking that the default should be:

- None for `EstimatorReport` and `CrossValidationReport` in both binary and multiclass.
- "estimator" for `ComparisonReport` also for both binary and multiclass.

WDYT @GaetandeCast? (it is what I have in `coefficients().plot()`)

GaetandeCast: Actually yes that makes more sense. I didn't put that much thought in the default choice initially. Let me change that.

glemaitre:
LTGM. Thanks @GaetandeCast

Issue #1957, opened 2025-07-31
Allow to set the style kwargs only in `set_style` and not in `plot`


glemaitre:
### Discussed in https://github.com/probabl-ai/skore/discussions/1875

<div type='discussions-op-text'>

<sup>Originally posted by **glemaitre** June 30, 2025</sup>
While working on https://github.com/probabl-ai/skore/pull/1805, it looks like there a possibility to hand up with several kwargs style. Let's provide an example:

```python
    def plot(
        self,
        *,
        x: str | None = None,
        y: str | None = None,
        hue: str | None = None,
        kind: Literal["dist", "corr"] = "dist",
        top_k_categories: int = 20,
        scatterplot_kwargs: dict[str, Any] | None = None,
        stripplot_kwargs: dict[str, Any] | None = None,
        boxplot_kwargs: dict[str, Any] | None = None,
        heatmap_kwargs: dict[str, Any] | None = None,
        histplot_kwargs: dict[str, Any] | None = None,
    ) -> None:
```

Switch `kind` or providing `x`, `y`, or `hue` will choose different type of visual representation and we need to dispatch which part to tweak. Currently, we have two ways to tweak:

```python
display = report.data.analyze()
display.plot(x="xxx", histplot_kwargs=dict(color="tab:orange"))
```

or

```python
display = report.data.analyze()
# example when we "persist" the style
display.set_style(histplot_kwargs=dict(color="tab:orange"))
display.plot(x="xxx")
# it can also be used in one liner
display.set_style(histplot_kwargs=dict(color="tab:orange")).plot(x="xxx")
```

The `set_style` was introduced to be able to persist the style across call to `display` at first.

Now, I'm wondering if we should not limit the API and allow only to call `set_style` to set the kwargs. The reasoning behind:

- it is not longer to write
- user benefit from the persisted style by default
- it bring clarity in the `plot` method because we reduce the number of parameters to only the element that will define what is shows on the plot
- we do not really introduce a new API here just give one way to do something

In short the API for the above case would become:

```python
    def plot(
        self,
        *,
        x: str | None = None,
        y: str | None = None,
        hue: str | None = None,
        kind: Literal["dist", "corr"] = "dist",
        top_k_categories: int = 20,
    ) -> None:
```

and 

```python
    def set_style(
        self,
        *,
        scatterplot_kwargs: dict[str, Any] | None = None,
        stripplot_kwargs: dict[str, Any] | None = None,
[...]

PR #2293, opened 2026-01-14
chore(skore): Moving confusion matrix display kwargs into set_style

glemaitre:
Partially address #1957 

This PR add a specialized `set_style` for the `ConfusionMatrixDisplay` such that we move the kwargs specifically in this function instead of putting it inside the `plot` one to simplify the user API.

The option `policy` is also set to `"update"` by default because it has a better behaviour by default by not overriding all the default parameter already set.

### DIFF
--- a/skore/tests/unit/displays/test_style.py
+++ b/skore/tests/unit/displays/test_style.py
@@ -21,7 +21,7 @@ def test_style_mixin():
     "initial_state, override_value, expected_result, use_explicit_policy",
     [
         (None, {"a": 1, "b": 2}, {"a": 1, "b": 2}, False),
-        ({"a": 1, "b": 2}, {"c": 3}, {"c": 3}, False),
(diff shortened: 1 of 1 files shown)


skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -605,3 +574,46 @@ def select_threshold_and_format(group):
...
+        ----------
+        policy : Literal["override", "update"], default="update"
+            Policy to use when setting the style parameters.
+            If "override", existing settings are set to the provided values.
+            If "update", existing settings are not changed; only settings that were
+            previously unset are changed.
+
+        heatmap_kwargs : dict, default=None
+            Additional keyword arguments to be passed to seaborn's
+            :func:`seaborn.heatmap`.
```
GaetandeCast, line 599: Not directly linked to the goal of this PR but this sphinx reference is a bit redundant:

Image

Since there are other occurrences of this, namely a few introduced in #2152, I can fix all of them in a subsequent PR.

PR #2294, opened 2026-01-15
chore(skore): Clean up docs references to seaborn

GaetandeCast:
In the documentation, some references to seaborn functions are redundant:

image
so let's remove the "seaborn's" mention.

 
An alternative fix would be to write 
```
seaborn's :func:`heatmap<seaborn.heatmap>`
```
instead of the current
```
seaborn's :func:`seaborn.heatmap`
```
but that feels a bit overkill.


auguste-probabl:
nice

auguste-probabl (approved):
Nice work!

PR #2301, opened 2026-01-16
docs(skore): Use coherent documentation style

auguste-probabl:


glemaitre:
@auguste-probabl Is it the only changes that you want to include before that I merge?

Issue #1780, opened 2025-05-28
feat(skore): Create a display for `feature_importance.permutation` for the different reports


glemaitre:
### What would you like to say?

Create a display for `feature_importance.permutation` for the different reports.

- [ ] Refactor and create display for `EstimatorReport` https://github.com/probabl-ai/skore/pull/2304
- [ ] Adapt display `CrossValidationReport`
- [ ] Adapt display for `ComparisonReport`

thomass-dev:
To notice, this is a breaking change: you change the return from dataframe to display, that is currently used by `skore-hub-project`.

divakaivan:
Hello, I am trying to get a better view on what type of graphs you'd expect from the different reports.

### For `EstimatorReport`

```python
disp1 =  report.feature_importance.permutation(
   scoring=["r2", "rmse"],
   n_repeats=2,
   seed=0,
)

disp2 = report.feature_importance.permutation(
   n_repeats=2,
   aggregate=["mean", "std"],
   seed=0,
)
```

`frame` would return

```
In [10]: disp1.frame()
Out[10]: 
Repeat             Repeat #0  Repeat #1
Metric Feature                         
r2     Feature #0   0.112065   0.048028
       Feature #1   1.616931   1.905097
       Feature #2   0.000000   0.016009
       Feature #3  -0.032018  -0.048028
rmse   Feature #0  -0.050333  -0.022699
       Feature #1  -0.431123  -0.481865
       Feature #2   0.000000  -0.007786
       Feature #3   0.016339   0.024947

In [11]: disp2.frame()
Out[11]: 
                      mean       std
Metric   Feature                    
accuracy Feature #0  0.020  0.011314
         Feature #1  0.440  0.050912
         Feature #2  0.002  0.002828
         Feature #3 -0.010  0.002828
```

and `plot`:

- for `disp1`

<img width="794" height="795" alt="Image" src="https://github.com/user-attachments/assets/ba3c7c35-9a4e-459e-9228-8a15eb2bacb2" />

- for `disp2`

<img width="752" height="607" alt="Image" src="https://github.com/user-attachments/assets/d3733ed7-d9e4-439f-ab49-970c9f201553" />

### For `CrossValidationReport

Would this be displaying a plot per cv split?

### For `ComparisonReport`

Reusing the above cases

divakaivan:
@auguste-probabl could you assign this to me, please?

PR #2304, opened 2026-01-16
feat(skore): Create `PermutationImportanceDisplay` for `permutation()` method

glemaitre:
Partially addressing https://github.com/probabl-ai/skore/issues/1780

This first step is to create a proper display named `PermutationImportanceDisplay`. It therefore requires to refactor and cleanup the code that is currently in `EstimatorReport.feature_importance.permutation`

On thing to be aware is that the method `feature_importance.permutation()` will change the way caching is working. The display itself is serving as the caching mechanism retaining the computed information. The report will therefore need to cache the object created based on the key.

TODO:

- [ ] Open an issue where we expose an attribute `facet_` in plot.

### DIFF
--- a/skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
+++ b/skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
@@ -400,7 +400,7 @@ class PermutationImportanceDisplay(DisplayMixin):
         boxplot_kwargs: dict[str, Any] | None = None,
         stripplot_kwargs: dict[str, Any] | None = None,
     ):
-        f"""Set the style parameters for the display.
+        """Set the style parameters for the display.
 
         Parameters
         ----------
--- a/sphinx/reference/report/displays.rst
+++ b/sphinx/reference/report/displays.rst
@@ -33,7 +33,7 @@ the API of each display.
    :template: base.rst
 
    CoefficientsDisplay
-   PermutationTestDisplay
+   PermutationImportanceDisplay
 
 .. rubric:: General API
 


skore/src/skore/_sklearn/_plot/feature_importance/utils.py
```diff
@@ -0,0 +1,44 @@
...
+    ----------
+    ax : plt.Axes
+        The matplotlib axis to decorate.
+    add_background_features : bool
+        Whether to add a background color for each group of features.
+    n_features : int
+        The number of features to displayed.
+    xlabel : str
+        The label for the x-axis.
+    """
```
GaetandeCast, line 28:
```suggestion
    ylabel : str
        The label for the y-axis.
    """
```

skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
```diff
@@ -0,0 +1,405 @@
...
+            else:
+                metric_name = metric.__name__.replace("_", " ")
+            scores = {metric_name: scores}
+
+        df_importances = []
+        for metric_name, metric_values in scores.items():
+            metric_importances = np.atleast_3d(metric_values["importances"])
+
+            df_metric_importances = []
+            for target_index, target_importances in enumerate(
```
GaetandeCast, line 97: It might be worth adding a comment explaining that this loops across labels/outputs, when there is a different metric for each. It took me a while to understand :)

skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
```diff
@@ -0,0 +1,405 @@
...
+            for col_index, ax in enumerate(row_axes):
+                if len(metrics) > 1:
+                    if row == "metric":
+                        xlabel = f"Decrease of {metrics[row_index]}"
+                    elif col == "metric":
+                        xlabel = f"Decrease of {metrics[col_index]}"
+                    else:
+                        xlabel = "Decrease of metric"
+                else:
+                    xlabel = f"Decrease of {metrics[0]}"
```
GaetandeCast:
```suggestion
                        xlabel = f"Decrease in {metrics[row_index]}"
                    elif col == "metric":
                        xlabel = f"Decrease in {metrics[col_index]}"
                    else:
                        xlabel = "Decrease in metric"
                else:
                    xlabel = f"Decrease in {metrics[0]}"
```
"Decrease in (e.g.) accuracy" is more correct I think

skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
GaetandeCast: Also there should be a check somewhere that all metrics are either per label/output or all "averaged" metrics. Otherwise the `plot()` crashes and the `frame()` is weird.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
```diff
@@ -263,14 +142,12 @@ def permutation(
         data_source: DataSource = "test",
```
GaetandeCast, line 142: We will want "both" at some point here too ;) since comparing train and test can help detect overfitting.

skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
```diff
@@ -0,0 +1,405 @@
...
+        aggregate : Aggregate | None, default=("mean", "std")
+            Aggregate the importances by the given metric.
+
+        Returns
+        -------
+        pd.DataFrame
+            Dataframe containing the importances.
+        """
+        if self.report_type == "estimator":
+            columns_to_drop = ["estimator"]
```
GaetandeCast, line 359: The renaming from `estimator_name` to `estimator` makes this unintuitive but oh well 🤷 Also this reminds me that we should apply the renaming to `estimator` to the whole code base.

skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
glemaitre: Actually I will add a parameter `metric` to `frame` or `plot` to be able to filter. If once filter, we are in a weird stage, then we raise an error.

skore/src/skore/_sklearn/_estimator/feature_importance_accessor.py
glemaitre: Yes, we already have inside an example where we do it by hand and it adds some boilerplate.

glemaitre:
@GaetandeCast it is good for another round. I added a new `metric` parameter to be able to quickly filter the frame and plot. I think it is handy because you might want to compute multiple metrics to create the display and just filter afterwards instead of calling several time the permutation which will recompute the predictions.

skore/src/skore/_sklearn/_plot/feature_importance/permutation.py
```diff
@@ -0,0 +1,435 @@
...
+    # ignore the type signature because we override kwargs by specifying the name of
+    # the parameters for the user.
+    def set_style(  # type: ignore[override]
+        self,
+        *,
+        policy: Literal["override", "update"] = "update",
+        boxplot_kwargs: dict[str, Any] | None = None,
+        stripplot_kwargs: dict[str, Any] | None = None,
+    ):
+        f"""Set the style parameters for the display.
```
GaetandeCast:
```suggestion
        """Set the style parameters for the display.
```

sphinx/reference/report/displays.rst
```diff
@@ -33,6 +33,7 @@ the API of each display.
    :template: base.rst
 
    CoefficientsDisplay
+   PermutationTestDisplay
```
GaetandeCast:
```suggestion
   PermutationImportanceDisplay
```

GaetandeCast:
Hmm we're hitting the timeout on tests with coverage again.

PR #2309, opened 2026-01-19
chore: Add ruff check for imports

auguste-probabl:


.pre-commit-config.yaml
```diff
@@ -27,6 +27,10 @@ repos:
       - id: ruff-check
         files: ^((skore|skore-hub-project|skore-local-project)/(hatch|src|tests))|(examples)/
         args: [--fix]
+      - id: ruff-check
+        alias: ruff check imports
+        files: ^((skore|skore-hub-project|skore-local-project)/(hatch|src|tests))|(examples)/
+        args: [--fix, --select, I]
```
thomass-dev: Can't you just add this option to the previous ruff-check job?

auguste-probabl: I'm not 100% sure; sorting is unintuitive with Ruff. As far as I knew `--select I` means "run *only* the check `I`" but I tried it and it seemed to work okay...

Issue #2183, opened 2025-12-01
Cannot plot PR curve with labels of type `np.str_`


auguste-probabl:
```python
from sklearn.datasets import load_iris
from sklearn.linear_model import LogisticRegression
from skore import train_test_split, ComparisonReport, CrossVal
idationReport
iris = load_iris(as_frame=True)
X, y = iris.data, iris.target
y = iris.target_names[y]
split_data = train_test_split(
    X=X, y=y, random_state=0, as_dict=True, shuffle=True
)
report_1 = CrossValidationReport(LogisticRegression(), X, y)
report_2 = CrossValidationReport(LogisticRegression(max_iter=500), X, y)
report = ComparisonReport([report_1, report_2])
display = report.metrics.precision_recall()
display.plot()
```
raises a strange-sounding error: `UndefinedVariableError: name 'np' is not defined`

The offending line is 
<https://github.com/probabl-ai/skore/blob/b2ed6df879e4654015d6f653a7449a4f327f1f1a/skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py#L657>

The issue is that labels are `np.str_` instances, so the `!r` in the f-string results in `"label == np.str_('setosa') & ..."` but it should be `"label == 'setosa' & ..."`. We use `!r` to ensure that normal Python strings are printed with surrounding quotes:
```
f"label == {label}" -> "label == setosa"
while
f"label == {label!r}" -> "label == 'setosa'".
```
We need the surrounding quotes for Pandas queries to work.

### Environment

Commit b2ed6df

DCchoudhury15:
@auguste-probabl  can i work on this ?

GaetandeCast:
@DCchoudhury15 Actually this was closed by #2198.

PR #2313, opened 2026-01-20
feat(skore): Refactor RocCurveDisplay to use seaborn and add subplot_by

### LINKED ISSUE #1445: enh(skore): Add an option to plot into several subplots some display information

When dealing with the `ComparatorReport` or `CrossValidationReport`, it is sometimes difficult to compare some curve or plot together: colors and types of marker or line are not enough to easily distinguish models or there is to much information on a single plot.

It would be nice to have a parameter to expand the display to multiplot in some cases. This option could be turn to `"auto"` such that we make a choice for the user.

### LINKED ISSUE #1957: Allow to set the style kwargs only in `set_style` and not in `plot`

### LINKED ISSUE #2190: refactor: Rewrite Displays using seaborn API
The different Displays are hard to maintain and extend; each new proposed option causes a new if-statement that makes the complexity of the code grow exponentially. The current occurrence of this is the `data_source="both"` option, proposed in https://github.com/probabl-ai/skore/issues/1874, and whose implementation has already begun; the next one after that is the `subplots_by` parameter proposed in https://github.com/probabl-ai/skore/issues/1445 and iterated on in https://github.com/probabl-ai/skore/pull/1720.

A proposed solution is inspired from https://github.com/probabl-ai/skore/pull/2152, which is to **refactor the plotting logic to rely on `seaborn`'s DataFrame-friendly API**. Indeed, Display data is already stored as DataFrames and exposed through `.frame()`.

One might further consider that `.plot()` could call `.frame()` itself, as is the case in https://github.com/probabl-ai/skore/pull/2177

Risk: This will not help us implement e.g. plotly representation.

Sub-tasks:
- [x] Refactor `PrecisionRecallCurveDisplay` using `seaborn` : #2193
- [x] Refactor `ConfusionMatrixDisplay` using `seaborn` : #2165 
- [x] Refactor `CoefficientsDisplay` using `seaborn` : #2152 
- [x] Refactor `PredictionErrorCurveDisplay` using `seaborn` : #2292 
- [x] Refactor `RocCurveDisplay` using `seaborn` : #2313

GaetandeCast:
As the last display on the list to refactor is `RocCurveDisplay`, this PR closes #1445, closes #2190 and closes #1957.

I put utility functions created for `PrecisionRecallCurveDisplay` in `._plot.utils` which inflates the diff quite a bit. Doing so I noticed some improvements to do in `precision_recall_curve.py` regarding kwargs.

screenshots:
image
image
image
image


waridrox:
oops, I was about to make a draft PR for the same, but I see you already did it 😅 @GaetandeCast 

PS: There were a lot of tests breaking as well...

GaetandeCast:
Yeah sorry @waridrox, we had to move forward with these since the refactors block the adding of `data_source="both"` to the `ComparisonReport` which is in our priority list. Thanks for the time you put in!

skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -124,20 +125,24 @@ def __init__(
     def plot(
         self,
         *,
-        subplot_by: Literal["auto", "label", "estimator"] | None = "auto",
+        subplot_by: Literal["auto", "label", "estimator", "data_source"]
+        | None = "auto",
```
auguste-probabl, line 129:
```suggestion
        subplot_by: Literal["auto", "label", "estimator", "data_source"] | None = "auto",
```

skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
```diff
@@ -785,21 +144,21 @@ def plot(
...
         Parameters
         ----------
-        estimator_name : str, default=None
-            Name of the estimator used to plot the ROC curve. If `None`, we use
-            the inferred name from the estimator.
-
-        roc_curve_kwargs : dict or list of dict, default=None
-            Keyword arguments to be passed to matplotlib's `plot` for rendering
-            the ROC curve(s).
+        subplot_by : Literal["auto", "label", "estimator", "data_source"] | None, \
```
auguste-probabl:
```suggestion
        subplot_by : {"auto", "label", "estimator", "data_source"} or None, \
```

skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
GaetandeCast: This makes it one character too long for ruff.

skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
```diff
@@ -785,21 +144,21 @@ def plot(
...
+            default="auto"
+            Column to use for creating subplots. Options:
+
+            - "auto": None for EstimatorReport and Cross-Validation Report, \
```
auguste-probabl:
```suggestion
            - "auto": None for EstimatorReport and CrossValidationReport, \
```

skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
auguste-probabl:
```suggestion
	is_cross_validation = "cross-validation" in self.report_type
```
And then inline the variable (don't define a new variable for this)

skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
```diff
@@ -814,109 +173,138 @@ def plot(
...
-                        estimator_name
-                        or self.roc_auc["estimator_name"].cat.categories.item()
-                    ),
-                    roc_curve_kwargs=roc_curve_kwargs,
-                    plot_chance_level=plot_chance_level,
-                    chance_level_kwargs=chance_level_kwargs,
+        if style:
+            relplot_kwargs["dashes"] = {"train": (5, 5), "test": ""}
+
+        is_cross_validation = "cross-validation" in self.report_type
```
auguste-probabl: inline

skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
```diff
@@ -1136,3 +524,48 @@ def frame(self, with_roc_auc: bool = False) -> DataFrame:
...
+        Parameters
+        ----------
+        policy : Literal["override", "update"], default="update"
+            Policy to use when setting the style parameters.
+            If "override", existing settings are set to the provided values.
+            If "update", existing settings are not changed; only settings that were
+            previously unset are changed.
+
+        relplot_kwargs : dict, default=None
+        Keyword arguments to be passed to :func:`seaborn.relplot` for rendering
```
auguste-probabl:
```suggestion
            Keyword arguments to be passed to :func:`seaborn.relplot` for rendering
```

skore/src/skore/_sklearn/_plot/metrics/roc_curve.py
```diff
@@ -814,109 +173,138 @@ def plot(
...
+                if self.pos_label is not None
+                else None
             )
         else:
-            raise ValueError(
-                "`report_type` should be one of 'estimator', 'cross-validation', "
-                "'comparison-cross-validation' or 'comparison-estimator'. "
-                f"Got '{self.report_type}' instead."
-            )
+            info_pos_label = None
```
auguste-probabl:
```suggestion
        if self.ml_task == "binary-classification" and self.pos_label is not None:
            info_pos_label = f"Positive label: {self.pos_label}"
        else:
            info_pos_label = None
```

auguste-probabl: Rendering is weird. If my suggestion doesn't work then leave it

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -359,3 +361,166 @@ def sample_mpl_colormap(
...
+
+    Returns (col, hue, style) tuple where each can be None if not applicable.
+    """
+    has_multiple_estimators = (
+        "estimator" in plot_data.columns and plot_data["estimator"].nunique() > 1
+    )
+    is_comparison = report_type in (
+        "comparison-estimator",
+        "comparison-cross-validation",
+    )
```
auguste-probabl:
```suggestion
    is_comparison = "comparison" in report_type
```

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -359,3 +361,166 @@ def sample_mpl_colormap(
...
+    is_multiclass = ml_task == "multiclass-classification"
+    has_both_data_sources = data_source == "both"
+
+    allowed_values: set[str | None] = {"auto"}
+    if is_multiclass:
+        allowed_values.add("label")
+        if report_type in ["estimator", "cross-validation"]:
```
auguste-probabl:
```suggestion
        if not is_comparison:
```

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -359,3 +361,166 @@ def sample_mpl_colormap(
...
+    statistic_column_name: Literal["average_precision", "roc_auc"],
+    statistic_acronym: Literal["AP", "AUC"],
+    chance_level_label: str | None = None,
+) -> None:
+    """Build custom legend with a custom statistic for a single axis."""
+    legend_labels = []
+    for hue_value in hue_order or [None]:
+        hue_value_str = (
+            f"'{hue_value}'" if isinstance(hue_value, str) else str(hue_value)
+        )
```
auguste-probabl, line 362:
```suggestion
        hue_value_str = f"{hue_value!r}"
```
not 100% sure

skore/tests/unit/displays/roc_curve/test_comparison_estimator.py
```diff
@@ -217,9 +205,37 @@ def test_multiple_roc_curve_kwargs_error(
...
+            assert mpl.colors.to_rgb(line.get_color()) != default_color
+
+    else:
+        display.set_style(relplot_kwargs={"color": "red"}).plot()
+        expected_colors = ["red"] * n_reports
+        for line, expected_color, default_color in zip(
+            display.lines_[:n_roc_lines], expected_colors, default_colors, strict=True
+        ):
+            assert line.get_color() == expected_color
+            assert mpl.colors.to_rgb(line.get_color()) != default_color
```
auguste-probabl:
```suggestion
    if multiclass:
        palette_colors = ["red", "blue", "green"]
        display.set_style(relplot_kwargs={"palette": palette_colors}).plot()
        expected_colors = palette_colors * n_reports
    else:
        display.set_style(relplot_kwargs={"color": "red"}).plot()
        expected_colors = ["red"] * n_reports

    for line, expected_color, default_color in zip(
        display.lines_[:n_roc_lines], expected_colors, default_colors, strict=True
    ):
        assert line.get_color() == expected_color
        assert mpl.colors.to_rgb(line.get_color()) != default_color

```

skore/tests/unit/displays/roc_curve/test_comparison_estimator.py
```diff
@@ -397,11 +410,79 @@ def test_multiclass_classification_constructor(
...
+    )
+    report_2 = EstimatorReport(
+        estimator, X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
+    )
+    report = ComparisonReport(
+        reports={"estimator_1": report_1, "estimator_2": report_2}
+    )
+    display = report.metrics.roc()
+    valid_values_str = ", ".join(valid_values)
+    err_msg = f"subplot_by must be one of {valid_values_str}. Got 'invalid' instead."
```
auguste-probabl:
```suggestion
    err_msg = f"subplot_by must be one of {', '.join(valid_values)}. Got 'invalid' instead."
```

sphinx/user_guide/displays.rst
```diff
@@ -40,29 +40,15 @@ not modify the display object nor require heavy computation.
...
 
 .. plot::
     :context: close-figs
     :align: center
 
     disp.set_style(
         chance_level_kwargs=dict(linestyle="-", linewidth=5, color="tab:purple")
-    )
-    disp.plot()
+    ).plot()
```
auguste-probabl: i think maybe leave it as `disp.plot()` to make it clear that `set_style` is stateful. we can do `set_style().plot()` in other places

skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
auguste-probabl: ouch. fair enough

skore/src/skore/_sklearn/_plot/utils.py
GaetandeCast: it works, I also inlined it

GaetandeCast: nevermind it broke the fix for issue #2183, reverting it.

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -65,95 +67,6 @@ def _validate_from_predictions_params(
         return pos_label
 
 
-class _ClassifierCurveDisplayMixin(_ClassifierDisplayMixin):
-    """Mixin class to be used in Displays requiring a classifier curve."""
-
-    def _validate_curve_kwargs(
```
GaetandeCast, line 71: the utils function `_validate_curve_kwargs` is no longer in use so we can remove the `_ClassifierCurveDisplayMixin` and use the `_ClassifierDisplayMixin` instead.

PR #2316, opened 2026-01-21
feat(skore): Add an HTML representation for interactive environment

### LINKED ISSUE #1104: feat(skore): create an HTML representation for `.help` in the differen reports

glemaitre:
closes #1104 

This is an experimental PR to show what could be done regarding having an HTML representation when we are in an interactive environment.

Check the documentation preview and the getting started example to see the rendering and behaviour.

This PR is doing a full refactoring of the `repr` module:

- we are creating a module specifically to handle rich representation
- we are creating a module specifically to handle HTML representation
- we have have jinja template for the html representation
- we create a data module to define what are the models used to inject the information into the rich and HTML representation. It contains as well all the utility to create those data model
- we create a base module containing a report, display, and accessor representation that delegate to the rich or HTML representation.
- finally, we tested each individual module which make it easier to ensure the proper behavior instead of an integrated tests for each report.

### DIFF
--- a/skore/src/skore/_utils/repr/rich_repr.py
+++ b/skore/src/skore/_utils/repr/rich_repr.py
@@ -2,9 +2,7 @@
 
 from abc import ABC, abstractmethod
 from dataclasses import asdict
-from io import StringIO
 
-from rich.console import Console
 from rich.panel import Panel
 from rich.tree import Tree
 
--- a/skore/src/skore/_sklearn/_cross_validation/data_accessor.py
+++ b/skore/src/skore/_sklearn/_cross_validation/data_accessor.py
@@ -1,4 +1,4 @@
-from typing import Any, Literal
+from typing import Literal
 
 import pandas as pd
 from skrub import _dataframe as sbd
--- a/skore/src/skore/_sklearn/_estimator/data_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/data_accessor.py
@@ -1,4 +1,4 @@
-from typing import Any, Literal
+from typing import Literal
 
 import pandas as pd
 from skrub import _dataframe as sbd
--- a/skore/src/skore/_sklearn/_plot/base.py
+++ b/skore/src/skore/_sklearn/_plot/base.py
@@ -1,4 +1,3 @@
-import inspect
 from collections.abc import Callable
 from functools import wraps
 from typing import Any, Literal, Protocol, runtime_checkable
--- a/skore/tests/unit/reports/comparison/test_report.py
+++ b/skore/tests/unit/reports/comparison/test_report.py
@@ -3,7 +3,6 @@ Tests of ComparisonReport which work regardless whether it holds EstimatorReport
 CrossValidationReports.
 """
 
-import re
 from io import BytesIO
 
 import joblib
--- a/skore/tests/unit/reports/cross_validation/test_report.py
+++ b/skore/tests/unit/reports/cross_validation/test_report.py
@@ -1,5 +1,4 @@
 import inspect
-import re
 
 import joblib
 import numpy as np
--- a/skore/tests/unit/reports/estimator/test_report.py
+++ b/skore/tests/unit/reports/estimator/test_report.py
@@ -1,5 +1,4 @@
 import inspect
-import re
 from copy import deepcopy
 from io import BytesIO
 


skore/src/skore/_sklearn/_base.py
```diff
@@ -18,8 +19,648 @@
...
+
+        return Panel(
+            content,
+            title=self._get_help_panel_title(),
+            expand=False,
+            border_style="orange1",
+        )
+
+
+class _HTMLHelpMixin(ABC):
```
rouk1: 😿 

Did you consider using an alternative approach that may be more maintainable ?
Two track to consider:
- a web component that gets hydrated by a json model 
- a templating language like Jinja (it is already available as a transient dependency through skrub iirc)

Both those strategy would avoid fragile escaping that made my eyes bleed 😆

glemaitre: Yes I think it is reasonable. It was to check the visualization but the solution is horrible to maintain.

I can think easily of a jinja. Do you have a resource where I could look at the first solution.

glemaitre:
I think it is good to go actually.

sphinx/Makefile
```diff
@@ -6,6 +6,7 @@ ifeq ($(SPHINX_DOMAIN),)
 	SPHINX_DOMAIN := docs.skore.probabl.ai
 endif
 
+export SPHINX_BUILD ?= 1
```
thomass-dev, line 9: Take a look on the line 17, isn't in conflict?

skore/src/skore/_utils/repr/help.css
```diff
@@ -0,0 +1,292 @@
+/* Light theme (default) */
+.skore-help-container,
```
rouk1: As you're using a shadow dom, you do not have to prefix all your class names. They will not leak to the host page.
Maybe you could only use semantic class name for readability.

skore/src/skore/_utils/repr/display_help-content.html.j2
```diff
@@ -0,0 +1,44 @@
+<div class="skore-help-container">
+    <div class="skore-help-title">{{ title }}</div>
```
rouk1:
```suggestion
    <header class="skore-help-title">{{ title }}</header>
```

skore/src/skore/_utils/repr/display_help-content.html.j2
```diff
@@ -0,0 +1,44 @@
...
+            <span class="skore-help-label">{{ root_node }}</span>
+        </div>
+        <div class="skore-help-root-content">
+            {# Methods section #}
+            {% if methods %}
+                <div class="skore-help-node skore-help-folder">
+                    <input type="checkbox" id="{{ methods_section.id }}" class="skore-help-toggle">
+                    <label for="{{ methods_section.id }}" class="skore-help-folder-label">
+                        <span class="skore-help-section-name">Methods</span>
+                    </label>
```
rouk1: You can avoid an id by nesting the input inside the label.


```suggestion
                    <label class="skore-help-folder-label">
                        <input type="checkbox" class="skore-help-toggle">
                        <span class="skore-help-section-name">Methods</span>
                    </label>
```

skore/src/skore/_utils/repr/display_help-content.html.j2
```diff
@@ -0,0 +1,44 @@
...
+                    <div class="skore-help-folder-content" id="{{ methods_section.branch_id }}">
+                        {% for method in methods %}
+                            <div class="skore-help-node skore-help-method">
+                                <span class="skore-help-method-name">.<a href="{{ method.doc_url }}" target="_blank" class="skore-help-method-tooltip">{{ method.name }}<span class="skore-help-tooltip-text">{{ method.description|e }}<br><br>Click to access the online documentation</span></a>{{ method.parameters }}</span>
```
rouk1: `<br/>` should in my opinion be avoid as much as possible. Formatting should happen at the CSS level.
Here you create a structure that could have been clearly represented with tags.

Also it is not 1998 anymore "click here" is a double information, your link is already in a anchor tag, user will understand that 😅

skore/src/skore/_utils/repr/display_help-content.html.j2
```diff
@@ -0,0 +1,44 @@
...
+            {% if attributes %}
+                <div class="skore-help-node skore-help-folder">
+                    <input type="checkbox" id="{{ attributes_section.id }}" class="skore-help-toggle">
+                    <label for="{{ attributes_section.id }}" class="skore-help-folder-label">
+                        <span class="skore-help-section-name">Attributes</span>
+                    </label>
+                    <div class="skore-help-folder-content" id="{{ attributes_section.branch_id }}">
+                        {% for attr in attributes %}
+                            <div class="skore-help-node skore-help-method">
+                                <span class="skore-help-method-name">.<a href="{{ attr.doc_url }}" target="_blank" class="skore-help-method-tooltip">{{ attr.name }}<span class="skore-help-tooltip-text">{{ attr.description|e }}<br><br>Click to access the online documentation</span></a></span>
```
rouk1: Same as above.

sphinx/Makefile
glemaitre: Nop, it is an environment variable to mention that we are building the documentation.

The one below is used by the script as an internal variable.

skore/src/skore/_utils/repr/display_help-content.html.j2
glemaitre: Let's remove the part of the code where I'm clearly demonstrated that I'm old fashion ;)

glemaitre:
I addressed the comment. I'll modify the documentation of accessor to be closer to the same pattern than report to be able to fold the methods. It will be more consistent.

glemaitre:
It is good on my side. You can have a look at the generated documentation to have a preview.
In another PR, I could reuse the HTML generated code to add it in the landing page.

glemaitre:
> I played with it a little in the docs preview and noticed that for CrossValidationReports and ComparisonReports, for non linear models, the feature_importance accessor appears but can't be interacted with since there is no available feature importance method in skore in these cases yet. Is it on purpose ? Shouldn't we rather hide the accessor if it can access nothing ?

The accessor is actually register programmatically but do not have method. I did not think about this case. I'll make sure that we append the accessor only if at least an item is available. If it makes it difficult then, we can just wait because it will bee soonish available.

glemaitre:
@GaetandeCast It was quite trivial to manage this corner case by avoiding adding an accessor if we did not have any method.

I added some non-regression test.

skore/src/skore/_utils/repr/display_help.html.j2
auguste-probabl, line 1: Since we depend explicitly on jinja2 now it should be added to pyproject.toml and the lock files should be regenerated

glemaitre: I added the dependencies and generated the lock files.

skore/src/skore/_utils/repr/display_help.html.j2
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    const template = document.getElementById('{{ container_id }}-template');
+    if (!template) {
+        return;
+    }
+
+    // Create shadow root with closed mode for better isolation
+    const shadowRoot = container.attachShadow({ mode: 'closed' });
+    shadowRoot.appendChild(template.content.cloneNode(true));
```
auguste-probabl, line 31: For the record, I still think this is an anti-pattern and we can achieve isolation with some simple prefixing of attributes (e.g. call every class `skore-help-*`). For instance, this could help us get rid of the Javascript code that just checks for dark mode: dark mode is now natively supported in CSS via `@media (prefers-color-scheme: dark)`

skore/src/skore/_utils/repr/help.css
```diff
@@ -0,0 +1,289 @@
...
+    --color-blue: #6b42f6;
+    --color-orange: #f97316;
+    --color-text: #fff;
+    --color-background: #000;
+    --color-orange-hover: rgba(249, 115, 22, 0.2);
+    --color-blue-hover: rgba(107, 66, 246, 0.2);
+}
+
+.container {
+    font-family: 'Monaco', 'Menlo', 'Ubuntu Mono', 'Consolas', monospace;
```
auguste-probabl: I'd prefer `monospace` to be the default, because then the user has control over it.

skore/src/skore/_utils/repr/html_repr.py
```diff
@@ -0,0 +1,90 @@
...
+from jinja2 import Environment, FileSystemLoader
+
+from skore._utils.repr.data import (
+    _AccessorHelpDataMixin,
+    _DisplayHelpDataMixin,
+    _ReportHelpDataMixin,
+)
+
+
+def get_jinja_env():
```
auguste-probabl, line 17: This could be cached

skore/src/skore/_utils/repr/rich_repr.py
```diff
@@ -0,0 +1,161 @@
...
+
+        base_methods = data.get("base_methods", [])
+        if base_methods:
+            methods_branch = tree.add("[bold cyan]Methods[/bold cyan]")
+            for method in base_methods:
+                displayed_name = f"{method['name']}(...)"
+                description = method["description"]
+                methods_branch.add(f".{displayed_name}".ljust(26) + f" - {description}")
+
+        attributes = data.get("attributes") or []
```
auguste-probabl:
```suggestion
        attributes = data.get("attributes", [])
```

skore/tests/unit/reports/estimator/feature_importance/test_accessor.py
```diff
@@ -1,3 +1,5 @@
+import pytest
```
auguste-probabl: Unused import?

skore/tests/unit/utils/repr/test_base.py
```diff
@@ -0,0 +1,218 @@
...
+    display_with_base_help.help()
+    mock_notebook.assert_called_once()
+    mock_html_cls.assert_called_once()
+    html_arg = mock_html_cls.call_args[0][0]
+    assert "skore-display-help-" in html_arg
+    mock_display.assert_called_once_with(sentinel_html)
+
+
+@patch("skore._utils.repr.base.is_environment_notebook_like", return_value=False)
+@patch("skore.console.print")
```
auguste-probabl: Nice; in the rest of the test suite we tend to use the pytest `monkeypatch` fixture but since this is builtin, I would tend to prefer it. We should eventually pick one and be consistent.

skore/tests/unit/utils/repr/test_data.py
```diff
@@ -0,0 +1,469 @@
...
+    ],
+)
+def test_build_attribute_text_fragment(obj, attribute_name, expected):
+    """_build_attribute_text_fragment returns encoded name,-type or encoded name."""
+    got = _build_attribute_text_fragment(obj, attribute_name)
+    if expected is not None:
+        expected = f"{quote(attribute_name, safe='')},-{quote(expected, safe='')}"
+    else:
+        expected = quote(attribute_name, safe="")
+    assert got == expected
```
auguste-probabl:
```suggestion
@pytest.mark.parametrize(
    "obj, attribute_name, expected",
    [
        (_ClassWithNumpydocAttrs(), "coef_", "coef_,-ndarray (n_features,)"),
        (_ClassWithNumpydocAttrs(), "param", "param_,-int"),
        (_ClassWithNoDocstring(), "coef_", "coef_"),
        (_ClassWithNumpydocAttrs(), "other_attr", "other_attr"),
    ],
)
def test_build_attribute_text_fragment(obj, attribute_name, expected):
    """_build_attribute_text_fragment returns encoded name,-type or encoded name."""
    assert _build_attribute_text_fragment(obj, attribute_name) == expected
```

skore/tests/unit/utils/repr/test_base.py
thomass-dev: Just a general note: prefer `pytest` when possible, it is used all over the repository.
We can reopen the debate `unittest` or `pytest`, but i don't think it is the right place 😄 .

skore/tests/unit/utils/repr/test_data.py
```diff
@@ -0,0 +1,469 @@
...
+    "method_name, expected",
+    [
+        ("documented", "First line of docstring."),
+        ("undocumented", "No description available"),
+    ],
+)
+def test_get_method_short_summary(method_name, expected):
+    """get_method_short_summary returns first docstring line or fallback."""
+    obj = _ClassWithDocstringMethod()
+    method = getattr(obj, method_name)
```
auguste-probabl:
```suggestion
    method = getattr(_ClassWithDocstringMethod(), method_name)
```

skore/tests/unit/utils/repr/test_base.py
thomass-dev: For the specific case when you need to assert that your mock has been called, combining `pytest.monkeypatch.setattr`  and `unittest.mock.Mock` is a good solution.

skore/tests/unit/utils/repr/test_data.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+def test_get_documentation_url_class_only(report_with_methods, display_with_methods):
+    """get_documentation_url returns base class URL with no accessor/method/attribute."""
+    for obj in (report_with_methods, display_with_methods):
+        url = get_documentation_url(obj=obj)
+        assert url.startswith("https://docs.skore.probabl.ai/")
+        assert "/reference/api/" in url
+        assert "skore." in url and obj.__class__.__name__ in url
+        assert url.endswith(".html")
```
auguste-probabl: Nit
```suggestion
        assert re.match(rf'https://docs.skore.probabl.ai/[^/]+/reference/api/skore\.[^/]+{obj.__class__.__name__}.html', url)
```

skore/tests/unit/utils/repr/test_html_repr.py
```diff
@@ -0,0 +1,69 @@
...
+@pytest.fixture
+def accessor_with_html(report_with_html):
+    """Accessor with HTML help mixin."""
+    return _AccessorWithHTML(parent=report_with_html)
+
+
+@pytest.fixture
+def display_with_html():
+    """Display with HTML help mixin."""
+    return _DisplayWithHTML()
```
auguste-probabl: Inline those

sphinx/Makefile
auguste-probabl: They are pretty similar which is confusing. Rename maybe to `IS_SPHINX_BUILD`, or even better: don't define a new variable and instead check for `SPHINX_DOMAIN`, `SPHINX_VERSION` or `SPHINX_RELEASE`

skore/src/skore/_utils/repr/data.py
```diff
@@ -0,0 +1,543 @@
...
+        sig = inspect.signature(method)
+        parameter_names = [
+            parameter_name
+            for parameter_name, _ in sig.parameters.items()
+            if parameter_name != "self"
+        ]
+        if parameter_names:
+            parameters = "(" + ", ".join(parameter_names) + ")"
+        else:
+            parameters = "()"
```
auguste-probabl:
```suggestion
        parameters = f"({", ".join(parameter_names)})"
```

skore/src/skore/_utils/repr/data.py
```diff
@@ -0,0 +1,543 @@
...
+
+        description = get_method_short_summary(method)
+        favorability = None
+        if hasattr(obj, "_get_favorability_text"):
+            favorability = obj._get_favorability_text(name)
```
auguste-probabl:
```suggestion
        favorability = obj._get_favorability_text(name) if hasattr(obj, "_get_favorability_text") else None
```

skore/src/skore/_utils/repr/display_help.html.j2
glemaitre: We exactly did way to avoid prepanding using the shadow-DOM. Here, I would still stick to this anti-pattern because the usual pattern troubled us so much before.

skore/src/skore/_utils/repr/data.py
```diff
@@ -0,0 +1,543 @@
...
+            AttributeHelp(
+                name=attr_name,
+                description=get_attribute_short_summary(self, attr_name),
+                doc_url=get_documentation_url(
+                    obj=self,
+                    attribute_name=attr_name,
+                ),
+            )
+            for attr_name in attrs_without_underscore + attrs_with_underscore
+        ]
```
auguste-probabl:
```suggestion
        items = [
            AttributeHelp(
                name=attr_name,
                description=get_attribute_short_summary(self, attr_name),
                doc_url=get_documentation_url(obj=self, attribute_name=attr_name),
            )
            for attr_name in attributes
        ]
```

auguste-probabl (changes requested):
Nearly there. There are some more deep stuff like the shadow DOM thing, that it would be nice to go over again. Otherwise it's looking good.

skore/src/skore/_utils/repr/display_help.html.j2
auguste-probabl: What's wrong with prepending? If it's just an issue of adding noise, I think it's worth it given the complexity that the shadow DOM brings.

skore/src/skore/_utils/repr/html_repr.py
glemaitre: If templating with `jinja` comes somewhere else then having a function local to the module is better because we load things relatively to it.

skore/tests/unit/reports/estimator/feature_importance/test_accessor.py
glemaitre: Indeed. I messed up the merge conflict.

skore/tests/unit/utils/repr/test_base.py
glemaitre: Let's keep `pytest` :)

skore/src/skore/_utils/repr/display_help.html.j2
rouk1: > dark mode is now natively supported in CSS via @media (prefers-color-scheme: dark)

This is true but there is also IDE (like vs code and its forks and Jupyter) which use different strategies to select themes. 

Also event if I agree that shadow dom adds complexity it also comes with greta isolation power which is nice especially in a notebook.

glemaitre:
Could we merge this PR :)

Issue #1573, opened 2025-04-22
feat(skore): Use critical difference diagrams in ComparisonReport


sylvaincom:
### Is your feature request related to a problem? Please describe.

As of skore v0.8.2, we have a `ComparisonReport` but the comparison is somewhat limited.

### Describe the solution you'd like

Use critical difference diagrams to compare several models, see scikit-learn's issue by @ogrisel at https://github.com/scikit-learn/scikit-learn/issues/20319

waridrox:
Sounds interesting, can I work on this? Also, what packages do you have in mind that can be used?

Something like this looks more descriptive: https://scikit-posthocs.readthedocs.io/en/latest/tutorial.html

auguste-probabl:
Please wait for #1512

sylvaincom:
Hi @waridrox, thanks for your interest, but this issue is not ready yet. Indeed, critical difference diagrams are already implemented in other packages and we could re-use their implementation if they allow us. See the discussion at https://github.com/scikit-learn/scikit-learn/issues/20319

glemaitre:
Just important points when it comes to the critical difference diagrams.

The visual representation is usually related to compare models performing on different datasets. However, in a cross-validation setting, there is a dependence between the different test sets which is not the case with multiple datasets. So we need to be extremely careful when it comes to the underlying statistical tests that we are using depending of the type of comparison that we are doing.

- for cross-validation, we should rely on the [following paper](https://proceedings.neurips.cc/paper_files/paper/1999/file/7d12b66d3df6af8d429c1a357d8b9e1a-Paper.pdf)
- in the case of multiple datasets, then we can fallback on the original [paper](https://www.jmlr.org/papers/volume7/demsar06a/demsar06a.pdf)

It also means that this methodological aspects could be discussed more broadly in the documentation.

PR #2317, opened 2026-01-21
feat: Get the best estimator from a comparison

auguste-probabl:
Implements `ComparisonReport.get_best_model`, which allows to extract the best report according to the given metric 

Example:
```python
from skore import ComparisonReport, EstimatorReport, train_test_split
report_1 = EstimatorReport(LogisticRegression(), **split_data)
report_2 = EstimatorReport(DecisionTreeClassifier(), **split_data)
report.get_best_model(metric="accuracy")  # the default
```

### DIFF
--- a/skore/src/skore/_sklearn/_comparison/report.py
+++ b/skore/src/skore/_sklearn/_comparison/report.py
@@ -7,13 +7,18 @@ from typing import TYPE_CHECKING, Any, Literal, cast
 
 import joblib
 import numpy as np
+import pandas as pd
 from numpy.typing import ArrayLike
 
 from skore._externals._pandas_accessors import DirNamesMixin
 from skore._sklearn._base import _BaseReport
 from skore._sklearn._cross_validation.report import CrossValidationReport
 from skore._sklearn._estimator.report import EstimatorReport
-from skore._sklearn.types import _DEFAULT, PositiveLabel
+from skore._sklearn.types import (
+    _DEFAULT,
+    Metric,
+    PositiveLabel,
+)
 from skore._utils._cache import Cache
 from skore._utils._progress_bar import progress_decorator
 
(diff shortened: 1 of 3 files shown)


GaetandeCast:
Nice! As a next step we can try to find a way to determine if the advantage of the best model is statistically significant when we have the std deviation of the metric (comparison of cross vals). We can take inspiration from this scikit-learn example:
https://scikit-learn.org/stable/auto_examples/model_selection/plot_grid_search_stats.html#sphx-glr-auto-examples-model-selection-plot-grid-search-stats-py

auguste-probabl:
For our consideration: https://scikit-learn.org/dev/auto_examples/model_selection/plot_grid_search_stats.html#comparing-two-models-frequentist-approach (implements the adjusted t-test mentioned [here](https://github.com/probabl-ai/skore/issues/1573#issuecomment-3628043580)). Thanks @ogrisel

auguste-probabl:
This brings out a lot of questions.

Statistical estimators such as the corrected t-statistic mentioned above might not be appropriate for our use-case, especially considering that:
- The CrossValidationReports are fitted on the same data
- The CrossValidationReports might have been made with different cross-validation strategies (e.g. 3-fold and 5-fold, or stratified and non-stratified)

If it were possible to come up with a different statistic for every situation, this would still be suboptimal as it would expose users to a fair amount of complexity.

So maybe the best call is to use a simple strategy such as "Compute mean and standard deviation, and check if the error bars overlap", and be explicit that this method is not necessarily the best statistical test?

But then, what should we do if the bars do overlap? I don't see us raising an error saying "Couldn't get a statistically significant outcome", same for printing a warning along with the result. The UX would be poor for a result that is not worth it.

Further, getting the "best" estimator by 1 metric might not be a desirable thing for a data scientist; indeed, it's often more a question of tradeoffs, e.g. inference time vs. accuracy. It might be more interesting to show a "Pareto front" with the desired metrics and let the user make their choice of what is the best.

With that I'll close this and we'll go back to the drawing board for this feature.

PR #2319, opened 2026-01-21
docs: Add `confusion_matrix()`

auguste-probabl:


sphinx/reference/report/comparison_report.rst
```diff
@@ -57,6 +57,7 @@ get the common performance metric representations.
     ComparisonReport.metrics.rmse
     ComparisonReport.metrics.roc_auc
     ComparisonReport.metrics.timings
+    ComparisonReport.metrics.confusion_matrix
```
glemaitre: I would expect this one to be next to the roc and precision_recall because it is a display out of the box.

GaetandeCast: After `custom_metric`, the metrics seem to be sorted in alphabetical order so I would expect `confusion_matrix` between `brier_score` and `log_loss`. 
But actually there are some inconsistencies: in `estimator_report.rst`, `timings` is above `accuracy` while its at the end of the list in the other cases.

Let's maybe decide a definite way of ordering things and make thing consistent.

auguste-probabl: would be nice to automate this list in the first place

auguste-probabl: Until we do, I've sorted things in alphabetical order, except for `summarize` and `help`.

PR #2321, opened 2026-01-22
fix(skore-hub-project/project): Slugify tenant and project name

thomass-dev:
Quote the path segments using `urllib.parse.quote(..., safe="")` is not sufficient, as they are automatically unquoted by uvicorn before being routed to the hub. In many case, if user uses special characters in the project name, especially with `/` (%2F) and `#` (%23), it breaks the hub.

For instance, if the lib calls the URL `/projects/my%2Fproject/artifacts`, the hub receives the URL `/projects/my/project/artifacts`. The meaning of the URL changes after uvicorn has decoded all path segments containing `%`.

In order to remain RESTful, we choose to restrict the user rather than use query parameters.

<ins>References:</ins>
https://github.com/Kludex/starlette/issues/826
https://github.com/Kludex/uvicorn/blob/main/uvicorn/protocols/http/httptools_impl.py#L258

### DIFF
--- a/skore-hub-project/src/skore_hub_project/project/project.py
+++ b/skore-hub-project/src/skore_hub_project/project/project.py
@@ -6,13 +6,14 @@ import itertools
 import re
 import warnings
 from collections.abc import Callable
-from functools import cached_property, wraps
+from functools import wraps
 from operator import itemgetter
 from re import sub as substitute
 from tempfile import TemporaryFile
 from typing import (
     TYPE_CHECKING,
     Any,
+    Literal,
     ParamSpec,
     Protocol,
     TypedDict,
@@ -20,7 +21,6 @@ from typing import (
     runtime_checkable,
 )
 from unicodedata import normalize
-from urllib.parse import quote
 
 import joblib
 import orjson
@@ -59,17 +59,19 @@ def slugify(string: str) -> str:
     """
     Slugify string.
 
-    The string can only contain unicode letters, digits, and characters ``.``, ``-``,
-    and ``_``.
+    The string must be lower-case and contain only ASCII letters, digits, and characters
+    ``.``, ``-``, and ``_``.
 
     In order:
+    - convert to ASCII and ignore characters in error,
     - replace characters that aren't alphanumerics, dots, dashes or underscores by dash,
     - convert repeated dots to single dots,
     - convert repeated dashes to single dash,
     - convert repeated underscores to single underscore,
     - strip leading and trailing dots, dashes, and underscores.
     """
-    string = normalize("NFKC", string)
+    string = normalize("NFKD", string).encode("ascii", "ignore").decode("ascii")
+
     string = string.lower()
     string = substitute(r"[^\w.-]", "-", string)
     string = substitute(r"[.]+", ".", string)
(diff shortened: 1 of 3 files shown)


skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -52,6 +55,30 @@ class Metadata(TypedDict):  # noqa: D101
         predict_time_mean: float | None
 
 
+def slugify(string: str) -> str:
```
rouk1, line 58: Questions ^^

- Why KC and not KD ?
- Does this strips unicode exotic stuff (emoji, non latin alphabet, ...)

For ref here is unicode only implementation taken from django.

```python

def slugify(value):
    value = str(value)
    value = (
        unicodedata.normalize("NFKD", value)
        .encode("ascii", "ignore")
        .decode("ascii")
    )
    value = re.sub(r"[^\w\s-]", "", value.lower())
    return re.sub(r"[-\s]+", "-", value).strip("-_")
```

skore-hub-project/src/skore_hub_project/project/project.py
```diff
@@ -136,10 +163,25 @@ def __init__(self, tenant: str, name: str):
         name : str
             The name of the project.
         """
-        self.created = False
+        if (slugified_tenant := slugify(tenant)) != tenant:
+            warnings.warn(
+                f"Your project will be addressed under the '{slugified_tenant}' tenant."
```
rouk1:
```suggestion
                f"Your workspace will be addressed under the '{slugified_tenant}' tenant."
```

skore-hub-project/tests/unit/project/test_project.py
```diff
@@ -78,41 +78,123 @@ def monkeypatch_table_report_representation(monkeypatch):
...
+            ("mytënant", "mytënant", False),
+            ("my.tenant", "my.tenant", False),
+            ("my-tenant", "my-tenant", False),
+            ("my_tenant", "my_tenant", False),
+            ("my tenant", "my-tenant", True),
+            ("my/tenant", "my-tenant", True),
+            ("my:tenant", "my-tenant", True),
+            ("my?tenant", "my-tenant", True),
+            ("my#tenant", "my-tenant", True),
+            ("my/:?#tënant", "my-tënant", True),
```
rouk1: slugify("")
```suggestion
            ("my/:?#tënant", "my-tënant", True),
            ("👽👾😸👨 ? # @ hello bßß œπ æµ∂ÏƒÏ", "'hello-b-ii'", True),
```

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: > Why KC and not KD ?

https://github.com/django/django/blob/b1ffa9a9d78b0c2c5ad6ed5a1d84e380d5cfd010/django/utils/text.py#L469

I assumed the hub accepted unicode characters, so as django, i used NFKC.

> Does this strips unicode exotic stuff (emoji, non latin alphabet, ...)

All characters that are not unicode characters, `.`, `-` or `_` are replaces by `-`.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: With this change,  you are mixing workspace notion and tenant, and you are saying in a way `tenant is under tenant`. I'm not convinced.

But fyi, in another PR i'm renaming tenant to workspace https://github.com/probabl-ai/skore/pull/2324.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: Ok so after an IRL discussion, we decided to allow only ASCII characters, even if Unicode characters are quoted using `urllib` before being sent as path segments. Otherwise it would implies the hub's front to do the same quoting.

thomass-dev: Fixed.

skore-hub-project/tests/unit/project/test_project.py
thomass-dev: Fixed.

PR #2322, opened 2026-01-22
feat: Extend .frame() API with format and query parameters

Premkumar-2004:
Extended `CoefficientsDisplay.frame() `with three new parameters:

`format ("long" | "wide" | "auto") `- Controls output shape

`long` (default): Original format, one row per coefficient
`wide`: Features as rows, labels/splits as columns
`auto`: Wide for EstimatorReport/CV, long for ComparisonReport
`aggregate (bool)` - For CV reports, shows `mean ± std `instead of individual splits

`query (dict)` - Filter by column values, e.g., `{"label": "setosa"}`

### DIFF
--- a/skore/src/skore/_sklearn/_plot/inspection/coefficients.py
+++ b/skore/src/skore/_sklearn/_plot/inspection/coefficients.py
@@ -83,7 +83,7 @@ class CoefficientsDisplay(DisplayMixin):
     }
 
     def __init__(self, *, coefficients: pd.DataFrame, report_type: ReportType):
-        if report_type not in ReportType.__args__:
+        if report_type not in ReportType.__args__:  # type: ignore
             raise TypeError(f"Unexpected report type: {report_type!r}")
 
         self.coefficients = coefficients
--- a/examples/use_cases/plot_feature_importance.py
+++ b/examples/use_cases/plot_feature_importance.py
@@ -452,33 +452,10 @@ print("Number of features after feature engineering:", n_features_engineered)
 # Let us display the 15 largest absolute coefficients:
 
 # %%
-<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< conflict 1 of 1
-+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++ ywulqxoo 5f4e2726 "feat: Extend .frame() API with format and query parameters" (rebased revision)
 engineered_ridge_report_coefficients = (
-    engineered_ridge_report.inspection.coefficients()
-    .frame()
-%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% diff from: znkxumpv 6faaf1f4 "Merge branch 'main' into issue-2161" (parents of rebased revision)
-\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\        to: rytokvlt e9252b12 "feat: Extend .frame() API with format and query parameters" (rebased revision)
- engineered_ridge_report_feature_importance = (
-     engineered_ridge_report.feature_importance.coefficients()
--    .frame()
-+    .frame(format="long")
-%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% diff from: zlqnrlzs b459ed23 "chore(dependencies): GITHUB-ACTIONS: Bump actions/checkout from 6.0.1 to 6.0.2 (#2349)"
-\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\        to: uyxuplwy 5eeef3f8 "ci: Increase the timeout on `pytest` workflow (#2344)"
- engineered_ridge_report_feature_importance = (
--    engineered_ridge_report.inspection.coefficients()
-+    engineered_ridge_report.feature_importance.coefficients()
-     .frame()
-%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%% diff from: pontloqt fe68d3ca "feat: Extend .frame() API with format and query parameters"
-\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\\        to: zqzptpvm 973258b5 "fix doctest" (rebased revision)
- engineered_ridge_report_feature_importance = (
--    engineered_ridge_report.feature_importance.coefficients()
-+    engineered_ridge_report.inspection.coefficients()
-     .frame(format="long")
->>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>> conflict 1 of 1 ends
-    .set_index("feature")
-    .sort_values(by="coefficient", key=abs, ascending=True)
-    .tail(15)
+    engineered_ridge_report.inspection.coefficients().frame(
+        select_k=15, sorting_order="descending"
+    )
 )
 
 engineered_ridge_report_coefficients.index = (


Premkumar-2004:
so i have modified the code as per your suggestion can you please review it now if still any changes required please suggest

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -84,75 +98,97 @@ def __init__(self, *, coefficients: pd.DataFrame, report_type: ReportType):
...
+            agg["value"] = agg.apply(
+                lambda r: f"{r['mean']:.3f} ± {r['std']:.3f}", axis=1
+            )
+            df = agg["value"].reset_index()
+            df.columns = list(group_cols) + ["coefficients"]
+            has_split = False
+
+        index = ["feature"] if not has_estimator else ["estimator", "feature"]
+        columns = (
+            [label_col, "split"] if has_split else [label_col] if has_label else None
```
GaetandeCast:
```suggestion
        index = ["feature"] if not has_label else [label_col, "feature"]
        columns = (
            ["estimator", "split"] if has_split else ["estimator"] if has_estimator else None
```
Personal preference but I would rather have the label as a row index and the estimator as a column index.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -54,7 +54,21 @@ class CoefficientsDisplay(DisplayMixin):
     ... )
     >>> report = EstimatorReport(LogisticRegression(), **split_data)
     >>> display = report.feature_importance.coefficients()
-    >>> display.frame()
+
+    Get coefficients in wide format (default):
+
+    >>> display.frame()  # doctest: +NORMALIZE_WHITESPACE
```
GaetandeCast:
```suggestion
    >>> display.frame()
```
The `NORMALIZE_WHITESPACE` is not needed I think

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -54,7 +54,21 @@ class CoefficientsDisplay(DisplayMixin):
     ... )
     >>> report = EstimatorReport(LogisticRegression(), **split_data)
     >>> display = report.feature_importance.coefficients()
-    >>> display.frame()
+
+    Get coefficients in wide format (default):
```
GaetandeCast:
```suggestion
    Get coefficients in wide format (default). This output is made to be human
    readable.
```

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -54,7 +54,21 @@ class CoefficientsDisplay(DisplayMixin):
...
+    >>> display.frame()  # doctest: +NORMALIZE_WHITESPACE
+    label                   setosa  versicolor  virginica
+    feature
+    Intercept                 9.2...       1.7...    -11.0...
+    sepal length (cm)        -0.4...       0.5...     -0.1...
+    sepal width (cm)          0.8...      -0.2...     -0.5...
+    petal length (cm)        -2.3...      -0.2...      2.5...
+    petal width (cm)         -0.9...      -0.7...      1.7...
+
+    Get coefficients in long format:
```
GaetandeCast:
```suggestion
    Get coefficients in long format. This output is made to be easily usable in code for
    further processing or plotting.
```

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
GaetandeCast: Actually when there is no available variable to index columns, like in the case of the example in the docstring, it is nice to have them indexed by label. Otherwise I would prefer label indexing rows.

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -84,75 +98,97 @@ def __init__(self, *, coefficients: pd.DataFrame, report_type: ReportType):
...
             columns_to_drop = []
-        else:
-            raise TypeError(f"Unexpected report type: {self.report_type!r}")
 
         if self.coefficients["label"].isna().all():
-            # regression problem
             columns_to_drop.append("label")
         if self.coefficients["output"].isna().all():
-            # classification problem
             columns_to_drop.append("output")
```
GaetandeCast, line 269: These inline comments can stay IMO. They help understand clean code better.

GaetandeCast:
Also the CI is red because you need to adapt the existing tests. Mostly by setting `format="long"` I think, since all existing tests expect long format and the default changed. You should also add a test for each type of report that asserts the correct behaviour of this new feature. Let me know if you need help reading the output of the CI to understand what test fails and why. You should also look into how to run tests locally with `pytest` to not be reliant on the CI.

Premkumar-2004:
i have Added `format="long"` to existing tests,Removed descriptive comments from docstring examples,Updated pivot logic per reviewer's preference,can you please review again .

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -54,7 +54,21 @@ class CoefficientsDisplay(DisplayMixin):
...
+
+    >>> display.frame(format="long")
```
auguste-probabl: This line should be moved to the docstring of `.frame()`; the example in the docstring of `.frame()` should not be deleted

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -84,75 +98,99 @@ def __init__(self, *, coefficients: pd.DataFrame, report_type: ReportType):
...
+            df = agg["value"].reset_index()
+            df.columns = list(group_cols) + ["coefficients"]
+            has_split = False
+
+        if has_estimator:
+            index = ["feature"] if not has_label else [label_col, "feature"]
+            columns = ["estimator", "split"] if has_split else ["estimator"]
+        else:
+            index = ["feature"]
+            columns = [label_col, "split"] if has_split else [label_col] if has_label else None
```
auguste-probabl: I think something like this would be cleaner:

```suggestion
        index = ([label_col] if "label" in df else []) + ["feature"]
        columns = (
            (["estimator"] if "estimator" in df else [])
            + ([label_col] if "label" in df else [])
            + (["split"] if "split" in df else [])
        )
```

do check that the tests pass though

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -84,75 +98,99 @@ def __init__(self, *, coefficients: pd.DataFrame, report_type: ReportType):
...
+            )
+            df = df.query(conditions)
+
+        if format == "long":
+            return df.reset_index(drop=True)
+
+        # Wide format: pivot the dataframe
+        has_label = "label" in df.columns
+        has_split = "split" in df.columns
+        has_estimator = "estimator" in df.columns
```
auguste-probabl: I don't think we need these variables; for example, everywhere `has_label` is used you can use `"label" in df` (you don't need the `.columns`)

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -84,75 +98,99 @@ def __init__(self, *, coefficients: pd.DataFrame, report_type: ReportType):
...
         if not include_intercept:
-            coefficients = coefficients.query("feature != 'Intercept'")
-        return coefficients
+            df = df.query("feature != 'Intercept'")
+
+        if query is not None:
+            conditions = " and ".join(
+                f"`{col}` == {repr(val)}" for col, val in query.items()
+            )
+            df = df.query(conditions)
```
auguste-probabl: There need to be tests for this (`query`)

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -84,75 +98,99 @@ def __init__(self, *, coefficients: pd.DataFrame, report_type: ReportType):
...
+        if format == "long":
+            return df.reset_index(drop=True)
+
+        # Wide format: pivot the dataframe
+        has_label = "label" in df.columns
+        has_split = "split" in df.columns
+        has_estimator = "estimator" in df.columns
+        label_col = "label" if has_label else "output"
+
+        if aggregate and has_split:
```
auguste-probabl: There need to be tests for this (`aggregate`)

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
GaetandeCast: With this suggestion, "label" is used to index rows and columns when present which is not desirable.

Premkumar-2004:
Added tests for query and aggregate parameters, fixed numpy type issues in test assertions.  now you can review it

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -137,22 +156,63 @@ def frame(self, *, include_intercept: bool = True):
...
+            conditions = " and ".join(
+                f"`{col}` == {repr(val)}" for col, val in query.items()
+            )
+            df = df.query(conditions)
+
+        if format == "long":
+            return df.reset_index(drop=True)
+
+        # Wide format: pivot the dataframe
+        label_col = "label" if "label" in df else "output"
```
GaetandeCast:
```suggestion
        label_col = "label" if "label" in df else "output" if "output" in df else None
```
"output" is not always present.
Actually the fact that I caught this because a doctest failed but not a unit test shows that the feature is not properly tested. Please add tests to ensure the behaviour of the feature in all conditions: all report types, all ml tasks etc.

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -137,22 +156,63 @@ def frame(self, *, include_intercept: bool = True):
...
+
+        if aggregate and "split" in df:
+            group_cols = [c for c in df.columns if c not in ("split", "coefficients")]
+            agg = df.groupby(group_cols, sort=False)["coefficients"].agg(["mean", "std"])
+            agg["value"] = agg.apply(
+                lambda r: f"{r['mean']:.3f} ± {r['std']:.3f}", axis=1
+            )
+            df = agg["value"].reset_index()
+            df.columns = list(group_cols) + ["coefficients"]
+
```
GaetandeCast, line 298:
```suggestion

        columns: list[str] | None
```
To make mypy happy.

skore/tests/unit/displays/coefficients/test_cross_validation.py
```diff
@@ -408,11 +408,62 @@ def test_include_intercept(
...
+    X = _convert_container(X, "dataframe", columns_name=columns_names)
+    splitter = 2
+
+    report = CrossValidationReport(clone(estimator), X, y, splitter=splitter)
+    display = report.feature_importance.coefficients()
+
+    # Filter by a specific split
+    df_filtered = display.frame(format="long", query={"split": 0})
+    assert df_filtered["split"].nunique() == 1
+    assert df_filtered["split"].iloc[0] == 0
```
auguste-probabl:
```suggestion
    assert set(df_filtered["split"]) == {0}
```

skore/tests/unit/displays/coefficients/test_cross_validation.py
```diff
@@ -408,11 +408,62 @@ def test_include_intercept(
...
+
+    labels = np.unique(y)
+    target_label = int(labels[0])
+    df_label = display.frame(format="long", query={"label": target_label})
+    assert df_label["label"].nunique() == 1
+    assert df_label["label"].iloc[0] == target_label
```
auguste-probabl:
```suggestion
    assert set(df_label["label"]) == target_label
```

skore/tests/unit/displays/coefficients/test_estimator.py
```diff
@@ -377,11 +377,44 @@ def test_include_intercept(
...
+        clone(estimator), X_train=X_train, y_train=y_train, X_test=X_test, y_test=y_test
+    )
+    display = report.feature_importance.coefficients()
+
+    labels = np.unique(y_train)
+    target_label = int(labels[0])
+    df_filtered = display.frame(format="long", query={"label": target_label})
+
+    assert df_filtered["label"].nunique() == 1
+    assert df_filtered["label"].iloc[0] == target_label
```
auguste-probabl:
```suggestion
    assert set(df_filtered["label"]) == target_label
```

skore/tests/unit/displays/coefficients/test_estimator.py
```diff
@@ -377,11 +377,44 @@ def test_include_intercept(
...
+
+    # Filter by feature
+    df_feature = display.frame(format="long", query={"feature": "Intercept"})
+    assert df_feature["feature"].unique().tolist() == ["Intercept"]
```
auguste-probabl:
```suggestion
    assert set(df_feature["feature"]) == {"Intercept"}
```

Premkumar-2004:
Added `format="long"` to two `.frame() `calls  that were crashing because they used `.set_index("feature")`,
Fixed lint issues: Changed `repr(val)` to `val!r` Split long lines for `group_cols` and `agg`,
Fixed lint issues: Shortened docstring Removed unused `pytest` import

Premkumar-2004:
`test_frame_format.py` - Removed the separate test file as requested.The actual pytest tests all passed locally (126 tests).

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
auguste-probabl, line 1: it would make more sense to show the default behaviour in the doctests

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -137,22 +156,72 @@ def frame(self, *, include_intercept: bool = True):
             columns_to_drop = ["estimator"]
         elif self.report_type == "comparison-estimator":
             columns_to_drop = ["split"]
-        elif self.report_type == "comparison-cross-validation":
+        else:  # comparison-cross-validation
             columns_to_drop = []
-        else:
-            raise TypeError(f"Unexpected report type: {self.report_type!r}")
```
auguste-probabl, line 262: this was fine, can you put it back?

auguste-probabl: actually, it's a style thing. you can leave it

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
Premkumar-2004: The original instruction was to add format="long" to the tests because the existing tests were written expecting long format output, and the new default is wide.

auguste-probabl: This is true for the unit tests (those in `test_*.py` files), but here I am talking about the doctests, meaning the examples in the docstrings. Since these are examples of how the library works, it makes more sense to show the default behaviour

Premkumar-2004: Updated the doctests to show `display.frame()` with the default wide format output.

glemaitre:
I would prefer that we exclude the `aggregate` and `query` feature for the moment. I would like to tackle them in a separate PR.

Regarding the `format`, I would expect `"wide"` to not work in the case of comparing cross-validation report with different split numbers.

If we support this case (I would need to test to be sure), then I would prefer to have a `format="auto"` that would choose `format="wide"` in priority and when not possible fallback on the "long" format.

Premkumar-2004:
Just to clarify - i was previously asked  to add the aggregate parameter test in `test_comparison_cross_validation.py, `but now you want to exclude `aggregate` and `query` for a separate PR. Should I remove the aggregate and query tests I added, along with removing those parameters from the `frame()` method?

glemaitre:
@GaetandeCast mention that there is a tight job between `format` and `aggregate`. Let me check the entire PR before to change anything.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -137,22 +156,72 @@ def frame(self, *, include_intercept: bool = True):
...
 
-        coefficients = self.coefficients.drop(columns=columns_to_drop)
+        df = self.coefficients.drop(columns=columns_to_drop)
+
         if not include_intercept:
-            coefficients = coefficients.query("feature != 'Intercept'")
-        return coefficients
+            df = df.query("feature != 'Intercept'")
+
+        if query is not None:
```
glemaitre: Now, I'm more convinced that we should not have the query in this PR.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -137,22 +156,72 @@ def frame(self, *, include_intercept: bool = True):
...
+
+        if aggregate and "split" in df:
+            group_cols = [
+                c for c in df.columns if c not in ("split", "coefficients")
+            ]
+            agg = df.groupby(group_cols, sort=False)["coefficients"].agg(
+                ["mean", "std"]
+            )
+            agg["value"] = agg.apply(
+                lambda r: f"{r['mean']:.3f} ± {r['std']:.3f}", axis=1
```
glemaitre: this format is annoying because it is a string an not number. We should have two columns `model1_mean` and `model1_std`.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
```diff
@@ -137,22 +156,72 @@ def frame(self, *, include_intercept: bool = True):
...
+
+        if format == "long":
+            return df.reset_index(drop=True)
+
+        # Wide format: pivot the dataframe
+        label_col: str | None = (
+            "label" if "label" in df else "output" if "output" in df else None
+        )
+
+        if aggregate and "split" in df:
```
glemaitre: I find the code quite complext to read here. I would follow something more in line with what we propose here: https://github.com/probabl-ai/skore/pull/2304/changes#diff-8a105da835a535d76786c5191c495c4e9b94d7edf1fbbd0c0a28c3f323c17110R336

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
glemaitre: So to provide an example only considering the `aggregate` part, it would give something like:

```python
        if self.report_type == "estimator":
            columns_to_drop = ["estimator", "split"]
            group_by = []
        elif self.report_type == "cross-validation":
            columns_to_drop = ["estimator"]
            group_by = ["feature"]
        elif self.report_type == "comparison-estimator":
            columns_to_drop = ["split"]
            group_by = []
        else:  # comparison-cross-validation
            columns_to_drop = []
            group_by = ["estimator", "feature"]

        if aggregate is not None and not group_by:
            raise ValueError

        if self.coefficients["label"].isna().all():
            columns_to_drop.append("label")
        else:
            group_by.append("label")
        if self.coefficients["output"].isna().all():
            columns_to_drop.append("output")
        else:
            group_by.append("output")

        frame = self.coefficients.drop(columns=columns_to_drop)

        if not include_intercept:
            frame = frame.query("feature != 'Intercept'")

        if aggregate is not None:
            frame = (
                frame.drop(columns=["split"])
                .groupby(group_by, sort=False)
                .aggregate(aggregate)
            ).reset_index()

            if isinstance(frame.columns, pd.MultiIndex):
                frame.columns = flatten_multi_index(frame.columns)
        return frame
```

Then, it means that we can make the pivoting after the potential aggregation.

Premkumar-2004:
2 sphinx CI fail because plot_getting_started.py uses .drop(columns=["split"]) which requires long format. Fixed by adding format="long" at lines 191 and 325.

skore/src/skore/_sklearn/_plot/feature_importance/coefficients.py
glemaitre: So for the wide format, it could go after the `aggregate` and I think that we can have:

```python
        if format == "wide":
            for col in ("split", "label", "output"):
                if col in frame.columns:
                    frame[col] = frame[col].apply(lambda x: f"{col}_{x}")
            index = ["feature"]
            values = frame.columns[
                frame.columns.str.startswith("coefficients")
            ].tolist()
            columns = frame.columns[~frame.columns.isin(index + values)].tolist()
            frame = frame.pivot_table(
                index=index,
                columns=columns,
                values=values,
                aggfunc="first",
                sort=False,
            )
            if isinstance(frame.columns, pd.MultiIndex):
                frame.columns = flatten_multi_index(frame.columns)
```

glemaitre: So to wrap-up I think that the implementation should look like:

```python
        if self.report_type == "estimator":
            columns_to_drop = ["estimator", "split"]
            group_by = []
        elif self.report_type == "cross-validation":
            columns_to_drop = ["estimator"]
            group_by = ["feature"]
        elif self.report_type == "comparison-estimator":
            columns_to_drop = ["split"]
            group_by = []
        else:  # comparison-cross-validation
            columns_to_drop = []
            group_by = ["estimator", "feature"]

        if aggregate is not None and not group_by:
            raise ValueError

        if self.coefficients["label"].isna().all():
            columns_to_drop.append("label")
        else:
            group_by.append("label")
        if self.coefficients["output"].isna().all():
            columns_to_drop.append("output")
        else:
            group_by.append("output")

        frame = self.coefficients.drop(columns=columns_to_drop)

        if not include_intercept:
            frame = frame.query("feature != 'Intercept'")

        if aggregate is not None:
            frame = (
                frame.drop(columns=["split"])
                .groupby(group_by, sort=False)
                .aggregate(aggregate)
            ).reset_index()

            if isinstance(frame.columns, pd.MultiIndex):
                frame.columns = flatten_multi_index(frame.columns)

        if format == "wide":
            if self.report_type.startswith(
                "comparison"
            ) and not self._has_same_features(frame=frame):
                raise ValueError
            for col in ("split", "label", "output"):
                if col in frame.columns:
                    frame[col] = frame[col].apply(lambda x: f"{col}_{x}")
            index = ["feature"]
            values = frame.columns[
                frame.columns.str.startswith("coefficients")
            ].tolist()
            columns = frame.columns[~frame.columns.isin(index + values)].tolist()
            frame = frame.pivot_table(
                index=index,
                columns=columns,
                values=values,
                aggfunc="first",
                sort=False,
            )
            if isinstance(frame.columns, pd.MultiIndex):
                frame.columns = flatten_multi_index(frame.columns)

        return frame
 ```

**Comment by GaetandeCast** [comment]
I agree with the points you make @glemaitre and the solutions you propose.

To answer your question, the solution might be to use an `"auto"` option for the `format` parameter that uses wide when the number of columns is limited (I would say <= 2, it's still readable in that situation imo) and that switches to long otherwise. I would still want to let the user override that to force wide even if it's hard to read. 

One could also make a point that in every case, the wide format is easier (but not necessarily easy) to read than long and therefore should be used by default. 

Let me know what you think.

**Comment by Premkumar-2004** [comment]
ok then,i would modify the code accordingly

**Comment by Premkumar-2004** [comment]
Added format parameter to `frame()` with `"wide"` (default) and `"long"` options.
->Removed `query` and `aggregate` parameters as requested (deferred to separate PR)
->Wide format pivots features as rows with labels/splits as columns
->Added validation to raise error when using wide format with estimators having different features
->Updated doctests to show default behavior

please can you review it @glemaitre @GaetandeCast @auguste-probabl

**Comment by Premkumar-2004** [comment]
sorry for the small error that let the ci pipeline test cases fail ,now i fixed them all the test should pass now (all test cases passed locally)
please can you review it @glemaitre @GaetandeCast @auguste-probabl

**Comment by Premkumar-2004** [comment]
I've completed the changes you requested (removing `query` and `aggregate` parameters, keeping only the `format `parameter with `"wide"/"long" `options). However, when I tried to rebase with the latest `main `branch, I encountered significant conflicts due to recent upstream changes:

PR #2346 renamed f`eature_importance` to inspection
PR #2305 added `select_k` and `sorting_order `parameters to the `frame() `method
These changes conflict with my implementation since the file path changed and the `frame() `method signature is now different.

Could you please advise on how to proceed? 
@GaetandeCast

**Comment by GaetandeCast** [comment]
@Premkumar-2004 I've solved merge conflicts for you. However there still are some unfinished discussions here. Please refrain from making too many changes and asking for reviews until we are settled on those. 

Let's tackle:
- the two comments from @glemaitre that are not resolved.
- if and how we should limit the wide format when there is too much information to display.
- how to interact with the newly added `sorting_order` parameter. For now I put the changes of this PR after the sorting and/or selecting of the features but it's probably not the best thing to do. For the long format, the current sorting makes sense but in wide format does sorting even make sense ? Maybe if we sort by the average coefficient value across columns. Also the `select_k` feature will break the pivot table when the selection is different across groups.

**Comment by Premkumar-2004** [comment]
For the wide format issues:
If there are too many columns(like if columns > 20), we can throw an `error `telling users to switch to `long `format.
For sorting in `wide `format, sorting by mean coefficient value across columns makes sense to me.
For `select_k and wide`, we should error out if different groups end up with different features since the pivot won't work

**Comment by auguste-probabl** [comment]
From a discussion with @GaetandeCast :

The most complex case we want to support is a ComparisonReport of CrossValidationReport:
```python
from skore import ComparisonReport, CrossValidationReport

X, y = load_iris(return_X_y=True)
estimator_1 = LogisticRegression(max_iter=10000, random_state=42)
estimator_2 = LogisticRegression(max_iter=10000, random_state=43)
comparison_report = ComparisonReport(
    [CrossValidationReport(estimator_1, X, y), CrossValidationReport(estimator_2, X, y)]
)
comparison_report.inspection.coefficients().frame()
```
with the output as of right now:
```
estimator  LogisticRegression_1                                         LogisticRegression_2  ...
split                         0         1         2         3         4                    0  ...
label                         0         0         0         0         0                    0  ...
feature                                                                                       ...
Intercept              9.862255  9.490178  9.203867  9.097505  9.657548             9.862255  ...
Feature #2            -2.350558 -2.357779 -2.395413 -2.364760 -2.364974            -2.350558  ...
Feature #3            -0.994196 -1.014208 -1.012199 -1.023142 -0.975954            -0.994196  ...
Feature #1             0.847198  0.858916  0.907798  0.941640  0.849863             0.847198  ...
Feature #0            -0.485465 -0.425253 -0.385390 -0.395168 -0.430971            -0.485465  ...
```

That table is hard to read because it's so wide, and also hard to query because the columns are a MultiIndex.

Some thoughts:
- When `format="wide"`, aggregate by split by default, like we do in `.metrics.summarize()`.
    - In the first iteration, force aggregation.
- Flatten MultiIndex, as already mentioned by @glemaitre.

More generally, we need to decide what `wide` is for:
- Prioritize readability (force more stuff, forbid more cases)? or
- Prioritize query-ability (more configurable, less readable by default)?

I think, along with @GaetandeCast, that we should go for readability. The customizability need is already addressed by `format="long"`.

So specifically for the Comparison[CV] case, it could look something like
```
estimator  LogisticRegression_1_label_0 LogisticRegression_2_label_0 LogisticRegression_1_label_1 ...
feature                                                                                           ...
Intercept              9.862255 ± 0.445             9.862255 ± 0.445             9.862255 ± 0.445 ...
Feature #2             9.862255 ± 0.445             9.862255 ± 0.445             9.862255 ± 0.445 ...
Feature #3             9.862255 ± 0.445             9.862255 ± 0.445             9.862255 ± 0.445 ...
Feature #1             9.862255 ± 0.445             9.862255 ± 0.445             9.862255 ± 0.445 ...
Feature #0             9.862255 ± 0.445             9.862255 ± 0.445             9.862255 ± 0.445 ...
```

There is a case to be made that the `±` thing goes a step too far, but it's much more readable this way.

As for the default value for `format`, I'm aligned with @GaetandeCast in thinking it should be `"wide"`. Displays are for displaying something meaningful and readable. For the most complex case, when the estimators have different features, it's a bit tough but I think we should make "wide" output the same thing as "long". The default parameters should not lead to an error if we can avoid it.

Later, the `query` and `aggregate` parameters can also help us get both: sane defaults, e.g. `aggregate=("mean", "std")`, but also customization for the user. This will be relevant e.g. when there are many classes.

@Premkumar-2004 Thanks for your work on this; I think it would be a bit faster to push on this PR to bring it to the finish line. Hope that's okay with you. Again your work is very much appreciated.

Premkumar-2004:
Sure, go ahead! Happy to help and thanks for the guidance on this. Let me know if you need anything else from my side.

Issue #1991, opened 2025-08-30
api(skore): Use keyword only argument for most of the cases


glemaitre:
### What would you like to say?

I see that we are not really consistent when it comes to requesting parameters: sometimes we use keywords only arguments (using the `*`) and sometimes we don't enforce it.

I think that in terms of API we should be consistent. I would push for the keyword-only arguments for the following reason:

- we made this change is scikit-learn a couple of years ago
- it makes the user code more explicit
- it bring of flexibility when adding new arguments to the signature without the need of deprecation or putting the argument last (because position do not matter anymore)

It does not mean that we cannot use position argument but it should only be restrain to the few first arguments of the signature where having them implicit does not impact the readability of the produce code.

To provide a concrete example, it would mean to change a signature from:

```python
    def analyze(
        self,
        data_source: Literal["train", "test", "all"] = "all",
        with_y: bool = True,
        subsample: int | None = None,
        subsample_strategy: Literal["head", "random"] = "head",
        seed: int | None = None,
    ) -> TableReportDisplay:
        """Plot dataset statistics.
```

to

```python
    def analyze(
        self,
        *,
        data_source: Literal["train", "test", "all"] = "all",
        with_y: bool = True,
        subsample: int | None = None,
        subsample_strategy: Literal["head", "random"] = "head",
        seed: int | None = None,
    ) -> TableReportDisplay:
        """Plot dataset statistics.
```

to avoid writing a cryptic call like:

```python
report.data.analyze("train", True, 1_000, "head", 0)
```

direkkakkar319-ops:
Hi @glemaitre i woud like to work on this, happy to audit the signatures and open a PR adding `* `   for keyword only enforcement across the codebase.

PR #2326, opened 2026-01-23
feat(skore): Create an EstimatorReport from a Comparison/CrossValidationReport

### LINKED ISSUE #2320: Retreive an `EstimatorReport` from a `CrossValidationReport`

In a typical ML workflow, such as the one presented in the reworked [`getting_started`](https://github.com/probabl-ai/skore/pull/2303/changes#diff-5531773835bbc8c38a516d05cf6a4bdb7bd04ab50a6d8f17c2a5f61bb0be77a4R258-R273), once I have chosen my model after comparing multiple ones using cross validated performance metrics, I'd like to make sure that the performance of my model holds up on previously unseen data. Currently in `skore`, this requires us to create a new instance of `EstimatorReport` by hand, with the right model, parameters and dataset.

It would be nice to have a way of creating an `EstimatorReport` from a `CrossValidationReport` that can be fed held out data, in one line. 

It would be fixed by implementing a new class method for `CrossValidationReport`, named e.g. `create_estimator_report()`, that creates a new instance of `EstimatorReport` using the model of the chosen `CrossValidationReport` and the full data set that was split for cross validation. It would also be (optionally) fed an external set of testing points, to see how the model performs when trained on the whole dataset and tested on unseen data.

Usage example:
```python
X_train, X_test, y_train, y_test = train_test_split(X, y)
bad_cv_report = CrossValidationReport(DummyClassifier(), X_train, y_train)
good_cv_report = CrossValidationReport(RandomForestClassifier(), X_train, y_train)
comparison_report = ComparisonReport([bad_cv_report, good_cv_report])
summary = comparison_report.metrics.summarize().frame()

# Notice that good_cv_report performs best
final_report = good_cv_report.create_estimator_report(X_test, y_test)
final_report.metrics.summarize().frame()

# Or
final_report = comparison_report.create_estimator_report("RandomForestClassifier", X_test, y_test)
final_report.metrics.summarize().frame()
```

_No response_

### Additional context

_No response_

GaetandeCast:
Closes #2320 

Adds `ComparisonReport.create_estimator_report()` and `CrossValidationReport.create_estimator_report()` methods to generate an estimator report for deployement/final evaluation.

### DIFF
--- a/examples/getting_started/plot_getting_started.py
+++ b/examples/getting_started/plot_getting_started.py
@@ -267,22 +267,16 @@ comparison.metrics.precision_recall().plot()
 # Now that we have chosen to deploy the linear model, we will train it on
 # the full experiment set and evaluate it on our held-out data: training on more data
 # should help performance and we can also validate that our model generalizes well to
-# new data.
+# new data. This can be done in one step with :meth:`~skore.ComparisonReport.create_estimator_report`.
 
 # %%
-from skore import EstimatorReport
 
-final_report = EstimatorReport(
-    simple_model,
-    X_train=X_experiment,
-    y_train=y_experiment,
-    X_test=X_holdout,
-    y_test=y_holdout,
-    pos_label="good",
+final_report = comparison.create_estimator_report(
+    name="Simple Linear Model", X_test=X_holdout, y_test=y_holdout
 )
 
 # %%
-# :class:`skore.EstimatorReport` has a similar API to the other report classes:
+# This returns a :class:`~skore.EstimatorReport` which has a similar API to the other report classes:
 
 # %%
 final_metrics = final_report.metrics.summarize()


skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -419,6 +419,62 @@ def get_predictions(
             for report in self.reports_.values()
         ]
 
+    def create_estimator_report(
+        self,
+        estimator_name: str,
```
glemaitre: Since we don't have any attribute called `estimator_name`, I'm thinking that we could shorten the name here.

```suggestion
        name: str,
```

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -419,6 +419,62 @@ def get_predictions(
...
+        X_test: ArrayLike | None = None,
+        y_test: ArrayLike | None = None,
+    ) -> EstimatorReport:
+        """
+        Create an estimator report from the comparison report.
```
glemaitre:
```suggestion
        """Create an estimator report from the comparison report.
```

glemaitre: Actually, we create the estimator by taking "one of" the estimator used in the comparison report.

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -419,6 +419,62 @@ def get_predictions(
             for report in self.reports_.values()
         ]
 
+    def create_estimator_report(
+        self,
```
glemaitre, line 424: I prefer keyword-only (we should probably work on https://github.com/probabl-ai/skore/issues/1991 to make it consistent).

```suggestion
        self,
        *,
```

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -419,6 +419,62 @@ def get_predictions(
...
+            Testing target.
+
+        Returns
+        -------
+        estimator_report : :class:`~skore.EstimatorReport`
+            The estimator report.
+        """
+        if estimator_name not in self.reports_:
+            raise ValueError(
+                f"Estimator name {estimator_name} not found in the comparison report."
```
glemaitre:
```suggestion
                f"Estimator name {estimator_name} not found in the comparison report."
                f"Available names are: {self.reports_.keys()}."
```

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -419,6 +419,62 @@ def get_predictions(
...
+
+        estimator_report = self.reports_[estimator_name]
+        return_report = EstimatorReport(
+            estimator_report.estimator,
+            X_train=np.concatenate(
+                [estimator_report._X_train, estimator_report._X_test]
+            ),
+            y_train=np.concatenate(
+                [estimator_report._y_train, estimator_report._y_test]
+            ),
```
glemaitre: If `X_train` is a dataframe, the call to `np.concatenate` will make it a NumPy array.

You will need instead to check whether you need to call `np.concatenate` or `pd.concat`.

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -419,6 +419,62 @@ def get_predictions(
...
+
+        X_test : {array-like, sparse matrix} of shape (n_samples, n_features) or None
+            Testing data. It should have the same structure as the training data.
+
+        y_test : array-like of shape (n_samples,) or (n_samples, n_outputs) or None
+            Testing target.
+
+        Returns
+        -------
+        estimator_report : :class:`~skore.EstimatorReport`
```
glemaitre: We can call the output `report`.

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -446,6 +446,41 @@ def get_predictions(
             for report in self.estimator_reports_
         ]
 
+    def create_estimator_report(
+        self, X_test: ArrayLike | None = None, y_test: ArrayLike | None = None
+    ) -> EstimatorReport:
+        """
+        Create an estimator report from the cross-validation report.
```
glemaitre:
```suggestion
        """Create an estimator report from the cross-validation report.
```

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -178,6 +178,7 @@ def __init__(
         self._y_test = y_test
         self._pos_label = pos_label
         self.fit_time_ = fit_time
+        self._parent_hash: np.int64 | None = None
```
glemaitre, line 181: we would need to create an issue for this new API since we could make sure that every report created from another (e.g. `ComparisonReport`) could reuse this key.

skore/tests/unit/reports/comparison/test_report.py
```diff
@@ -127,3 +129,107 @@ def test_pos_label_mismatch(report):
     err_msg = "Expected all estimators to have the same positive label."
     with pytest.raises(ValueError, match=err_msg):
         ComparisonReport(reports)
+
+
+def test_create_estimator_report_from_estimator_reports():
```
glemaitre: we need to parametrize by using numpy array and dataframe.

You can look in the test suite for `_convert_container` that is a utility from scikit-learn where we can just parametrize and convert the array into a dataframe or array (no-op in this case).

skore/tests/unit/reports/comparison/test_report.py
```diff
@@ -127,3 +129,107 @@ def test_pos_label_mismatch(report):
...
+        X_experiment, y_experiment, test_size=0.2, random_state=42, shuffle=False
+    )
+    comparison_report = ComparisonReport(
+        {
+            name: EstimatorReport(
+                est, X_train=X_train, X_test=X_test, y_train=y_train, y_test=y_test
+            )
+            for name, est in estimators.items()
+        }
+    )
```
glemaitre: Could we double check if there are already `fixture` in `conftest.py` that already define a comparison report.

At the end, we only want to extract a new estimator from it and it would reduce the code.

skore/tests/unit/reports/comparison/test_report.py
```diff
@@ -127,3 +129,107 @@ def test_pos_label_mismatch(report):
...
+    est_report_w_test = comparison_report.create_estimator_report(
+        "estimator_2", X_test=X_heldout, y_test=y_heldout
+    )
+
+    assert isinstance(est_report_w_test, EstimatorReport)
+    assert est_report_w_test._parent_hash == comparison_report._hash
+    assert np.array_equal(est_report_w_test.X_train, X_experiment)
+    assert np.array_equal(est_report_w_test.y_train, y_experiment)
+    assert np.array_equal(est_report_w_test.X_test, X_heldout)
+    assert np.array_equal(est_report_w_test.y_test, y_heldout)
```
glemaitre: Since we are going to have dataframe and numpy array, I'm thinking that computing the hash would be easier to compare probably.

skore/tests/unit/reports/comparison/test_report.py
```diff
@@ -127,3 +129,107 @@ def test_pos_label_mismatch(report):
...
+
+
+def test_create_estimator_report_from_cross_validation_reports():
```
glemaitre: I'll have the same remarks than above.

skore/tests/unit/reports/cross_validation/test_report.py
```diff
@@ -255,3 +256,31 @@ def test_no_y():
     instance"""
     report = CrossValidationReport(KMeans(), X=np.random.rand(100, 5))
     assert isinstance(report, CrossValidationReport)
+
+
+def test_create_estimator_report(forest_binary_classification_data):
```
glemaitre: And the same here.

skore/tests/unit/reports/comparison/test_report.py
GaetandeCast: The existing fixtures for comparison reports don't return the dataset unfortunately. Anyways since we now convert the type between array and dataframe, we need to build the comparison by hand.

skore/src/skore/_sklearn/_estimator/report.py
GaetandeCast: For `ComparisonReport`, should `parent_hash` be a list of all the hashes of the reports it was created with?

GaetandeCast:
I applied your suggestions @glemaitre, ready for a second round.

skore/src/skore/_sklearn/_estimator/report.py
glemaitre: maybe a `dict` because we could associate the name of the report as a key since we already have it.

PR #2277, opened 2026-01-09
feat(skore): Create a "custom" metrics registry under each report

thomass-dev:
> [!CAUTION]
> Just a notebook for reflection for now.

Points to cover:
- redefine `__call__` in each metric subclass to provide a more user-friendly API
    - equivalent of the current implementation with `r2` calling `_r2`
- show the docstr from the metric `r2 = R2(); help(r2)`
- show the helper
    - standard metrics
    - custom metrics
- deal with CrossValidation (only metrics)

---

Closes https://github.com/probabl-ai/skore-hub/issues/940.

PR #2331, opened 2026-01-23
chore(skore/estimator-report): Clean some parts of the metrics accessor

auguste-probabl:
In preparation for a larger refactor of `_MetricsAccessor`

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -453,12 +453,10 @@ class _MetricsAccessor(
 
             scores.append(score_df)
 
-        has_multilevel = any(
+        if any(
             isinstance(df, pd.DataFrame) and isinstance(df.index, pd.MultiIndex)
             for df in scores
-        )
-
-        if has_multilevel:
+        ):
             # Convert single-level dataframes to multi-level
             for i, df in enumerate(scores):
                 if hasattr(df, "index") and not isinstance(df.index, pd.MultiIndex):
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -459,7 +459,7 @@ class _MetricsAccessor(
         ):
             # Convert single-level dataframes to multi-level
             for i, df in enumerate(scores):
-                if hasattr(df, "index") and not isinstance(df.index, pd.MultiIndex):
+                if not isinstance(df.index, pd.MultiIndex):
                     if "regression" in self._parent._ml_task:
                         name_index = ["Metric", "Output"]
                     else:
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -553,10 +553,10 @@ class _MetricsAccessor(
 
             if isinstance(score, np.ndarray):
                 score = score.tolist()
-            if hasattr(score, "item"):
-                score = score.item()
+
             if isinstance(score, list) and len(score) == 1:
                 score = score[0]
+
             if isinstance(score, list) and "classification" in self._parent._ml_task:
                 score = dict(
                     zip(self._parent._estimator.classes_.tolist(), score, strict=False)
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -551,6 +551,9 @@ class _MetricsAccessor(
 
             score = metric_fn(y_true, y_pred, **kwargs)
 
+            if hasattr(score, "item"):
+                score = score.item()
+
             if isinstance(score, np.ndarray):
                 score = score.tolist()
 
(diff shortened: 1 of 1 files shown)


thomass-dev:
Please note that #2277 will result in significant changes to the metrics accessor.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -556,19 +551,17 @@ def _compute_metric_scores(
...
-
-        if isinstance(score, list):
-            if "classification" in self._parent._ml_task:
-                return dict(
+            if isinstance(score, list) and len(score) == 1:
+                score = score[0]
+            if isinstance(score, list) and "classification" in self._parent._ml_task:
+                score = dict(
                     zip(self._parent._estimator.classes_.tolist(), score, strict=False)
                 )
```
thomass-dev: Please use if/elif when the condition are exclusive. Same for your change on line 554.

thomass-dev (changes requested):
IMO, this PR looks like a PR that changes only the "styling code". This should be avoided when possible, or included in other changes when really needed.

Otherwise, we could do the same thing for all parts of the code...

auguste-probabl:
> IMO, this PR looks like a PR that changes only the "styling code". This should be avoided when possible, or included in other changes when really needed.

I did these changes while working on a larger refactor, and I wanted to make it as simple to review and merge as possible

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -556,19 +551,17 @@ def _compute_metric_scores(
...
 
-            if len(score) == 1:
-                return score[0]
+            self._parent._cache[cache_key] = score
+
```
GaetandeCast: to be honest I don't really see the point of this change. I'm not sure it makes stuff clearer.

GaetandeCast (commented):
I'm not against cleaning up the code even if it does not impact functionality. There is just one change I'm not convinced by.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: the point is that now the function looks like
```python
def f():
	cache_key = get_cache_key()
	if cache_key in cache:
		output = cache[cache_key]
	else:
		output = do_stuff()
		cache[cache_key] = output
	return output
```
whereas before it was
```python
def f():
	cache_key = get_cache_key()
	if cache_key in cache:
		output = cache[cache_key]
	else:
		output = do_stuff()
		cache[cache_key] = output

	output = actually_do_more_stuff(output)

	return output
```

The actual logic of the function was split into different places.

GaetandeCast: okay that makes sense

GaetandeCast:
So I'm ok with this, although I agree with @thomass-dev 's remark about exclusive if/if statements.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: ~@GaetandeCast @thomass-dev I've simplified the conditional but~ the conditions are not quite exclusive enough to go further. Please advise.

auguste-probabl: reverted as I misunderstood what `array.item()` does (I thought it gave you the element of a 1-element array, but it also converts numpy-typed numbers to native types).

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -558,19 +553,20 @@ def _compute_metric_scores(
...
+            if isinstance(score, list) and len(score) == 1:
+                score = score[0]
 
-        if isinstance(score, list):
-            if "classification" in self._parent._ml_task:
-                return dict(
+            if isinstance(score, list) and "classification" in self._parent._ml_task:
+                score = dict(
                     zip(self._parent._estimator.classes_.tolist(), score, strict=False)
                 )
```
```suggestion

                if len(score) == 1:
                    score = score[0]
                if "classification" in self._parent._ml_task:
                    score = dict(
                        zip(self._parent._estimator.classes_.tolist(), score, strict=False)
                    )
            elif hasattr(score, "item"):
                score = score.item() 
```

auguste-probabl: I don't think that will work because if score is `[1]` and we're doing classification then it will be converted to 1 at line 558 and then we'll go into line 561, which will fail because `score` is no longer a list.

GaetandeCast: then let's do `elif "classification" in ...`

auguste-probabl: done. i'm not convinced it's clearer tbh

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: @thomass-dev please give it another look

thomass-dev (approved):
LGTM now, thanks.

PR #2336, opened 2026-01-28
chore: Replace ad-hoc tracking with Google Tag Manager

tuscland:
We enhance our visitor tracking abilities by replacing ad-hoc configuration of trackers with a unified solution based on Google Tag Manager. Specifically, this modification has the effect of displaying a cookie consent banner, allowing the user to opt-in or out of usage tracking.


tuscland:
For some reason my commit is not considered signed, but is on my local machine.

Issue #2327, opened 2026-01-23
Refactor `skore` authentication by exposing a new `skore.login()` function


rouk1:
This new function should implement this flowchart:

```mermaid
flowchart TB
    A(["skore.login()"]) --> n14["Is skore-hub-project installed ?"]
    B{"is SKORE_API_KEY environement variable set ?"} --> C["YES"] & D["NO"]
    C --> n2["Use given API key or token to authenticate"]
    D --> n4["Display the link to the login flow. (Usefull in JupyterLite)<br>Also display a message. that says:<br>To avoid an interactive authentication process create an API key <u>here</u>."]
    n14 --> n15["YES"] & n16["NO"]
    n15 --> B
    n16 --> n17["Invite user to install skore-hub-project then exit."]
    n4 --> n18["Poll for succesfull auth and bearer token."]
    n18 --> n2

    n14@{ shape: diam}
    n4@{ shape: rect}
    n15@{ shape: rect}
    n16@{ shape: proc}
    n17@{ shape: proc}
    n18@{ shape: out-in}
```

- [ ] Remove CLI skore-hub-login & skore-hub-logout
- [ ] Do not persist authentication token to user's machines, it is now the responsibility of user to chose there login strategy

thomass-dev:
@rouk1 @glemaitre Are we clear that this new design requires users to always have a call to `skore.login()` in their script, even when using an API key?

Before, the authentication was invisible when detecting an API key. I suggest maintaining this behavior: when setting an API key in your environment, a call to `skore.login` in your script is not mandatory.

rouk1:
Yes ! If a user wants to store there report remotely they must call `skore.login`.

At least for now 😅

thomass-dev:
Ok, i'm really not convinced by this, because if a user wants to store a report in the hub, he must specify it in the project prototype anyway.

glemaitre:
I'm for calling `skore.login` as well as this is already a similar pattern in other library like `wandb`.

Issue #1845, opened 2025-06-18
feat(skore-hub-project): expose the equivalent of `skore-hub-login` CLI through `skore` Python API


glemaitre:
### What would you like to say?

Right now, we can connect to the hub by launching a terminal and executing the command `skore-hub-login`. This command is a Python entry point and we could expose in `skore` the login function directly.

The benefit would be that I can have a cell where I can launch the connection in the notebook or interactive environment without the need to launch an additional terminal.

Right now, it just mean that we need to expose the following import:

```python
from skore_hub_project.authentication.login import login
```

and add it in the documentation.

glemaitre:
The same should be done for the `logout` function

Issue #2330, opened 2026-01-23
hub:// prefix should not be the discriminent for remote project anymore


rouk1:
Currently prefixing a project name with `hub://` allow skore to manage a project as a remote project.
We want to remove this and rely on the mode kwarg that already exist.

> [!NOTE]
> Wether this kwarg should be renamed might be discuss here. (or later ?)

PR #2340, opened 2026-01-28
feat(skore)!: Refactor hub authentication

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/2327 and https://github.com/probabl-ai/skore/issues/1845.

Add a new `skore.login()` function that must be called at the top of your script, before interacting with the project API.

### DIFF
--- a/skore/src/skore/__init__.py
+++ b/skore/src/skore/__init__.py
@@ -1,14 +1,15 @@
 """Configure logging and global settings."""
 
-import logging
-import warnings
+from logging import INFO, NullHandler, getLogger
+from warnings import warn
 
-import joblib
+from joblib import __version__ as joblib_version
 from rich.console import Console
 from rich.theme import Theme
 
 from skore._config import config_context, get_config, set_config
 from skore._externals._sklearn_compat import parse_version
+from skore._login import login
 from skore._sklearn import (
     ComparisonReport,
     ConfusionMatrixDisplay,
@@ -30,33 +31,47 @@ from skore._utils._patch import setup_jupyter_display
 from skore._utils._show_versions import show_versions
 from skore.project import Project
 
+# Configure jupyter display for VS Code compatibility
+setup_jupyter_display()
+
+
+if parse_version(joblib_version) < parse_version("1.4"):
+    set_config(show_progress=False)
+    warn(
+        "Because your version of joblib is older than 1.4, some of the features of "
+        "skore will not be enabled (e.g. progress bars). You can update joblib to "
+        "benefit from these features.",
+        stacklevel=2,
+    )
+
+
 __all__ = [
-    "CrossValidationReport",
+    "CoefficientsDisplay",
     "ComparisonReport",
     "ConfusionMatrixDisplay",
+    "CrossValidationReport",
     "Display",
     "EstimatorReport",
+    "MetricsSummaryDisplay",
+    "PermutationImportanceDisplay",
     "PrecisionRecallCurveDisplay",
     "PredictionErrorDisplay",
     "Project",
     "RocCurveDisplay",
-    "MetricsSummaryDisplay",
-    "show_versions",
-    "train_test_split",
+    "TableReportDisplay",
     "config_context",
     "get_config",
+    "login",
     "set_config",
-    "TableReportDisplay",
-    "CoefficientsDisplay",
-    "PermutationImportanceDisplay",
+    "show_versions",
+    "train_test_split",
 ]
 
-logger = logging.getLogger(__name__)
-logger.addHandler(logging.NullHandler())  # Default to no output
-logger.setLevel(logging.INFO)
 
-# Configure jupyter display for VS Code compatibility
-setup_jupyter_display()
+logger = getLogger(__name__)
+logger.addHandler(NullHandler())  # Default to no output
+logger.setLevel(INFO)
+
 
 skore_console_theme = Theme(
     {
@@ -66,14 +81,5 @@ skore_console_theme = Theme(
     }
 )
 
-console = Console(theme=skore_console_theme, width=88)
 
-joblib_version = parse_version(joblib.__version__)
-if joblib_version < parse_version("1.4"):
-    warnings.warn(
-        "Because your version of joblib is older than 1.4, some of the features of "
-        "skore will not be enabled (e.g. progress bars). You can update joblib to "
-        "benefit from these features.",
-        stacklevel=2,
-    )
-    set_config(show_progress=False)
+console = Console(theme=skore_console_theme, width=88)
--- a/skore/tests/unit/test_init.py
+++ b/skore/tests/unit/test_init.py
@@ -1,5 +1,5 @@
-import importlib
-import sys
+from importlib import import_module
+from sys import modules
 from unittest.mock import Mock, patch
 
 from pytest import warns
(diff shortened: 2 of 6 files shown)


glemaitre:
In the prompt, could show a tiny URL instead of the long URL?

thomass-dev:
> In the prompt, could show a tiny URL instead of the long URL?

@auguste-probabl had some trouble when we wrapped the URL 😅 https://github.com/probabl-ai/skore/pull/2066.

glemaitre:
Basically, I would prefer to have a shorter URL instead of wrapping because as a naive user, I don't really care about the URL. If you give me a `https://skore.probabl.ai/auth/xxxx` I would be fine and it up to us to make the redirection.

I would not consider it as a blocker however for the moment because I have no clue how expensive is it to deal with the redirection.

thomass-dev:
> Basically, I would prefer to have a shorter URL instead of wrapping because as a naive user, I don't really care about the URL. If you give me a `https://skore.probabl.ai/auth/xxxx` I would be fine and it up to us to make the redirection.
> 
> I would not consider it as a blocker however for the moment because I have no clue how expensive is it to deal with the redirection.

@retromechs what is the complexity to alias the ory authorization URL on the hub-side, before sending it to the library? The hub will have to manage the redirection itself if we want to avoid an third-party service for security issue.

IMHO, it's not worth the cost.

retromechs:
> Basically, I would prefer to have a shorter URL instead of wrapping because as a naive user, I don't really care about the URL. If you give me a `https://skore.probabl.ai/auth/xxxx` I would be fine and it up to us to make the redirection.
> 
> I would not consider it as a blocker however for the moment because I have no clue how expensive is it to deal with the redirection.

I’m not in favor of using an external url shortener for authorization urls. They contain a unique token used in the authentication flow, and even if the risk is very low, sending it to a 3rd party service increases the security risk.

We could build an internal shortener (for example with a dedicated endpoint that redirects the short url to the ORY one), but that would require storing auth tokens in the db, which we don’t do today and isn’t great security wise. The dev effort is low, my concern is mainly about security.

glemaitre:
OK so let's go with the long URL then. Thanks @retromechs @thomass-dev

skore/src/skore/__init__.py
```diff
@@ -62,14 +81,59 @@
...
-if joblib_version < parse_version("1.4"):
-    warnings.warn(
-        "Because your version of joblib is older than 1.4, some of the features of "
-        "skore will not be enabled (e.g. progress bars). You can update joblib to "
-        "benefit from these features.",
-        stacklevel=2,
-    )
-    set_config(show_progress=False)
+
+def login(*, mode: Literal["local", "hub"] = "hub", **kwargs):
```
glemaitre: Could we move this into a `_auth` module instead of the `__init__.py` directly. We can just make the import here.

skore/src/skore/__init__.py
```diff
@@ -62,14 +81,59 @@
...
+    recommend that you set up an API key via [url](coming soon) and use it to log in.
+
+    .. rubric:: Local mode
+
+    Otherwise, it configures the credentials used to communicate with the ``local``
+    backend, persisting projects on the user machine.
+
+    Parameters
+    ----------
+    mode : Literal["local", "hub"], optional
```
glemaitre: Usually we have the following.

```suggestion
    mode : {"local", "hub"}, default="hub"
```

skore/src/skore/__init__.py
```diff
@@ -62,14 +81,59 @@
...
+        The mode of the storage backend to log in, default hub.
```
glemaitre:
```suggestion
        The mode of the storage backend to log in.
```

skore/src/skore/__init__.py
glemaitre: We don't modify any rst file. So I think that we should be updating the documentation as well. In this PR, should let's focus on the API reference. We will manage examples aside to not delay the merge.

skore/src/skore/__init__.py
```diff
@@ -62,14 +81,59 @@
...
+    **kwargs : dict
+        Extra keyword arguments passed to the login function, depending on its mode.
+
+        timeout : int, mode:hub only, optional
```
glemaitre:
```suggestion
        timeout : int, default=600
```

skore/src/skore/__init__.py
```diff
@@ -62,14 +81,59 @@
...
+            The maximum time in second before raising an error, default 600.
```
glemaitre:
```suggestion
            The maximum time in second before raising an error when
            communicating with the hub. Only available when `mode="hub"`.
```

skore/src/skore/__init__.py
```diff
@@ -62,14 +81,59 @@
...
+
+    See Also
+    --------
+    :class:`~skore.Project` :
+        Refer to the :ref:`project` section of the user guide for more details.
+    """
+    if mode == "local":
```
glemaitre: I assume that this PR does not touch the API for the moment. The choice between `mode = "hub"` and `remote = True` comes in another PR?

skore/src/skore/__init__.py
```diff
@@ -62,14 +81,59 @@
...
+        return
+
+    MODE = "hub"
```
glemaitre: Why do we use capital later in the scope of the function?

skore/tests/unit/test_init.py
glemaitre, line 1: I would then recall this one `test_auth.py`.

skore/src/skore/__init__.py
thomass-dev: It will be covered in another PR by https://github.com/probabl-ai/skore/issues/2330.

skore/src/skore/__init__.py
thomass-dev: Fixed in c80bdf6d.

skore/src/skore/__init__.py
thomass-dev: Yes i assumed that the documentation needs a dedicated PR.
As the documentation currently only uses local projects, it doesn't need to be modified.

skore/src/skore/__init__.py
thomass-dev: Fixed in dae9afa6.

skore/src/skore/__init__.py
thomass-dev: Fixed in dae9afa6.

skore/src/skore/__init__.py
thomass-dev: Fixed in dae9afa6.

skore/src/skore/__init__.py
thomass-dev: Fixed in dae9afa6.

skore/src/skore/__init__.py
thomass-dev: > Could we move this into a _auth module instead of the __init__.py directly. We can just make the import here.

Fixed in 8dc172f8

skore/tests/unit/test_init.py
thomass-dev: Fixed in 8dc172f8.

thomass-dev:
All comments have been addressed, ready for a next review round.

rouk1 (approved):
This looks good to me ! 👍🏻

PR #2344, opened 2026-01-29
ci: Increase the timeout on `pytest` workflow

thomass-dev:
The duration of the `skore` tests has increased significantly recently, from 2 to 4 minutes to 6 to 7 minutes.
The coverage increases it even more (>10 minutes) and reaches the timeout.

While waiting to investigate and/or optimize the tests, increase the timeout from 10 to 15 minutes.


thomass-dev:
cc @GaetandeCast

PR #2350, opened 2026-02-02
docs(skore/api): Replace TOC with summary table

auguste-probabl:
Edit the API landing page such that we can easily search the component available in `skore`.

### DIFF
--- a/sphinx/reference/index.rst
+++ b/sphinx/reference/index.rst
@@ -83,16 +83,6 @@ Inspection
      - Display for permutation feature importance
 
 
-General
-"""""""
-
-.. list-table::
-   :widths: 30 70
-
-   * - :class:`Display`
-     - Base class for all display objects
-
-
 Configuration
 ^^^^^^^^^^^^^
 


sphinx/reference/index.rst
```diff
@@ -9,9 +9,125 @@ This page lists all the public functions and classes of the skore package.
...
+     - Set skore configuration values
+
+
+Utilities
+^^^^^^^^^
+
+.. list-table::
+   :widths: 30 70
+
+   * - :func:`show_versions`
```
GaetandeCast, line 106: There is no hyperlink here in the preview. `show_versions` should be added to one of the rst files, probably utils.rst

sphinx/reference/index.rst
```diff
@@ -9,9 +9,125 @@ This page lists all the public functions and classes of the skore package.
...
+
+.. list-table::
+   :widths: 30 70
+
+   * - :class:`EstimatorReport`
+     - Comprehensive report for evaluating a single scikit-learn estimator
+   * - :class:`CrossValidationReport`
+     - Report for evaluating an estimator using cross-validation
+   * - :class:`ComparisonReport`
+     - Report for comparing multiple estimator or cross-validation reports
```
glemaitre: I would merge those two sections and put the reports first and the function after. They are all helping at ML assistance.

sphinx/reference/index.rst
```diff
@@ -9,9 +9,125 @@ This page lists all the public functions and classes of the skore package.
...
+
+Project Management
+^^^^^^^^^^^^^^^^^^
+
+.. list-table::
+   :widths: 30 70
+
+   * - :class:`Project`
+     - Main class for managing a skore project and its reports
```
glemaitre: I would move the project first. I can think that in other examples than the getting started, we could always start by creating a project and then use it.

sphinx/reference/index.rst
```diff
@@ -9,9 +9,118 @@ This page lists all the public functions and classes of the skore package.
...
+     - Display for permutation feature importance
+
+
+General
+"""""""
+
+.. list-table::
+   :widths: 30 70
+
+   * - :class:`Display`
```
GaetandeCast: Is it relevant to advertise the base class here ? I'm not sure what to think about it.

auguste-probabl: Probably not, i'll remove it

PR #2354, opened 2026-02-03
ci: Upload coverage to codecov

### LINKED ISSUE #1840: ci: Add codecov to better interpret python coverage
## Context

We would like to use a tool such as codecov to improve the interpretation of python test coverage.  
Here is an example of [output](https://github.com/scikit-learn/scikit-learn/pull/8206#issuecomment-279553465).  

## What would you like to say?

https://about.codecov.io/pricing/

Codecov is actually free with unlimited seats for open source software.  

## TODO  

- [ ] implement codecov

auguste-probabl:
Closes #1840

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -232,12 +232,11 @@ jobs:
         env:
           PACKAGE: ${{ matrix.package }}
 
-      - name: Upload coverage reports
+      - name: Upload coverage reports to Codecov
         if: ${{ matrix.coverage && (github.event_name == 'pull_request') }}
-        uses: actions/upload-artifact@b7c566a772e6b6bfb58ed0dc250532a479d7789f # v6.0.0
+        uses: codecov/codecov-action@671740ac38dd9b0130fbe1cec585b89eea48d3de # v5.5.2
         with:
-          name: ${{ matrix.package }}-coverage
-          path: ${{ matrix.package }}/coverage/
+          files: ${{ matrix.package }}/coverage/pytest-coverage.xml
 
   pytest:
     needs:


auguste-probabl:
To really test this, the PR needs to be merged to main; I need to test this in a copy of the repo.

thomass-dev (changes requested):
~~I don't think `codecov` will be able to write comments on the PR, because there is no write permission in the `pytest.yml` workflow to avoid security issue in a fork context (the GITHUB_TOKEN is restricted and there is no secrets inheritance).~~

~~I think you should instead use `codecov` in `.github/workflows/pr-display-code-coverage.yml` in replacement of `MishaKav/pytest-coverage-comment`.~~

auguste-probabl:
> I don't think `codecov` will be able to write comments on the PR

It just did though, is there something I'm not getting?

thomass-dev:
Indeed, i went too fast, it is an external bot that writes directly to the PR.

thomass-dev:
In that case, if codecov is in charge of displaying the coverage, please remove the `.github/workflows/pr-display-code-coverage.yml` workflow.

thomass-dev:
Note: please ensure that the coverage of each package (in our mono-repo/multi-package environment) is well managed/identified.

auguste-probabl:
@thomass-dev I did some experimenting in a copy of the repo, because I need the Codecov step to have run on `main` to get the diff for a PR. Here is what I did in main: https://github.com/auguste-probabl/skore-copy/compare/4091b54ed62ba0a978cebf341ffe0ad343c4af7e...main

And then I opened a PR: https://github.com/auguste-probabl/skore-copy/pull/3

But it's not necessarily iso with our current setup, with the segmentation between maintainers and external contributors, etc.

Would you mind reviewing what I did on my copy here? Do you think this would work for us? And also we'll have to get a token from Codecov.

auguste-probabl:
Superseded by #2929

PR #2355, opened 2026-02-03
chore(skore): Reduce test run time

### LINKED ISSUE #2351: CI: reduce the runtime of the test suite

The run time of the test suite has multiplied by 2 to 3 times in the past few months, see: [tests with coverage taking 5 minutes in november](https://github.com/probabl-ai/skore/pull/2125) when they now take over 10 minutes. That forced us increase the timeout of the pytest workflow in #2344. 

Since November, the bulk of the changes in the tests is linked to changes in the displays of skore: extension of `ConfusionMatrixDisplay`, addition of `PermutationImportanceDisplay`, refactor of other misc. display to use seaborn, addition of new features (`subplot_by`, `data_source="both"`, `sorting_order` & `select_k` ) etc.

This made the test suite grow in a disorganized way and potentially leading to the observed slowdown of the testing jobs.

Here are the some leads to investigate:
- I noticed that the tests for the `CoefficientsDisplay` take a significant amount of time to run compared to other displays.
Ex on my machine: `pytest skore/tests/unit/displays/coefficients` takes 62 seconds vs 9.3 seconds for `skore/tests/unit/displays/roc_curve`
- Using parametrized tests and optimized fixtures when possible should optimize the runtime of many tests and make them easier to maintain in the future.
- It is possible that part of the slowdown is due to `seaborn` being slower than pure matplotlib. This is just an intuition.

If anyone has any suggestions, remarks or insights on this matter, feel free to post them here.

GaetandeCast:
Fixes #2351 & reverts #2344.

The general ideas of this refactoring to speed up the test suite are the following:
- Reduce the number of features in the fixtures. This greatly increases the speed of the computations.
- Create new fixtures for each of the different reports and that are persisted for the whole display module. Calling an accessor does not modify the report object so this is fine so long as we are careful to not modify the attributes of the report in the tests. When needed, e.g. to test cache behavior, a report has to be created manually.
These fixtures use dummy models by default for speed which can be changed to forests or linear models by overwriting the `estimator_type` fixture in the module's `conftest.py`.  
- Create display level fixtures that are also persisted across the module and that precompute the default appearance of the display plot. Calling `Display.plot()` is by far the slowest action in the tests so the idea is to do it as little as possible and re-use the pre-computed default look as much as possible. Of course to test the behavior of e.g. the `subplot_by` argument, one must call `Displat.plot()` 

Additionally, this PR refactors the general organization of the tests in each display modules to follow what was introduced for confusion matrix: a `test_common.py` file that tests all general functionality, parametrized for all report types and ml tasks and specialized `test_*_report.py` and `test_comparison_*.py` files for tests that would be too complicated to parametrize e.g. those that test `subplot_by`. 
This should make the test suite easier to read, maintain and expand.
 
If you have suggestions or disagree with some of the choices made here, please let me know.

### DIFF
--- a/skore/tests/unit/displays/confusion_matrix/conftest.py
+++ b/skore/tests/unit/displays/confusion_matrix/conftest.py
@@ -5,6 +5,11 @@ mpl.rc("figure", max_open_warning=False)
 
 
 @pytest.fixture(scope="module")
+def estimator_type():
+    return "linear"
+
+
+@pytest.fixture(scope="module")
 def estimator_reports_binary_classification_figure_axes(
     pyplot, estimator_reports_binary_classification
 ):
--- a/skore/tests/unit/displays/confusion_matrix/test_common.py
+++ b/skore/tests/unit/displays/confusion_matrix/test_common.py
@@ -93,10 +93,12 @@ class TestConfusionMatrixDisplay:
             "data_source",
         ]
         n_classes = len(display.display_labels)
-        n_thresholds = len(display.thresholds)
-        n_splits = 2 if "cross_validation" in fixture_prefix else 1
-        n_reports = 2 if "comparison" in fixture_prefix else 1
-        expected_rows = n_thresholds * n_classes * n_classes * n_splits * n_reports
+        expected_rows = 0
+        for _, group in display.confusion_matrix.groupby(
+            ["split", "estimator"], dropna=False
+        ):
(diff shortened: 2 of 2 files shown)


thomass-dev (approved):

thomass-dev:
Thanks for this huge work @GaetandeCast!!!

Issue #2353, opened 2026-02-02
bug(skore-hub-project): Flaky test with metrics thread computing


thomass-dev:
https://github.com/probabl-ai/skore/actions/runs/21593410998/job/62219176829

```python-traceback
================================== FAILURES ===================================
_______ TestCrossValidationReportPayload.test_model_dump_classification _______

self = <tests.unit.report.test_cross_validation_report.TestCrossValidationReportPayload object at 0x0000013037E3A0D0>
small_cv_binary_classification = CrossValidationReport(estimator=RandomForestClassifier(random_state=42), ...)
payload = CrossValidationReportPayload(key='<key>', groups=None, estimator_class_name='RandomForestClassifier', dataset_fingerpr...ckle=Pickle(content_type='application/octet-stream', checksum='blake3-HAz+yoDboHjBs83NuDn8jVR7ILDJMrg/MMPl1Pgo4tE='))])

    @mark.filterwarnings(
        # ignore precision warning due to the low number of labels in
        # `small_cv_binary_classification`, raised by `scikit-learn`
        "ignore:Precision is ill-defined.*:sklearn.exceptions.UndefinedMetricWarning",
        # ignore deprecation warnings generated by the way `pandas` is used by
        # `searborn`, which is a dependency of `skore`
        "ignore:The default of observed=False is deprecated.*:FutureWarning:seaborn",
    )
    @mark.usefixtures("monkeypatch_artifact_hub_client")
    @mark.usefixtures("monkeypatch_upload_routes")
    def test_model_dump_classification(self, small_cv_binary_classification, payload):
        _, checksum = serialize(small_cv_binary_classification)
    
>       payload_dict = payload.model_dump()
                       ^^^^^^^^^^^^^^^^^^^^

tests\unit\report\test_cross_validation_report.py:392: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
venv\Lib\site-packages\pydantic\main.py:464: in model_dump
    return self.__pydantic_serializer__.to_python(
C:\hostedtoolcache\windows\Python\3.11.9\x64\Lib\functools.py:1001: in __get__
    val = self.func(instance)
          ^^^^^^^^^^^^^^^^^^^
venv\Lib\site-packages\skore_hub_project\report\report.py:127: in metrics
    task.result()
[...]

Issue #1792, opened 2025-06-04
Limitation of the current implementation of progress bar


glemaitre:
### What would you like to say?

Currently, the progress bars are implemented by using a decorator on the method of a report. When this report has underlying report, the progress bar manager is passed to those nested reports such that they can attach there own progress bar.

This pattern is problematic in the case that the execution around the nested reports is using a parallel loop: the nested reports will be pickled but the progress bar manager contain a lock that cannot be pickled. The temporary workaround is to disable the progress bars in this particular case of the nested reports.

An alternative implementation that would work in all cases would be as designed in scikit-learn: https://github.com/scikit-learn/scikit-learn/pull/28760. We would need the manager to live in its own thread with shared memory. The reports will then push event into a queue in the manager and the manager itself is in charge of updating the display. It avoids the pickling issue because the manager is not attached to the reports anymore.

However, it required a good amount of work to make this implementation and it should not be straightforward.

auguste-probabl:
May be closed by #2357

PR #2357, opened 2026-02-03
fix(skore): Make the progress thread-safe

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/2353 and https://github.com/probabl-ai/skore/issues/1792.

---

The current problem is that, when using multi-threading to compute the metrics/displays of a report, two threads can work on the same progress-bar. At the end of the first thread, the progress bar is deleted before the second thread has finished. If the second thread tries to access the progress bar, an exception is raised -> thread-safety issue.

Refactor and simplify the progress-bar system to make it thread-safe: remove decorator in favor of a dedicated class, with only one dedicated task per progress bar. This progress-bar is directly instantiated where we want, instead of being a member variable at the report level that is managed by the decorator.

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -35,7 +35,7 @@ dependencies = [
   "numpy",
   "pandas",
   "plotly",
-  "rich>=14",
+  "rich>=14.1",
   "scikit-learn",
   "skore[local]",
   "skrub",


skore/tests/unit/utils/test_progress_bar.py
```diff
@@ -7,73 +10,91 @@ def test_standalone_progress():
...
     result = task.run()
 
     assert result == "done"
-    assert task._standalone_n_calls == 4
-    assert task._progress_info is None
+
+    assert task.progress.description == "Standalone"
+    assert task.progress.total == 2
+
+    assert task.progress._ProgressBar__progress.tasks[0].description == "Standalone"
```
auguste-probabl: Please, just call it `_progress`. Double underscore is overkill.

skore/src/skore/_utils/_progress_bar.py
```diff
@@ -12,12 +12,63 @@
...
+    transient=True,
+    disable=(not get_config()["show_progress"]),
+)
+
+
+class ProgressBar:
+    """Simplified progress bar based on ``rich.Progress``."""
+
+    def __init__(self, description=None, total=None):
+        self.__description = description
```
auguste-probabl: single underscores please.

skore/src/skore/_utils/_progress_bar.py
```diff
@@ -38,72 +89,11 @@ def progress_decorator(
     def decorator(func: Callable[..., T]) -> Callable[..., T]:
```
auguste-probabl: Can you document that `func` needs to take `progress: ProgressBar` as an argument for the decorator to work? Maybe in the typing, or by catching the exception?

auguste-probabl:
@jeremiedbb Making a mental note for us to talk about this.

skore/src/skore/_utils/_progress_bar.py
thomass-dev: I didn't get good typing even with ChatGPT... I added an assert and docstr.
Should be fixed in 6e4dbd6b.

skore/tests/unit/utils/test_progress_bar.py
thomass-dev: We're not going to have this discussion again (the conclusion was that we continue to use `__`), but honestly i don't understand.
It's not overkill, it's just a way to "privatize" member variables provided by python itself.

To respect the `skore` convention, i replaced all `__` by `_` in a343c664.
However, i will continue to use the `__` in other packages.

skore/src/skore/_utils/_progress_bar.py
auguste-probabl: assert is probably the best, it would be hard to type this. Although, if we want to priviledge speed maybe we can just put this in the docstring as a big `Note:` or something. I'll let you decide.

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -289,6 +287,8 @@ def cache_predictions(
             "auto", "predict", "predict_proba", "decision_function"
         ] = "auto",
         n_jobs: int | None = None,
+        *,
+        progress: ProgressBar,
```
glemaitre: It is weird for me to expose a `progress` publicly to the user. One should not care about the progress bar apart of setting the config. Would it be possible to have always a private method, e.g. `_cache_predictions` that take the `progress` parameter while the public one does not.

Internally, we can call the private to pass whatever we want and create a progress bar if we wish but the user will never face it.

thomass-dev: All other decorated functions have the same behavior you describe.
I will do the same here.

thomass-dev: I also asked myself another question yesterday: is the decorator still pertinent?
I could simply change it to 


```python
def my_function():
    [...]
    with ProgressBar(description="blabla", total=10) as progress:
        [...]
        progress.advance()
    [...]
```

thomass-dev: For instance

```
diff --git a/skore/src/skore/_sklearn/_estimator/report.py b/skore/src/skore/_sklearn/_estimator/report.py
index 30eac39b..37155abe 100644
--- a/skore/src/skore/_sklearn/_estimator/report.py
+++ b/skore/src/skore/_sklearn/_estimator/report.py
@@ -216,13 +216,10 @@ class EstimatorReport(_BaseReport, DirNamesMixin):
         """
         self._cache = Cache()
 
-    @progress_decorator(description="Caching predictions")
     def cache_predictions(
         self,
         response_methods: Literal["auto"] | list[str] = "auto",
         n_jobs: int | None = None,
-        *,
-        progress: ProgressBar,
     ) -> None:
         """Cache estimator's predictions.
 
@@ -288,15 +285,17 @@ class EstimatorReport(_BaseReport, DirNamesMixin):
         )
         # trigger the computation
         total_iterations = len(response_methods) * len(pos_labels) * len(data_sources)
-        progress.total = total_iterations
 
         # do not mutate directly `self._cache` during the execution of Parallel
         results_to_cache: dict[tuple[Any, ...], Any] = {}
-        for results in generator:
-            results_to_cache.update(
-                (key, value) for key, value, is_cached in results if not is_cached
-            )
-            progress.advance()
+
+        with ProgressBar(description="Caching predictions", total=total_iterations):
+            for results in generator:
+                results_to_cache.update(
+                    (key, value) for key, value, is_cached in results if not is_cached
+                )
+
+                progress.advance()

```

auguste-probabl: If it means we can take `progress` out of the parameter list then I'm all for it

thomass-dev: I totally removed the decorator in 24e3f2bc6c3ef2a7f0afaf304e213dab30827a85.

glemaitre:
I quickly try to serialize a `ComparisonReport` with the following:

```python
import joblib
from skore import ComparisonReport, CrossValidationReport
X, y = make_classification(random_state=42, n_samples=10_000, n_features=100)
estimator_1 = RandomForestClassifier()
estimator_2 = RandomForestClassifier()
report_1 = CrossValidationReport(estimator_1, X, y, n_jobs=2)
report_2 = CrossValidationReport(estimator_2, X, y, n_jobs=2)
report = ComparisonReport([report_1, report_2], n_jobs=2)
report.cache_predictions()
joblib.dump(report, "report.joblib")
```

So it solves the problem in https://github.com/probabl-ai/skore/issues/1792. The progress bar is also working as expected. So I'm pretty happy with what we have here.

skore/src/skore/_utils/_progress_bar.py
```diff
@@ -12,99 +13,73 @@
...
+        """Advance the progress bar by one step."""
+        self._progress.update(self._task, advance=1, refresh=True)
+
+
+def track(
+    sequence: Iterable[Any], description: str, total: float | None = None
+) -> Iterable[Any]:
+    """Track progress by iterating over a sequence."""
+    if total is None:
+        total = float(length_hint(sequence)) or None
```
auguste-probabl, line 80: Today I Learned that `length_hint` exists ^^' Honestly, I would just `len` or even require that `total` is not None. After all, `rich`'s progress bar can take `total=None`.  We tested `length_hint` on a generator and got 0, which is not desirable.

thomass-dev: This is a rich behavior: when `total` is not passed, and the total can't be automatically computed, the progress-bar renders a pulsing animation.

thomass-dev:
> We're looking over this with @jeremiedbb and we don't understand yet how it's possible for this to work for nested progress bars. Why don't you get the infamous `rich.LiveError` that we saw so often before?

There is no more `LiveError` since `14.0.0`:

https://rich.readthedocs.io/en/latest/live.html#nesting-lives

> Nesting Lives
> 
> If you create a Live instance within the context of an existing Live instance, then the content of the inner Live will be displayed below the outer Live.
> 
> Prior to version 14.0.0 this would have resulted in a LiveError exception.

Just test with a simple rich example (`track` uses `Live` instances) 😉

https://rich.readthedocs.io/en/latest/progress.html#nesting-progress-bars

```python
from rich.progress import track
from time import sleep


for count in track(range(10), transient=True):
    for letter in track("ABCDEF", transient=True):
        print(f"Stage {count}{letter}")

```

skore/pyproject.toml
auguste-probabl, line 1: Extra nitpick: The nested Progress capability was released in 14.1.0

PR #2359, opened 2026-02-03
docs(api): Add accessor function tables

### LINKED ISSUE #1407: Flatten the API doc page

The API page:
https://skore.probabl.ai/dev/reference/index.html

I don't see all the functions available, so I can't:
- Do a "crtl-F" to find a function
- Easily find the doc of a function that I know that I'm looking for, as I need to figure out in which subpage it is

### Suggested improvement

Fold in the subpages in the main API page

### Additional context

_No response_

auguste-probabl:
Adds collapsed-by-default tables containing the methods for every accessor of every Report class.
Text search still works when the dropdowns are collapsed (this is normal behavior in HTML dropdowns).

Automate the writing of tables with a Sphinx extension.

AI disclosure:
I used AI to generate the extension, and then heavily edited it.

Closes #1407


auguste-probabl:
Not sure if this is ready to merge.

You can see in the dropdown tables that some of the methods should be in the docs, but aren't, like `ComparisonReport.{roc,precision_recall,prediction_error}`. This is something we already talked about in https://github.com/probabl-ai/skore/pull/2319. Maybe we can reuse the automation here to address that too?

Also, https://github.com/probabl-ai/skore/pull/2316 already goes through every class to recover methods and attributes. It's a long shot as the two pieces of code have different targets (docs readers vs. notebook users), but maybe we could factorize some of the logic?

auguste-probabl:
~Ouch, the auto-generated table doesn't pass linting. I really want this to be automatic, so I'll do my best to make this work.~

On second thought, maybe this file could go in `reference/api` and not be checked into the repo? It'll be regenerated every time anyway.

sphinx/_templates/accessor_summary.rst
auguste-probabl, line 1: Maybe this jinja template can go directly in the script, to keep everything together.

sphinx/sphinxext/generate_accessor_tables.py
```diff
@@ -0,0 +1,151 @@
...
+def setup(app: Sphinx) -> dict[str, Any]:
+    """Setup the extension."""
+    app.add_config_value("accessor_summary_classes", [], "html")
+    app.connect("config-inited", generate_accessor_tables)
+
+    return {
+        "version": "0.1",
+        "parallel_read_safe": True,
+        "parallel_write_safe": True,
+    }
```
auguste-probabl, line 181: Discovered this: I think we should make our other scripts follow this same convention.

sphinx/conf.py
```diff
@@ -39,9 +39,17 @@
...
+    "generate_accessor_tables",
 ]
 exclude_patterns = ["build", "Thumbs.db", ".DS_Store"]
 
+# Configuration for generate_accessor_tables extension
+accessor_summary_classes = [
+    "skore.EstimatorReport",
+    "skore.CrossValidationReport",
+    "skore.ComparisonReport",
+]
```
auguste-probabl, line 51: Not 100% sure this config should look like this, but it works for now.

sphinx/_templates/accessor_summary.rst
```diff
@@ -0,0 +1,14 @@
+{% for class_info in classes %}
```
glemaitre: I really like it. Do you think it is possible to actually document the class as the first item and to have a block: main class followed by all the accessors as you have now and remove the table that we have above and only keep the function.

sphinx/_templates/accessor_summary.rst
```diff
@@ -0,0 +1,14 @@
...
+.. dropdown:: {{ class_info.name }} accessors
+   :icon: chevron-down
+   {% for accessor_name, accessor_data in class_info.accessors.items() %}
+   **{{ accessor_name.capitalize() }}**: :class:`{{ class_info.name }}.{{ accessor_name }}`
+
+   .. list-table::
+      :widths: 30 70
+      {% for method_name, method_doc in accessor_data.methods %}
+      * - :func:`~{{ class_info.name }}.{{ accessor_name }}.{{ method_name }}`
```
glemaitre: We have an issue for the display function:

Image

The `roc`, `precision_recall`, etc. are not appearing with an hyperling. I assume something is wrong with the link that is created.

glemaitre (commented):
As a summary I really like what you did. I think that it can get through quickly it is a net improvement.

sphinx/_templates/accessor_summary.rst
auguste-probabl: I think it's because we haven't actually added these methods to the docs:

https://github.com/probabl-ai/skore/blob/4091b54ed62ba0a978cebf341ffe0ad343c4af7e/sphinx/reference/report/comparison_report.rst

GaetandeCast (commented):
This looks great! 

One nitpick I have is that `train_test_split` feels a bit out of place now, stuck between the classes and accessors. Maybe it's just me.

Image

sphinx/_templates/accessor_summary.rst
glemaitre: Oh it could be because it does trigger building the API reference then. It is a nice catch then.

sphinx/sphinxext/generate_accessor_tables.py
```diff
@@ -0,0 +1,151 @@
...
+        logger.info(f"Collected accessor data for {class_name}")
+
+    template_dir = Path(app.confdir) / "_templates"
+    env = Environment(loader=FileSystemLoader(str(template_dir)))
+    template = env.get_template("accessor_summary.rst")
+
+    rst_content = template.render(classes=classes_data)
+
+    (Path(app.srcdir) / "reference" / "api").mkdir(exist_ok=True)
+    output_path = Path(app.srcdir) / "reference" / "api" / "accessor_tables.rst"
```
auguste-probabl, line 156: Maybe it would be better to put this file in `reference/` and have it tracked in Git. That way we can easily see when a change in code effects a change in the docs, without having to go check the rendered docs for every PR.

But if we do this, then we'd still have to be careful in reviews to tell contributors to build the docs locally, to regenerate the file; and that would be very annoying. So keeping it in `reference/api` is probably better.

sphinx/sphinxext/generate_accessor_tables.py
auguste-probabl: #2361

auguste-probabl:
> One nitpick I have is that `train_test_split` feels a bit out of place now, stuck between the classes and accessors. Maybe it's just me.

Yep agreed. https://github.com/probabl-ai/skore/pull/2359#discussion_r2763902640 is proposing to put the class link as the dropdown title, like this:
image

sphinx/_templates/accessor_summary.rst
auguste-probabl: If we want to do this well there should be more automation to fetch all public API and put it in the right `reference/report/.*_report.rst` file. I think this should be done in a follow-up PR.

glemaitre: OK but can you open an issue for it in case you don't want to do the PR right now.

sphinx/sphinxext/generate_accessor_tables.py
```diff
@@ -0,0 +1,157 @@
...
+    doc = getattr(obj, "__doc__", "")
+    if doc:
+        first_line = doc.strip().split("\n")[0].strip()
+        doc = first_line.rstrip(".")
+    return doc
+
+
+def get_accessor_methods(cls: type, accessor_name: str) -> list[tuple[str, str]]:
+    """
+    Get methods from an accessor by introspecting the accessor class.
```
glemaitre: Can you generate the strings using the NumPyDoc convention.

glemaitre (dismissed):
Otherwise LGTM. I would like to be able to have a table inside the dropdown title but I don't think that we can easily tweak this.

glemaitre:
I tweaked a bit for a bit and open: https://github.com/auguste-probabl/skore/pull/5/

In short, it is only a tweak to have a "summary" of the details much closer to the normal table:

image

glemaitre (approved):

PR #2375, opened 2026-02-05
fix(skore/CoefficientsDisplay): Change default ordering to `None`

### LINKED ISSUE #2369: bug(skore): By default `CoefficientsDisplay.plot()` should ordered of `feature_names`

Given the following code:

```python
estimator = LogisticRegression()

report = CrossValidationReport(estimator, X=X, y=y, splitter=2)
report.inspection.coefficients().plot()
```

would provide:

Image

Without any sorting, I would expect the coefficients to be ordered by feature names which in this case should be by lexical order because it is the original ordered.

Premkumar-2004:
Fixes #2369

Changed the default value of `sorting_order `from `"descending"` to `None `in `CoefficientsDisplay.frame()` and `CoefficientsDisplay.plot()`.

Features now keep their original order by default, same as `estimator.feature_names_in_`.if you want them sorted by magnitude, just pass `sorting_order="descending".`


auguste-probabl:
What issue does this PR address? If there is one open, please add a link; otherwise add a paragraph to explain why you feel this change is needed. Thanks.

Premkumar-2004:
should i update the tests?

Premkumar-2004:
updated the tests !!

GaetandeCast (approved):
LGTM, thanks!

Issue #2280, opened 2026-01-12
Try out jupyter-sphinx to render the Summary widget in the docs


auguste-probabl:
### Which part of the documentation needs improvement?

The Getting started example.

### Describe the problem found in the documentation

Currently we display a screenshot.

### Suggested improvement

https://jupyter-sphinx.readthedocs.io/en/latest/

auguste-probabl:
We tried this with @lesteve and the result was not conclusive.

Inserting a jupyter-sphinx directive inside of a `sphinx-gallery`-managed example does not work: the jupyter-sphinx code does not have access to the variables defined in the sphinx-gallery example; the two have different contexts and the code they define is run at different times during the docs building process.

Future investigation could include writing a full example with jupyter-sphinx to validate that it renders HTML correctly. Then, think about how to integrate this example with our current docs architecture (replace sphinx-gallery entirely? keep sphinx-gallery, and have a new pipeline to then build the jupyter-sphinx examples?)

Finally, keep in mind that even if the widget is rendered as HTML, interacting with it still will not change the rest of the notebook's progression. In this sense, it is not much better than the current screenshot.

PR #2388, opened 2026-02-07
docs(skore): Embed interactive Project summary plot

Adrien-1997:
Instead of relying on ipywidget interactivity (#2280), this PR takes an alternative route: the widget is executed at Sphinx build time, the underlying Plotly figure is extracted, and an HTML fragment is generated and embedded directly in the landing template.

This keeps the plot fully interactive (native Plotly), avoids Jupyter/ipywidgets at runtime, and works with static hosting.
parallel_coords

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -187,7 +187,6 @@ sphinx/build/
 sphinx/auto_examples/
 sphinx/reference/api/
 sphinx/sg_execution_times.rst
-sphinx/_templates/landing/project_summary_plot.html
 examples/plot_*.png
 sphinx/_templates/demo_report_help_generated.html
 
--- a/sphinx/_scripts/generate_landing_project_summary_plot.py
+++ b/sphinx/_scripts/generate_landing_project_summary_plot.py
@@ -1,30 +1,51 @@
-"""
-Sphinx extension that generates the Plotly HTML fragment embedded on the landing page.
-
-This writes a template snippet to sphinx/_templates/landing/project_summary_plot.html
-    {% include "landing/project_summary_plot.html" %}
-"""
+# Generate the Plotly HTML fragment used by the custom landing page.
+#
+# Output:
+# - sphinx/_templates/landing/project_summary_plot.html
+#
+# Executed during the Sphinx build (builder-inited), then included from the custom
+# landing template via:
+#
+#    {% include "landing/project_summary_plot.html" %}
 
 from __future__ import annotations
 
 from pathlib import Path
+import os
+import sys
 import tempfile
 import uuid
 
 import pandas as pd
 import plotly.io as pio
-import skrub
-import skore
-from skore import CrossValidationReport
 
 from sklearn.datasets import make_regression
 from sklearn.ensemble import HistGradientBoostingRegressor
 from sklearn.linear_model import Ridge
 
 
+HERE = Path(__file__).resolve()
+
+# Ensure we import local repo code (skore/src) instead of site-packages.
+# File is: repo_root/sphinx/_scripts/generate_landing_project_summary_plot.py
+REPO_ROOT = HERE.parents[2]
+SKORE_SRC = REPO_ROOT / "skore" / "src"
+sys.path.insert(0, str(SKORE_SRC))
+
+# Now it is safe to import local skore/skrub.
+import skrub  # noqa: E402
+import skore  # noqa: E402
+from skore import CrossValidationReport  # noqa: E402
+
+
+SPHINX_DIR = HERE.parents[1]
+OUT_TEMPLATE = SPHINX_DIR / "_templates" / "landing" / "project_summary_plot.html"
+OUT_TEMPLATE.parent.mkdir(parents=True, exist_ok=True)
+
+
 def _build_project_summary():
-    """Build a small Project + Summary, fast and deterministic."""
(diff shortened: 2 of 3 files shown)


auguste-probabl:
(Also you're really close to having your commits signed, I believe you just need to tell GitHub your signature in the settings)

adrien-morel:
Thanks!
I could make this work on a standard Sphinx-Gallery page (`getting_started`), but not on the landing page which is a custom template. I’ll rework it following the pattern you mentioned.
For the signed commits: I’ve set up a key but something is still missing, I’ll fix that.

adrien-morel:
Done, I moved the code into `sphinx/sphinxext` and wired it as a proper Sphinx extension 🙂

Also, since there was no reusable snippet for `project.summarize()`, I rebuilt a minimal, self-contained example to generate the plot at build time.

sphinx/sphinxext/landing_project_summary_plot.py
```diff
@@ -0,0 +1,144 @@
...
+    df["division"] = pd.cut(df["f0"], bins=4, labels=["A", "B", "C", "D"]).astype(str)
+
+    report_ridge = CrossValidationReport(Ridge(alpha=1.0), df_num, y)
+    report_hgbdt = CrossValidationReport(
+        skrub.tabular_pipeline(HistGradientBoostingRegressor(random_state=0)),
+        df,
+        y,
+    )
+
+    project = skore.Project(name=f"adult_census_survey_{uuid.uuid4().hex[:8]}")
```
auguste-probabl, line 64:
```suggestion
    project = skore.Project(name=f"adult_census_survey_{uuid.uuid4().hex[:8]}", workspace=tmp_dir)
```

sphinx/sphinxext/landing_project_summary_plot.py
```diff
@@ -0,0 +1,144 @@
...
+from sklearn.ensemble import HistGradientBoostingRegressor
+from sklearn.linear_model import Ridge
+
+
+def _ensure_local_imports() -> None:
+    """Prefer importing skore from the local repository (skore/src) during docs builds."""
+    here = Path(__file__).resolve()
+    repo_root = here.parents[2]  # repo_root/sphinx/sphinxext/<this_file>
+    skore_src = repo_root / "skore" / "src"
+    sys.path.insert(0, str(skore_src))
```
auguste-probabl: I'm not sure this is necessary. The `report_help.py` script does `import skore` too and it seems to work

sphinx/sphinxext/landing_project_summary_plot.py
```diff
@@ -0,0 +1,144 @@
+"""
+Sphinx extension that generates the Plotly HTML fragment embedded on the landing page.
+
+This writes a template snippet to:
+
+- sphinx/_templates/landing/project_summary_plot.html
```
auguste-probabl: This file path needs to be in .gitignore

auguste-probabl:
```suggestion
This writes a template snippet to sphinx/_templates/landing/project_summary_plot.html
```

sphinx/sphinxext/landing_project_summary_plot.py
```diff
@@ -0,0 +1,144 @@
...
+
+The landing template includes it with:
+
+    {% include "landing/project_summary_plot.html" %}
```
auguste-probabl:
```suggestion
```

sphinx/sphinxext/landing_project_summary_plot.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+def _set_checkboxes_by_description(root, descriptions_to_true: set[str]) -> None:
+    for w in _walk_widgets(root):
+        if hasattr(w, "description") and hasattr(w, "value"):
+            if str(getattr(w, "description", "")) in descriptions_to_true:
+                try:
+                    w.value = True
+                except Exception:
+                    pass
```
auguste-probabl, line 86: do we need to be that defensive? Or at least be more precise about the exception type

auguste-probabl:
You'll need to take into account #2406. Let me know if I can help

adrien-morel:
Thanks for the review! I’ve addressed the comments and adjusted accordingly.

adrien-morel:
One difficulty on my side is that I don’t have the code snippet that produced the current screenshot on the landing page. What I tried was to reconstruct an equivalent from the API, to check whether we could avoid a static image.

If you’re okay with it, I can retry starting strictly from the exact snippet that generated the screenshot. If that still doesn’t yield something faithful to the notebook experience, then keeping the screenshot makes sense.

auguste-probabl:
> One difficulty on my side is that I don’t have the code snippet that produced the current screenshot on the landing page.

You can find an example usage of the Project [here](https://docs.skore.probabl.ai/dev/auto_examples/getting_started/plot_getting_started.html#tracking-our-work-with-a-skore-project); even if the model names aren't exactly right that's fine, for example we can could change the landing page code snippet.

But TBH I don't really get why this is a blocker with respect to getting all the interactive parts (checkboxes, dropdowns) in the landing page.

adrien-morel:
Closing this PR for now. I wasn’t able to find a clean and robust way to embed the full Project.summarize() widget output in the landing page.

PR #2391, opened 2026-02-09
chore(skore): Add pytest --durations option to show the durations of the slowest tests

jeromedockes:
related to #2351

### DIFF
--- a/skore-hub-project/pyproject.toml
+++ b/skore-hub-project/pyproject.toml
@@ -63,6 +63,7 @@ addopts = [
   "--no-header",
   "--verbosity=2",
   "--dist=loadscope",
+  "--durations=25",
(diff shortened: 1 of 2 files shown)


thomass-dev:
Duplicated with https://github.com/probabl-ai/skore/pull/2355?

glemaitre:
Nop this one, is only to have the statistics shown for the 20 slower tests each time. We have this in scikit-learn and it is quite useful.

thomass-dev:
In the @GaetandeCast PR, it's not in the right place, but it is

image

jeromedockes:
> In the @GaetandeCast PR, it's not in the right place, but it is

right, sorry I missed that somehow

PR #2396, opened 2026-02-09
refactor(plot): return plotting objects instead of storing attributes

### LINKED ISSUE #2374: api(skore): Change the return values of `.plot()` and the attribute of `Display`

Currently `Display` create some attributes when calling `.plot()`.

First, it is a bit weird that those attributes does not exist before calling `.plot`.
Then, it also gives the possibilities to user to alter the internal of the displays which might not be the best.

So an alternative is to make `display.plot()` return the matplotlib figure or seaborn facet grid such that the user can further modify it but without storing this element as an attribute.

Sharkyii:
This PR refactors all Display.plot() methods to return the plotting object instead of storing them as attributes on the display instance.

Previously, calling .plot() created attributes such as figure_, ax_,. These attributes did not exist before plotting, which introduced hidden state and allowed external mutation of internal display objects.

Fixes #2374

Follow up -
Edit other test file to work with the new pattern


skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -462,7 +463,7 @@ def _plot_single_estimator(
         barplot_kwargs: dict[str, Any],
         boxplot_kwargs: dict[str, Any],
         stripplot_kwargs: dict[str, Any],
-    ) -> None:
+    ) -> Any:
```
auguste-probabl:
```suggestion
    ) -> seaborn.FacetGrid:
```
Same change everywhere.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -206,11 +191,11 @@ def _plot_matplotlib(
         facet_grid_kwargs_validated = _validate_style_kwargs(
             {"col": subplot_by_validated, **self._default_facet_grid_kwargs}, {}
         )
-        self.facet_ = sns.FacetGrid(
+        facet_ = sns.FacetGrid(
             data=frame,
             **facet_grid_kwargs_validated,
         )
-        self.figure_, self.ax_ = self.facet_.figure, self.facet_.axes.flatten()
+        figure_, ax_ = facet_.figure, facet_.axes.flatten()
```
auguste-probabl: In another part of the PR you instead write 

```suggestion
        _, ax_ = facet_.axes.flatten()
```
and then you use `facet_.figure` e.g. when you do `suptitle`.

I think I prefer doing it that way, without creating a new variable? Not sure. It's up to you, just make sure it's consistent across the whole PR.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -243,10 +228,10 @@ def plot_heatmap(data, **kwargs):
             if threshold_value is None:
                 threshold_value = 0.5 if self.response_method == "predict_proba" else 0
             title = f"{title}\nDecision threshold: {threshold_value:.2f}"
-        self.figure_.suptitle(f"{title}\n{info_data_source}")
+        figure_.suptitle(f"{title}\n{info_data_source}")
 
-        for ax in self.ax_:
-            ax.set(
+        for axis in ax_:
```
auguste-probabl:
```suggestion
        for ax in ax_:
```
Changing the name doesn't make it clearer in this case.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -278,8 +263,9 @@ def plot_heatmap(data, **kwargs):
                     },
                 )
 
-        if len(self.ax_) == 1:
-            self.ax_ = self.ax_[0]
+        if len(ax_) == 1:
+            ax_ = ax_[0]
```
auguste-probabl: I wonder if this block still matters now

skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
```diff
@@ -263,24 +252,24 @@ def _plot_matplotlib(
...
-            labels = [t.get_text() for t in self.facet_._legend.get_texts()]
-            self.facet_._legend.remove()
+        if facet_._legend is not None:
+            handles = list(facet_._legend.legend_handles)
+            labels = [t.get_text() for t in facet_._legend.get_texts()]
+            facet_._legend.remove()
             if hue == "split":
                 labels = [f"Split #{label}" for label in labels]
-        handles.append(ax.lines[0])
+        handles.append(ax_.flatten()[0].lines[0])
```
auguste-probabl: Why is this line changing this much?

skore/tests/unit/displays/coefficients/test_comparison_estimator.py
```diff
@@ -88,22 +88,16 @@ def test_binary_classification(
...
+
+    assert result.axes[0, 0].get_xlabel() == "Magnitude of coefficient"
+    assert result.axes[0, 0].get_ylabel() == ""
+
+    result = display.plot(subplot_by="estimator")
+    assert isinstance(result.axes.flatten(), np.ndarray)
+    assert len(result.axes.flatten()) == len(report.reports_)
+    for report_name, ax in zip(report.reports_, result.axes.flatten(), strict=True):  # noqa: B007
         # report_name is used in the pandas.query as an string which is not detected by
         # ruff as a used variable
```
auguste-probabl:
```suggestion
    for report_name, ax in zip(report.reports_, result.axes.flatten(), strict=True):
```
while we're at it

auguste-probabl: same for other occurrences of this

Sharkyii:
@auguste-probabl  could you review this PR? 
If the direction looks good, I’ll continue updating the remaining tests.

auguste-probabl:
> If the direction looks good, I’ll continue updating the remaining tests.

Yep, I think this works fine, thanks!

Sharkyii:
@auguste-probabl could you please review this when you have a moment? I’ll double-check again in the meantime to ensure all tests are passing.

Sharkyii:
@auguste-probabl All tests are passing on my side - ready for review.

glemaitre:
@Sharkyii Could you solve the conflicts?

Sharkyii:
@glemaitre Solved!

skore/src/skore/_sklearn/_plot/data/table_report.py
```diff
@@ -289,7 +291,7 @@ def _plot_matplotlib(
         hue: str | None = None,
         kind: Literal["dist", "corr"] = "dist",
         top_k_categories: int = 20,
-    ) -> None:
+    ) -> tuple[Figure, Axes]:
```
glemaitre, line 294: I think that we should be only returning FacetGrid or Figure (when we don't have a FacetGrid)

- One can access to the figure or axis from facet
- One can access to the axis from a figure

In this way only return a single object, it will make the API simpler.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -153,7 +153,7 @@ def _plot_matplotlib(
         normalize: Literal["true", "pred", "all"] | None = None,
         threshold_value: float | None = None,
         subplot_by: Literal["split", "estimator", "auto"] | None = "auto",
-    ) -> None:
+    ) -> sns.axisgrid.FacetGrid:
```
glemaitre: Uhm this is weird. `sns.FaceGrid` seems to be the write import

skore/src/skore/_sklearn/_plot/metrics/precision_recall_curve.py
```diff
@@ -176,7 +176,7 @@ def _plot_matplotlib(
         subplot_by: Literal["auto", "label", "estimator", "data_source"]
         | None = "auto",
         despine: bool = True,
-    ) -> None:
+    ) -> sns.axisgrid.FacetGrid:
```
glemaitre: same here.

skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
```diff
@@ -190,7 +190,7 @@ def _plot_matplotlib(
             "actual_vs_predicted", "residual_vs_predicted"
         ] = "residual_vs_predicted",
         despine: bool = True,
-    ) -> None:
+    ) -> sns.axisgrid.FacetGrid:
```
glemaitre: you can make a find and replace for this occurrence.

skore/tests/unit/displays/coefficients/conftest.py
```diff
@@ -171,8 +171,8 @@ def comparison_cross_validation_reports_multioutput_regression_figure_axes(
 ):
     report = comparison_cross_validation_reports_multioutput_regression
     display = report.inspection.coefficients()
-    display.plot()
-    return display.figure_, display.ax_
+    facet = display.plot()
+    return facet
```
glemaitre: Could you replace here and elsewhere to

```suggestion
    return display.plot()
```

Let's directly return instead to store in a variable that we don't need.

glemaitre (commented):
I made a first pass that should change a couple of lines before to do another round of review.

Sharkyii:
@glemaitre I have updated it from my end

PR #2404, opened 2026-02-10
fix(skore): Prevent memory overflow from dangerous arange

cakedev0:
There is a dangerous code path in `_column_is_classification`: `sequential = np.arange(y_unique[0], y_unique[-1] + 1)` can create arbitrary big arrays. It's honestly unlikely to happen, but if it happens, it could be "expensive" (e.g. filling all the RAM and making other jobs on the same machine crash).

Minimal reproducer (using only public APIs):

```Python
import numpy as np
import pandas
import skore, skrub

y = np.repeat(np.array([0, 1, 10**12], dtype=np.int64), 1000)
n = y.size
X = pandas.DataFrame(np.random.rand(n, 2), columns=['A', 'B'])
y = pandas.Series(y)

estimator = skrub.tabular_pipeline("classification")
X_train, X_test, y_train, y_test = skore.train_test_split(X, y)
```

CI failing for the new test (before the fix was pushed): https://github.com/probabl-ai/skore/actions/runs/21878956616/job/63155763165?pr=2404

### DIFF
--- a/skore/tests/unit/sklearn/test_utils.py
+++ b/skore/tests/unit/sklearn/test_utils.py
@@ -118,7 +118,10 @@ def test_find_ml_task_pandas():
 
 
 def test_find_ml_task_pandas_large_label_gap():
-    # non regression-test for PR
+    """"
+    Check that `_find_ml_task` does not fail on large arrays.
+    Non-regression test for https://github.com/probabl-ai/skore/pull/2404
+    """
     y = pandas.Series(numpy.repeat(numpy.array([0, 1, 10**12], dtype=numpy.int64), 100))
     assert _find_ml_task(y, None) == "regression"
 


skore/tests/unit/sklearn/test_utils.py
```diff
@@ -117,5 +117,11 @@ def test_find_ml_task_pandas():
     assert _find_ml_task(y, None) == "multiclass-classification"
 
 
+def test_find_ml_task_pandas_large_label_gap():
+    # non regression-test for PR
```
auguste-probabl:
```suggestion
    """"
    Check that `_find_ml_task` does not fail on large arrays.
    Non-regression test for https://github.com/probabl-ai/skore/pull/2404
    """
```

PR #2427, opened 2026-02-11
fix(skore/reports): Fix auto-completion in `IPython`

### LINKED ISSUE #2386: fix: IPython completion doesn't work consistently for EstimatorReports
### Describe the bug

I'm having issues with tab completion of EstimatorReport instances. Strangely enough it doesn't seem to affect ComparisonReports the same way, although it's hard to isolate what might be going wrong. Would appreciate if readers of this could try to reproduce the issue.

### Steps/Code to Reproduce

```python
from skore import train_test_split, EstimatorReport

X, y = load_iris(return_X_y=True, as_frame=True)
split_data = train_test_split(X=X, y=y, shuffle=False, as_dict=True)
report = EstimatorReport(RandomForestClassifier(), **split_data)

report.<tab> # nothing happens
```

### Expected Behavior

A list of available methods and attributes

### Actual Behavior

Nothing

### Environment

```shell
ipython 9.9.0

System:
    python: 3.13.9 (main, Oct 14 2025, 13:52:31) [GCC 14.3.0]
executable: /home/auguste/Desktop/work/skore/skore/.venv/bin/python3
   machine: Linux-6.12.59-x86_64-with-glibc2.40

Python dependencies:
        skore: 0.0.0+unknown (main)
          pip: 25.2
    anywidget: 0.9.21
      ipython: 9.9.0
   ipywidgets: 8.1.8
       joblib: 1.5.3
   matplotlib: 3.10.8
        numpy: 2.4.0
       pandas: 2.3.3
       plotly: 6.5.2
         rich: 14.2.0
 scikit-learn: 1.8.0
      seaborn: 0.13.2
 skore[local]: None
        skrub: 0.7.1
skore-hub-project: 0.0.0+unknown (main)
```

jeromedockes:
closes #2386 

the proposed workaround for https://github.com/davidhalter/jedi/issues/1990 is to slightly relax the properties' annotation from Literal to str

I can add a test; it would require depending on `jedi` for the tests WDYT?

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -53,8 +53,9 @@ test-base = [
   "pytest-xdist",
   "scipy-stubs",
   "xdoctest",
+  "jedi",
 ]
-test = ["skore[test-base]", "polars", "pyarrow", "jedi"]
+test = ["skore[test-base]", "polars", "pyarrow"]
 test-lts-cpu = ["skore[test-base]", "polars-lts-cpu"]
 sphinx-base = [
   "altair",


thomass-dev:
> I can add a test; it would require depending on jedi for the tests WDYT?

Please do it; you will also need to update the lockfiles using `bash ci/pip-compile.sh --test-requirements skore`.

thomass-dev:
It would be interesting to have a generic test to ensure that there is no future regression on the future properties of reports.

jeromedockes:
> It would be interesting to have a generic test to ensure that there is no future regression on the future properties of reports.

the test collects suggested completions using jedi, so if a new property is added that causes it to fail it should cause the test to fail

skore/tests/unit/reports/cross_validation/test_report.py
auguste-probabl, line 1: Maybe we could have 1 test, parametrized with the report type? We do this in a number of other places (search for `test_common.py`)

skore/pyproject.toml
```diff
@@ -54,7 +54,7 @@ test-base = [
   "scipy-stubs",
   "xdoctest",
 ]
-test = ["skore[test-base]", "polars", "pyarrow"]
+test = ["skore[test-base]", "polars", "pyarrow", "jedi"]
```
thomass-dev: Please put it in the `test-base` section, just above.

skore/pyproject.toml
```diff
@@ -54,7 +54,7 @@ test-base = [
   "scipy-stubs",
   "xdoctest",
 ]
-test = ["skore[test-base]", "polars", "pyarrow"]
```
auguste-probabl, line 57: Add it to `test-base` instead

skore/pyproject.toml
auguste-probabl: duplicate

PR #2424, opened 2026-02-11
feat(skore-hub-project/put): Improve progress tracking

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/2418 and https://github.com/probabl-ai/skore/issues/2342.

PR #2431, opened 2026-02-12
fix(skore/configuration)!: Make the configuration inheritable between threads

thomass-dev:
Part of https://github.com/probabl-ai/skore/pull/2424.

The configuration has been refactored to be inheritable between threads and to be a self-contained class.
The parallel module has been simplified, with a remaining delayed class in charge to reset the configuration (to the parent state) for each joblib task.

---

https://docs.python.org/3.14/library/threading.html#thread-local-data
https://github.com/python/cpython/blob/main/Lib/_threading_local.py#L82
https://stackoverflow.com/questions/1408171/thread-local-storage-in-python
https://stackoverflow.com/questions/64058938/javas-inheritablethreadlocal-equivalent-in-python

### DIFF
--- a/skore/src/skore/_utils/_parallel.py
+++ b/skore/src/skore/_utils/_parallel.py
@@ -7,6 +7,24 @@ from skore._config import LocalConfiguration
 
 
 def delayed(function):
+    """Capture the arguments of a function to delay its execution.
+
+    This alternative to ``joblib.delayed`` is meant to be used in conjunction
+    with ``joblib.Parallel``.
+
+    It ensures each task has its own copy of the ``skore`` configuration.
+
+    Parameters
+    ----------
+    function : callable
+        The function to be delayed.
+
+    Returns
+    -------
+    output: tuple
+        Tuple containing the delayed function, the positional arguments, and the
+        keyword arguments.
+    """
     local = configuration.local.__dict__.copy()
 
     @wraps(function)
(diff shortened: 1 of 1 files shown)


sphinx/reference/index.rst
```diff
@@ -79,20 +79,6 @@ Inspection
...
-   :widths: 30 70
-
-   * - :func:`config_context`
-     - Context manager for temporarily modifying skore configuration
-   * - :func:`get_config`
-     - Get current skore configuration values
-   * - :func:`set_config`
-     - Set skore configuration values
-
-
```
thomass-dev, line 95: I need help to change this section:

- `get_config` is now directly the global variable `skore.configuration`
- `set_config` is splitted in two property setters: `skore.configuration.show_progress` and `skore.configuration.plot_backend`
- `config_context` is now `skore.configuration()`

thomass-dev:
@glemaitre could you take a look? The docstr are missing and sphinx should be fixed, but the PR is otherwise ready to be reviewed.

skore/src/skore/_config.py
```diff
@@ -2,167 +2,94 @@
...
+            yield
+        finally:
+            self.show_progress = show_progress_copy
+            self.plot_backend = plot_backend_copy
+
+
+configuration = Configuration()
+
+
+def _change_configuration_for_testing():
```
glemaitre: I would suggest to move it in the `_testing.py` module where we have the same trick for different functions that we know that we need pickling at test time.

thomass-dev: Fixed in 57dd071d.

glemaitre:
I open https://github.com/probabl-ai/skore/pull/2502 for the documentation.

sphinx/reference/index.rst
thomass-dev: Resolved by https://github.com/probabl-ai/skore/pull/2445.

glemaitre:
LGTM. Thanks @glemaitre

PR #2436, opened 2026-02-12
fix(skore-hub-project): Raise earlier an error when pos_label not provided before serialization

### LINKED ISSUE #2417: Binary classification: wasted computation if pos_label is not set

Put binary classif:
pos_label is not specified check is done too late the computation is wasted & it is not immediately clear where 1
should pass pos_label

glemaitre:
closes #2417 
requires https://github.com/probabl-ai/skore/pull/2438

Force to raise an earlier when `pos_label` is not set at the level of the report when trying to serialize. Provide a way to make the proper change.

### DIFF
--- a/skore-hub-project/src/skore_hub_project/project/project.py
+++ b/skore-hub-project/src/skore_hub_project/project/project.py
@@ -25,6 +25,7 @@ from unicodedata import normalize
 import joblib
 import orjson
 from httpx import HTTPStatusError
+from sklearn.utils.validation import _check_pos_label_consistency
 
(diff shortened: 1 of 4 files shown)


glemaitre:
I think that I will dissociate this PR into two PRs: one modifying the `CrossValidationReport` and one modifying the `skore-hub-project` (which will be this one).

Here, I think that I'm interested to know if we want to have a separate function to validate the report before to start the upload.

glemaitre:
I would say that it is ready for review now

PR #2438, opened 2026-02-12
feat(skore): Allow to set pos_label after creating report

glemaitre:
Partially address #2417 

To offer a solution to user to change `pos_label`, we should be less restrictive and allow changing `pos_label` after instantiation. Since we do it with the `EstimatorReport`, it is pretty straightforward. We only need to invalidate the cache and set hashes to be consistent with the way we handle it in `EstimatorReport`.

Potentially, I think that we could have a better handling of the cache by not removing everything.


auguste-probabl (approved):
LGTM

Issue #2421, opened 2026-02-11
Upload progress bar is broken in jupyter


rouk1:
### What would you like to say?

When it goes above default cell height it becomes invisible and then an error message is displayed.

`Error displaying widget: model not found`

rouk1:
Is this fixed by #2140  ?

thomass-dev:
Minimal reproducer:

```python
import sklearn
import skore
import skrub
from contextlib import contextmanager

@contextmanager                                                                         
def switch_mpl_backend() -> Iterator[None]:                                             
    import matplotlib as mpl                                                            
    import matplotlib.pyplot as plt                                                     
                                                                                        
    original_mpl_backend = mpl.get_backend()                                            
    original_plt_backend = plt.get_backend()                                            
                                                                                        
    try:                                                                                
        mpl.use("agg")                                                                  
        plt.switch_backend("agg")                                                       
        yield                                                                           
    finally:                                                                            
        mpl.use(original_mpl_backend)                                                   
        plt.switch_backend(original_plt_backend)

data, target = sklearn.datasets.fetch_openml(data_id=42125, as_frame=True, return_X_y=True)
data_train, data_test, target_train, target_test = sklearn.model_selection.train_test_split(data, target)
estimator = skrub.tabular_pipeline("regressor")
report = skore.EstimatorReport(
    estimator,
    X_train=data_train,
    y_train=target_train,
    X_test=data_test,
    y_test=target_test,
)

with switch_mpl_backend():
    report.metrics.prediction_error().plot()

report.metrics.prediction_error().plot()
```

The issue is between the `matplotlib` backend dynamic switch and `seaborn`.

PR #2439, opened 2026-02-12
fix(skore-hub-project): Replace `switch_mpl_backend` by `switch_mpl_ioff`

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/2421.

Problem: the plots were no longer displayed after a put, because rolling back the `matplotlib` backend from "agg" to "interactive" has no effect after `seaborn` has been imported (ie `pyplot`). `seaborn` continues to use the "agg" backend.

Solution: instead of changing dynamically the `matplotlib` backend, which appears to have serious side-effects especially when `pyplot` is already imported, disable temporarily the interactive mode.

---

https://matplotlib.org/stable/users/explain/figure/backends.html#selecting-a-backend
https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.ioff.html

### DIFF
--- a/sphinx/sphinxext/github_link.py
+++ b/sphinx/sphinxext/github_link.py
@@ -92,5 +92,5 @@ def setup(app):
     """Configure linkcode_resolve for GitHub links."""
     app.config["linkcode_resolve"] = make_linkcode_resolve(
         package="skore",
-        url_fmt="https://github.com/Soham-47/skore/blob/{revision}/skore/{path}#L{lineno}",
+        url_fmt="https://github.com/probabl-ai/skore/blob/{revision}/skore/{path}#L{lineno}",
     )


glemaitre:
The error is just due to a matching error in `show_versions`.

glemaitre:
Uhm it changes as well the checksum from blake3. Hope that it is deterministic between call.

thomass-dev:
@glemaitre and @rouk1 (originally, the agg was intended to avoid your macOS pop-up windows), can you both test it on your machines? It works on mine.

rouk1 (approved):
Tested in Jupyter lab and in vscode interactive and it works !

glemaitre (dismissed):

glemaitre:
Is is good for me. I tried locally and it solves the problem with matplotlib.

PR #2445, opened 2026-02-12
fix(skore-hub-project): Do not aggregate importance when sending to hub

glemaitre:
We should not aggregate the values before to send to the hub. The hub will be in charge to have a plot allowing to display the right information.

### DIFF
--- a/skore-hub-project/tests/unit/artifact/media/test_inspection.py
+++ b/skore-hub-project/tests/unit/artifact/media/test_inspection.py
@@ -1,3 +1,4 @@
+import inspect
 from functools import partialmethod
 
 from pydantic import ValidationError
@@ -17,7 +18,12 @@ def serialize(result) -> bytes:
     import orjson
 
     if hasattr(result, "frame"):
-        result = result.frame()
+        # FIXME: in the future, all inspection methods should have an aggregate
+        # parameter and we should be sending unaggregated data to the hub.
+        if "aggregate" in inspect.signature(result.frame).parameters:
+            result = result.frame(aggregate=None)
+        else:
+            result = result.frame()
 
     return orjson.dumps(
         result.fillna("NaN").to_dict(orient="tight"),


glemaitre:
It is not surprising that the checksum is different because we send different data.

I think that we should sync with @rouk1 and @thomass-dev just to be sure what we want to send to the hub and what the hub is supposed to do.

glemaitre:
@thomass-dev What is the proper way to update the hash computed by `blake3`?

thomass-dev:
> @thomass-dev What is the proper way to update the hash computed by `blake3`?

In `test_inspection.py`, update the `serialize` function with your changes.

skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
```diff
@@ -27,7 +28,13 @@ def content_to_upload(self) -> bytes | None:  # noqa: D102
             return None
 
         display = function()
-        frame = display.frame()
+        # FIXME: in the future, all inspection methods should have an aggregate
+        # parameter and we should be sending unaggreagted data to the hub.
+        sig = inspect.signature(display.frame)
+        if "aggregate" in sig.parameters:
```
thomass-dev: Nit: 
```suggestion
        if "aggregate" in inspect.signature(display.frame).parameters:
```

skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
```diff
@@ -27,7 +28,13 @@ def content_to_upload(self) -> bytes | None:  # noqa: D102
             return None
 
         display = function()
-        frame = display.frame()
+        # FIXME: in the future, all inspection methods should have an aggregate
+        # parameter and we should be sending unaggreagted data to the hub.
```
thomass-dev:
```suggestion
        # parameter and we should be sending unaggregated data to the hub.
```

glemaitre:
@thomass-dev What would be the way to take care about the failure?

PR #2447, opened 2026-02-13
docs(skore): Mention `split` in class docstring

### LINKED ISSUE #2443: docs(skore): Documentation of mean decrease impurity is out dated

We recently merged the support for `CrossValidationReport` in `ImpurityDecreaseDisplay` but we did not mention that we have a new column named `split`.

GaetandeCast:
Fixes #2443 & adds missing blank lines between parameters docstring.


GaetandeCast:
It's what we do in the rest of the code base AFAIK. So it's to respect our own convention.

Issue #2409, opened 2026-02-11
Running many put in parallel does not use a headless ploting backend


rouk1:
Following script never finishes.

```python
import asyncio
import uuid

import skore
from skore import EstimatorReport, Project, train_test_split
from sklearn.datasets import make_regression
from sklearn.linear_model import LinearRegression


skore.login()


async def put_report(report):
    random_project_name = f"P-{str(uuid.uuid4())[:8]}"
    small_project = Project(f"hub://spam-spam-spam/{random_project_name}")
    # Run blocking network / IO call in a thread so we can parallelize
    await asyncio.to_thread(small_project.put, "r", report)


async def main():
    little_dataset = make_regression(
        n_samples=1_000,
        n_features=10,
        random_state=42,
    )
    split_data = train_test_split(
        X=little_dataset[0],
        y=little_dataset[1],
        random_state=0,
        as_dict=True,
    )
    er = EstimatorReport(LinearRegression(), **split_data)
    er.inspection.permutation_importance(data_source="train")
    er.inspection.permutation_importance(data_source="test")

    # Launch 1000 put_report calls in parallel
    tasks = [put_report(er) for _ in range(1_000)]
    await asyncio.gather(*tasks)


if __name__ == "__main__":
    asyncio.run(main())
```

```
Uploading... ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00

^C^Cobject address  : 0x117fc3e80
object refcount : 877
object type     : 0x102cceda8
object type name: RuntimeError
object repr     : RuntimeError("Cannot create a GUI FigureManager outside the main thread using the MacOS backend. Use a non-interactive backend like 'agg' to make plots on worker threads.")
lost sys.stderr
^C^C^C^C
```

thomass-dev:
This shouldn't be included in the board for public launch. Being thread-safe/multi-process-safe/async-safe is a feature request in its own.

PR #2114, opened 2025-10-30
feat(skore-hub-project): Reduce `put` time

thomass-dev:
Partially addresses https://github.com/probabl-ai/skore/issues/2140:

- Use the hash generated by `skore` to upload a report
- Thread the compute of report media
    - All media are computed in "parallel",  divided in chunk of ~10mb and sent to the upload queue
    - The upload queue is emptied in "parallel" 6 chunks of 6, to avoid flooding the hub

PR #2459, opened 2026-02-13
fix(skore): Do not start matplotlib GUI in non-main thread

glemaitre:
related #2409 

When putting in parallel, we trigger Matplotlib GUI outside of the main thread which seems to be tricky.

Here, `plt.tight_layout` is triggering such behaviour.

However, it does not address all the problem detected in #2409. Indeed, because we are using seaborn that call `plt.figure`, then the Matplotlib GUI is called from outside skore. The solution, would be to switch to a non-interactive backend.

But we end-up in the same issue pointed out by @thomass-dev. I assume that putting object in parallel will have exactly the same problem. So while using `plt.ioff` to deactivate the interactive mode allowing to not break matplotlib in a notebook, this solution does not solve really the problem of the backend.

So we might finally need to find a way to switch the plotting backend even once `matplotlib` has been imported because we don't have the hand on what `seaborn` is doing.

### DIFF
--- a/skore/src/skore/_sklearn/_plot/base.py
+++ b/skore/src/skore/_sklearn/_plot/base.py
@@ -1,3 +1,4 @@
+import threading
 from collections.abc import Callable
 from functools import wraps
 from typing import Any, Literal, Protocol, runtime_checkable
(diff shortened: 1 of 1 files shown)


thomass-dev:
> So we might finally need to find a way to switch the plotting backend even once matplotlib has been imported because we don't have the hand on what seaborn is doing.


You can do something like

image

But i gave up this method because it will drastically impact the user by forgetting all the settings he could have made manually with `plt.blabla`.

thomass-dev:
Note that in https://github.com/probabl-ai/skore/pull/2114 https://github.com/probabl-ai/skore/pull/2114/changes/ba12bb879e283a7338c3336214bb2553fb8819d0 (WIP) in a threading context (so similar to the asyncio issue), i ended up by spawning a dedicated matplotlib plotter process.

PR #2461, opened 2026-02-14
chore(skore): Switch from `mypy` to `ty`

Sharkyii:
Since ty surfaced a large number of pre-existing diagnostics, I added scoped ignore rules in each project’s pyproject.toml to establish a stable baseline:

• skore/: 12 rules ignored
• skore-hub-project/: 5 rules ignored
• skore-local-project/: 5 rules ignored

Fixes: #2376, replaces https://github.com/probabl-ai/skore/pull/2455.


Sharkyii:
@thomass-dev When you have some time, I would appreciate another review. I’ve updated the PR to address the earlier concerns and verified the CI behavior.

thomasloux:
@Sharkyii you probably tagged the wrong person :)

.pre-commit-config.yaml
```diff
@@ -32,47 +32,52 @@ repos:
...
+  - repo: local
     hooks:
+    - id: ty
+      alias: ty-skore
+      name: ty skore/
+      entry: ty check --project skore skore
+      language: python
+      pass_filenames: false
+      additional_dependencies:
+        - ty
```
glemaitre: Can you explain why pre-commit would need this line?

glemaitre: OK it is because it is not a `pre-commit` hook and thus you need to install `ty`.

skore/pyproject.toml
```diff
@@ -160,18 +160,19 @@ convention = "numpy"
...
-  "jinja2.*",
-  "joblib.*",
-  "pandas.*",
-  "plotly.*",
-  "seaborn.*",
-  "sklearn.*",
-  "skrub.*",
-]
+[tool.ty.rules]
+unresolved-import = "ignore"
```
glemaitre: I see that we have many ignore rules compared to the original setup with `mypy`. @Sharkyii could you point out the reasons. Are those specific to `ty` only and there where not activated with `mypy` before?

.pre-commit-config.yaml
Sharkyii: Hi @glemaitre ,
Just saw the message since it’s a local hook (repo: local), pre-commit creates its own isolated environment and doesn’t provide ty by default

skore/pyproject.toml
Sharkyii: I am not 100% sure but mostly ty is stricter than our previous mypy setup

Sharkyii: I will point the exact problem soon as possible

Sharkyii:
### **Why ty flags errors that mypy didn’t**

**Import & attribute resolution**

`Errors: unresolved-import, unresolved-attribute, possibly-missing-attribute
`
mypy often supresses missing or dynamic attributes that are used in the repo as Any.
Docs:
mypy:  [here](https://mypy.readthedocs.io/en/stable/dynamic_typing.html)

**Function call & signature checking**

`Errors: unknown-argument, invalid-argument-type, missing-argument, no-matching-overload, call-non-callable
`
mypy skips checking many untyped function bodies and treats unknown types as Any, so invalid calls often pass silently. ty performs stricter call-site checking even when types are incomplete.

Docs:
mypy untyped defs not checked → [here](https://mypy.readthedocs.io/en/stable/common_issues.html)
mypy Any behavior → [here](https://mypy.readthedocs.io/en/stable/dynamic_typing.html)
ty diagnostics philosophy → [here](https://docs.astral.sh/ty/reference/rules)

**Assignment & return typing**

`Errors: invalid-assignment, invalid-return-type, invalid-type-form
`
mypy widens many inferred types to avoid false positives and accepts more loose assignments. ty keeps stricter inference and flags mismatches instead of widening automatically.

Docs:
mypy assignment rules → [here](https://mypy.readthedocs.io/en/stable/error_code_list.html)

**Hygiene / redundancy checks**

`Errors: redundant-cast, unused-type-ignore-comment
`
mypy only reports these if explicitly enabled. ty reports them by default to keep annotations clean.

Docs:
mypy unused ignore behavior → [here](https://mypy.readthedocs.io/en/stable/error_code_list2.html)
ty ignore handling → [here](https://docs.astral.sh/ty/reference/configuration)

skore/pyproject.toml
Sharkyii: @glemaitre i have commented it below in the PR i hope it addresses your concern

skore-hub-project/pyproject.toml
```diff
@@ -109,18 +109,25 @@ convention = "numpy"
...
+
+[tool.ty.overrides.analysis]
+respect-type-ignore-comments = true
+
+[tool.ty.rules]
+unresolved-import = "ignore"
+unknown-argument = "ignore"
+invalid-argument-type = "ignore"
+unused-type-ignore-comment = "ignore"
+invalid-type-form = "ignore"
```
auguste-probabl: Can you try to remove some of these and fix the code instead?

auguste-probabl: Same for the other packages

.pre-commit-config.yaml
```diff
@@ -32,47 +32,52 @@ repos:
       - id: ruff-format
         files: ^((skore|skore-hub-project|skore-local-project)/(hatch|src|tests))|(examples)/
 
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.19.1
+  - repo: local
```
auguste-probabl: There is also https://github.com/allganize/ty-pre-commit. But I guess defining it locally is ok

.pre-commit-config.yaml
```diff
@@ -32,47 +32,52 @@ repos:
...
+        - matplotlib
+        - numpy
+        - pandas
+        - plotly
+        - rich
+        - scikit-learn
+        - skrub
+        - seaborn
+        - polars
+        - pyarrow
```
auguste-probabl: Can we find a way around this? Maybe using `uv`?

.pre-commit-config.yaml
```diff
@@ -32,47 +32,52 @@ repos:
...
         files: ^((skore|skore-hub-project|skore-local-project)/(hatch|src|tests))|(examples)/
 
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.19.1
+  - repo: local
     hooks:
+    - id: ty
+      alias: ty-skore
+      name: ty skore/
+      entry: ty check --project skore skore
```
auguste-probabl:
```suggestion
      entry: ty check --project skore
```
I think this is enough.

Sharkyii: Hey @auguste-probabl it's because the ty follows this pattern -
`ty check [OPTIONS] [PATH]...`

Sharkyii: Oh Okay, i checked this and this worked as it automatically detects the path
Will change it!

skore-hub-project/pyproject.toml
Sharkyii: Looks like removing this rule affects 83 fixtures

.github/workflows/lint.yml
```diff
@@ -47,4 +47,4 @@ jobs:
           pre-commit run --all-files typos
           pre-commit run --all-files nbstripout
           pre-commit run --all-files ruff-check
-          pre-commit run --all-files mypy
+          pre-commit run --all-files ty
```
auguste-probabl: We could just have
```
run: pre-commit run --all-files
```

@thomass-dev WDYT?

thomass-dev: As i see, we don't do `ruff-format`.
But now that we have no-more UI pre-commit hooks, i think it is good to just run pre-commit.

skore-hub-project/pyproject.toml
Sharkyii: I’ve solved the failing rules in skore-local-project and skore-hub-project, and a few in skore.

Sharkyii:
I have solved most of the conflicts by removing the rules and used ` # type: ignore[assignment]` in some of them

.github/workflows/lint.yml
auguste-probabl:
```suggestion
        run: pre-commit run --all-files
```

auguste-probabl: I mean, replace the whole `run` block.

skore-local-project/src/skore_local_project/storage.py
```diff
@@ -28,7 +28,7 @@ class DiskCacheStorage:
 
     Attributes
     ----------
-    storage : Cache
+    storage : StorageCache
         The underlying diskcache Cache object.
```
auguste-probabl, line 32: This docstring is wrong, the class only has `directory` as an attribute

skore-local-project/src/skore_local_project/storage.py
```diff
@@ -7,7 +7,7 @@
 
 from diskcache import Cache
 
-Cache = partial(Cache, size_limit=float("inf"), cull_limit=0, eviction=None)
+StorageCache = partial(Cache, size_limit=float("inf"), cull_limit=0, eviction=None)
```
auguste-probabl: Would this work:
```suggestion
Cache = partial(diskcache.Cache, size_limit=float("inf"), cull_limit=0, eviction=None)
```

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -135,7 +135,7 @@ def permutation_importance(
         max_samples: float = 1.0,
         n_jobs: int | None = None,
         seed: int | None = None,
-    ) -> pd.DataFrame:
+    ) -> pd.DataFrame | PermutationImportanceDisplay:
```
auguste-probabl: This will never be a DataFrame

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -503,7 +503,7 @@ def _compute_metric_scores(
         # to check if the metric has already been computed
         cache_key_parts: list[Any] = [
             self._parent._hash,
-            metric_fn.__name__,
+            metric_fn.__name__,  # type: ignore[assignment]
```
auguste-probabl: I think this needs a specific check that `__name__` is defined

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -256,7 +256,7 @@ def cache_predictions(
                     response_methods += ["predict_proba"]
                 if hasattr(self._estimator, "decision_function"):
                     response_methods += ["decision_function"]
-            pos_labels = self._estimator.classes_.tolist() + [None]
+            pos_labels = self._estimator.classes_.tolist() + [None]  # type: ignore[assignment]
```
auguste-probabl: can you `cast` here instead?

skore/src/skore/_sklearn/_plot/data/table_report.py
```diff
@@ -533,7 +533,9 @@ def _plot_distribution_2d(
                 raise ValueError(
                     "If 'x' and 'y' are categories, 'hue' must be continuous."
                 )
-
+            assert x is not None
+            assert y is not None
+            assert hue is None or isinstance(hue, pd.Series)
```
auguste-probabl: `cast` instead

skore/src/skore/_sklearn/train_test_split/train_test_split.py
```diff
@@ -237,7 +237,7 @@ class labels.
     from skore import console  # avoid circular import
 
     for warning_class in TRAIN_TEST_SPLIT_WARNINGS:
-        warning = warning_class.check(**kwargs)
+        warning = cast(Any, warning_class).check(**kwargs)
```
auguste-probabl:
```suggestion
        warning = cast(TrainTestSplitWarning, warning_class).check(**kwargs)
```

skore/src/skore/_sklearn/feature_names.py
```diff
@@ -31,18 +32,19 @@ def _get_feature_names(
         # list and so we need to check if we have an array-like before to make any
         # conversion
         if hasattr(feature_names, "tolist"):
-            feature_names = feature_names.tolist()
-        return feature_names
+            return cast(Callable[[], list[str]], feature_names.tolist)()
```
auguste-probabl: Does this work?
```suggestion
            feature_names = cast(list[str], feature_names.tolist())
```
The current formulation is not readable

.pre-commit-config.yaml
```diff
@@ -32,47 +32,25 @@ repos:
...
-        pass_filenames: false
-        alias: mypy-skore-local-project
-        name: mypy skore-local-project/
-        files: ^skore-local-project/
-        args: [--config-file=skore-local-project/pyproject.toml, skore-local-project]
-        additional_dependencies: [skore]
+    - id: ty
+      alias: ty-skore
+      name: ty skore/
+      entry: uv run ty check --project skore
```
auguste-probabl: This needs to specify a version of `ty`; the latest is 0.0.17

auguste-probabl:
Working on it, hope that's okay ^^'

Sharkyii:
Yupp, works for me.

skore-local-project/src/skore_local_project/metadata.py
```diff
@@ -120,9 +119,12 @@ def metric(report: EstimatorReport, name: str) -> float | None:
 
     def __post_init__(self, report: EstimatorReport | CrossValidationReport) -> None:
         """Initialize dynamic fields."""
-        report = cast(EstimatorReport, report)
         super().__post_init__(report)
 
+        from skore import EstimatorReport
```
thomass-dev, line 124: This is redundant with the import in `TYPE_CHECKING`, isn't?

auguste-probabl: No, I got test failures when this wasn't imported out of TYPE_CHECKING. So it looks like `cast` statements are not simply removed at runtime...

auguste-probabl:
I'm having a lot of trouble with this because I never get the same check results between CI and my machine. I'm going to let this sit for a while and hope that `ty` matures more and provides a more robust solution.

skore-hub-project/pyproject.toml
```diff
@@ -109,18 +109,8 @@ convention = "numpy"
 [tool.ruff.lint.per-file-ignores]
 "tests/*" = ["D"]
 
-[tool.mypy]
-exclude = ["hatch/", "tests/"]
-strict = true
```
thomass-dev, line 114: This change has to be discussed, as `ty` doesn't have a strict mode for now and we would to be strict as much as possible in `skore-hub-project` and `skore-local-project`.

https://docs.astral.sh/ty/reference/typing-faq/#does-ty-have-a-strict-mode

Could you please list me the rules that will not be applied with this change?

skore-local-project/src/skore_local_project/metadata.py
```diff
@@ -118,10 +117,14 @@ def metric(report: EstimatorReport, name: str) -> float | None:
 
         return cast_to_float(getattr(report.metrics, name)(data_source="test"))
 
-    def __post_init__(self, report: EstimatorReport) -> None:  # type: ignore[override]
+    def __post_init__(self, report: EstimatorReport | CrossValidationReport) -> None:
```
thomass-dev, line 120: I know it violates the Liskov substitution principle, but your change is not true. Please revert.

skore-local-project/src/skore_local_project/metadata.py
```diff
@@ -164,10 +167,14 @@ def timing(report: CrossValidationReport, label: str) -> float | None:
 
         return cast_to_float(series.iloc[0])
 
-    def __post_init__(self, report: CrossValidationReport) -> None:  # type: ignore[override]
+    def __post_init__(self, report) -> None:
```
thomass-dev, line 170: Please revert.

skore-local-project/src/skore_local_project/metadata.py
thomass-dev: We can fix the violation in a dedicated issue as we have done in https://github.com/probabl-ai/skore/blob/main/skore-hub-project/src/skore_hub_project/report/estimator_report.py, but it is not important here.

PR #2463, opened 2026-02-14
feat(skore/ConfusionMatrixDislay): Introduce `threshold_value="all"` option

glemaitre:
This PR is introducing `threshold_value="default"` such that `threshold_value=None` would returns a dataframe with all the thresholds.

The behaviour `decision_thresholds=None` is particularly useful if one wants to combine the the output of `frame` with the one of the ROC and the PR curve.

It should not be the default but we should make is possible.

### DIFF
--- a/skore/tests/unit/displays/confusion_matrix/test_comparison_estimator.py
+++ b/skore/tests/unit/displays/confusion_matrix/test_comparison_estimator.py
@@ -51,7 +51,7 @@ def test_frame_default_threshold(
     comparison_report = ComparisonReport([report_1, report_2])
     display = comparison_report.metrics.confusion_matrix()
 
-    frame = display.frame(threshold_value="default")
+    frame = display.frame()
     assert isinstance(frame, pd.DataFrame)
     n_classes = len(display.display_labels)
     n_estimators = 2
--- a/skore/tests/unit/displays/confusion_matrix/test_cross_validation.py
+++ b/skore/tests/unit/displays/confusion_matrix/test_cross_validation.py
@@ -90,7 +90,7 @@ def test_frame_default_threshold(
     report = CrossValidationReport(estimator, X=X, y=y, splitter=cv)
     display = report.metrics.confusion_matrix()
 
-    frame = display.frame(threshold_value="default")
+    frame = display.frame()
     assert isinstance(frame, pd.DataFrame)
     n_classes = len(display.display_labels)
     assert frame.shape == (n_classes * n_classes * cv, 7)
--- a/skore/tests/unit/displays/confusion_matrix/test_estimator.py
+++ b/skore/tests/unit/displays/confusion_matrix/test_estimator.py
@@ -71,7 +71,7 @@ def test_frame_default_threshold(
     )
     display = report.metrics.confusion_matrix()
 
-    frame = display.frame(threshold_value="default")
+    frame = display.frame()
     assert isinstance(frame, pd.DataFrame)
     n_classes = len(display.display_labels)
     assert frame.shape == (n_classes * n_classes, 7)


skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -20,6 +20,8 @@
     YPlotData,
 )
 
+ThresholdValue = float | Literal["default"] | None
```
auguste-probabl:
```suggestion
ThresholdValue = float | Literal["default"]
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -505,25 +510,46 @@ def frame(
...
-            If None and thresholds are available, returns the confusion matrix at the
-            default threshold (0.5 for `predict_proba` response method, 0 for
-            `decision_function` response method).
+        threshold_value : float, "default" or None, default="default"
+            The decision threshold(s) to use when applicable (binary classification
+            only). If ``"default"``, returns the confusion matrix at the default
+            threshold (0.5 for `predict_proba` response method, 0 for
+            `decision_function` response method). If ``None``, returns all available
+            thresholds without filtering. If a float, returns the confusion matrix at
+            the closest available threshold to the given value.
```
auguste-probabl, line 525:
```suggestion
            The decision threshold(s) to use when applicable (binary classification
            only). If `"default"`, returns the confusion matrix at the default
            threshold (0.5 for `predict_proba` response method, 0 for
            `decision_function` response method). If `None`, returns all available
            thresholds without filtering. If a float, returns the confusion matrix at
            the closest available threshold to the given value.
```
Same in other places

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -505,25 +510,46 @@ def frame(
...
         Returns
         -------
         frame : pandas.DataFrame
             The confusion matrix as a dataframe.
         """
         normalize_col = "normalized_by_" + normalize if normalize else "count"
-        if threshold_value is not None and self.ml_task != "binary-classification":
+        if (
+            threshold_value is not None
+            and threshold_value != "default"
```
auguste-probabl:
```suggestion
            threshold_value not in (None, "default")
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -505,25 +510,46 @@ def frame(
...
+                [
+                    "true_label",
+                    "predicted_label",
+                    normalize_col,
+                    "threshold",
+                    "split",
+                    "estimator",
+                    "data_source",
+                ]
+            ].rename(columns={normalize_col: "value"})
```
auguste-probabl: Can you refactor to avoid duplication with line 554?

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -505,25 +510,46 @@ def frame(
             conditions or all the population. If None, the confusion matrix will not be
             normalized.
 
-        threshold_value : float or None, default=None
-            The decision threshold to use when applicable.
-            If None and thresholds are available, returns the confusion matrix at the
-            default threshold (0.5 for `predict_proba` response method, 0 for
-            `decision_function` response method).
+        threshold_value : float, "default" or None, default="default"
```
auguste-probabl: `"all"` would be better than `None`

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
GaetandeCast: +1

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
glemaitre: True, let's go for `"all"`

skore/tests/unit/displays/confusion_matrix/test_comparison_cross_validation.py
```diff
@@ -86,7 +86,7 @@ def test_frame_default_threshold(
     report = ComparisonReport([cv_report_1, cv_report_2])
     display = report.metrics.confusion_matrix()
 
-    frame = display.frame(threshold_value=None)
+    frame = display.frame(threshold_value="default")
```
auguste-probabl:
```suggestion
    frame = display.frame()
```

skore/tests/unit/displays/confusion_matrix/test_comparison_estimator.py
```diff
@@ -51,7 +51,7 @@ def test_frame_default_threshold(
     comparison_report = ComparisonReport([report_1, report_2])
     display = comparison_report.metrics.confusion_matrix()
 
-    frame = display.frame(threshold_value=None)
+    frame = display.frame(threshold_value="default")
```
auguste-probabl:
```suggestion
    frame = display.frame()
```

skore/tests/unit/displays/confusion_matrix/test_cross_validation.py
```diff
@@ -90,7 +90,7 @@ def test_frame_default_threshold(
     report = CrossValidationReport(estimator, X=X, y=y, splitter=cv)
     display = report.metrics.confusion_matrix()
 
-    frame = display.frame(threshold_value=None)
+    frame = display.frame(threshold_value="default")
```
auguste-probabl:
```suggestion
    frame = display.frame()
```

skore/tests/unit/displays/confusion_matrix/test_estimator.py
```diff
@@ -71,7 +71,7 @@ def test_frame_default_threshold(
     )
     display = report.metrics.confusion_matrix()
 
-    frame = display.frame(threshold_value=None)
+    frame = display.frame(threshold_value="default")
```
auguste-probabl:
```suggestion
    frame = display.frame()
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -107,15 +109,15 @@ def plot(
         self,
         *,
         normalize: Literal["true", "pred", "all"] | None = None,
-        threshold_value: float | None = None,
+        threshold_value: ThresholdValue = "default",
```
auguste-probabl, line 112: So now "default" means what `None` used to mean? I would rather leave it to `None`, or use `"auto"` which is more similar to the rest of the repo. Preference for `None`.

GaetandeCast: I disagree, we are in a situation where `"default"` is meaningful since it is dependent on the type of estimator (SVM with decision function vs classifier with predict proba). I don't like `None` since we can't plot with "no" threshold (unlike e.g. `subplot_by=None` where it makes sense). As for `"auto"`, to me it conveys the idea that we choose a "good" default for the user. In this case, `threshold=0.5` is almost never the good choice but there needs to be a default value.

auguste-probabl: Fair enough.

PR #2470, opened 2026-02-15
fix(skore): Fix _verbose_name and add aggregate parameter to inspection displays

moksha-hub:
## Summary


glemaitre:
Your PR is tackling several issues not associated with your description.

I'm not closing right away but you need to focus the PR to a specific issue, add a proper description in the summary of the PR, make sure to propose a fix and a non-regression tests or tests in general for the patch to be reviewed.

Issue #2441, opened 2026-02-12
docs(skore): Wrong find and replace from feature_importance to inspection


glemaitre:
### What would you like to say?

I see an occurrences where we messed up the find and replace from `feature_importances` to `inspection`:

<img width="734" height="44" alt="Image" src="https://github.com/user-attachments/assets/802ffba4-04b2-4053-8251-8bd040ee462a" />

So we should have another look in the documentation where we got this issue.

PR #2471, opened 2026-02-15
fix(skore): Change _verbose_name from 'feature_importance' to 'inspection'

moksha-hub:
## Summary

This PR fixes issue **#2441: Wrong find and replace from `feature_importance` to `inspection`**.

---

### Problem

The inspection accessor was registered with the name `"inspection"` (via `_register_accessor("inspection", ...)`), but the `_verbose_name` attribute was incorrectly set to `"feature_importance"`.

This caused inconsistencies in:

* Rich / HTML display output
* Help system documentation

---

### Solution

Changed `_verbose_name` from `"feature_importance"` to `"inspection"` in three files:

* `skore/src/skore/_sklearn/_cross_validation/inspection_accessor.py`
* `skore/src/skore/_sklearn/_estimator/inspection_accessor.py`
* `skore/src/skore/_sklearn/_comparison/inspection_accessor.py`

---

### Changes

Each file had a one-line update:

```python
# Before
_verbose_name: str = "feature_importance"

# After
_verbose_name: str = "inspection"
```

This aligns `_verbose_name` with the registered accessor name, ensuring consistent display in rich/HTML output and the help system.

---

### Testing

A non-regression test was added in:

```
skore/tests/unit/utils/test_accessors.py
```

The test verifies that all three inspection accessor classes have `_verbose_name` set to `"inspection"`.

### DIFF
--- a/skore/tests/unit/utils/test_accessors.py
+++ b/skore/tests/unit/utils/test_accessors.py
@@ -184,4 +184,34 @@ def test_check_has_feature_importance():
 
     err_msg = "Estimator 'hello' is not a supported estimator by the function called."
     with pytest.raises(AttributeError, match=err_msg):
-        assert _check_estimator_has_feature_importances()(accessor)
+        _check_estimator_has_feature_importances()(accessor)
+
+
+def test_inspection_accessor_verbose_name():
+    """Test that inspection accessor _verbose_name matches registered name.
+
+    This is a non-regression test for issue #2441 to ensure that the
+    _verbose_name attribute of inspection accessors is aligned with
+    the name used to register the accessor.
+    """
+    from skore._sklearn._comparison.inspection_accessor import (
+        _InspectionAccessor as ComparisonInspectionAccessor,
+    )
+    from skore._sklearn._cross_validation.inspection_accessor import (
+        _InspectionAccessor as CVInspectionAccessor,
+    )
+    from skore._sklearn._estimator.inspection_accessor import (
+        _InspectionAccessor as EstimatorInspectionAccessor,
+    )
+
+    # All inspection accessors should have _verbose_name set to "inspection"
+    # to match the registered accessor name
+    assert (
+        ComparisonInspectionAccessor._verbose_name == "inspection"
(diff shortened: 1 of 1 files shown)


glemaitre:
Things that you need to fix for further contribution and that you can edit in this PR:

- you need to sign your commit: https://github.com/probabl-ai/skore/pull/2471#issuecomment-3904427612
- You need to edit the description of what you try to achieve in this PR to provide context: https://github.com/probabl-ai/skore/pull/2471#issue-3944066318

This said, nice catch for the inconsistencies but it does not solve #2441. However, it means that the display of the `rich` or `html` display was wrong.

So it also means that we should add a non-regression test for this behaviour such that the `_verbose_name` of the accessor are aligned with the name used to register the accessor.

skore/tests/unit/utils/test_accessors.py
```diff
@@ -184,4 +184,34 @@ def __init__(self):
 
     err_msg = "Estimator 'hello' is not a supported estimator by the function called."
     with pytest.raises(AttributeError, match=err_msg):
-        assert _check_estimator_has_feature_importances()(accessor)
+        _check_estimator_has_feature_importances()(accessor)
```
glemaitre, line 187: It is not a good test. What I want is to be sure that the parameter that we assigned as name in:

https://github.com/probabl-ai/skore/blob/main/skore/src/skore/_sklearn/_comparison/__init__.py#L8-L9

is corresponding to the `_verbose_name` to be sure that both parameters are aligned.

Also your test was just generating by providing my previous comment to an LLM and only pasting committing the changes. If it your plan, I would advise to not contribute to this project.

What I expect here is the following:

- engage with the reviewer in case some aspects are not clear
- follow the code standard of the project

So here:

- we should not have local import inside a test function
- we should have a limited number of comments (comments should exist when we fail to make the code self-explicit).
- we should have a test that take into accounts the three type of reports, the different accessors and check the that the way we register the accessor is aligne with the `_verbose_name`. The current test only check that the value of `_verbose_name` is associated with a constant value set by hand.

PR #2476, opened 2026-02-15
refactor(skore): Introduce _report_type to avoid hard coding when needed

glemaitre:
Introduce to have a `_report_type` argument such that we don't hard code the string each time we need it.

It was first proposed in https://github.com/probabl-ai/skore/pull/2032#discussion_r2695386896


skore-local-project/src/skore_local_project/metadata.py
```diff
@@ -33,14 +33,12 @@ def report_type(
...
     """Human readable type of a report."""
-    from skore import CrossValidationReport, EstimatorReport
-
-    if isinstance(report, CrossValidationReport):
-        return "cross-validation"
-    if isinstance(report, EstimatorReport):
-        return "estimator"
-
-    raise TypeError
+    # _report_type is defined on skore._sklearn._base._BaseReport; both report
```
glemaitre: @thomass-dev any idea why I have CI failure?

Do we install `skore` in editable mode in the CI or we actually installed from `PyPI` which explain `mypy` to scream and at the same time the failure on the CI?

thomass-dev: For the `skore` CI, we install `skore-local-project` from `PyPI`.
For the `skore-local-project` CI, we install `skore` from `PyPI`.

thomass-dev: I recommend you to revert this change, and do it in a future PR after the release of skore.
Otherwise, you can make a retro-compatible change with something like: `if hasattr(report, "_report_type"): [...]`.

glemaitre: OK let me do the backward compatibility. It seems good.

glemaitre:
I implemented the backward compatibility and added a `# TODO` to know that we can remove it some days :)

skore/src/skore/_sklearn/_base.py
```diff
@@ -24,6 +24,12 @@ class _BaseReport(ReportHelpMixin):
     """
 
     _ACCESSOR_CONFIG: dict[str, dict[str, str]]
+    _report_type: Literal[
+        "cross-validation",
+        "estimator",
```
auguste-probabl:
```suggestion
        "estimator",
        "cross-validation",
```

skore-local-project/src/skore_local_project/metadata.py
```diff
@@ -33,13 +33,18 @@ def report_type(
     report: EstimatorReport | CrossValidationReport,
 ) -> Literal["cross-validation", "estimator"]:
     """Human readable type of a report."""
-    from skore import CrossValidationReport, EstimatorReport
```
thomass-dev, line 36: Just rollback this change to green the CI.

Issue #2267, opened 2026-01-08
docs(skore): Revisit the example from technical section


glemaitre:
### What would you like to say?

We should review and add new examples to provide insights regarding `skore-lib` internal in the technical section.

glemaitre:
This part has been revised already. We can now iterate over time.

PR #2481, opened 2026-02-15
docs(skore): Add small example to illustrate the usage of the skore.Project in local

glemaitre:
Towards closing https://github.com/probabl-ai/skore/issues/2267

It is a really short example to explain a couple of technical details regarding the `summarize()` methods and what it returns when used. locally.

### DIFF
--- a/examples/technical_details/plot_skore_project.py
+++ b/examples/technical_details/plot_skore_project.py
@@ -24,8 +24,9 @@ from tempfile import TemporaryDirectory
 
 from skore import Project
 
-tmpdir = Path(TemporaryDirectory().name)
-project = Project("example-project", workspace=tmpdir)
+tmp_dir = TemporaryDirectory()
+tmp_path = Path(tmp_dir.name)
+project = Project("example-project", workspace=tmp_path)
 
 # %%
 from sklearn.datasets import load_breast_cancer
@@ -92,4 +93,5 @@ reports.metrics.summarize().frame()
 reports.metrics.roc().plot(subplot_by=None)
 
 # %%
-project.delete("example-project", workspace=tmpdir)
+project.delete("example-project", workspace=tmp_path)
+tmp_dir.cleanup()


examples/technical_details/plot_skore_project.py
```diff
@@ -0,0 +1,95 @@
...
+# We use a temporary directory as the workspace so the example is self-contained.
+# In practice you can omit ``workspace`` to use the default (e.g. a ``skore/``
+# directory in your user cache).
+from pathlib import Path
+from tempfile import TemporaryDirectory
+
+from skore import Project
+
+tmpdir = Path(TemporaryDirectory().name)
+project = Project("example-project", workspace=tmpdir)
```
thomass-dev: Since we don't necessarily want the user to find this project, should you not delete tmpdir at the end of the example?

examples/technical_details/plot_skore_project.py
```diff
@@ -0,0 +1,95 @@
+"""
+.. _example_skore_project:
+
+===================
+Local skore Project
+===================
+
+This example shows how to use :class:`~skore.Project` in **local** mode: store
+reports on your machine and inspect them. A key point is that
+:meth:`~skore.Project.summarize` returns a :class:`~skore.project._summary.Summary`,
```
GaetandeCast, line 10: the `Summary` hyperlink does not render on the preview.

glemaitre: Yep it is a known issue because we are not exposing it publicly if I'm not wrong.

examples/technical_details/plot_skore_project.py
glemaitre: Yep, let's do that. I was hesitating.

examples/technical_details/plot_skore_project.py
glemaitre: So the reason is that we don't expose the `Summary` in the `skore.project` module. Since it is in a `_summary` private module, we never list it in the API documentation.

We could decide to expose it and then list it. But I think that it should be done in another PR then.

PR #2482, opened 2026-02-15
chore(skore): Make ipython and ipywidget optional dependencies

glemaitre:
This PR proposes to remove `ipywidgets` and `ipython` as mandatory libraries and instead to make them optional.

Basically they are only used inside the `SummaryDisplay` which can default to a dataframe representation. Also `rich` is doing similar style of optional dependencies.

It might be nice to not have those when dealing with a minimal setup.


thomass-dev (approved):
LGTM.

PR #2456, opened 2026-02-13
fix(skore): Don't create a tuple from a string for the cache key

cakedev0:
Fixes #2450 

Failing non-regression test before the fix: https://github.com/probabl-ai/skore/actions/runs/21992666989/job/63543866467?pr=2456

PR #2483, opened 2026-02-16
refactor(skore): Uniform cache key sanitizing

cakedev0:
@glemaitre said:

> I assume that we should have a better way to handle creating the key for caching but it is for later.

I propose to make this now, because I feel the more we wait the more we'll loose time on handling it manually and chasing bugs.

I propose to keep explicit tuples as key, and not a deep hash as I initially proposed in https://github.com/probabl-ai/skore/pull/2456#issuecomment-3897970542. This should allow for easier debugging (keys are human-readable, except for the parts that gets hashed by joblib).

Fix: While doing it I noticed that we were sometimes using python functions/class-instances as keys (not hashed with `joblib`), I don't think this was the intended behavior (in other parts of the code, similar objects were hashed).

### DIFF
--- a/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
@@ -195,7 +195,7 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
         aggregate: Aggregate | None = ("mean", "std"),
         **metric_kwargs: Any,
     ):
-        is_cv_report = self._parent._reports_type == "comparison-cross-validation"
+        is_cv_report = self._parent._report_type == "comparison-cross-validation"
 
         cache_key = deep_key_sanitize(
             (


thomass-dev:
Note: keep in mind that changing the cache key form is a breaking-change.

cakedev0:
> Note: keep in mind that changing the cache key form is a breaking-change.

What do you mean? Does it interact with the hub?

(anyway that feels to me like an additional reason not to wait to do it, it will be even worse later)

thomass-dev:
> What do you mean? Does it interact with the hub?

Yes.

https://github.com/probabl-ai/skore/blob/main/skore-hub-project/src/skore_hub_project/artifact/media/inspection.py#L42

cakedev0:
Hum I see, thanks.

Well, I propose to remove the coupling between this skore-hub piece of code and the private cache of skore report: https://github.com/probabl-ai/skore/pull/2483/commits/5874826a7fb115ef9637c77b2f50ebde9cdaad56

thomass-dev:
In this PR, don't change `skore-hub-project` because with our layout mono-repo/multi-package and the breaking-change, your CI will remains red.

We can iterate here on `skore`, release, and make the update afterwards.

cakedev0:
(I'm on my phone, I'll improve formatting of this comment later)

Ok I need to split this PR in 3 PRs then I think, in a kind of "green-blue" migration pattern. 

First expose "get_cached_permutation_..." in skore, then call it from skore-hub once it's released, and then I can change the private behavior of the cache without impacting skore-hub.

cakedev0:
> > Does it interact with the hub?
> 
> Yes: https://github.com/probabl-ai/skore/blob/main/skore-hub-project/src/skore_hub_project/artifact/media/inspection.py#L42

I tested and the new cache keys are still compatible with this piece of logic because what it reads (`parent_hash, name, data_source, at_step, _, metric, *_ = key`) is unchanged with the new cache key:

```Python
deep_key_sanitize((
                self._parent._hash,
                "permutation_importance",
                data_source,
                at_step,
                data_source_hash,
                metric,
                kwargs,
))
```

The only big difference will be in kwargs, but this isn't checked. So this PR should not even break anything :smile:

skore/src/skore/_utils/_cache_key.py
```diff
@@ -0,0 +1,52 @@
...
+            raise TypeError("Only string keys are support for mappings")
+        kv_pairs = ((key, deep_key_sanitize(item)) for key, item in value.items())
+        return ("mapping", tuple(sorted(kv_pairs, key=lambda kv: kv[0])))
+    if isinstance(value, Set):
+        items = (deep_key_sanitize(item) for item in value)
+        return ("set", tuple(sorted(items, key=hash)))
+    if isinstance(value, Sequence) or hasattr(value, "__iter__"):
+        return tuple(deep_key_sanitize(item) for item in value)
+
+    raise TypeError(f"Unsupported type: {type(value)}")
```
thomass-dev, line 52: Have you considered using a JSON serializer with a custom default fallback?

skore/src/skore/_utils/_cache_key.py
thomass-dev, line 1: Also, i think we should think creating a proper cache class that extends the dict properties (such as `UserDict`).

By doing it this way, we should be able to encapsulate `deep_key_sanitize` call in a custom `__getitem__`/`__setitem__`.

skore/src/skore/_utils/_cache_key.py
```diff
@@ -0,0 +1,52 @@
...
+from collections.abc import Mapping, Sequence, Set
+from typing import Any
+
+import joblib
+import numpy as np
+
+from skore._sklearn.types import _DEFAULT
+
+
+def deep_key_sanitize(value: Any) -> Any:
```
thomass-dev, line 12: @glemaitre couldn't we fix the len/content of the `key`?

I'm thinking out loud, i don't have all the use-cases in mind, but something like:

```python
@dataclass
class CacheKey:
    hash
    type

@dataclass
class MetricCacheKey(CacheKey):
   data_source
   at_step
   data_source_hash
   metric
   kwargs

@dataclass
class PredictionCacheKey(CacheKey):
    ...
```

skore/src/skore/_utils/_cache_key.py
cakedev0: Hum, I think it's convenient to have tuples, to raise for unexpected types, etc. Plus this happens to be compatible with the current `skore-hub-project` logic which I find convenient.

I think the few extra lines of code are worth the control/predictability.

Also I think mixing a big default fallback (`_DEFAULT, callable, array, set` and maybe others) with JSON serialization might be harder to read.

skore/src/skore/_utils/_cache_key.py
cakedev0: We have that in `skore/src/skore/_utils/_cache.py` (it does some locking). I hesitated to add it there. I think I find it easier to read the code when I see the `deep_key_sanitize` explicitly, but I very ok with hiding it into `Cache` if you think it's better.

skore/src/skore/_utils/_cache_key.py
cakedev0: I think we first need to have a high-level brainstorm on the cache before making it more powerful/complex.

Like:
- why we cache?
- what do we cache? IMO we should only cache at the finest grain, so that if you compute `[a, b]` in one call then you can read just `a` or just `b` from the cache (on a call that only returns `a`/`b`). It's not what we do today, which from my user experience so far reduces a lot the value of the cache.

glemaitre:
I would be in favor of merging this quick fix and take the time needed to discuss the caching in general in `skore`.

I think that we are in a better place today to make an actual call on the feature: it is not anymore only a Python dict, we know some requirements related to the `put` and thread safety, and have more understanding on where we would benefit from it. We also some additional thoughts related to what `skrub` and `scikit-learn` would be trying to do in this space.

PR #2489, opened 2026-02-17
chore(skore): Disallow clustering models in reports

GaetandeCast:
We do not plan to support clustering models in the skore reports for the moment so this PR add an early error when trying to build a report with a clustering model.

This way, X and y have to both be not None which simplifies things in some places.

### DIFF
--- a/skore/tests/unit/reports/comparison/cross_validation/test_report.py
+++ b/skore/tests/unit/reports/comparison/cross_validation/test_report.py
@@ -71,7 +71,7 @@ def test_clustering():
     """Check that we cannot create a report with a clustering model."""
     with pytest.raises(
         ValueError,
-        match="skore does not support clustering models yet. Please use a "
+        match="Clustering models are not supported yet. Please use a "
         "classification or regression model instead.",
     ):
         ComparisonReport(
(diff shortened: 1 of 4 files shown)


skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -154,6 +154,12 @@ def __init__(
         splitter: int | SKLearnCrossValidator | Generator | None = None,
         n_jobs: int | None = None,
     ) -> None:
+        if is_clusterer(estimator):
+            raise ValueError(
+                "skore does not support clustering models yet. Please use a"
+                " classification or regression model instead."
+            )
```
auguste-probabl: Is this necessary given that we delegate to EstimatorReport later? If it is then maybe factorize the error message. While we're at it, I would prefer to formulate it without "skore":

```suggestion
        if is_clusterer(estimator):
            raise ValueError(
                "Clustering models are not supported yet. Please use a"
                " classification or regression model instead."
            )
```

GaetandeCast: Yeah you're right, since all report types delegate to `EstimatorReport`, we can raise the error in a single place. Removed duplicate error messages.

PR #2503, opened 2026-02-19
ci(pre-commit): Remove nbstripout

auguste-probabl:
The hook is never run because we don't have Jupyter notebooks checked in anymore.


thomass-dev:
> We should add caching / switch to https://github.com/pre-commit/action

Looking at the related action, it should be better to just add manually the cache, and avoid a dependency to an extra action (the action is quiet simple).

https://github.com/pre-commit/action/blob/main/action.yml

PR #2510, opened 2026-02-19
feat(skore-hub-project): Send inspection media for cross-validation reports

glemaitre:
Lib is now computing all the dataframe required to show the importances in cross-validation.

So we need to send them.


thomass-dev:
We have to release `skore` before updating the lockfile + merging this PR.

PR #2511, opened 2026-02-20
feat(skore/ComparisonReport): Add permutation importance

### LINKED ISSUE #2469: feat(skore): Add `inspection.permutation_importance` in `ComparisonReport`

Once `inspection.permutation_importance` is added to `EstimatorReport` and `CrossValidationReport` then, it can be added to `ComparisonReport`.

GaetandeCast:
Closes #2469


skore/src/skore/_sklearn/_plot/inspection/permutation_importance.py
```diff
@@ -257,20 +246,97 @@ def _plot_matplotlib(
...
+                "The estimators have different features and should be plotted on "
+                "different axis using `subplot_by='estimator'`."
+            )
+
+        if hue is None:
+            stripplot_kwargs.pop("palette", None)
+            # we don't need the palette and we are at risk of raising an error or
+            # deprecation warning if passing palette without a hue
+
+        self._categorical_plot(
```
GaetandeCast, line 294: We could "inline" `_categorical_plot` here since it's only used once. I'm not sure if it would improve or degrade code readability.

glemaitre:
Could we slightly revisit the routines. I open a PR that refactor all the over routines: https://github.com/probabl-ai/skore/pull/2516

Basically, it will be easy to rework what you already have (an LLM will get the exact pattern that we use in the other PR) and the test do not need to change.

The advantage is that we are going to cache the results of the underlying estimators that makes that we can later benefit from it when sending the results to the hub without the need to recreate dataframe and fill the cache by hand.

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
auguste-probabl, line 1: Refactor using #2516

skore/src/skore/_sklearn/_plot/inspection/permutation_importance.py
```diff
@@ -398,13 +464,28 @@ def frame(
...
+                metric = [metric]
+            available_metrics = self.importances["metric"].unique()
+            for m in metric:
+                if m not in available_metrics:
+                    raise ValueError(
+                        f"The metric {m!r} is not available. Please select metrics from"
+                        f" the following list: {', '.join(available_metrics)}. "
+                        "Otherwise, use the `metric` parameter of the "
+                        "`.permutation_importance()` method to specify the metrics to"
+                        " use for computing the importances."
```
auguste-probabl: Some inconsistencies between spaces at the end or at the start of each line. I tend to put it at the end, but i just want it consistent

skore/src/skore/_sklearn/_plot/inspection/permutation_importance.py
auguste-probabl: i have a tiny preference for keeping it extracted for consistency with the other displays.

skore/tests/unit/displays/permutation_importance/test_common.py
```diff
@@ -139,3 +120,20 @@ def test_frame_aggregate(
...
+        display = report.inspection.permutation_importance(n_repeats=2, seed=0)
+
+        available_metric = "r2" if "regression" in task else "accuracy"
+        err_msg = (
+            "The metric 'invalid-metric' is not available. Please select metrics from"
+            f" the following list: {available_metric}. "
+            "Otherwise, use the `metric` parameter of the "
+            "`.permutation_importance\(\)` method to specify the metrics to"
+            " use for computing the importances."
+        )
```
auguste-probabl:
```suggestion
        err_msg = "The metric 'invalid-metric' is not available."
```

skore/tests/unit/displays/permutation_importance/test_comparison_cross_validation.py
```diff
@@ -0,0 +1,116 @@
...
+    ],
+)
+def test_invalid_subplot_by(pyplot, task, request):
+    report = request.getfixturevalue(f"comparison_cross_validation_reports_{task}")
+    display = report.inspection.permutation_importance(seed=0, n_repeats=2)
+    valid_values = ["estimator", "split", "auto", "None"]
+    err_msg = (
+        "The column 'incorrect' is not available for subplotting. You can use "
+        f"the following values to create subplots: {', '.join(valid_values)}"
+    )
```
auguste-probabl: nit
```suggestion
    err_msg = "The column 'incorrect' is not available for subplotting."
```

skore/tests/unit/displays/permutation_importance/test_comparison_cross_validation.py
```diff
@@ -0,0 +1,116 @@
...
+        "regression",
+        "multioutput_regression",
+    ],
+)
+@pytest.mark.parametrize(
+    "subplot_by, expected_len",
+    [
+        ("estimator", 2),
+        ("split", 2),
+        ("auto", 0),
```
auguste-probabl: nit: Shouldn't it be
```suggestion
        ("auto", 1),
```
(or None)

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
```diff
@@ -1,23 +1,32 @@
...
     _check_comparison_report_sub_estimators_have_coef,
     _check_comparison_report_sub_estimators_have_feature_importances,
 )
+from skore._utils._cache_key import deep_key_sanitize
 
 if TYPE_CHECKING:
     from skore import ComparisonReport
     from skore._sklearn._cross_validation.report import CrossValidationReport
 
+Metric = str | Callable | list[str] | tuple[str] | dict[str, Callable] | None
```
auguste-probabl: Maybe we can factorize this somewhere

skore/src/skore/_sklearn/_plot/inspection/permutation_importance.py
GaetandeCast: same

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
GaetandeCast: Actually we already have a `Metric` type in `_sklearn/types.py`. Let me reuse it here.

GaetandeCast: Ok so I reused the parameter signature and docstring from the metrics accessors: 
```
        metric: Metric | list[Metric] | dict[str, Metric] | None = None,
```
```
        metric : str, callable, scorer, or list of such instances or dict of such \
            instances, default=None
            The metrics to report. The possible values (whether or not in a list) are:

            - if a string, either one of the built-in metrics or a scikit-learn scorer
              name. You can get the possible list of string using
              `report.metrics.help()` or :func:`sklearn.metrics.get_scorer_names` for
              the built-in metrics or the scikit-learn scorers, respectively.
            - if a callable, it should take as arguments `y_true`, `y_pred` as the two
              first arguments. Additional arguments can be passed as keyword arguments
              and will be forwarded with `metric_kwargs`. No favorability indicator can
              be displayed in this case.
            - if the callable API is too restrictive (e.g. need to pass
              same parameter name with different values), you can use scikit-learn
              scorers as provided by :func:`sklearn.metrics.make_scorer`. In this case,
              the metric favorability will only be displayed if it is given explicitly
              via `make_scorer`'s `greater_is_better` parameter.
```

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
GaetandeCast: Waiting for it to be merged

PR #2516, opened 2026-02-20
refactor(skore): Calling underlying estimator method instead of factory

glemaitre:
This PR proposes to refactor slightly the way we call the inspection methods for the cross-validation and comparison reports.

Instead to make that `_compute_data_for_display` over-complicated by accepting complex sequence, we made it only called for a single-estimator. The inner loop are put directly in the inspection method directly where we just alter the name of estimator (for the comparison) and the split index (for cross-validation).

It makes the code a bit less complex to understand and it has the advantages that for cross-validation, we call the public inspection method (e.g. permutation importance) and trigger the cache to the underlying estimator that will be stored and send subsequently to the hub.

If we don't do that, we need to make a complex reconstruction at the time of `put` while it is pretty easy to make it during the looping here. It looks like a quick win-win.

The previous tests ensure that we don't change anything in terms of user behaviour.

### DIFF
--- a/skore/src/skore/_sklearn/_comparison/inspection_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/inspection_accessor.py
@@ -82,15 +82,16 @@ class _InspectionAccessor(_BaseAccessor["ComparisonReport"], DirNamesMixin):
         21    report big alpha  Feature #9        0.4...
         >>> display.plot() # shows plot
         """
-        frames = []
-        for name, report in self._parent.reports_.items():
-            display = report.inspection.coefficients()
-            df = display.coefficients.copy()
-            df["estimator"] = name
-            frames.append(df)
-        coefficients = pd.concat(frames, ignore_index=True)
         return CoefficientsDisplay(
-            coefficients=coefficients,
+            coefficients=pd.concat(
+                [
+                    report.inspection.coefficients()
+                    .coefficients.copy()
+                    .assign(estimator=name)
+                    for name, report in self._parent.reports_.items()
+                ],
+                ignore_index=True,
+            ),
             report_type=self._parent._report_type,
         )
 
@@ -145,15 +146,16 @@ class _InspectionAccessor(_BaseAccessor["ComparisonReport"], DirNamesMixin):
         7    big trees   petal width (cm)       0.4...
         >>> display.plot() # shows plot
         """
-        frames = []
-        for name, report in self._parent.reports_.items():
-            display = report.inspection.impurity_decrease()
-            df = display.importances.copy()
-            df["estimator"] = name
-            frames.append(df)
-        importances = pd.concat(frames, ignore_index=True)
         return ImpurityDecreaseDisplay(
-            importances=importances,
+            importances=pd.concat(
+                [
+                    report.inspection.impurity_decrease()
+                    .importances.copy()
+                    .assign(estimator=name)
+                    for name, report in self._parent.reports_.items()
+                ],
+                ignore_index=True,
+            ),
             report_type=self._parent._report_type,
         )
 
--- a/skore/src/skore/_sklearn/_cross_validation/inspection_accessor.py
+++ b/skore/src/skore/_sklearn/_cross_validation/inspection_accessor.py
@@ -63,15 +63,16 @@ class _InspectionAccessor(_BaseAccessor[CrossValidationReport], DirNamesMixin):
         7       1  Feature #2       17.1...
         >>> display.plot() # shows plot
         """
-        frames = []
-        for split_idx, report in enumerate(self._parent.estimator_reports_):
-            display = report.inspection.coefficients()
-            df = display.coefficients.copy()
-            df["split"] = split_idx
-            frames.append(df)
-        coefficients = pd.concat(frames, ignore_index=True)
         return CoefficientsDisplay(
-            coefficients=coefficients,
+            coefficients=pd.concat(
+                [
+                    report.inspection.coefficients()
+                    .coefficients.copy()
+                    .assign(split=split_idx)
+                    for split_idx, report in enumerate(self._parent.estimator_reports_)
+                ],
+                ignore_index=True,
+            ),
             report_type=self._parent._report_type,
         )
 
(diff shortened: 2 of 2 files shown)


skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -688,23 +687,23 @@ def _plot_comparison(
     def _compute_data_for_display(
         cls,
         *,
-        estimators: Sequence[BaseEstimator],
-        names: list[str],
-        splits: list[int | float],
+        estimator: BaseEstimator,
+        name: str,
+        split: int | float,
```
auguste-probabl:
```suggestion
        split: int | Literal[np.nan],
```

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -715,38 +714,34 @@ def _compute_data_for_display(
...
-            coefficients.append(np.concatenate([intercept, coef]))
+        coef_data = np.concatenate([intercept, coef])
 
-            feat_names = ["Intercept"] + _get_feature_names(
-                predictor, transformer=preprocessor, n_features=coef.shape[0]
-            )
-            feature_names.extend(feat_names)
-            est_names.extend([name] * len(feat_names))
-            split_indices.extend([split] * len(feat_names))
+        feat_names = ["Intercept"] + _get_feature_names(
```
auguste-probabl:
```suggestion
        feature_names = ["Intercept"] + _get_feature_names(
```

auguste-probabl: other occurrences too

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
auguste-probabl: other occurrences too

glemaitre: Technically, NaN are floats:

```python
In [1]: import numpy as np

In [2]: type(np.nan)
Out[2]: float
```

Not sure that the type checker will be happy :) (xref: https://stackoverflow.com/questions/73231699/how-to-type-hint-np-nan#:~:text=Type%20checkers%20will,str%29%20-%3E%20str%3A&text=Option%204.%20If,or%20personal%20experience.)

auguste-probabl: Thanks for the reference. Then, could we at least do
```python
NaN = float

def f(x: int | NaN):
	"""
	x : int or np.nan
	"""
```

skore/src/skore/_sklearn/_cross_validation/inspection_accessor.py
```diff
@@ -299,39 +301,39 @@ def permutation_importance(
...
+                        data_source=data_source,
+                        X=X,
+                        y=y,
+                        at_step=at_step,
+                        metric=metric,
+                        n_repeats=n_repeats,
+                        max_samples=max_samples,
+                        n_jobs=n_jobs,
+                        seed=seed,
                     )
```
GaetandeCast: This can be factored if we rename the return of `_get_X_y_and_data_source_hash` to `X` and `y` instead of `X_` and `y_true`

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
```diff
@@ -83,32 +83,29 @@ def coefficients(self) -> CoefficientsDisplay:
...
-                splits=splits,
-                report_type=self._parent._report_type,
-            )
+                    display = estimator_report.inspection.coefficients()
+                    df = display.coefficients.copy()
+                    df["estimator"] = name
+                    df["split"] = split_idx
+                    frames.append(df)
+        coefficients = pd.concat(frames, ignore_index=True)
+        return CoefficientsDisplay(
```
GaetandeCast, line 85: We can avoid this if/else if we use the `CrossValidationReport's` inspection accessor instead of looping on the split indices.

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
```diff
@@ -161,30 +158,27 @@ def impurity_decrease(self) -> ImpurityDecreaseDisplay:
         7    big trees   petal width (cm)       0.4...
         >>> display.plot() # shows plot
         """
+        frames = []
         if self._parent._report_type == "comparison-estimator":
```
GaetandeCast: Same remark as above.

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
```diff
@@ -83,32 +83,29 @@ def coefficients(self) -> CoefficientsDisplay:
...
-                names=names,
-                splits=splits,
-                report_type=self._parent._report_type,
-            )
+                    display = estimator_report.inspection.coefficients()
+                    df = display.coefficients.copy()
+                    df["estimator"] = name
+                    df["split"] = split_idx
+                    frames.append(df)
+        coefficients = pd.concat(frames, ignore_index=True)
```
auguste-probabl: You should use the `keys` argument of `pd.concat` for this

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
GaetandeCast: The split argument is actually useless now that we only use `_compute_data_for_display` for single estimator reports. So I removed it and we assign a non `np.nan` value in the cross validation report's constructor.

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
GaetandeCast: Not sure what you mean exactly here.

auguste-probabl: something like

```suggestion
        coefficients = pd.concat(
			[
				report.inspection.coefficients().coefficients.copy()
				for report in enumerate(self._parent.estimator_reports_)
			],
			keys=range(len(self._parent.estimator_reports_)),
			names=["split"],
	    )
```

GaetandeCast: But as far as I understand it, the `keys` parameter of `pd.concat` is used to create a multi index on the rows, which I don't think is useful here. If the objective of your suggestion is to avoid the temporary storing of the frames, then this worked:
```python
        coefficients = pd.concat(
            [
                report.inspection.coefficients()
                .coefficients.copy()
                .assign(estimator=name)
                for name, report in self._parent.reports_.items()
            ],
            ignore_index=True,
        )
```
I've applied this across inspection methods and for both comparison reports and cross validation reports. Tell me if that works for you.

auguste-probabl: You can `reset_index` to take care of the multiindex. Regardless, your snippet works for me too, just remember to add the "split" column

glemaitre:
Thanks @auguste-probabl and @GaetandeCast for finishing this one.

PR #2517, opened 2026-02-22
feat(skore): add Pythonic dictionary interface and individual item deletion

KaranSinghDev:
**Description:**

Hello, 
This PR is an attempt on enhancing the `Project` class by adding a native Python dictionary interface and the ability to delete individual items/experiments.

Currently, users can use `put()` and `get()` methods, and there is no way to remove a specific experiment from a project without deleting the entire project.

**Implementation**
1. **Frontend (`skore`):** Implemented `__getitem__`, `__setitem__`, `__delitem__`, and `__contains__`, to allow users to interact with projects using standard Python syntax:
   - `project["model_v1"] = report`
   - `report = project["model_v1"]`
   - `del project["model_v1"]`
   - `"model_v1" in project`
2. **Backend (`skore-local-project`):** Implemented `delete_item()` logic. This includes a garbage collection step that only deletes binary artifacts if they are no longer referenced by any other key or project.

## For Test
I have put a dedicated test suite to verify the full lifecycle:
```bash
# Test frontend logic
python -m pytest skore/tests/unit/project/test_dict_interface.py

# Test backend deletion and garbage collection
python -m pytest skore-local-project/tests/unit/test_delete_item.py
```

## Checklist
-   Tests have been added to reflect the changes.
-   Documentation (docstrings) updated to match NumPy style.
-   Self-review conducted: Confirmed reference counting for artifact deletion.
-   All CI checks passed locally (Full suite: 1699+ tests).

I will be waiting if you have any thoughts on this  .Thanks

### DIFF
--- a/skore-local-project/src/skore_local_project/project.py
+++ b/skore-local-project/src/skore_local_project/project.py
@@ -248,6 +248,55 @@ class Project:
         )
 
     @ensure_project_is_not_deleted
+    def delete_item(self, key: str) -> None:
+        """
+        Delete a key-report pair from the local project.
+
+        Parameters
+        ----------
+        key : str
+            The key of the report to delete.
+
+        Raises
+        ------
+        TypeError
+            If the key is not a string.
+        KeyError
+            If the key does not exist in the project.
+        """
+        if not isinstance(key, str):
+            raise TypeError(f"Key must be a string (found '{type(key)}')")
+
+        # Find the specific metadata entries for THIS project and THIS key
+        metadata_ids_to_delete = []
+        artifact_ids_to_check = set()
+
+        for meta_id, value in self.__metadata_storage.items():
+            if value["project_name"] == self.name and value["key"] == key:
+                metadata_ids_to_delete.append(meta_id)
+                artifact_ids_to_check.add(value["artifact_id"])
+
+        if not metadata_ids_to_delete:
+            raise KeyError(f"Key '{key}' not found in project '{self.name}'.")
+
+        # Delete the metadata entries
+        for meta_id in metadata_ids_to_delete:
+            del self.__metadata_storage[meta_id]
+
+        # Scan ALL metadata to see if these artifacts are still referenced
+        active_artifacts = {
+            val["artifact_id"] for val in self.__metadata_storage.values()
+        }
+
+        for artifact_id in artifact_ids_to_check:
+            # No one else is using this file. Safe to delete.
+            if (
+                artifact_id not in active_artifacts
+                and artifact_id in self.__artifacts_storage
+            ):
+                del self.__artifacts_storage[artifact_id]
+
+    @ensure_project_is_not_deleted
     def get(self, id: str) -> EstimatorReport | CrossValidationReport:
         """Get a persisted report by its id."""
         if id in self.__artifacts_storage:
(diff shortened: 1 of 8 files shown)


glemaitre:
It is a tricky contribution to handle because while we could forward with the `local` backend, we need to have the same `hub` backend.

KaranSinghDev:
@glemaitre I understand the concern to keep the API unified and consistent, i would like to work on it (I have some methods but have to refine them ). So let me know if you are open to it since I think this would be really good for users . I would understand if you want to keep it on hold for now, or try to implement through a different apporach which I can try  .

glemaitre:
It is even a bit more complicated because we want to go beyond the Python dictionary like API because we have the concept of tracking reports over time and access them.

There is more discussion in this issue: https://github.com/probabl-ai/skore/issues/1269

We probably evolved from what it was at the time. But in short, we still would like to have an implicit way to store reports. For sure nowadays, we would need the `Project` because it would help for different reason and having also an explicit way to store element would be nice.

I think that we are going to comeback to this issue soon'ish indeed.

thomass-dev:
Let's draft the PR for now.

KaranSinghDev:
Ok @thomass-dev , I will try to be in loop , let me know if there is anything I could do on this

jeromedockes:
It's a matter of personal taste but I'm not sure I would want the dictionary interface for something that is so different from an in-memory mapping of some kind. For example I could be surprised by the fact that `something[key]` takes a long time or can fail for other reasons than a keyerror; I don't think a dictionary is the right mental image for projects given that they can be remote on the skore hub. In general abusing operator overloading for objects that don't closely resemble built-in types can result in confusing interfaces in my experience.

jeromedockes:
having a way to remove reports from projects does sound useful :)

KaranSinghDev:
@jeromedockes I have been thinking about it and I think it would be better with dropping that part of the interface. Hiding potential network latency and remote failure modes behind standard dictionary operators seems to violates the mental model of an in-memory mapping. 

About removal of reports, would stripping out the dict magic methods (__getitem__, __setitem__, etc.) and working on the explicit .delete_item(key) method would be good ? This is something which came up but I would like to refine before working on .

jeromedockes:
> About removal of reports, would stripping out the dict magic methods (getitem, setitem, etc.) and working on the explicit .delete_item(key) method would be good ?

I don't know what the maintainers think but I can see that it might be useful. In any case we can separate the 2 issues as they are quite independent: (i) adding delete functionality (with an explicit method rather than an operator) and (ii) introducing the mapping interface

thomass-dev:
There is a design question with the hub to tackle before going further with the delete (delete all versions of a report, delete the last one, delete one in particular etc.). And in fact, it's the same question for the local projects, as we get report by id, not by key.

Personally i'm not convinced by a delete report feature <ins>for now</ins>, as the management/display/ux of the versions of a report is not clear both on local and hub projects. Note that we want a unified UX, between local and hub projects.

Wait before we comeback to you, thanks!

In addition of the @jeromedockes comment, you should also split your PR by concern (skore/skore-local-project).

thomass-dev:
> introducing the mapping interface

For this part, let's iterate on the related issue https://github.com/probabl-ai/skore/issues/1269 mentioned by @glemaitre .

PR #2523, opened 2026-02-24
chore(docs): Auto-generate accessor method lists

### LINKED ISSUE #2390: docs: Automate API documentation of `metrics` methods

As mentioned in https://github.com/probabl-ai/skore/pull/2359#issuecomment-3842770958, the documentation for the `metrics` accessor is maintained manually for each report, which can lead to inconsistencies (see for example #2319). Since #2359 we now have some logic to go through the public methods that could be re-used to craft rST files automatically.

DCchoudhury15:
# docs: Automate API documentation for report accessors

Closes #2390

## Summary

This PR automates the maintenance of the accessor method lists (like `metrics`, `inspection`, and `data`) in our documentation.

Previously, these lists were hardcoded in the `.rst` files for `EstimatorReport`, `CrossValidationReport`, and `ComparisonReport`. This was fragile because adding a new metric in the code meant we had to remember to manually update the docs, leading to potential sync issues.

## Implementation Details

I updated the `generate_accessor_tables` Sphinx extension to handle this automatically.

- **Sentinel System:** The script now scans the target `.rst` files for specific comments (`.. # generated-accessor-start: metrics`, etc.).
- **Auto-Injection:** During the build, it inspects the report classes to find all public methods and injects an up-to-date `autosummary` block between the sentinels.
- **Safety:** The script checks if the content has actually changed before writing to disk (idempotency), so we don't trigger unnecessary rebuilds. It also disables parallel writes to avoid race conditions.

I've also updated `conf.py` with a new `accessor_rst_reports` config to map the classes to their respective documentation files.

## verification

I ran `make html` locally and confirmed that:
- [x] The `.rst` files are correctly updated with the full list of methods.
- [x] The indentation in the generated blocks matches the existing 4-space style.
- [x] Repeated builds do not modify the files if nothing has changed.


auguste-probabl (changes requested):
Thanks for the PR! The `.. # generated-accessor-start` mechanic seems a bit too complex, it could be replaced by generating files, like we do in `generate_accessor_tables`.

DCchoudhury15:
@auguste-probabl i was unaware of  the complexity. The regex/sentinel approach was definitely overkill. I'll refactor it to just generate separate RST files for the accessors and then .. include them in the reports. Much cleaner that way. Updating it now!

sphinx/sphinxext/generate_accessor_tables.py
```diff
@@ -173,9 +205,8 @@ def setup(app: Sphinx) -> dict[str, Any]:
...
     app.add_config_value("accessor_summary_classes", [], "html")
     app.connect("config-inited", generate_accessor_tables)
-
     return {
-        "version": "0.1",
+        "version": "0.3",
         "parallel_read_safe": True,
-        "parallel_write_safe": True,
+        "parallel_write_safe": False,
     }
```
auguste-probabl: Can you explain what this does?

DCchoudhury15: Sorry about that, I added it thinking the file writes during the build could cause issues in parallel builds, but looking at it more carefully the writes happen in config-inited before any of that, so it doesn't actually do anything useful here. Reverting it now.

DCchoudhury15:
@auguste-probabl apologies for the same error i am investigating and will put up a working solution asap

auguste-probabl:
Okay awesome, the docs build!

So I went and looked at the rendered docs, and the sidebar is different:

| Before | After |
| ------------- | ------------- |
| image | image |




I think it's nice to only put the method name, it's lighter; however there are conflicts, like `help()`, and also it's clear that the methods belong to accessors. I think it could work if we could put the accessor bases together with their methods, e.g.

```
CrossValidationReport
CrossValidationReport.help
CrossValidationReport.cache_predictions
...
CrossValidationReport.data
help
analyze
CrossValidationReport.inspection
help
coefficients
impurity_decrease
permutation_importance
CrossValidationReport.metrics
help
summarize
accuracy
brier_score
...
```

@DCchoudhury15 if you want to try to make it work this way that would be cool to see; otherwise, you should make the docs render the same as they are in `main`.

sphinx/sphinxext/generate_accessor_tables.py
```diff
@@ -128,35 +159,86 @@ def generate_accessor_tables(app: Sphinx, config: Any) -> None:
         The Sphinx configuration object.
     """
     classes_to_process = config.accessor_summary_classes
-
     if not classes_to_process:
-        raise ValueError("accessor_summary_classes not found")
+        return
```
auguste-probabl: I'm curious: why was this changed?

DCchoudhury15: You're right, no good reason to change that — it should fail loudly if the config is missing. Reverted it back.

DCchoudhury15:
@auguste-probabl i have done the changes, pr up for review.

Issue #2134, opened 2025-11-02
feat(skore): Integration with MLFlow


glemaitre:
### What would you like to say?

MLFlow is a reference when it comes to model registry. They also have their own experiment tracker that focus on a single model experiment tracking.

In terms of integration with `skore` and the `skore-*-project`, it looks like there is two possible integrations:

- A sort of back tracking models that have been stored by MLFlow. It means that skore could provide a way to load those models, create any report that could be then stored if needed.
- Experiments could be carried with skore and at the time of storing reports, some artifacts could be stored in the MLFlow instance.

So this issue would definitely needs a bunch of work to investigate which use case a user of MLFlow would be expecting when using skore.

glemaitre:
@cakedev0 is currently looking at this issue when the potential integration. We are going to provide more details on sub-tasking once we know more.

cakedev0:
We are moving forward with the following option:

> Experiments can be run with skore and, when storing reports, some artifacts can be stored in the MLflow instance.

More details can be found in:
- Step 1 described here: https://github.com/probabl-ai/skore-hub/discussions/1192
- Option 1 described here: https://github.com/probabl-ai/ml-flow-poc/blob/main/investigation_report.md
- The POC implementation here (especially the `skomlflow` folder): https://github.com/probabl-ai/ml-flow-poc/

**The goal is to build a `skore-mlflow-project`.**

The current implicit contract between `skore` and Project plugins reduces the scope
of choices we can make (which is good for a first step, but will likely lack flexibility in the long run; we will revisit this later).

**Proposed interface:**
- `Project.__init__(name=None, *, tracking_uri=None)`:
  - Do not require an experiment name; apparently it is already set by default in some cases (like MLflow in Azure ML).
  - Do not require a tracking URI for the same reasons.
- `Project.put(key: str, report)`: maps report concepts to MLflow concepts and logs them:
  - The mapping used in the POC is fairly easy to read in the [POC](https://github.com/probabl-ai/ml-flow-poc/skomlflow/) (`metrics.py` and `reports.py`).
    - It contains a number of arbitrary choices. It should be fine to iterate on those based on user feedback.
    - For now, it won't try inspecting the cache to upload custom metrics or the long-to-compute permutation importance.
      This should be addressed later.
  - Logging is fairly simple: just call the appropriate `mlflow.log_xxx(...)` function.
  - **logging context handling:**
    - By default, it creates a new run, using `run_name=key`.
    - If a run is active, two options:
[...]

PR #2527, opened 2026-02-25
feat(skore-mlflow-project): Add integration with MLflow

cakedev0:
Implements #2134

#### Initial interface proposition

https://github.com/probabl-ai/skore/issues/2134#issuecomment-3959634747
Not much has changed, as anyway, the current plugin interface constraints the choices quite a lot (which is convenient for a first implementation)

#### Remaining questions/TODOs:
- [X] handling of logging context: decide what do if there is already an active run:
  - (selected behavior :heavy_check_mark:) crash
  - OR: log only artifacts (i.e. only plots and HTMLs) and ignore key.
  - OR: create a nested run
  - OR (initial behavior): ignore it and create a new independent run

- [x] Documentation/Examples: in PR https://github.com/probabl-ai/skore/pull/2532.

#### Review guide

The diff is scary, but it mainly comes from CI things, the actual code diff is something like +600 (excluding tests).

- `skore-mlflow-project/`: the main part of the implementation, it mainly:
  - inspects report objects and call all the metrics/plots/...
  - logs those metrics/plots/... into MLFlow 
  - @glemaitre tested this part, I think we can consider it :heavy_check_mark: 
- `skore/`: :heavy_check_mark: 
  - small changes to plug `skore.Project` with `skore-mlflow-project/`
  - @thomass-dev has quickly reviewed this part, and I addressed his two comments
- CI changes :heavy_check_mark: 
  - basic changes to run tests `skore-mlflow-project` => validated by @thomass-dev 
  - more complex changes to cover many MLFlow versions are in a separate PR: #2531 
#### Coverage report from local coverage run:

```
Name                                      Stmts   Miss  Cover   Missing
-----------------------------------------------------------------------
src/skore_mlflow_project/__init__.py          5      0   100%
src/skore_mlflow_project/_matplotlib.py      15      0   100%
src/skore_mlflow_project/project.py         191      5    97%   200, 231-232, 377, 379
src/skore_mlflow_project/protocol.py         36      0   100%
src/skore_mlflow_project/reports.py         145      3    98%   203-204, 264
-----------------------------------------------------------------------
TOTAL                                       392      8    98%
```

### DIFF
--- a/examples/use_cases/plot_employee_salaries.py
+++ b/examples/use_cases/plot_employee_salaries.py
@@ -156,7 +156,7 @@ data_display.plot(kind="corr")
 
 # %%
 hgbt_split_1 = hgbt_model_report.estimator_reports_[0]
-hgbt_split_1.metrics.summarize(favorability=True).frame()
+hgbt_split_1.metrics.summarize().frame(favorability=True)
 
 # %%
 # The favorability of each metric indicates whether the metric is better
--- a/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
@@ -169,7 +169,9 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
                 results.index = results.index.str.replace(
                     r"\((.*)\)$", r"\1", regex=True
                 )
-        return MetricsSummaryDisplay(results)
+        return MetricsSummaryDisplay(
+            data=results, report_type=self._parent._report_type
+        )
 
     def _compute_metric_scores(
         self,
@@ -199,15 +201,32 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
                 )
             )
 
-            kwargs = dict(
-                data_source=data_source,
-                **metric_kwargs,
-            )
+            kwargs = dict(data_source=data_source, **metric_kwargs)
             if is_cv_report:
                 kwargs["aggregate"] = None
 
+            # FIXME(#1837)
+            # "favorability" and "flat_index" are passed to `.frame()` for
+            # EstimatorReports while they are passed to `.summarize()` for
+            # CrossValidationReports
+            frame_kwargs = {}
+            if not is_cv_report:
+                for key in ("favorability", "flat_index"):
+                    if key in kwargs:
+                        frame_kwargs[key] = kwargs.pop(key)
+
+            if "favorability" in kwargs:
+                favorability = kwargs["favorability"]
+            elif "favorability" in frame_kwargs:
+                favorability = frame_kwargs["favorability"]
+            else:
+                favorability = False
+            favorability = cast(bool, favorability)
+
             individual_results = [
-                result.frame() if report_metric_name == "summarize" else result
+                result.frame(**frame_kwargs)
+                if report_metric_name == "summarize"
+                else result
                 for result in track(
                     parallel(
                         joblib.delayed(getattr(report.metrics, report_metric_name))(
@@ -224,14 +243,14 @@ class _MetricsAccessor(_BaseMetricsAccessor, _BaseAccessor, DirNamesMixin):
                 results = _combine_estimator_results(
                     individual_results,
                     estimator_names=self._parent.reports_.keys(),
-                    favorability=metric_kwargs.get("favorability", False),
+                    favorability=favorability,
                     data_source=data_source,
                 )
             else:  # "CrossValidationReport"
                 results = _combine_cross_validation_results(
                     individual_results,
                     estimator_names=self._parent.reports_.keys(),
-                    favorability=metric_kwargs.get("favorability", False),
+                    favorability=favorability,
                     aggregate=aggregate,
                 )
 
--- a/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
@@ -168,7 +168,7 @@ class _MetricsAccessor(
                 results.index = results.index.str.replace(
                     r"\((.*)\)$", r"\1", regex=True
                 )
-        return MetricsSummaryDisplay(summarize_data=results)
+        return MetricsSummaryDisplay(data=results, report_type="cross-validation")
 
     def _compute_metric_scores(
         self,
@@ -196,8 +196,15 @@ class _MetricsAccessor(
                 )
             )
 
+            frame_kwargs = {}
+            for key in ("favorability", "flat_index"):
+                if key in metric_kwargs:
+                    frame_kwargs[key] = metric_kwargs.pop(key)
+
             results = [
-                result.frame() if report_metric_name == "summarize" else result
+                result.frame(**frame_kwargs)
+                if report_metric_name == "summarize"
+                else result
                 for result in track(
                     parallel(
                         delayed(getattr(report.metrics, report_metric_name))(
@@ -220,7 +227,7 @@ class _MetricsAccessor(
             # Pop the favorability column if it exists, to:
             # - not use it in the aggregate operation
             # - later to only report a single column and not by split columns
-            if metric_kwargs.get("favorability", False):
+            if frame_kwargs.get("favorability", False):
                 favorability = results.pop("Favorability").iloc[:, 0]
             else:
                 favorability = None
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -1,12 +1,12 @@
 import inspect
-from collections.abc import Callable, Iterable
+from collections.abc import Callable
 from functools import partial
 from typing import Any, Literal, cast
 
 import numpy as np
 import pandas as pd
-from numpy.typing import ArrayLike, NDArray
-from sklearn import metrics as sklearn_metrics
+import sklearn
+from numpy.typing import ArrayLike
 from sklearn.metrics._scorer import _BaseScorer
 from sklearn.utils.metaestimators import available_if
 
@@ -40,7 +40,6 @@ from skore._utils._accessor import (
     _get_ys_for_single_report,
 )
 from skore._utils._cache_key import deep_key_sanitize
-from skore._utils._index import flatten_multi_index
 
 
 class _MetricsAccessor(
@@ -67,8 +66,6 @@ class _MetricsAccessor(
         metric_kwargs: dict[str, Any] | None = None,
         response_method: str | list[str] | None = None,
         pos_label: PositiveLabel | None = _DEFAULT,
-        favorability: bool = False,
-        flat_index: bool = False,
     ) -> MetricsSummaryDisplay:
         """Report a set of metrics for our estimator.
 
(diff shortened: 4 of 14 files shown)


skore/src/skore/_project/project.py
```diff
@@ -305,7 +322,20 @@ def delete(name: str, *, mode: ProjectMode = "local", **kwargs):
...
         """
+        if mode not in Project.__DELETE_KWARGS_BY_MODE:
+            raise ValueError(
+                f'`mode` must be "hub", "local" or "mlflow" (found {mode})'
+            )
+
+        unsupported_kwargs = set(kwargs) - Project.__DELETE_KWARGS_BY_MODE[mode]
+        if unsupported_kwargs:
+            unsupported = ", ".join(sorted(f"{key!r}" for key in unsupported_kwargs))
+            raise TypeError(f"Unexpected keyword argument(s): {unsupported}")
```
thomass-dev: Please do not control the kwargs here.
This creates another dependency between `skore` and the plugins.
Let each plugin manage its own kwargs as before, and raise an exception.
The docstr section on kwargs should be sufficiently explicit.

skore-mlflow-project/LICENSE
thomass-dev, line 1: This file is not needed, as the LICENSE is copied from root by the hatch hook during building.

You have to add an entry in the `.gitignore` file:

```
.gitignore:skore/LICENSE
.gitignore:skore-hub-project/LICENSE
.gitignore:skore-local-project/LICENSE
```

skore/src/skore/_project/project.py
cakedev0: Oops sorry, removed that.

skore-mlflow-project/LICENSE
cakedev0: > You have to add an entry in the .gitignore file:

Thx! Did that.

glemaitre:
How much work would it be to include a small examples in the "technical details" section to show that we can push into MLflow using this backend.

I would naively think that it should be feasible by starting the MLflow server before starting the generation of the examples.

I'll start to do a review.

examples/technical_details/plot_mlflow_backend.py
```diff
@@ -0,0 +1,69 @@
...
+
+from sklearn.datasets import make_classification
+from sklearn.linear_model import LogisticRegression
+from skore import EstimatorReport, Project, train_test_split
+from skrub import tabular_pipeline
+
+# %%
+# Build one report to persist
+# ===========================
+X, y = make_classification(
```
glemaitre: I used locally the getting starting example:

<details>

```python
# %%
from sklearn.datasets import fetch_openml
from skrub import TableReport

german_credit = fetch_openml(data_id=31, as_frame=True, parser="pandas")
X, y = german_credit.data, german_credit.target
TableReport(german_credit.frame)

# %%

estimator = tabular_pipeline(LogisticRegression())
report = CrossValidationReport(
    estimator, X, y
)

# %%
from skore import Project

project = Project(
    name="german_credit", mode="mlflow", tracking_uri="http://localhost:5001"
)

# %%
project.put("log_reg", report)

```

</details>

It fails with

```pytb
ValueError: y_true takes value in {'bad', 'good'} and pos_label is not specified: either make y_true take value in {0, 1} or {-1, 1} or pass pos_label explicitly.
```

It is normal but I had the impression that we did quite some computation before failing. In the hub, backend we added an early failures in `put`:

```python
        if report.ml_task == "binary-classification":
            # check that pos_label is either specified or can be inferred from the data
            if isinstance(report, EstimatorReport):
                target = report.estimator_.classes_
            else:  # CrossValidationReport
                target = report.estimator_reports_[0].estimator_.classes_

            try:
                _check_pos_label_consistency(report.pos_label, target)
            except ValueError as exc:
                raise ValueError(
                    "For binary classification, the positive label must be specified. "
                    "You can set it using `report.pos_label = <positive_label>`."
                ) from exc
```

cakedev0: It doesn't feel great to re-implement this logic a third time... :sweat_smile: Should be only done once on the skore's side IMO...

cakedev0: And btw, I use `"sqlite:///mlflow.db"` and it fails after 2s, I'm not sure how to have it work with `"http://localhost:5001"` for the tacking URI...

cakedev0:
I think the current biggest problem is the number of warnings that mlflow tiggers, though I already had that with just using mlflow, so I guess MLFlow users are used to it :sweat_smile: but I'll still try improving it.

skore/src/skore/_project/login.py
```diff
@@ -18,9 +18,9 @@ def login(*, mode: ProjectMode = "hub", **kwargs):
 
     It must be called at the top of your script.
 
-    Two storage backends are available and their credentials can be configured using the
-    ``mode`` parameter. Please note that both can be configured in the same script
-    without side-effect.
+    Three storage backends are available and their credentials can be configured
```
auguste-probabl:
```suggestion
    Several storage backends are available and their credentials can be configured
```

examples/technical_details/plot_mlflow_backend.py
```diff
@@ -0,0 +1,63 @@
...
+    # example page focuses on skore usage rather than backend startup details.
+    with redirect_stdout(io.StringIO()), redirect_stderr(io.StringIO()):
+        project = Project(
+            "technical-details-mlflow",
+            mode="mlflow",
+            tracking_uri=backend_store_uri,
+        )
+        project.put("linreg-baseline", report)
+        summary = project.summarize()
+    summary
```
auguste-probabl: That doesn't seem to print anything in the rendered docs

auguste-probabl (commented):
Thanks, excited to see this in action!

.github/workflows/sphinx.yml
```diff
@@ -222,7 +222,11 @@ jobs:
 
       - name: Install
         working-directory: skore
-        run: python -m pip install --force-reinstall --no-deps .
+        run: |
+          set -eu
+          python -m pip install --force-reinstall --no-deps .
+          python -m pip install --upgrade "mlflow==3.10.0"
+          python -m pip install --force-reinstall --no-deps ../skore-mlflow-project
```
thomass-dev: I'm okay to see this during the coding process, but i am less so for merging it. It's an unstable change, and that goes against how the CI currently works: all tests are made with released dependencies to replicate the user environment. Also, especially for the documentation, we ensure that way we communicate on feature that is available for the user.

With this change, we could easily end up in a situation where a new example of `skore-mlflow-project` is deployed on the documentation, but `skore-mlflow-project` is not yet released.

Instead, you should add `skore-mlflow-project` in `skore/pyproject.toml` as a `sphinx` dependencies, and regenerate the `sphinx` lockfile.

But it implies to do it in two time/split your PR in two: 
1. we merge the PR with the new package, and we release the package,
2. we merge the PR with the new example.

thomass-dev: Otherwise, LGTM for all changes outside the `skore-mlflow-project/` directory (i reviewed only CI and `skore` parts).

cakedev0: Splitting in another PR makes perfect sense :+1: , that was my initial plan to do example(s) in a separate PR.

Did that.

examples/technical_details/plot_mlflow_backend.py
cakedev0: Indeed, I need to make a dataframe from it. I moved the example in a separate PR: https://github.com/probabl-ai/skore/pull/2532

examples/technical_details/plot_mlflow_backend.py
glemaitre: > Should be only done once on the skore's side IMO...

Fine with this. We can just come this in the `put` in `skore` then.

glemaitre: > And btw, I use "sqlite:///mlflow.db" and it fails after 2s, I'm not sure how to have it work with "http://localhost:5001" for the tacking URI...

I just started the server and UI with:

```shell
mlflow ui --host 0.0.0.0 --port 5001 --disable-security-middleware
```

and it uses the sqlite database by default. I used 5001 port because I had something that was already up on the 5000 port.

Not sure that it answers the question.

glemaitre:
I see the following things that have been uploaded when doing a cross-validation:

image

Is there a way to avoid to drop the single metric CSV since we already have this info in the `all_metrics.csv`. To give some perspective, in the future, we intend to have a metric registry which mean that this "metric summary" should contain all necessary metric defined for the project.

When looking at a split, I see the following artifacts:

image

It makes me think that we could organize the dump data aligned with the name of the report accessor (data, inspection, metrics). Then, for dropping the info of a specific display, I would reuse `frame` and `plot` folders that would recall the methods of the display.

thomass-dev:
Note for the next iteration (and re-transcription of an informal discussion):

If the plugin/multi-package layout is holding back your development, you could develop the `mlflow` backend in a python module. For instance inside `skore` : `skore/skore/src/skore/mlflow`, and you call it directly in `skore.Project.__init__ ` with something like `if mode == 'mlflow': self.__project = skore.mlflow.Project(...)`.
 
We can live very well with both during the PoC (`mlflow` in module, other backends in plugin) if it helps you/makes you faster, as long as you comply with the project API.

No need to stick to the current architecture for now.

cakedev0:
@glemaitre 

> Is there a way to avoid to drop the single metric CSV since we already have this info in the all_metrics.csv

Yes, did that :+1: (and renamed all_metrics.csv to simply metrics.csv)

Also, now naming of artifact is as follow:
image

PS: a metrics registry sounds like a great idea. MLFlow has something a bit like that, could be used as an inspiration: https://mlflow.org/docs/latest/ml/evaluation/#custom-metrics--visualizations

glemaitre:
> PS: a metrics registry sounds like a great idea. MLFlow has something a bit like that, could be used as an inspiration

It is really close to what @thomass-dev starting to do: https://github.com/probabl-ai/skore/pull/2277

The only major difference is that we request the response method because we are in a higher abstraction level than MLflow in this case.

skore-mlflow-project/src/skore_mlflow_project/_matplotlib.py
```diff
@@ -0,0 +1,32 @@
...
+from __future__ import annotations
+
+from collections.abc import Iterator
+from contextlib import contextmanager
+
+import matplotlib.pyplot as plt
+
+
+@contextmanager
+def switch_mpl_backend(backend: str = "agg") -> Iterator[None]:
```
rouk1, line 12: As this function exists in `skore-hub-project` aswell maybe this could be factorized in `skore` itself wdyt ?

cakedev0:
Thanks for giving it a try. I don't know why it doesn't work for you, I tried reproducing the failure:

In one terminal, from repo root `uvx mlflow server`
In a second one, from repo root too:
```
uv venv .matt-mlflow
source .matt-mlflow/bin/activate
uv pip install -e "./skore[test,jupyter]" -e "./skore-hub-project[test]" -e "./skore-mlflow-project"
uv run matt.py  # I copied your snippet in there
```

It worked and I could find the run/report (and sub-runs for each split) in the UI.

----

Note that `project = Project("simple-project", mode="mlflow")` lets `mlflow` select a default `tracking_uri` (it happened to work as expected in this case). I except that people with a mlflow setup will already have import `mlflow.set_tracking_uri(...)` in their script/notebook already, or use the kwarg: `Project("simple-project", mode="mlflow", tracking_uri=...)`. `tracking_uri` is documented in the docstring but it's easy to miss because it's at the end after the hub things, I think it's a big inconvenient of having the same class for every backend (and not sub-classes or "sub-interfaces").

skore-mlflow-project/src/skore_mlflow_project/_matplotlib.py
thomass-dev: In the next iteration ;).

cakedev0:
Nice catch, fixed the glitch! It was a mlflow thing.

examples/technical_details/plot_mlflow_backend.py
cakedev0: Ah I missed your last answer on this thread.

Yes that was the question, it possible that using "sqlite:///mlflow.db" for the tracking URI is fastest than using "http://localhost:5001" (less serialization steps maybe?). Anyway, "real users" are rather expected to use a non-local MLFlow server.

I'll open an issue so we don't forget about it this thread.

PR #2530, opened 2026-02-26
fix(skore): Ensure figure-level legend is included when saving PredictionErrorDisplay plot

### LINKED ISSUE #2505: The legend of `PredictionErrorDisplay` is cut off when saved with `savefig`
### Describe the bug

For the `PredictionErrorDisplay`, when one wants to save the figure with `plt.savefig()`, the resulting file shows the plot with the legend cut off. This is solved by doing `plt.savefig("name.png", bbox_inches="tight")` but it should not be the user's responsibility to do this, rather ours to correctly plot.

### Steps/Code to Reproduce

```python
X, y = make_regression()
split_data = train_test_split(X, y, test_size=0.2, random_state=0, as_dict=True)
report_1 = EstimatorReport(RandomForestRegressor(), **split_data)

display_1 = report_1.metrics.prediction_error(data_source="both")
display_1.plot()
plt.savefig("pred_error.png")
```

### Expected Behavior

Image

### Actual Behavior

Image

### Environment

```shell
System:
    python: 3.14.0 (main, Nov 19 2025, 22:48:15) [Clang 21.1.4 ]
executable: /home/gdc/dev/skore/.venv/bin/python
   machine: Linux-6.14.0-37-generic-x86_64-with-glibc2.39

Python dependencies:
        skore: 0.0.0+unknown
          pip: None
       jinja2: 3.1.6
       joblib: 1.5.3
   matplotlib: 3.10.8
        numpy: 2.4.2
       pandas: 2.3.3
       plotly: 6.5.2
         rich: 14.3.2
 scikit-learn: 1.8.0
      seaborn: 0.13.2
 skore[local]: None
        skrub: 0.7.2
skore-hub-project: 0.0.16
```

MuditAtrey:
Fixes #2505

## Problem
`PredictionErrorDisplay.plot()` creates a figure-level legend via `self.figure_.legend()` with `bbox_to_anchor=(0.5, 0)`, placing it below the figure boundary. matplotlib's `savefig` excludes figure-level legends from boundary calculation by default, causing the legend to be clipped in saved output.

Confirmed via inspecting the legend's bounding box after `plot()`:
```python
print(display.figure_.legends[0].get_window_extent())
# Bbox(x0=211.875, y0=-67.0, x1=392.125, y1=0.0)
```
`y1=0.0` confirms the legend sits exactly at the bottom edge — below the figure boundary.

Note: `tight_layout(rect=...)` and `subplots_adjust` were both investigated but neither accounts for figure-level legends — the bounding box remained at `y0=-67, y1=0` in both cases.

## Fix
Move the legend from figure level to axis level via `ax_[len(ax_)//2].legend()`, so it is included in `tight_layout` computation. Increase figure height by `1.2 inches` to make room for the legend below the plot. This avoids patching `savefig` and prevents pickling issues.

## Testing
Verified across:
- Basic save: `plt.savefig("file.png")`
- `kind="actual_vs_predicted"`
- Single data source
- High DPI: `dpi=300`
- PDF output
- Multiple estimators comparison
- Small figure size

## Before / After
image
image

### DIFF
--- a/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
+++ b/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
@@ -297,7 +297,7 @@ class PredictionErrorDisplay(DisplayMixin):
 
         labels.append("Perfect predictions")
 
-        self.ax_[len(self.ax_)//2].legend(
+        self.ax_[len(self.ax_) // 2].legend(
             handles,
             labels,
             loc="upper center",
@@ -305,8 +305,10 @@ class PredictionErrorDisplay(DisplayMixin):
             ncol=1,
             frameon=True,
         )
+
         if len(self.ax_) == 1:
             self.ax_ = self.ax_[0]
+
         w, h = self.figure_.get_size_inches()
         self.figure_.set_size_inches(w, h + 0.25 * len(labels))
         self.figure_.tight_layout()
--- a/skore/tests/unit/displays/prediction_error/test_comparison_cross_validation.py
+++ b/skore/tests/unit/displays/prediction_error/test_comparison_cross_validation.py
@@ -11,7 +11,7 @@ def test_legend(pyplot, task, legend_prefix, request):
     figure, _ = request.getfixturevalue(
         f"comparison_cross_validation_reports_{task}_figure_axes"
     )
-    legend = figure.axes[len(figure.axes)//2].get_legend()
+    legend = figure.axes[len(figure.axes) // 2].get_legend()
     assert legend is not None
     legend_texts = [t.get_text() for t in legend.get_texts()]
     assert len(legend_texts) == 3
@@ -29,7 +29,8 @@ def test_legend_actual_vs_predicted(pyplot, task, legend_prefix, request):
     report = request.getfixturevalue(f"comparison_cross_validation_reports_{task}")
     display = report.metrics.prediction_error()
     display.plot(kind="actual_vs_predicted")
-    legend_texts = [t.get_text() for t in display.figure_.axes[len(display.figure_.axes)//2].get_legend().get_texts()]
+    legend = display.figure_.axes[len(display.figure_.axes) // 2].get_legend()
+    legend_texts = [t.get_text() for t in legend.get_texts()]
     assert len(legend_texts) == 3
     assert legend_texts[0] == f"{legend_prefix} #0"
     assert legend_texts[1] == f"{legend_prefix} #1"
--- a/skore/tests/unit/displays/prediction_error/test_comparison_estimator.py
+++ b/skore/tests/unit/displays/prediction_error/test_comparison_estimator.py
@@ -10,7 +10,7 @@ def test_legend(pyplot, task, n_legend_entries, request):
     figure, _ = request.getfixturevalue(
         f"comparison_estimator_reports_{task}_figure_axes"
     )
-    legend = figure.axes[len(figure.axes)//2].get_legend()
+    legend = figure.axes[len(figure.axes) // 2].get_legend()
     assert legend is not None
     legend_texts = [t.get_text() for t in legend.get_texts()]
     assert len(legend_texts) == n_legend_entries
@@ -29,7 +29,8 @@ def test_legend_actual_vs_predicted(pyplot, task, n_legend_entries, request):
     report = request.getfixturevalue(f"comparison_estimator_reports_{task}")
     display = report.metrics.prediction_error()
     display.plot(kind="actual_vs_predicted")
-    legend_texts = [t.get_text() for t in display.figure_.axes[len(display.figure_.axes)//2].get_legend().get_texts()]
+    legend = display.figure_.axes[len(display.figure_.axes) // 2].get_legend()
+    legend_texts = [t.get_text() for t in legend.get_texts()]
     assert len(legend_texts) == n_legend_entries
     if task == "multioutput_regression":
         assert legend_texts[0] == "Output #0"
@@ -94,7 +95,8 @@ def test_subplot_by_data_source(pyplot, task, request):
     else:
         display.plot(subplot_by="data_source")
         assert len(display.ax_) == 2
-        legend_texts = [t.get_text() for t in display.figure_.axes[len(display.figure_.axes)//2].get_legend().get_texts()]
+        legend = display.figure_.axes[len(display.figure_.axes) // 2].get_legend()
+        legend_texts = [t.get_text() for t in legend.get_texts()]
         assert len(legend_texts) == 3
         assert legend_texts[0] == "DummyRegressor_1"
         assert legend_texts[1] == "DummyRegressor_2"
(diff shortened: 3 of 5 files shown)


skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
```diff
@@ -305,6 +305,15 @@ def _plot_matplotlib(
...
         )
+        
+        # figure-level legends are excluded from savefig's boundary calculation by default;
+        # patching savefig to use bbox_inches="tight" ensures the legend is included when saving.
+
+        _original_savefig = self.figure_.savefig
+        def _savefig_with_legend(*args, **kwargs):
+            kwargs.setdefault("bbox_inches", "tight")
+            _original_savefig(*args, **kwargs)
+        self.figure_.savefig = _savefig_with_legend
```
jeromedockes: sorry for arriving late to the discussion but I do not think we should patch a method on the figure, especially not with a local function as it will cause issues for pickling. Instead I would suggest attaching the legend to the axis so it is taken into account in the tight_layout computation and increasing slightly the figure height to make room for it

```diff
diff --git a/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py b/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
index ad35a22cf..f06cc86f9 100644
--- a/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
+++ b/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
@@ -297,11 +297,11 @@ class PredictionErrorDisplay(DisplayMixin):
 
         labels.append("Perfect predictions")
 
-        self.figure_.legend(
+        self.ax_[-1].legend(
             handles,
             labels,
             loc="upper center",
-            bbox_to_anchor=(0.5, 0),
+            bbox_to_anchor=(0.5, -.2),
             ncol=1,
             frameon=True,
         )
@@ -309,6 +309,10 @@ class PredictionErrorDisplay(DisplayMixin):
         if len(self.ax_) == 1:
             self.ax_ = self.ax_[0]
 
+        w, h = self.figure_.get_size_inches()
+        self.figure_.set_size_inches(w, h + 1.2)
+        self.figure_.tight_layout()
+
     def _get_plot_columns(
         self,
         subplot_by: Literal["auto", "estimator", "data_source", "split", "output"]
```

jeromedockes: @GaetandeCast does it help in your environment? seems to work ok for me

GaetandeCast: Yeah I don't like overwriting `savefig` either. Your solution works well on my side too!
I suggest changing:
```
self.ax_[-1].legend(
```
to:
```
self.ax_[len(self.ax_)//2].legend(
```
To have the legend on (one of) the center ax.

GaetandeCast: @MuditAtrey please implement this change instead. This will require updating tests since the location of the legend will have changed from figure to axis level. If you're not comfortable doing all this, I can take over the PR if you want.

jeromedockes: > To have the legend on (one of) the center ax.

good idea -- just a question are they always on a single row?

if we want perfectly centered in the case of even number of axes we could add a zero-height invisible axis at the bottom of the grid and use that to set the legend but probably not worth the hassle

MuditAtrey: Thanks for the review @jeromedockes and @GaetandeCast! Moving the legend to axis level avoids the pickling issue entirely. I'll implement the changes with `ax_[len(self.ax_)//2]` and the figure height adjustment this evening. Please give me a few hours.

GaetandeCast: @jeromedockes Yes, currently there is only a single row. If we eventually get around letting the user set their `(row, col)` choice, we will have to update this.
>if we want perfectly centered in the case of even number of axes we could add a zero-height invisible axis at the bottom of the grid and use that to set the legend but probably not worth the hassle.

Maybe at a second stage :)

No worries @MuditAtrey, take your time.

MuditAtrey: @jeromedockes @GaetandeCast `h + 1.2` adds a noticeable height increase even for single subplot cases, would a smaller value like `0.8` be more appropriate, or should this be dynamic based on legend height?
one note though: `0.8` makes the legend clip with the axes
the extra height seems to be also visible in the normal display out in jupyter.
one of the test script below: 
```python
from sklearn.datasets import *
from skore import *
from sklearn.ensemble import *
import matplotlib.pyplot as plt

X, y = make_regression()
split_data = train_test_split(X, y, test_size=0.2, random_state=0, as_dict=True)
report_1 = EstimatorReport(RandomForestRegressor(), **split_data)

display_1 = report_1.metrics.prediction_error(data_source="both")
display_1.plot()
plt.savefig("test_basic.png")
```


suggested solution vs the my current proposal:

image
image

----

image
image




Tested with `0.8`, legend gets clipped. `1.2` seems to be the minimum needed to fully show the legend, the height increase is necessary.



Posting this comment for transparency and approval by the maintainers before committing to the changes

MuditAtrey: working on updating the tests

GaetandeCast: From testing on my side, using `bbox_to_anchor=(0.5, -0.3),` with `self.figure_.set_size_inches(w, h + 0.25*len(labels))` seems to be the most robust in terms of clipping. Even for regression with 10 outputs or for 10 cv folds it look ok.

image
image
image

MuditAtrey: @GaetandeCast Updated, now using `bbox_to_anchor=(0.5, -0.3)` with `h + 0.25 * len(labels)` for height as you suggested. Tests updated for axis-level legend and all 117 passing.

skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
```diff
@@ -297,17 +297,19 @@ def _plot_matplotlib(
...
+        self.ax_[len(self.ax_)//2].legend(
             handles,
             labels,
             loc="upper center",
-            bbox_to_anchor=(0.5, 0),
+            bbox_to_anchor=(0.5, -0.3),
             ncol=1,
             frameon=True,
         )
-
```
GaetandeCast, line 308: keep this blank line

skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
```diff
@@ -297,17 +297,19 @@ def _plot_matplotlib(
...
         if len(self.ax_) == 1:
             self.ax_ = self.ax_[0]
+        w, h = self.figure_.get_size_inches()
```
GaetandeCast, line 312: add an empty line between the blocks

skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
MuditAtrey: I installed the pre-commit AFTER doing the commit, running the pre-commit locally first, before committing, apologise for the inconvenience I am causing

GaetandeCast: No worries, it's a learning process :wink:

PR #2537, opened 2026-02-27
feat(skore): Remove the datasource=xy option

### LINKED ISSUE #2526: Discussion: disallowing using different data with the same report?

I wonder if we could reconsider the possibility to pass new datasets to an existing report (either by setting `X_train` or `X_test` in an EstimatorReport, or by using `data_source="X_y"`). The drawbacks I see are:

- It "muddies" a bit the nature of the report: it is not only a report about a past experiment, but it can actually be used like a fitted estimator, augmented with caching, for making predictions on new data. I think that can make it a bit harder to form a mental picture of what the EstimatorReport is and how it should be used
- It complexifies the interface of `get_predictions` and many other functions such as `accuracy` by having the additional parameters X and y, which interact with each other and the `data_source` parameter ("X can only be passed if data_source is X_y" ...)
- It complexifies the caching logic, due to handling multiple datasets and the possibility that the cache needs to be flushed when the datasets are mutated

When we have new data, the only thing that can really be reused in the report is the fitted estimator. This is easily done in any case by creating a new report, as it can accept a fitted estimator:

```
new_report = EstimatorReport(old_report.estimator_, fit=False, X_test=..., y_test=...)
```

(briefly discussed IRL with @glemaitre )

jeromedockes:
closes #2526 

#### Change description

Still WIP, a first pass to see what it would look like to remove the `data_source="X_y"` option

#### Contribution checklist

requires updating tests)

#### AI usage disclosure

AI tools were involved for:

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/inspection_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/inspection_accessor.py
@@ -306,6 +306,12 @@ class _InspectionAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
                 "permutation_importance",
                 data_source,
                 at_step,
+                #
+                # skore-hub-project expects an item for data_source_hash (but
+                # ignores its value). Until skore-hub-project is updated we
+                # insert None as a placeholder.
+                None,
+                #
                 metric,
                 kwargs,
             )
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -541,8 +541,6 @@ class _MetricsAccessor(
             Whether to cast the numbers to floats. If `False`, the return value
             is `None` when the predictions have never been computed.
         """
-        X, _ = self._get_X_y(data_source=data_source)
-
         predict_time_cache_key = (
             self._parent._hash,
             data_source,
--- a/skore/src/skore/_sklearn/_comparison/inspection_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/inspection_accessor.py
@@ -341,6 +341,12 @@ class _InspectionAccessor(_BaseAccessor["ComparisonReport"], DirNamesMixin):
                 "permutation_importance",
                 data_source,
                 at_step,
+                #
+                # skore-hub-project expects an item for data_source_hash (but
+                # ignores its value). Until skore-hub-project is updated we
+                # insert None as a placeholder.
+                None,
+                #
                 metric,
                 kwargs,
             )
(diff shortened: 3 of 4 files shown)


auguste-probabl:
image

you love to see it XD

thomass-dev:
🔥

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -321,7 +307,6 @@ def permutation_importance(
                 "permutation_importance",
                 data_source,
                 at_step,
-                data_source_hash,
```
cakedev0, line 323: Currently, cache structure is inspected by `skore-hub-project` for the inspection accessor:

`parent_hash, name, data_source, at_step, _, metric, *_ = ...`

So let's put a `None` there and add a comment about compatibility with `skore-hub-project`. 

Same in CV reports.

cakedev0 (changes requested):
A small important comment below. And otherwise, I prompted a LLM to check if you forgot to remove a few things, and it seems so:
1. Remove dead "X_y" branches in plot title rendering. These branches are now unreachable and only add complexity:
  - roc_curve.py:269
  - precision_recall_curve.py:261
  - prediction_error.py:248
  - confusion_matrix.py:232

2. Clean stale docstrings that still mention X_y / external sets. Examples:
  - roc_curve.py:57
  - precision_recall_curve.py:59
  - prediction_error.py:57
  - confusion_matrix.py:47
  - comparison/utils.py:29

3. Clean stale docstring parameter sections that still describe removed X/y args. Examples:
  - cross_validation/metrics_accessor.py:319
  - cross_validation/metrics_accessor.py:377
  - comparison/metrics_accessor.py:401

4. Remove stale hash-related wording that no longer applies.
  - _get_X_y docstring still mentions hash logic: _base.py:65
  - Example note still says metrics rely on hashing input data: plot_estimator_report.py:186

5. Finish the test cleanup for display titles. One test still expects "X_y" as a valid display data source: test_common.py:151

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -592,15 +541,11 @@ def _predict_time(
             Whether to cast the numbers to floats. If `False`, the return value
             is `None` when the predictions have never been computed.
         """
-        if data_source_hash is None:
-            X, _, data_source_hash = self._get_X_y_and_data_source_hash(
-                data_source=data_source, X=X, y=y
-            )
+        X, _ = self._get_X_y(data_source=data_source)
```
cakedev0: X is not used.

```suggestion
        _ = self._get_X_y(data_source=data_source)
```

Or remove this line completely? It's not a no-op as it can raise, but I don't think we want/need this behavior because the rest of the function supports a cache miss (`self._parent._cache.get(..., (float("nan") if cast else None))`) so I'd say we don't except this function to raise but rather to return None/NaN.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
jeromedockes: thanks!

PR #2539, opened 2026-03-01
feat(skore): Add aggregate parameter to ImpurityDecreaseDisplay

### LINKED ISSUE #2464: api(skore): Add aggregate parameter to ImpurityDecreaseDisplay

When dealing with `CrossValidationReport`, it is handy to have an `aggregate` option. We should implement this option for the `ImpurityDecreaseDisplay`.

MuditAtrey:
#### Change description

# Closes #2464.

Adds an `aggregate` parameter to `ImpurityDecreaseDisplay.frame()`, following the same pattern as `PermutationImportanceDisplay`. When `aggregate` is not `None`, per-split rows are aggregated and the `split` column is replaced by `importance_mean` and `importance_std` columns. `_plot_matplotlib` now passes `aggregate=None` explicitly when calling `frame()`.

opening as a draft because I'm unsure about the expected behavior for the `comparison-cross-validation` report type.

my current implementation aggregates over splits per estimator in this case too, which causes `test_with_pipeline` in `test_comparison_cross_validation.py` to fail. 
not sure whether to update that test or leave `comparison-cross-validation` unaggregated, any guidance would be appreciated.

#### Contribution checklist

with pre-commit run --all-files)
requires updating tests)
#### AI usage disclosure

AI tools were involved for:

this is what opened when drafting the PR for this

### DIFF
--- a/skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
+++ b/skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
@@ -151,8 +151,8 @@ class ImpurityDecreaseDisplay(DisplayMixin):
         aggregate : {"mean", "std"}, ("mean", "std") or None, default=("mean", "std")
             Aggregate the importances over splits. Only relevant when
             ``report_type`` is ``"cross-validation"`` or
-            ``"comparison-cross-validation"``. If ``None``, the raw per-split
-            values are returned.
+            ``"comparison-cross-validation"``; ignored otherwise. If ``None``,
+            the raw per-split values are returned.
 
         Returns
         -------
--- a/skore/tests/unit/displays/impurity_decrease/test_common.py
+++ b/skore/tests/unit/displays/impurity_decrease/test_common.py
@@ -45,18 +45,31 @@ class TestImpurityDecreaseDisplay:
             expected.add("estimator")
         assert set(frame.columns) == expected
 
-    def test_frame_aggregate_none(self, fixture_prefix, task, request):
+    def test_frame_aggregate(self, fixture_prefix, task, request):
         report = request.getfixturevalue(f"{fixture_prefix}_{task}")
         if isinstance(report, tuple):
             report = report[0]
         display = report.inspection.impurity_decrease()
-        frame = display.frame(aggregate=None)
 
-        if "cross_validation" in fixture_prefix:
-            assert "split" in frame.columns
-        assert "importance" in frame.columns
-        assert "importance_mean" not in frame.columns
-        assert "importance_std" not in frame.columns
+        if "cross_validation" not in fixture_prefix:
+            # situation 1: aggregate is ignored for non-CV report types
+            assert set(display.frame().columns) == set(
+                display.frame(aggregate=None).columns
+            )
+        else:
+            # situation 2: aggregate=None does not aggregate
+            frame_none = display.frame(aggregate=None)
+            assert "split" in frame_none.columns
+            assert "importance" in frame_none.columns
+            assert "importance_mean" not in frame_none.columns
+            assert "importance_std" not in frame_none.columns
+
+            # situation 3: default aggregate aggregates properly
+            frame_agg = display.frame()
+            assert "split" not in frame_agg.columns
+            assert "importance_mean" in frame_agg.columns
+            assert "importance_std" in frame_agg.columns
+            assert "importance" not in frame_agg.columns
 
     def test_internal_data_structure(self, fixture_prefix, task, request):
         report = request.getfixturevalue(f"{fixture_prefix}_{task}")


MuditAtrey:
For context on what I've tried: the `comparison-cross-validation` case falls into `columns_to_drop = []` in `frame()`, meaning both `estimator` and `split` are kept before aggregation. The current implementation then groups by `["estimator", "feature"]` and aggregates over splits, which produces `importance_mean` and `importance_std` per estimator per feature, dropping `split` in the process.

`test_with_pipeline` in `test_comparison_cross_validation.py` currently asserts the old columns `["estimator", "split", "feature", "importance"]`, so it fails with the new behavior. Two options as I see it:
1. Update the test to expect `["estimator", "feature", "importance_mean", "importance_std"]` meaning `comparison-cross-validation` also aggregates
2. Guard the aggregation block so it only runs for `"cross-validation"` and not `"comparison-cross-validation"` which leaves the comparison case unaggregated

Happy to implement whichever direction makes more sense, this is one of the first issues I've worked on in this project, so I want to make sure I'm aligned with the intended design before making further decisions.

MuditAtrey:
Here's what's working:

`EstimatorReport` remains unchanged, returns raw importance per feature:
```
            feature  importance
0  sepal length (cm)    0.101454
1   sepal width (cm)    0.034117
2  petal length (cm)    0.469279
3   petal width (cm)    0.395150
```

`CrossValidationReport` with default `aggregate=("mean", "std")` gives aggregates over splits:
```
            feature  importance_mean  importance_std
0  sepal length (cm)         0.092147        0.007942
1   sepal width (cm)         0.024870        0.004717
2  petal length (cm)         0.434609        0.016254
3   petal width (cm)         0.448374        0.014374
```

`CrossValidationReport` with `aggregate=None` returns raw per-split values:
```
    split            feature  importance
0       0  sepal length (cm)    0.096871
...
```

the only open question is the `comparison-cross-validation` case, as described above.

GaetandeCast:
>1- Update the test to expect `["estimator", "feature", "importance_mean", "importance_std"]` meaning comparison-cross-validation also aggregates

Yes go with this option, we want to aggregate for comparisons too.

skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
```diff
@@ -142,38 +143,25 @@ def _compute_data_for_display(
...
-                     feature  importance
-        0  sepal length (cm)     0.1...
-        1   sepal width (cm)     0.0...
-        2  petal length (cm)     0.4...
-        3   petal width (cm)     0.3...
+            Dataframe containing the mean decrease in impurity. When
+            ``aggregate`` is not ``None`` and the report type involves
+            cross-validation splits, the ``split`` column is removed and
+            ``importance`` is replaced by aggregated columns (e.g.
+            ``importance_mean``, ``importance_std``).
```
GaetandeCast:
```suggestion
            ``importance`` is replaced by aggregated columns:
            ``importance_mean`` and ``importance_std``.
```
The only possible aggregation is mean and std so no need to say "e.g."

skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
```diff
@@ -184,7 +172,19 @@ def frame(self) -> pd.DataFrame:
         else:  # comparison-cross-validation
             columns_to_drop = []
 
-        return self.importances.drop(columns=columns_to_drop)
+        frame = self.importances.drop(columns=columns_to_drop)
+
+        if aggregate is not None and "split" in frame.columns:
+            group_by = [c for c in ["estimator", "feature"] if c in frame.columns]
```
GaetandeCast, line 178:
```suggestion
            group_by = [c for c in ["estimator", "feature", "label", "output"] if c in frame.columns]
```

skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
```diff
@@ -184,7 +172,19 @@ def frame(self) -> pd.DataFrame:
...
+            frame = (
+                frame.groupby(group_by, sort=False, dropna=False).aggregate(
+                    {"importance": aggregate}
```
GaetandeCast:
```suggestion
                    aggregate
```

MuditAtrey: applied this change but noticed it now aggregates all numeric columns including `split`, producing `split_mean` and `split_std` columns, the fix might be to drop `split` before the groupby. Is there a cleaner way to handle this?

GaetandeCast: yes indeed, we need to add `.drop(columns=["split"])` before the `.groupby()`.

skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
GaetandeCast: Actually I got confused with the `CoefficientsDisplay`. There are no "label" or "output" columns for the `ImpurityDecreaseDisplay` so you can revert this change.

skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
GaetandeCast: the block should look like :
```
        if aggregate is not None and "split" in frame.columns:
            group_by = [c for c in ["estimator", "feature"] if c in frame.columns]
            frame = (
                frame.drop(columns=["split"])
                .groupby(group_by, sort=False, dropna=False)
                .aggregate(aggregate)
            ).reset_index()
            if isinstance(frame.columns, pd.MultiIndex):
                frame.columns = flatten_multi_index(frame.columns)
```

MuditAtrey:
~~all tests passing now, is there anything else needed before I mark this ready for review?~~

thanks for marking it as `ready for review` 😀

MuditAtrey:
How do I resolve these below? especially the `changelog` one, I couldn't find `CHANGELOG.rst`

requires updating tests)

skore/tests/unit/displays/impurity_decrease/test_common.py
```diff
@@ -35,13 +35,29 @@ def test_frame_structure(self, fixture_prefix, task, request):
...
         if "cross_validation" in fixture_prefix:
-            expected.add("split")
+            expected |= {"importance_mean", "importance_std"}
+        else:
+            expected.add("importance")
         if "comparison" in fixture_prefix:
             expected.add("estimator")
         assert set(frame.columns) == expected
 
+    def test_frame_aggregate_none(self, fixture_prefix, task, request):
```
GaetandeCast: Let's test the full behavior of the parameter here:
- test that changing the value of the parameter does nothing for non cross validation report types
- when there are splits, test that `aggregate=None` does not aggregate
- when there are splits, test that the default aggregates properly

Let's try to keep things minimal as we are in `test_common.py` but still we should go through the different situations

skore/src/skore/_sklearn/_plot/inspection/impurity_decrease.py
```diff
@@ -142,38 +143,25 @@ def _compute_data_for_display(
...
         """Get the mean decrease in impurity in a dataframe format.
 
-        The returned dataframe is not going to contain constant columns or columns
-        containing only NaN values.
+        Parameters
+        ----------
+        aggregate : {"mean", "std"}, ("mean", "std") or None, default=("mean", "std")
+            Aggregate the importances over splits. Only relevant when
+            ``report_type`` is ``"cross-validation"`` or
+            ``"comparison-cross-validation"``. If ``None``, the raw per-split
```
GaetandeCast: mention that the argument is ignored otherwise

GaetandeCast:
> How do I resolve these below? 

The documentation has built properly so the first point is ok, see https://github.com/probabl-ai/skore/pull/2539#issuecomment-3985102754
For the changelog, it has not been added yet so you can ignore this point for now.

Issue #2538, opened 2026-03-01
bug(skore): TypeError when calling .data.analyze() with scipy.sparse matrices


KaranSinghDev:
### Describe the bug

As mentioned in previous issues (#2422), there is no specific guardrail for scipy.sparse matrices in its data accessors. While #2433 updated the  logic to handle list/tuple inputs, it uses `accept_sparse=False` internally. 

If a user initializes an `EstimatorReport` or `CrossValidationReport` with sparse data the initialization succeeds, but calling the `.data.analyze()` method triggers a fatal TypeError.

### Steps/Code to Reproduce

```Python
import scipy.sparse as sp
import numpy as np
from sklearn.linear_model import Ridge
from skore import EstimatorReport

# 1, Spare Matrix Creation
X_sparse = sp.csr_matrix(np.random.rand(100, 10))
y = np.random.rand(100)

# 2. Initialize report 
report = EstimatorReport(Ridge(), X_train=X_sparse, y_train=y)

# 3. Trigger analysis (crashes)
report.data.analyze(data_source="train")
````

### Expected Behavior

Should not have crashed with a low-level TypeError. It should have detected the sparse input and provide a graceful response (like 'NotImplementedError`). The current error message suggests using `.toarray()` to convert the data to a dense array. On typical large-scale NLP or genomics datasets , following this suggestion could also lead to out of memory crash. 

### Actual Behavior

```Python
Traceback (most recent call last):
  File "sparse_type_E.py", line 15, in <module>
    report.data.analyze(data_source="train")
  File ".../src/skore/_sklearn/_estimator/data_accessor.py", line 186, in analyze
    X, y = self._retrieve_data_as_frame(data_source, with_y)
  File ".../src/skore/_sklearn/_estimator/data_accessor.py", line 63, in _retrieve_data_as_frame
    X = _normalize_X_as_dataframe(X)
  File ".../src/skore/_utils/_dataframe.py", line 14, in _normalize_X_as_dataframe
    X = check_array(X, accept_sparse=False, ensure_2d=True, ensure_all_finite=False)
  File ".../venv/lib/python3.12/site-packages/sklearn/utils/validation.py", line 981, in check_array
    array = _ensure_sparse_format(...)
  File ".../venv/lib/python3.12/site-packages/sklearn/utils/validation.py", line 602, in _ensure_sparse_format
    raise TypeError(
TypeError: Sparse data was passed, but dense data is required. Use '.toarray()' to convert to a dense numpy array.
```
[...]

PR #2540, opened 2026-03-01
fixing the issue number #2538

direkkakkar319-ops:
Problem
When a user initializes EstimatorReport or CrossValidationReport with a scipy.sparse matrix, initialization succeeds but calling .data.analyze() later crashes with a confusing low-level TypeError from sklearn internals, suggesting .toarray() as a fix — which can cause out-of-memory crashes on large datasets like NLP or genomics data.

Root Cause
In skore/_utils/_dataframe.py, _normalize_X_as_dataframe() calls:
pythoncheck_array(X, accept_sparse=False, ...)
This check happens too late — only when .data.analyze() is called — rather than at the point data enters skore.

Fix
Add an explicit sparse guard at the top of _normalize_X_as_dataframe(), before check_array is reached:
pythonif sp.issparse(X):
    raise NotImplementedError(
        "Sparse matrices are not yet supported in .data.analyze(). "
        "Convert to dense only if memory allows."
    )
sp.issparse() is used because it covers all sparse formats (csr, csc, coo, etc.).
Why NotImplementedError?
It signals this is a missing feature, not a user mistake — and keeps the door open for future sparse support.

Changes
Added sparse guard in _normalize_X_as_dataframe()
Added unit tests for both report types
Updated docstring to document sparse as unsupported

### DIFF
--- a/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
+++ b/skore/src/skore/_sklearn/_plot/metrics/prediction_error.py
@@ -301,7 +301,7 @@ class PredictionErrorDisplay(DisplayMixin):
             handles,
             labels,
             loc="upper center",
-            bbox_to_anchor=(0.5, -0.3),
+            bbox_to_anchor=(0.5, -0.15),
             ncol=1,
             frameon=True,
         )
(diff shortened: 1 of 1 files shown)


GaetandeCast:
Closing as what this changes has already been fixed in #2530 and is unrelated to the PR description.

PR #2543, opened 2026-03-02
fix(skore): Exit early if sparse matrices are passed in data accessors

### LINKED ISSUE #2538: bug(skore): TypeError when calling .data.analyze() with scipy.sparse matrices

KaranSinghDev:
#### Change description

Fixes #2538 
I added early-exit guardrails in skore/_utils/_dataframe.py using scipy.sparse.issparse(). If sparse data is detected, the SDK now raises a clear NotImplementedError. 

#### Contribution checklist

- [x] Unit tests were added or updated ( test_dataframe )

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@
 🙌 Thanks for contributing a pull request!
 
 If this is your first contribution, take a look at our contribution guidelines:
-https://docs.skore.probabl.ai/dev/contributing.html
+https://docs.skore.probabl.ai/stable/contributing.html
 -->
 #### Change description
 <!--
@@ -36,12 +36,12 @@ requires updating tests)
 - [ ] All the commits in the PR are signed (more information
 [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits))
 - [ ] The pull request title respects the Conventional Commits convention (more information
-[here](https://docs.skore.probabl.ai/dev/contributing.html#pull-requests))
+[here](https://docs.skore.probabl.ai/stable/contributing.html#pull-requests))
 
 #### AI usage disclosure
 <!--
 If AI tools were involved in creating this PR, please check all boxes that apply
-below and make sure you understand our [automated contributions policy](https://docs.skore.probabl.ai/dev/contributing.html#automated-contributions-policy)
+below and make sure you understand our [automated contributions policy](https://docs.skore.probabl.ai/stable/contributing.html#automated-contributions-policy)
 -->
 AI tools were involved for:
 - [ ] Code generation (e.g., when writing an implementation or fixing a bug)
--- a/.github/workflows/pr-commits-verification.yml
+++ b/.github/workflows/pr-commits-verification.yml
@@ -36,7 +36,7 @@ jobs:
           message: |
             > [!CAUTION]
             > Some commits in the pull request are not signed, or GitHub is not able to verify the signature.
-            > Please sign all your commits; you can find more information [here](https://docs.skore.probabl.ai/dev/contributing.html#signing-commits).
+            > Please sign all your commits; you can find more information [here](https://docs.skore.probabl.ai/stable/contributing.html#signing-commits).
             > Please note that when you activate commit signing, you'll need to retroactively sign your previous commits.
 
       - name: Delete previous comment on success
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -150,6 +150,12 @@ jobs:
         run: |
           set -u
 
+          if [[ "${GITHUB_EVENT_NAME}" == "pull_request" ]]; then
+              echo "SPHINX_VERSION=pull-${GITHUB_EVENT_NUMBER}" >> "${GITHUB_OUTPUT}"
+              echo "SPHINX_RELEASE=0.0.0+pull-${GITHUB_EVENT_NUMBER}" >> "${GITHUB_OUTPUT}"
+              exit 0
+          fi
+
           if [[ "${GITHUB_EVENT_NAME}" != "push"  ]] || [[ "${GITHUB_REF_TYPE}" != "tag" ]]; then
               echo "SPHINX_VERSION=dev" >> "${GITHUB_OUTPUT}"
               echo "SPHINX_RELEASE=0.0.0+dev" >> "${GITHUB_OUTPUT}"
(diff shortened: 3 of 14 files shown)


skore/src/skore/_utils/_dataframe.py
```diff
@@ -27,6 +34,12 @@ def _normalize_X_as_dataframe(X: ArrayLike) -> pd.DataFrame:
 
 def _normalize_y_as_dataframe(y: ArrayLike) -> pd.DataFrame:
     """Normalize target data as a pandas DataFrame with predictable column names."""
+    if sp.issparse(y):
+        raise NotImplementedError(
+            "Data analysis via skrub is currently not supported for sparse matrices. "
+            "Please use dense data."
+        )
+
```
auguste-probabl, line 42: Is the check necessary for y? We don't run `check_array` here.

skore/src/skore/_utils/_dataframe.py
```diff
@@ -10,6 +11,12 @@
 
 def _normalize_X_as_dataframe(X: ArrayLike) -> pd.DataFrame:
     """Normalize feature data as a pandas DataFrame with string column names."""
+    if sp.issparse(X):
+        raise NotImplementedError(
+            "Data analysis via skrub is currently not supported for sparse matrices. "
+            "Please use dense data."
+        )
+
```
auguste-probabl, line 19: @cakedev0 Do you remember why `check_array` is run with `accept_sparse=False` in the first place?

skore/tests/unit/utils/test_dataframe.py
auguste-probabl, line 1: I'd say only the first 2 tests are needed, and they can go with the existing `_normalize_{X,y}_as_dataframe` tests

skore/src/skore/_utils/_dataframe.py
cakedev0: Because `pd.DataFrame(X)` does not work / does not work as expected on sparse matrices. And so we indeed can't use `TableReport` on it.

cakedev0: Btw, if we want to add support for sparse inputs in `skrub.TableReport`, I have some ideas for that. I've worked a lot with sparse matrices ^^ 

But it might not make sense, at least for tabular ML, tables should not really be sparse before pre-processing.

skore/tests/unit/utils/test_dataframe.py
KaranSinghDev: @auguste-probabl @cakedev0  I understand I have removed the integration tests and put only the unit tests for _normalize_X and _normalize_y in the test_accessor.py (let me know if you meant by another file) let me know your thoughts on it .

skore/src/skore/_utils/_dataframe.py
cakedev0: I think we can have this check here. It seems scikit-learn accept sparse y in some cases (multi-label), so it could happen to have a sparse y in reports.

KaranSinghDev: Without the check sparse y was falling through to y as np.asarray(y),  as it is sparse matrix, np.asarray() wraps it into a 0-dimensional object array with an empty shape tuple (). when tried to access y.shape[1], it crashes with IndexError: tuple index out of range

skore/tests/unit/reports/estimator/data/test_accessor.py
```diff
@@ -171,3 +172,21 @@ def test_analyze_subsampling(
         assert display.summary["dataframe"].index.to_list() == list(range(10))
     else:
         assert display.summary["dataframe"].index.to_list() != list(range(10))
+
+
+def test_normalize_X_as_dataframe_sparse():
+    """Check that _normalize_X_as_dataframe rejects sparse matrices."""
+    from skore._utils._dataframe import _normalize_X_as_dataframe
```
auguste-probabl: Move these imports to the top level

PR #2524, opened 2026-02-24
ci(skore): Create example pushing on public `Skore Hub` project

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/2274.

---

Add two examples pushing on public `Skore Hub` project:
- Modify `plot_getting_started.py`
- Add `plot_skore_hub_project.py`
    
The examples are pushed on:
```
https://skore.probabl.ai/skore/example-getting-started-<version>
https://skore.probabl.ai/skore/example-skore-hub-project-<version>
```
    
Also, change the sphinx workflow to:    
- Build examples pushing on `Skore Hub`  **<ins>only</ins>** on main/tags/maintainer pull-requests for security issues
  Disable such examples otherwise (no preview in fork pull-requests)
- Build gallery using `parallel=True` / `-j auto`

PR #2550, opened 2026-03-03
fix(skore-hub-project/splitting_strategy): Use `_safe_indexing` with `report.y`

thomass-dev:
Needed for https://github.com/probabl-ai/skore/pull/2524/: fix `splitting_strategy` when `report.y` is an indexed `pandas.Series`.

### DIFF
--- a/skore-hub-project/src/skore_hub_project/report/cross_validation_report.py
+++ b/skore-hub-project/src/skore_hub_project/report/cross_validation_report.py
@@ -205,11 +205,15 @@ class CrossValidationReportPayload(ReportPayload[CrossValidationReport]):
         of 200 buckets, and averaging the number of samples belonging to the test-set in
         each of these buckets. @TODO: find a better representation of the distribution.
         """
+        from skore._externals._sklearn_compat import (  # type: ignore[attr-defined]
+            _safe_indexing,
+        )
+
         splits = []
 
         for train_indices, test_indices in self.report.split_indices:
-            train_y = self.report.y[train_indices]
-            test_y = self.report.y[test_indices]
+            train_y = _safe_indexing(self.report.y, train_indices)
+            test_y = _safe_indexing(self.report.y, test_indices)
             train_target_distribution: list[float] = []
             test_target_distribution: list[float] = []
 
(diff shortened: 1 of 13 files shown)


skore-hub-project/src/skore_hub_project/report/cross_validation_report.py
```diff
@@ -205,11 +205,15 @@ def splitting_strategy(self) -> dict[str, Any]:
         of 200 buckets, and averaging the number of samples belonging to the test-set in
         each of these buckets. @TODO: find a better representation of the distribution.
         """
+        from skore._externals._sklearn_compat import (  # type: ignore[attr-defined]
```
glemaitre, line 208: You don't want to move this on the top of the file?

thomass-dev: I don't want to have `skore` top-level import, kind of "`skore-hub-project` is always used in conjunction with `skore`, but `skore` should not be a direct dependency".

skore-hub-project/tests/unit/report/test_cross_validation_report.py
```diff
@@ -244,6 +245,54 @@ def test_splits_for_regression(self, project):
             assert all(isinstance(value, float) for value in train_target_distribution)
             assert all(isinstance(value, float) for value in test_target_distribution)
 
+    @mark.respx(assert_all_called=False)
+    def test_splitting_strategy_with_y_serie(self, project):
```
glemaitre: Do you think that it could be good to parametrize the test with array and series using this function:

 https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_testing.py#L964-L971

You could parametrize using two strings "array" and "series".

skore-hub-project/src/skore_hub_project/report/cross_validation_report.py
glemaitre: OK I see.

glemaitre: Maybe put a small comment to acknowledge then.

skore-hub-project/tests/unit/report/test_cross_validation_report.py
thomass-dev: Array is covered by other tests, do you want an explicit test for array?

glemaitre: No it is fine then.

PR #2561, opened 2026-03-05
chore(skore)!: Rename the `name` parameter of `create_estimator_report()` to `report_key`

GaetandeCast:
Renames the `name` parameter of `ComparisonReport.create_estimator_report()` to ~`key` to be consistent with the rest of the codebase~ `report_key` to be more explicit than `name`, and adds examples in the docstring.

cc @MarieSacksick

### DIFF
--- a/examples/getting_started/plot_getting_started.py
+++ b/examples/getting_started/plot_getting_started.py
@@ -270,7 +270,7 @@ comparison.metrics.precision_recall().plot()
 # %%
 
 final_report = comparison.create_estimator_report(
-    key="Simple Linear Model", X_test=X_holdout, y_test=y_holdout
+    report_key="Simple Linear Model", X_test=X_holdout, y_test=y_holdout
 )
 
 # %%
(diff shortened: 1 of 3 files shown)


examples/getting_started/plot_getting_started.py
```diff
@@ -270,7 +270,7 @@
 # %%
 
 final_report = comparison.create_estimator_report(
-    name="Simple Linear Model", X_test=X_holdout, y_test=y_holdout
+    key="Simple Linear Model", X_test=X_holdout, y_test=y_holdout
```
glemaitre: `key` might be a bit too much generic. Since the required key could be taken from `reports.keys()`, we could call it `report_key`

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -432,32 +432,57 @@ def create_estimator_report(
 
         Parameters
         ----------
-        name : str
-            The name of the estimator to create a report for.
+        key : str
+            The key associated with the estimator to create a report for, as stored in
+            the :attribute:`reports_` attribute of the :class:`~skore.ComparisonReport`.
```
glemaitre: Here, we might want to mention that one can use `reports_.keys()` to list them the potential key.

GaetandeCast: done

PR #2573, opened 2026-03-06
feat(skore): Add evaluate() dispatcher function

### LINKED ISSUE #2520: feat: add top-level `skore.evaluate(estimator, X_test, y_test)` for final test-set evaluation
### The gap

skore already has the excellent `CrossValidationReport` for **during development** (unfitted estimator + full training data + CV).

What is still missing is the **final honest evaluation** after you have trained your model on all training data:

```python
# Current way (works but verbose + requires knowing internal names)
report = EstimatorReport(
    estimator=final_model,
    X_test=X_test,
    y_test=y_test,
    fit=False
)
report.metrics.summarize()

from skore import evaluate

report = evaluate(final_model, X_test, y_test)   
report.metrics.summarize()
# or even auto-rich output on creation



_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by raotalha71**
#### Change description

Closes #2520

## What this PR does

Adds a public `skore.evaluate()` function that dispatches to the right report class based on the `splitter` parameter:

| `splitter` value | Returns |
|---|---|
| `float` (default `0.2`) | `EstimatorReport` (train-test split via skore's `train_test_split`) |
| `"prefit"` | `EstimatorReport` (already fitted model, test-only evaluation) |
| `int` | `CrossValidationReport` (k-fold CV) |
| CV object (`KFold`, etc.) | `CrossValidationReport` |
| `estimator=list` | `ComparisonReport` (wraps individual reports) |

## Design decisions (confirmed with @jeromedockes)

- `splitter="prefit"` uses a string sentinel (not `None`)
- `splitter=int` maps to KFold via `CrossValidationReport`'s `check_cv()`
- Uses **skore's** `train_test_split` (not sklearn's) to get smart warnings for free
- For multiple estimators with float splitter, data is split once and reused so `ComparisonReport` sees identical test targets

## Example usage

```python
from skore import evaluate

# Train-test split (default)
report = evaluate(LogisticRegression(), X, y)

# 5-fold cross-validation
report = evaluate(LogisticRegression(), X, y, splitter=5)

# Pre-fitted model
fitted = LogisticRegression().fit(X, y)
report = evaluate(fitted, X, y, splitter="prefit")

# Compare models
report = evaluate([LogisticRegression(), DecisionTreeClassifier()], X, y, splitter=5)

### DIFF
--- a/skore/tests/unit/sklearn/test_evaluate.py
+++ b/skore/tests/unit/sklearn/test_evaluate.py
@@ -4,8 +4,6 @@ from sklearn.model_selection import KFold, StratifiedKFold
 
 from skore import ComparisonReport, CrossValidationReport, EstimatorReport, evaluate
 
-# --- splitter="prefit" ---
-
 
 def test_evaluate_prefit_estimator(regression_data):
     """A fitted estimator with splitter='prefit' returns an EstimatorReport."""
@@ -26,9 +24,6 @@ def test_evaluate_prefit_unfitted_raises(regression_data):
         evaluate(LinearRegression(), X, y, splitter="prefit")
 
 
-# --- splitter=float (train-test split) ---
-
-
 def test_evaluate_float_splitter(regression_data):
     """A float splitter triggers a single split and returns an EstimatorReport."""
     X, y = regression_data
@@ -45,9 +40,6 @@ def test_evaluate_default_splitter(regression_data):
     assert len(report.X_test) == 0.2 * len(X)
 
 
-# --- splitter=int (cross-validation) ---
-
-
 def test_evaluate_int_splitter(regression_data):
     """An int splitter triggers cross-validation."""
     X, y = regression_data
@@ -56,18 +48,12 @@ def test_evaluate_int_splitter(regression_data):
     assert len(report.estimator_reports_) == 3
 
 
-# --- splitter=CV object ---
-
-
 def test_evaluate_cv_object_splitter(regression_data):
     """A CV object splitter triggers cross-validation."""
     X, y = regression_data
     report = evaluate(LinearRegression(), X, y, splitter=KFold(n_splits=4))
     assert isinstance(report, CrossValidationReport)
     assert len(report.estimator_reports_) == 4
-
-
-# --- multiple estimators ---
 
 
 def test_evaluate_multiple_estimators(regression_data):
@@ -102,17 +88,11 @@ def test_evaluate_multiple_estimators_multiple_X(regression_data):
     assert isinstance(report, ComparisonReport)
 
 
-# --- invalid splitter ---
-
-
 def test_evaluate_invalid_splitter_string(regression_data):
     """An invalid string splitter raises ValueError."""
     X, y = regression_data
     with pytest.raises(ValueError, match="Invalid string value for splitter"):
         evaluate(LinearRegression(), X, y, splitter="invalid")
-
-
-# --- classification ---
 
 
 def test_evaluate_classification(binary_classification_data):
@@ -128,9 +108,6 @@ def test_evaluate_classification_cv(binary_classification_data):
     report = evaluate(LogisticRegression(), X, y, splitter=StratifiedKFold(n_splits=3))
     assert isinstance(report, CrossValidationReport)
     assert len(report.estimator_reports_) == 3
-
-
-# --- pos_label ---
 
 
 def test_evaluate_pos_label(binary_classification_data):
--- a/skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
+++ b/skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
@@ -285,9 +285,6 @@ def test_train_test_split_check_dict_no_X_no_y():
     assert list(keys) == ["z_train", "z_test"]
 
 
-# --- TrainTestSplit ---
-
-
 class TestTrainTestSplit:
     """Tests for the TrainTestSplit splitter class."""
 

### CODE REVIEW

File: skore/src/skore/_sklearn/evaluate.py [hunk 1]
```diff
@@ -0,0 +1,209 @@
...
+
+    >>> report = evaluate(LogisticRegression(), X, y, splitter=5)
+
+    Evaluate a pre-fitted estimator:
+
+    >>> fitted_model = LogisticRegression().fit(X, y)
+    >>> report = evaluate(fitted_model, X, y, splitter="prefit")
+    """
+    if isinstance(estimator, list):
+        return _evaluate_multiple(
```
glemaitre: I don't think that we need to have this function:

```python
return ComparisonReport(
    [
        _evaluate_single(est, X=X, y=y, splitter=splitter, pos_label=pos_label, n_jobs=n_jobs)
        for est in estimators
    ]
)
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+            estimator, X, y, splitter=splitter, pos_label=pos_label, n_jobs=n_jobs
```
glemaitre: Actually `X` could be a list as specified by @GaetandeCast in https://github.com/probabl-ai/skore/issues/2520#issuecomment-3989801576. It will be the case that one did some feature engineering and will attempt to assess the impact of it. So we actuaally need to get both `X` and `estimator` in the loop

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+    y: ArrayLike | None = None,
+    *,
+    splitter: float | int | str | SKLearnCrossValidator | Generator | None = 0.2,
+    pos_label: int | float | bool | str | None = None,
+    n_jobs: int | None = None,
+) -> EstimatorReport | CrossValidationReport | ComparisonReport:
+    """Evaluate one or more estimators on the given data.
+
+    This function is a dispatcher that creates the appropriate report based on
+    the ``splitter`` parameter:
```
glemaitre: I would avoid to name "dispatcher" here. It is a really technical terms.

I would only simplify that passing several `estimator` will provide a report to compare them while the `splitter` parameter will enable whether or not a cross-validation.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+      evaluates it on ``X`` and ``y`` as test data.
+    - An integer or cross-validation object triggers cross-validation and
+      returns a :class:`~skore.CrossValidationReport`.
+    - When ``estimator`` is a list, each estimator is evaluated individually
+      and the results are wrapped in a :class:`~skore.ComparisonReport`.
+
+    Parameters
+    ----------
+    estimator : estimator object or list of estimator objects
+        A scikit-learn compatible estimator, or a list of estimators to compare.
```
glemaitre:
```suggestion
        A scikit-learn compatible estimator, or a list of such estimators to compare.
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+    y: Any,
+    *,
+    splitter: Any,
+    pos_label: Any,
+    n_jobs: Any,
+) -> EstimatorReport | CrossValidationReport:
+    """Evaluate a single estimator based on the splitter type."""
+    from skore._sklearn._cross_validation.report import CrossValidationReport
+    from skore._sklearn._estimator.report import EstimatorReport
+    from skore._sklearn.train_test_split.train_test_split import train_test_split
```
glemaitre: Apart if you have a circular import, you should not import those inside the function.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+
+    if isinstance(splitter, str):
+        if splitter != "prefit":
+            raise ValueError(
+                f"Invalid string value for splitter: {splitter!r}. "
+                "The only supported string value is 'prefit'."
+            )
+        check_is_fitted(estimator)
```
glemaitre: You don't need this part of the code. It will be handled by the `EstimatorReport` itself.

```python
return EstimatoReport(estimator, X_test=X, y_test=y, pos_label=pos_label)
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+            estimator, X, y, pos_label=pos_label, splitter=splitter, n_jobs=n_jobs
+        )
+
+    # Duck-type check for CV splitter objects (has split and get_n_splits)
+    if hasattr(splitter, "split") and hasattr(splitter, "get_n_splits"):
+        return CrossValidationReport(
+            estimator, X, y, pos_label=pos_label, splitter=splitter, n_jobs=n_jobs
+        )
+
+    raise TypeError(
```
glemaitre: Let's not raise the error ourself. Let's just pass the splitter and the let `CrossValidationReport` raise the error (or the `check_cv`).

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+                "The only supported string value is 'prefit'."
+            )
+        check_is_fitted(estimator)
+        return EstimatorReport(
+            estimator, fit=False, X_test=X, y_test=y, pos_label=pos_label
+        )
+
+    if isinstance(splitter, (float, type(None))):
+        test_size = splitter if splitter is not None else 0.2
+        split_data = train_test_split(X, y, test_size=test_size, as_dict=True)
```
glemaitre: Here, I would prefere to always rely on splitter and create a `TrainTestSplit` splitter. We can make this splitter private for the moment.

```python
if isinstance(splitter, float):
    splitter = TrainTestSplit(test_size=test_size)
```

We need to create a splitter that internally call the `sklearn.model_selection.train_test_split` that would return the training and testing indices with the `split` method and `1` for the `get_n_split()`.
Then, it means, that we can always call the `CrossValidationReport`. We only need to unpack the report for splitter with a single split:

```python
report = CrossValidationReport(..., splitter=splitter, ...)
if splitter.get_n_splits() == 1:
    return report.estimator_reports_[0]
return report
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+            raise ValueError(
+                f"Invalid string value for splitter: {splitter!r}. "
+                "The only supported string value is 'prefit'."
+            )
+        check_is_fitted(estimator)
+        return EstimatorReport(
+            estimator, fit=False, X_test=X, y_test=y, pos_label=pos_label
+        )
+
+    if isinstance(splitter, (float, type(None))):
```
glemaitre: We should not support `None`

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,209 @@
...
+    estimator : estimator object or list of estimator objects
+        A scikit-learn compatible estimator, or a list of estimators to compare.
+
+    X : array-like of shape (n_samples, n_features)
+        Feature matrix.
+
+    y : array-like of shape (n_samples,) or None, default=None
+        Target vector. Can be ``None`` for unsupervised tasks.
+
+    splitter : float, int, str, cross-validation object, or None, default=0.2
```
glemaitre: From what I understand, we would like to not have the option `None` because we are using `"prefit"`. @jeromedockes did I understand properly?

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. Inlined the ComparisonReport creation directly in  evaluate() .Recursive calls handle each estimator.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. When estimator is a list, X can now also be a list (zipped with strict=True). When X is not a list, the same X is reused for all estimators.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. Simplified the docstring  no longer says "dispatcher".

glemaitre:
@raotalha71 please do not force push in PR. We are going to squash commits at the end. Not force pushing allow to see the increment between commit.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. Updated the estimator parameter docstring accordingly.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. All imports moved to module level ,verified no circular import exists.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. Removed check_is_fitted. For splitter="prefit", we now just call EstimatorReport(estimator, X_test=X, y_test=y, pos_label=pos_label) and let it handle the fitted/unfitted logic internally.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. Removed the explicit TypeError ,invalid splitter types now fall through to CrossValidationReport which handles the error.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. Created _TrainTestSplit class that wraps sklearn.model_selection.train_test_split, implements split/get_n_splits, and caches the split result for comparison consistency. Float splitter now always goes through CrossValidationReport, and single-split results are unwrapped via get_n_splits() == 1.

skore/src/skore/_sklearn/evaluate.py
raotalha71: Done. Removed None from the splitter type annotation.

raotalha71:
Ohh my bad sorry , will not force push again

glemaitre:
Also no need to answer in each comment repeating the requested change. It is enough to push your changes and only comment when something is unclear or that we should pay attention during the next review. A single message to mention that it is ready for a next review is enough 😉

raotalha71:
Haha , I will avoid spam replies...😁

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,177 @@
...
+    This private splitter wraps ``sklearn.model_selection.train_test_split`` and
+    exposes ``split`` / ``get_n_splits`` so that it can be passed to
+    :class:`~skore.CrossValidationReport`.
+
+    The split result is cached so that the same indices are reused when
+    comparing multiple estimators.
+    """
+
+    def __init__(self, test_size: float = 0.2) -> None:
+        self.test_size = test_size
```
glemaitre: You need to add all the parameters of the `train_test_split` funciton from scikit-learn.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,177 @@
...
+        self._cached_split: tuple[np.ndarray, np.ndarray] | None = None
```
glemaitre: We don't need to cache the split because the `random_state` will be used for such case.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,177 @@
...
+            n_samples = X.shape[0] if hasattr(X, "shape") else len(X)
+            indices = np.arange(n_samples)
+            train_idx, test_idx = sklearn_train_test_split(
+                indices, test_size=self.test_size
+            )
+            self._cached_split = (train_idx, test_idx)
+        yield self._cached_split
+
+
+def evaluate(
```
GaetandeCast, line 22: We should probably document the usage of this function in the documentation to make it discoverable. Perhaps in a separate PR?

raotalha71: Thanks! I agree, I can open a separate PR to add documentation for evaluate() once this one is merged.

GaetandeCast:
@raotalha71 there are some typing issues with mypy that need to be resolved. Install pre-commit hooks to have mypy run locally each time you commit.

raotalha71:
> @raotalha71 there are some typing issues with mypy that need to be resolved. Install pre-commit hooks to have mypy run locally each time you commit.

Done! I've installed pre-commit hooks locally so mypy and other checks run automatically before each commit. The mypy type errors were already fixed in commit 71b7c04 , used cast to narrow the reports list type for ComparisonReport, and added a type guard for X when estimator is not a list.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,186 @@
...
+    def get_n_splits(self, X: Any = None, y: Any = None, groups: Any = None) -> int:
+        """Return the number of splits (always 1)."""
+        return 1
+
+    def split(self, X: Any, y: Any = None, groups: Any = None):
+        """Generate a single train-test split of indices."""
+        if self._cached_split is None:
+            n_samples = X.shape[0] if hasattr(X, "shape") else len(X)
+            indices = np.arange(n_samples)
+            train_idx, test_idx = sklearn_train_test_split(
```
GaetandeCast: We should fix the seed in this call, and be explicit in the documentation about what we do when `splitter` is a `float`: i.e. the default value of shuffle, the value of the seed etc.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,186 @@
...
+                evaluate(
+                    est,
+                    X,
+                    y,
+                    splitter=splitter,
+                    pos_label=pos_label,
+                    n_jobs=n_jobs,
+                )
+                for est in estimator
+            ]
```
GaetandeCast:
```suggestion
        if not isinstance(X, list):
            X = [X] * len(estimator)
        reports = [
            evaluate(
                est,
                x,
                y,
                splitter=splitter,
                pos_label=pos_label,
                n_jobs=n_jobs,
            )
            for est, x in zip(estimator, X, strict=True)
        ]
```
Let's avoid repetition.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,186 @@
...
+    >>> report = evaluate(LogisticRegression(), X, y)
+
+    Cross-validation with 5 folds:
+
+    >>> report = evaluate(LogisticRegression(), X, y, splitter=5)
+
+    Evaluate a pre-fitted estimator:
+
+    >>> fitted_model = LogisticRegression().fit(X, y)
+    >>> report = evaluate(fitted_model, X, y, splitter="prefit")
```
GaetandeCast: I would not advertise reusing the train set for testing.

GaetandeCast: i.e. add a manual train_test before the fit

raotalha71:
Addressed all review comments in commit 2ed2b84:

Added all train_test_split parameters to _TrainTestSplit (test_size, train_size, random_state, shuffle, stratify)
Removed caching using random_state=0 by default for reproducibility
Simplified list X handling with if not isinstance(X, list): X = [X] * len(estimator) as suggested
Fixed prefit docstring example to use a proper train/test split before fitting
Documented float splitter behavior in docstring (shuffle enabled, fixed seed random_state=0)

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+
+from typing import TYPE_CHECKING, Any, cast
+
+import numpy as np
+from sklearn.base import BaseEstimator
+from sklearn.model_selection import train_test_split as sklearn_train_test_split
+
+from skore._sklearn._comparison.report import ComparisonReport
+from skore._sklearn._cross_validation.report import CrossValidationReport
+from skore._sklearn._estimator.report import EstimatorReport
```
GaetandeCast, line 11:
```suggestion
from skore import ComparisonReport, CrossValidationReport, EstimatorReport
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+    def get_n_splits(self, X: Any = None, y: Any = None, groups: Any = None) -> int:
+        """Return the number of splits (always 1)."""
+        return 1
+
+    def split(self, X: Any, y: Any = None, groups: Any = None):
+        """Generate a single train-test split of indices."""
+        n_samples = X.shape[0] if hasattr(X, "shape") else len(X)
+        indices = np.arange(n_samples)
+        train_idx, test_idx = sklearn_train_test_split(
+            indices,
```
GaetandeCast:
```suggestion
        train_idx, test_idx = sklearn_train_test_split(
            np.arange(X.shape[0] if hasattr(X, "shape") else len(X)),
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+            test_size=self.test_size,
+            train_size=self.train_size,
+            random_state=self.random_state,
+            shuffle=self.shuffle,
+        )
```
GaetandeCast:
```suggestion
            stratify=self.stratify,
        )
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+    if isinstance(splitter, str):
+        if splitter != "prefit":
+            raise ValueError(
+                f"Invalid string value for splitter: {splitter!r}. "
+                "The only supported string value is 'prefit'."
+            )
+        return EstimatorReport(estimator, X_test=X, y_test=y, pos_label=pos_label)
+
+    if isinstance(splitter, float):
+        splitter = _TrainTestSplit(test_size=splitter)
```
GaetandeCast: Actually why don't we use a `ShuffleSplit` with `n_split=1` directly here? Is there any benefit to using `train_test_split` that is just a wrapper to `ShuffleSplit`? cc @glemaitre

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+    Parameters
+    ----------
+    estimator : estimator object or list of estimator objects
+        A scikit-learn compatible estimator, or a list of such estimators
+        to compare.
+
+    X : array-like of shape (n_samples, n_features) or list of array-like
+        Feature matrix. When ``estimator`` is a list, ``X`` can also be a
+        list of feature matrices (one per estimator) to assess the impact
+        of different feature engineering pipelines.
```
GaetandeCast:
```suggestion
        list of feature matrices (one per estimator) to e.g. compare models with
        different preprocessing pipelines.
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+
+    y : array-like of shape (n_samples,) or None, default=None
+        Target vector. Can be ``None`` for unsupervised tasks.
```
GaetandeCast:
```suggestion
        Target vector.
```
Actually we recently dropped support for unsupervised learning, until we make proper adjustments for it.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+            random_state=self.random_state,
+            shuffle=self.shuffle,
+        )
+        yield (train_idx, test_idx)
+
+
+def evaluate(
+    estimator: BaseEstimator | list[BaseEstimator],
+    X: ArrayLike | list[ArrayLike],
+    y: ArrayLike | None = None,
```
GaetandeCast:
```suggestion
    y: ArrayLike,
```
Following the previous comment.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+        to the underlying report.
+
+    n_jobs : int or None, default=None
+        Number of jobs for parallel execution. Forwarded to
+        :class:`~skore.CrossValidationReport` or
+        :class:`~skore.ComparisonReport`.
+
+    Returns
+    -------
+    report : EstimatorReport, CrossValidationReport, or ComparisonReport
```
GaetandeCast:
```suggestion
    report : :class:`~skore.EstimatorReport`, :class:`~skore.CrossValidationReport` \
        or :class:`~skore.ComparisonReport`
```

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,188 @@
...
+    ValueError
+        If ``splitter`` is a string other than ``"prefit"``.
+
+    Examples
+    --------
+    >>> from sklearn.datasets import make_classification
+    >>> from sklearn.linear_model import LogisticRegression
+    >>> from skore import evaluate
+    >>> X, y = make_classification(random_state=42)
+    >>> report = evaluate(LogisticRegression(), X, y)
```
GaetandeCast, line 94:
```suggestion

    Default 80/20 train-test split:

    >>> report = evaluate(LogisticRegression(), X, y)
```

skore/tests/unit/sklearn/test_evaluate.py
```diff
@@ -0,0 +1,149 @@
...
+
+@pytest.fixture
+def regression_Xy():
+    return make_regression(n_samples=80, n_features=5, random_state=42)
+
+
+@pytest.fixture
+def classification_Xy():
+    return make_classification(n_samples=80, n_features=5, random_state=42)
+
```
GaetandeCast: Please use the existing fixtures defined in `skore/tests/conftest.py` instead of creating new ones.

skore/tests/unit/sklearn/test_evaluate.py
```diff
@@ -0,0 +1,149 @@
...
+# --- splitter="prefit" ---
+
+
+def test_evaluate_prefit_estimator(regression_Xy):
+    """A fitted estimator with splitter='prefit' returns an EstimatorReport."""
+    X, y = regression_Xy
+    fitted = LinearRegression().fit(X, y)
+    report = evaluate(fitted, X, y, splitter="prefit")
+    assert isinstance(report, EstimatorReport)
+
```
GaetandeCast, line 18:
```suggestion
    assert isinstance(report, EstimatorReport)
    assert report.X_train is None
    assert report.y_train is None
    assert report.X_test is X
    assert report.y_test is y
```

skore/tests/unit/sklearn/test_evaluate.py
```diff
@@ -0,0 +1,149 @@
...
+
+def test_evaluate_float_splitter(regression_Xy):
+    """A float splitter triggers a single split and returns an EstimatorReport."""
+    X, y = regression_Xy
+    report = evaluate(LinearRegression(), X, y, splitter=0.3)
+    assert isinstance(report, EstimatorReport)
+    assert report.X_train is not None
+    assert report.X_test is not None
+
```
GaetandeCast, line 33:
```suggestion
    assert len(report.X_test) == 0.3 * len(X)
```

skore/tests/unit/sklearn/test_evaluate.py
```diff
@@ -0,0 +1,149 @@
...
+
+def test_evaluate_default_splitter(regression_Xy):
+    """Calling without splitter uses the default and returns an EstimatorReport."""
+    X, y = regression_Xy
+    report = evaluate(LinearRegression(), X, y)
+    assert isinstance(report, EstimatorReport)
+    assert report.X_train is not None
```
GaetandeCast:
```suggestion
    assert len(report.X_test) == 0.2 * len(X)
```

skore/tests/unit/sklearn/test_evaluate.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+# --- pos_label ---
+
+
+def test_evaluate_pos_label(classification_Xy):
+    """pos_label is forwarded to the underlying report."""
+    X, y = classification_Xy
+    report = evaluate(LogisticRegression(), X, y, splitter=0.2, pos_label=1)
+    assert isinstance(report, EstimatorReport)
```
GaetandeCast:
```suggestion
    report = evaluate(LogisticRegression(), X, y, splitter=0.2, pos_label=0)
    assert isinstance(report, EstimatorReport)
    assert report.pos_label == 0
```
Let's use the non default value and make sure it is correctly forwarded.

GaetandeCast (commented):
Here is an other pass.

skore/src/skore/_sklearn/evaluate.py
raotalha71: This change causes a circular import — skore.__init__ imports from skore._sklearn, which imports evaluate.py, so importing from skore at the top level in evaluate.py fails with ImportError: cannot import name 'ComparisonReport' from partially initialized module 'skore'. Keeping the internal imports for now.
image

raotalha71:
Addressed all review comments in commit 5f8ad9e:
Inlined n_samples/indices into a single expression in split()
Added missing stratify=self.stratify to the train_test_split call
Updated X docstring: "to e.g. compare models with different preprocessing pipelines"
Removed unsupervised mention, made y a required parameter
Used Sphinx cross-references in Returns section
Simplified docstring examples with "Default 80/20 train-test split:" label
Replaced custom fixtures with existing regression_data and binary_classification_data from conftest.py
Improved prefit test assertions (X_train is None, X_test is X, etc.)
Added len(report.X_test) == 0.3 * len(X) and 0.2 * len(X) assertions
Changed pos_label test to use non-default value (0) and assert report.pos_label == 0
Regarding the import suggestion (from skore import ...): this causes a circular import since skore.__init__ ->  skore._sklearn -> evaluate.py -> skore (not fully loaded yet). Keeping the internal imports.

skore/src/skore/_sklearn/evaluate.py
glemaitre: You have the possibility to have `shuffle=False` with the `train_test_split`.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,191 @@
...
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+    from numpy.typing import ArrayLike
+
+    from skore._sklearn.types import SKLearnCrossValidator
+
+
+class _TrainTestSplit:
```
GaetandeCast: Let's actually expose this class in the public API. This will require proper testing and documentation, and it should also be relocated to `train_test_split.py`.
`evaluate()` should then mention that `splitter=float` is a shortcut to using `TrainTestSplit`, which allows more control over the parameters.

raotalha71:
Addressed in 1c3f913:

Renamed _TrainTestSplit to TrainTestSplit and moved it from evaluate.py to train_test_split/train_test_split.py with a full docstring (Parameters, Examples).
Exported TrainTestSplit in train_test_split/__init__.py, _sklearn/__init__.py, and skore/__init__.py.
Updated evaluate.py to import from the new location and documented that splitter=float is a shortcut for TrainTestSplit(test_size=float).
Added 7 tests for TrainTestSplit in test_train_test_split.py.
You can review it

skore/src/skore/_sklearn/train_test_split/train_test_split.py
```diff
@@ -16,6 +17,98 @@
...
+class TrainTestSplit:
+    """Single train-test split implementing the cross-validation protocol.
+
+    This splitter wraps :func:`sklearn.model_selection.train_test_split` and
+    exposes ``split`` / ``get_n_splits`` so that it can be passed as the
+    ``splitter`` argument of :func:`~skore.evaluate`.
+
+    Passing ``splitter=float`` to :func:`~skore.evaluate` is a shortcut for
+    ``splitter=TrainTestSplit(test_size=float)``.  Use this class directly
+    when you need more control over the splitting parameters.
```
GaetandeCast:
```suggestion
    This splitter wraps :func:`sklearn.model_selection.train_test_split` and
    exposes ``split`` / ``get_n_splits`` so that it can be passed as the
    ``splitter`` argument of any `skore` or `scikit-learn` function.
```
Let's keep the scope general, not restricted to being used in `evaluate`.

skore/src/skore/_sklearn/train_test_split/train_test_split.py
```diff
@@ -16,6 +17,98 @@
...
+        as the class labels.
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from skore import TrainTestSplit
+    >>> splitter = TrainTestSplit(test_size=0.3, random_state=0)
+    >>> X = np.arange(20).reshape(10, 2)
+    >>> for train, test in splitter.split(X):
+    ...     print(len(train), len(test))
```
GaetandeCast:
```suggestion
    >>> for train, test in splitter.split(X):
    ...     train, test
```
Let's show the default a bit more explicitly. You'll have to update the expected output.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -0,0 +1,154 @@
...
+    splitter : float, int, str, or cross-validation object, default=0.2
+        Determines how the data is split:
+
+        - ``float``: shortcut for
+          ``TrainTestSplit(test_size=float)`` — a single train-test split
+          (e.g. ``0.2`` means 80 %% train / 20 %% test).  The data is
+          shuffled before splitting with a fixed seed
+          (``random_state=0``) for reproducibility.  Pass a
+          :class:`~skore.TrainTestSplit` instance directly for more
+          control over the splitting parameters.
```
GaetandeCast:
```suggestion
        - ``float``: perform a single train-test split where the data is shuffled before
          splitting with a fixed seed (``random_state=0``) for reproducibility.
          Pass a :class:`~skore.TrainTestSplit` instance for more control over the
          splitting parameters.
```
Let's be more concise.

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -281,3 +282,60 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+
+class TestTrainTestSplit:
+    """Tests for the TrainTestSplit splitter class."""
+
+    def test_get_n_splits(self):
+        splitter = TrainTestSplit()
+        assert splitter.get_n_splits() == 1
+
+    def test_split_yields_one_pair(self):
+        import numpy as np
```
GaetandeCast: Don't import inside tests, do it once at the top of the file.

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -281,3 +282,60 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+        assert len(test_idx) == 30
+
+    def test_split_reproducible(self):
+        import numpy as np
+
+        X = np.arange(200).reshape(100, 2)
+        splits_a = list(TrainTestSplit(random_state=42).split(X))
+        splits_b = list(TrainTestSplit(random_state=42).split(X))
+        np.testing.assert_array_equal(splits_a[0][0], splits_b[0][0])
+        np.testing.assert_array_equal(splits_a[0][1], splits_b[0][1])
```
GaetandeCast: We should not test scikit-learn: our tests should only cover what skore does ie the init and its default, and the two class methods. So you can remove those tests.

skore/tests/unit/sklearn/train_test_split/test_train_test_split.py
```diff
@@ -281,3 +282,60 @@ def test_train_test_split_check_dict_no_X_no_y():
...
+
+    def test_default_random_state_is_zero(self):
+        assert TrainTestSplit().random_state == 0
+
+    def test_default_test_size(self):
+        assert TrainTestSplit().test_size == 0.2
+
```
GaetandeCast, line 309: Let's test all defaults in a single test.

raotalha71:
Applied in ecda1b3e, thanks for the detailed review.
Generalized TrainTestSplit docstring scope to “any skore or scikit-learn function” (not evaluate-specific).
Updated the example to show train, test arrays directly, with expected output adjusted.
Simplified the splitter=float description in evaluate() docstring to the concise wording you suggested.
Moved numpy import to the top of the test module.
Removed tests that were validating scikit-learn internals (test_size, reproducibility).
Consolidated defaults checks into a single test_defaults test.

Issue #2135, opened 2025-11-02
feat(skore): Add model calibration display


glemaitre:
### What would you like to say?

We should provide a wrapper around the `CalibrationDisplay` of scikit-learn. It should follow our `Display` API where `.frame` should provide a dataframe and `.plot` is a figure.

The more difficult part to answer is to decide in which accessor to make this display available. I can think of two potential locations:

- `.model`
- `.diagnosis`

Having the former would then question why `feature_importance` is not inside it. For `diagnosis`, we should think a bit more what other tools could go there.

So I think that we first need to answer those questions.

GaetandeCast:
Suggestion: we rename the `feature_importance` module to `inspection` and include the calibration display, along the feature importance displays/methods.

GaetandeCast:
After discussing IRL with @glemaitre, we agreed that it would be nice to rename the `feature_importance` accessor to `inspection`, as well as renaming the `.permutation()` method (soon to be returning a display, #2304) to `permutation_importance()` and perhaps renaming `mean_impurity_decrease` to `impurity_decrease`. This API change could be done first and the calibration display in a second time.

snath-xoc:
ping @glemaitre I would be happy to get started on implementation of calibration display (agreed it should be in ```.diagnosis``` more so)

PR #2595, opened 2026-03-14
feat(skore): Implement CalibrationDisplay for EstimatorReport

snath-xoc:
#### Change description
Initial minimal code attempt at adding CalibrationDisplay into skore. Linked to issue #2135. Still in a draft stage.

#### Contribution checklist
requires updating tests)

#### AI usage disclosure
AI tools were involved for:

Some to-dos still remain:

- [ ] Test whether there is a better way of porting the sklearn ```CalibrationDisplay```/```_BinaryClassifierCurveDisplayMixin``` as current code unpacks them to use the base components in skore
- [ ] Include the inspection tool in comparison and cross-validation reports
- [ ] Handle multi-class  classifiers (already almost there with the keyword argument ```pos_label```
- [ ] Streamline plotting code
- [ ] Add tests


snath-xoc:
As a preliminary check one can use the following code:

```python

X, y = make_classification(
    n_samples=100_000, n_features=20, n_informative=2, n_redundant=10, random_state=42
)

    X, y, test_size=0.99, random_state=42)
lr = LogisticRegression(C=1.0)

logisticregression_report = EstimatorReport(
    lr,
    X_train=X_train,
    X_test=X_test,
    y_train=y_train,
    y_test=y_test,
)
logisticregression_report.inspection.calibration_curve().frame()
```
Output:
<table border="1" class="dataframe">
  <thead>
    <tr style="text-align: right;">
      <th></th>
      <th>predicted_probability</th>
      <th>fraction_of_positives</th>
      <th>data_source</th>
      <th>label</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <th>0</th>
      <td>0.049821</td>
      <td>0.067349</td>
      <td>test</td>
      <td>1</td>
    </tr>
    <tr>
      <th>1</th>
      <td>0.293844</td>
      <td>0.346451</td>
      <td>test</td>
      <td>1</td>
    </tr>
    <tr>
      <th>2</th>
      <td>0.501131</td>
      <td>0.553564</td>
      <td>test</td>
      <td>1</td>
    </tr>
    <tr>
      <th>3</th>
      <td>0.707339</td>
      <td>0.726359</td>
      <td>test</td>
      <td>1</td>
    </tr>
    <tr>
      <th>4</th>
      <td>0.944889</td>
      <td>0.936064</td>
      <td>test</td>
      <td>1</td>
    </tr>
  </tbody>
</table>
</div>

```python
logisticregression_report.inspection.calibration_curve().plot()
```
Output:
image

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +341,61 @@ def permutation_importance(
 
         return display
 
+    def calibration_curve(
+        self,
+        *,
+        data_source: DataSource = "test",
+        n_bins: int = 5,
+        strategy: Literal["uniform", "quantile"] = "uniform",
+        response_method: Literal["auto", "predict_proba", "decision_function"] = "auto",
```
glemaitre: I would make that we remove this option and only use `predict_proba`

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +341,61 @@ def permutation_importance(
 
         return display
 
+    def calibration_curve(
+        self,
+        *,
+        data_source: DataSource = "test",
+        n_bins: int = 5,
+        strategy: Literal["uniform", "quantile"] = "uniform",
```
glemaitre: Let's default to "quantile"

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +341,61 @@ def permutation_importance(
...
+        response_method: Literal["auto", "predict_proba", "decision_function"] = "auto",
+        pos_label: int | str | None = None,
```
glemaitre: The `pos_label` will depend on the way forward in `https://github.com/probabl-ai/skore/pull/2663`. The way we think right now is to expose it in `plot` and `frame` after some discussion and compute the calibration for both possibility if the default `pos_label` was not defined before.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +341,61 @@ def permutation_importance(
...
+    ) -> CalibrationDisplay:
+        """Display the calibration curve.
+
+        This method is available for binary classification reports. It computes
+        the calibration curve data for each report and returns a display that
+        can be plotted with :meth:`CalibrationDisplay.plot`.
+
+        Parameters
+        ----------
+        data_source : DataSource | None, optional
```
glemaitre: Here, the style for the type does not really follow the scikit-learn convention. I would suggest to use it.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +341,61 @@ def permutation_importance(
...
+            The data source for the calibration curve. If None, the default data
+            source is used.
+        n_bins : int, optional
```
glemaitre: It would mean something like:

```suggestion
        n_bins : int, default=5
```

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,216 @@
...
+    Attributes
+    ----------
+    facet_ : seaborn FacetGrid
+        FacetGrid containing the coefficients.
+
+    figure_ : matplotlib Figure
+        Figure containing the plot.
+
+    ax_ : ndarray of matplotlib Axes
+        Array of matplotlib Axes with the different matplotlib axis.
```
glemaitre: Now, we don't store those anymore: https://github.com/probabl-ai/skore/pull/2660

Basically, it just mean that the `.plot` will return the matplotlib figure instead.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,216 @@
...
+        X: ArrayLike,
+        y: ArrayLike,
+        report_type: ReportType,
+        n_bins: int = 5,
+        strategy: Literal["uniform", "quantile"] = "uniform",
+        response_method: Literal["auto", "predict_proba", "decision_function"] = "auto",
+        pos_label: int | str | None = None,
+    ) -> CalibrationDisplay:
+        """Compute the data for the calibration curve display."""
+        # Get predicted probabilities
```
glemaitre: You can remove the comments here and after

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,216 @@
...
+        y_pred, pos_label = _get_response_values_binary(
```
glemaitre: Here, you need to use the cached prediction.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,216 @@
...
+        df = pd.DataFrame(
+            {
+                "predicted_probability": predicted_probability,
+                "fraction_of_positives": fraction_of_positives,
+            }
+        )
+        df["estimator"] = name
+        df["data_source"] = data_source
+        df["split"] = np.nan
+        df["label"] = pos_label
```
glemaitre: So actually with my aforementioned comment, then it might make sense to store `pos_label` after all :)

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,216 @@
...
+        # We always have to show the legend for at least the reference line
+        self.ax_.legend(loc="lower right")
+        add_background_features = hue is not None
+
+        self.ax_.set_xlabel(f"Mean predicted probability {info_pos_label}")
+        self.ax_.set_ylabel(f"Fraction of positives {info_pos_label}")
+        self.ax_.set_xlim(0, 1)
+        self.ax_.set_ylim(0, 1)
+
+        if add_background_features:
```
glemaitre: This part is indeed not needed

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,216 @@
...
+            self.ax_.plot([0, 1], [0, 1], "k:", label=ref_line_label)
+
+        # We always have to show the legend for at least the reference line
+        self.ax_.legend(loc="lower right")
+        add_background_features = hue is not None
+
+        self.ax_.set_xlabel(f"Mean predicted probability {info_pos_label}")
+        self.ax_.set_ylabel(f"Fraction of positives {info_pos_label}")
+        self.ax_.set_xlim(0, 1)
+        self.ax_.set_ylim(0, 1)
```
glemaitre: Usually, you have a `.set` with matplotlib where you can pass those as kwargs. I prefer it.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,216 @@
...
+        self.figure_, self.ax_ = self.plot_.figure, self.plot_.axes
+        ref_line_label = "Perfectly calibrated"
+        if ref_line:
+            self.ax_.plot([0, 1], [0, 1], "k:", label=ref_line_label)
+
+        # We always have to show the legend for at least the reference line
+        self.ax_.legend(loc="lower right")
+        add_background_features = hue is not None
+
+        self.ax_.set_xlabel(f"Mean predicted probability {info_pos_label}")
```
glemaitre: Could you force to have a squared axis.

glemaitre (commented):
Just the couple of comments that I forgot to post from last week.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
snath-xoc: I suggested a change for now where we just use the pos_label from "_parent"

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +342,61 @@ def permutation_importance(
...
+        n_bins : int, default=5
+            The number of bins to use for the calibration curve. Default is 5.
+        strategy : str, optional
+            The strategy to use for binning. Default is "uniform".
+        response_method : str, optional
+            The method to use for obtaining prediction probabilities. Default is
+            "auto".
+        pos_label : int | str | None, optional
+            The label of the positive class. If None, the default positive label
+            is used.
```
GaetandeCast:
```suggestion
        data_source : {"test", "train"}, default="test"
            The data source for the calibration curve. If None, the default data
            source is used.

        n_bins : int, default=5
            The number of bins to use for the calibration curve. Default is 5.

        strategy : {"uniform", "quantile"}, default="quantile"
            The strategy to use for binning. Default is "uniform".
```
To follow conventions in skore. Also remove `response_method` and `pos_label`.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +342,61 @@ def permutation_importance(
...
+    def calibration_curve(
+        self,
+        *,
+        data_source: DataSource = "test",
+        n_bins: int = 5,
+        strategy: Literal["uniform", "quantile"] = "quantile",
+    ) -> CalibrationDisplay:
+        """Display the calibration curve.
+
+        This method is available for binary classification reports. It computes
```
GaetandeCast, line 346: When you add support for multiclass classification, make sure to look at the recent development in the probability metric displays (PR/ROC/confusion matrix), such as in #2728 and #2663. Now we always compute the data to display for all labels and decide which curve to plot with the following rules:
If `label` is set in `.plot`, only show the curve of the corresponding class. In binary that means treating the `label` class as the positive class. If `label` is not set, use the report's `pos_label` in binary classif if it was set, otherwise show all classes. The same logic is applied to pick which class to query, if any, in `.frame`.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -338,6 +342,61 @@ def permutation_importance(
...
+        Returns
+        -------
+        :class:`CalibrationDisplay`
+            The calibration curve display. :meth:`CalibrationDisplay.plot` for
+            more details.
+
+        Returns
+        -------
+        :class:`CalibrationDisplay`
+            The calibration curve display.
```
GaetandeCast: This is duplicated.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+
+        # Get predicted probabilities
+        y_pred_cache = _get_cached_response_values(
+            cache=cache,
+            estimator_hash=int(estimator_hash),
+            estimator=estimator,
+            X=X,
+            response_method="predict_proba",
+            pos_label=pos_label,
+        )
```
GaetandeCast: Let's move that in the accessor file and do:

```
results = _get_cached_response_values(
    cache=self._parent._cache,
    estimator_hash=int(self._parent._hash),
    estimator=self._parent.estimator_,
    X=X,
    response_method=response_method,
    pos_label=pos_label,
    data_source=data_source,
)
for key_tuple, value, is_cached in results:
    if not is_cached:
        self._parent._cache[key_tuple] = value
    if key_tuple[-1] != "predict_time":
        y_pred = value
```
Then just pass `y_true` and `y_pred` to `_compute_data_for_display` which will then just assume the role of calling `sklearn.calibration_curve` and constructing the dataframe for storing the display data. 
Snippet stolen from `.../_estimator/metrics_accessor.py` which is a good reference on how to use our internal tooling.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+        df["estimator"] = name
+        df["data_source"] = data_source
+        df["pos_label"] = pos_label
+        df["split"] = np.nan
+        return cls(calibration_report=df, report_type=report_type)
+
+    @DisplayMixin.style_plot
+    def plot(self) -> None:
+        """Plot the mean decrease in impurity for the different features.
```
GaetandeCast: This needs updating.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+        >>> display.plot()
+        """
+        return self._plot()
+
+    def _plot_matplotlib(self) -> None:
+        """Dispatch the plotting function for matplotlib backend.
+
+        This method creates a bar plot showing the mean decrease in impurity for each
+        feature using seaborn's catplot. For cross-validation reports, it uses a
+        strip plot with boxplot overlay to show the distribution across splits.
```
GaetandeCast: Same

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+
+    _default_line_kwargs = {"marker": "s", "linestyle": "-", "color": "blue"}
+
+    def __init__(self, *, calibration_report: pd.DataFrame, report_type: ReportType):
+        self.calibration_report = calibration_report
+        self.report_type = report_type
+
+    def frame(self) -> pd.DataFrame:
+        """Return the data used for the display as a DataFrame."""
```
GaetandeCast: This should document what the method returns.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+            }
+        )
+        df["estimator"] = name
+        df["data_source"] = data_source
+        df["pos_label"] = pos_label
+        df["split"] = np.nan
+        return cls(calibration_report=df, report_type=report_type)
+
+    @DisplayMixin.style_plot
+    def plot(self) -> None:
```
GaetandeCast: As mentioned above, this will eventually need a `label` parameter to select which class to plot. Same in `.frame`.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+            lineplot_kwargs=lineplot_kwargs,
+        )
+
+    def _plot_calibration_curve_single_estimator(
+        self,
+        *,
+        frame: pd.DataFrame,
+        estimator_name: str,
+        hue: str | None = None,
+        ref_line: bool = True,
```
GaetandeCast: I'm wondering if this parameter is useful. In the prediction error display, there is no toggle on the perfect prediction line but in the ROC & PR displays there is a `plot_chance_level` param, so we are inconsistent in what we expose. I would argue in favor of removing the toggle in ROC and PR since I don't think imposing the line on the user is problem. Any opinion @glemaitre @jeromedockes ?

Also in the current state of the code, `ref_line` is not exposed to the user.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+        _despine_matplotlib_axis(
+            self.ax_,
+            axis_to_despine=("top", "right"),
+            remove_ticks=True,
+            x_range=None,
+            y_range=None,
+        )
+        self.figure_.suptitle(f"Calibration Curve of {estimator_name}")
+
+        return self
```
GaetandeCast: You can remove this, skore displays return `None`.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def _plot_calibration_curve_single_estimator(
+        self,
+        *,
+        frame: pd.DataFrame,
+        estimator_name: str,
+        hue: str | None = None,
+        ref_line: bool = True,
+        lineplot_kwargs: dict[str, Any],
+    ):
```
GaetandeCast, line 220:
```suggestion
    ) -> None:
```

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+
+        _despine_matplotlib_axis(
+            self.ax_,
+            axis_to_despine=("top", "right"),
+            remove_ticks=True,
+            x_range=None,
+            y_range=None,
+        )
+        self.figure_.suptitle(f"Calibration Curve of {estimator_name}")
+
```
GaetandeCast, line 258: You should implement the `set_style` override.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,222 @@
...
+        estimator_name: str,
+        hue: str | None = None,
+        ref_line: bool = True,
+        lineplot_kwargs: dict[str, Any],
+    ):
+        self.plot_ = sns.lineplot(
+            data=frame,
+            x="predicted_probability",
+            y="fraction_of_positives",
+            markers=True,
```
GaetandeCast: Default values for seaborn plotting parameters should be set in `_default_line_kwargs`, otherwise users can't change their values with `set_style`. Right now as a user I could not remove markers for instance.

GaetandeCast (commented):
Looking solid! Here is a set of suggestions. They are mostly about aligning with the functionalities and conventions of the code base.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
snath-xoc: I added in the pos_label option, hope it looks O.K.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
snath-xoc: Tried to add this with the ```report_pos_label``` argument, hope it looks O.K.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
snath-xoc: I now use ```self._parent.get_prediction``` I'm guessing the``` _get_cached_response_values``` is no longer available (could not see it after updating the branch)?

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
snath-xoc: Removed the parameter now but kept the line in the plot... unless we want to remove the line altogether?

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
snath-xoc: Given the previous comment, it seems like now I have to return Figure?

snath-xoc:
I now am able to plot multiple labels to get a figure as follows:

image

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
GaetandeCast: Indeed, I forgot we switched to that behaviour x)

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
GaetandeCast: No let's keep the line, it's helpful. I'll open the discussion somewhere else on whether it's relevant to expose params such as `plot_chance_level` in the other displays.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
GaetandeCast: Yes that's perfect, sorry I was looking at an old version of the code.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
GaetandeCast: It does :)

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -327,6 +331,56 @@ def permutation_importance(
 
         return display
 
+    @available_if(
+        _check_supported_ml_task(supported_ml_tasks=["binary-classification"])
+    )
```
GaetandeCast, line 349:
```suggestion
    @available_if(_check_estimator_has_method("predict_proba"))
    def calibration_curve(
```
We need to assert that the classifier has a `predict_proba`

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,300 @@
...
+            lineplot_kwargs = {k: v for k, v in lineplot_kwargs.items() if k != "color"}
+            facet = sns.lineplot(
+                data=frame,
+                x="predicted_probability",
+                y="fraction_of_positives",
+                hue="label",
+                **lineplot_kwargs,
+            )
+        else:
+            facet = sns.lineplot(
```
GaetandeCast: Let us switch to a `sns.relplot`. The reason is threefold:
- It will be needed when we add support for comparisons and cross validations
- relplot creates its own figure. Currently, calling `calibration_curve().plot()` twice in a row overlaps the plots in the same figure.
- relplot has a `height` parameter that lets us define the size of the figure. Our default for this parameter in skore is 6, which is a bit larger than the current view (whose size can't be changed rn).

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -327,6 +331,56 @@ def permutation_importance(
...
+            The strategy to use for binning. Default is "uniform".
+
+        Returns
+        -------
+        :class:`CalibrationDisplay`
+            The calibration curve display. :meth:`CalibrationDisplay.plot` for
+            more details.
+
+        """
+        y_pred = self._parent.get_predictions(
```
GaetandeCast:
```suggestion
        y_pred = self._parent._get_predictions(
```
The public version of `get_predictions` will use the report's `pos_label` when it is not `None` and return a  `(n_sample,)` array. We need the full `(n_sample, n_classes)` array, in case the `label` in `plot` is different from the `parent_pos_label`. So let's use the private `_get_prediction`

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,300 @@
...
+        """Return the data used for the display as a DataFrame.
+
+        Parameters
+        ----------
+        label : int, float, bool, str or None, default=report pos_label
+            The class whose curve to select. Use ``None`` to show all classes.
+
+        Returns
+        -------
+        DataFrame
```
GaetandeCast, line 106: Maybe add a little more details on the content of the dataframe.

GaetandeCast (commented):
Here is an other pass, we are getting close! Once we have support for multiclass, I think we can merge and look at comparisons and cross validations in subsequent PRs. I'll also need to make a pass on the tests.

snath-xoc:
I can't reproduce the ```test_kwargs``` and ```test_plot_structure``` failures locally, any ideas as to why they happen?

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,316 @@
...
+            facet = sns.relplot(
+                data=frame,
+                x="predicted_probability",
+                y="fraction_of_positives",
+                hue="label",
+                **lineplot_kwargs,
+            )
+        else:
+            facet = sns.relplot(
+                data=frame,
```
GaetandeCast:
```suggestion
            facet = sns.relplot(
                data=frame,
                kind="line",
                x="predicted_probability",
                y="fraction_of_positives",
                hue="label",
                **lineplot_kwargs,
            )
        else:
            facet = sns.relplot(
                data=frame,
                kind="line",
```
The lines don't show currently.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,316 @@
...
+    6               0.290888               0.287212        test      1
+    7               0.500070               0.526809        test      1
+    8               0.707157               0.706303        test      1
+    9               0.941637               0.940048        test      1
+    """
+
+    _default_line_kwargs = {
+        "marker": "s",
+        "linestyle": "-",
+        "color": "blue",
```
GaetandeCast:
```suggestion
        "palette": "tab10",
```
Let's use the same palette as in other displays.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,316 @@
...
+        *,
+        frame: pd.DataFrame,
+        estimator_name: str,
+        lineplot_kwargs: dict[str, Any],
+        ax_set_kwargs: dict[str, Any],
+        label: PositiveLabel = None,
+    ):
+        if "label" in frame.columns:
+            # Multiple classes: let seaborn handle colours via hue
+            lineplot_kwargs = {k: v for k, v in lineplot_kwargs.items() if k != "color"}
```
GaetandeCast: ~Then this is can be removed.~
This can move to the else branch with `"color"` changed to `"palette"`

skore/tests/unit/displays/calibration_curve/conftest.py
```diff
@@ -0,0 +1,44 @@
...
+
+@pytest.fixture(scope="module")
+def estimator_reports_binary_classification_figure_axes(
+    pyplot, estimator_reports_binary_classification
+):
+    report = estimator_reports_binary_classification[0]
+    display = report.inspection.calibration_curve()
+    fig = display.plot()
+    return fig, fig.axes
+
```
GaetandeCast, line 15: We can add fixtures for multiclass classification too now. And use them in `test_common`

GaetandeCast (commented):
Let's also add an entry to the changelog and then it's good for me!
cc @glemaitre, want to have another look?

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -328,6 +333,57 @@ def permutation_importance(
 
         return display
 
+    @available_if(
+        _check_supported_ml_task(supported_ml_tasks=["binary-classification"])
```
GaetandeCast: Hmm strange I realize that I was able to try the display for multiclass data locally without this check stopping me.

snath-xoc: Hmmmm same for me, let me check.

skore/src/skore/_sklearn/_plot/inspection/calibration_curve.py
```diff
@@ -0,0 +1,318 @@
...
+        else:
+            facet = sns.relplot(
+                data=frame,
+                kind="line",
+                x="predicted_probability",
+                y="fraction_of_positives",
+                label=estimator_name,
+                hue=None,
+                **lineplot_kwargs,
+            )
```
GaetandeCast:
```suggestion
        if "label" in frame.columns:
            lineplot_kwargs["hue"] = "label"
        else:
            lineplot_kwargs = {
                k: v for k, v in lineplot_kwargs.items() if k != "palette"
            }
            lineplot_kwargs["label"] = estimator_name
            lineplot_kwargs["hue"] = None

        facet = sns.relplot(
            data=frame,
            kind="line",
            x="predicted_probability",
            y="fraction_of_positives",
            **lineplot_kwargs,
        )
```
Last suggestion, it avoids a warning when label is not None.
```bash
UserWarning: Ignoring `palette` because no `hue` variable has been assigned.
```

GaetandeCast (changes requested):
Final touch ! I'll investigate the `_check_supported_ml_task` issue separately.

skore/src/skore/_sklearn/_estimator/inspection_accessor.py
```diff
@@ -328,6 +333,57 @@ def permutation_importance(
...
+    )
```
GaetandeCast:
```suggestion
        _check_all_checks(
            [
                _check_supported_ml_task(
                    supported_ml_tasks=[
                        "binary-classification",
                        "multiclass-classification",
                    ]
                ),
                _check_estimator_has_method("predict_proba"),
            ]
        )
    )
```
Found how to make multiple checks :)

snath-xoc: Aha great thanks for that :)

GaetandeCast (approved):
LGTM, thanks a lot @snath-xoc, great work!

Issue #2544, opened 2026-03-02
chore: Remove caching logic from `CrossValidationReport` and `ComparisonReport`


auguste-probabl:
### What would you like to say?

Since these are essentially lists of `EstimatorReports`, they already benefit from the caching at the `EstimatorReport` level, so the added complexity adds very little performance benefits.

thomass-dev:
Please note that the cache is currenlty used to retrieve permutation importance for both `EstimatorReport` and `CrossValidationReport` reports.

cakedev0:
I 100% agree with this issue.

----

@thomass-dev 

> Please note that the cache is currenlty used to retrieve permutation importance for both EstimatorReport and CrossValidationReport reports.

In `skore-hub-project` right? The logic could be adapted for `CrossValidationReport`: we could inspect the cache of children `EstimatorReport`s, and then call `cv_report.inspection.permutation_importance()` if we see that this will just read from cache.

And in the long term, the plan is to stop inspecting a private attribute of reports in `skore-hub-project`, right? (what you proposed in https://github.com/probabl-ai/skore/issues/2200 should go in this direction).

thomass-dev:
I also agree with the issue. My previous message was just a point to consider and address during the development.

> In skore-hub-project right? 

Yes, and we can easily change `skore-hub-project` without breaking anything: at worst, the permutation is not sent to the hub.

> the plan is to stop inspecting a private attribute of reports in skore-hub-project, right?

In the long term, yes, totally. But at short term, the cache could also become a public object.

cakedev0:
Can I start working on that? It seems we all agree, and the changes will not be very important I think.

Once the changes merged, I'll open an issue to detail the steps that we'll have to make `skore-hub-project` to continue uploading permutation. 

Edit: as per offline discussion with @auguste-probabl, I'm moving forward.

cakedev0:
Actually, the cache cannot be blindly removed, indeed for metrics displays, we don't rely on children reports computations (so we don't benefit from children's cache): https://github.com/probabl-ai/skore/blob/a201e6049e7ba2fa06ac9e4e71010869e504e868/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py#L855

As discussed offline with @auguste-probabl I will:
- remove it where it can be removed (=> just from summarize, which relies on children reports summarize, which itself is cached)
[...]

PR #2600, opened 2026-03-16
chore(skore): Remove caching from CV/Comp where it has no value

cakedev0:
Towards https://github.com/probabl-ai/skore/issues/2544

#### Change description

In CV/Comparison reports, cache is in 3 places:
- `metrics.summarize()` : can be removed without perf loss, because it calls children report, and thus benefits from their cache => **removed**
- `inspection.permutation_importance()`: same => **removed** too.
- `metrics._get_display` => doesn't benefit fully from children cache, so this caching is kept in this PR.

This will **break `skore-hub-project` cache inspection** (used for permutation importance). I tried to find a fix, but it's not really possible to preserve the current behavior entirely without important changes in `skore`. Conclusion: let's move forward with https://github.com/probabl-ai/skore/discussions/2587 first, it will make the fix easier.

Note: It could be nice to **release skore before merging this PR.** This would release `skore[mlflow]` (for now, it can only be installed using `skore-mlflow-project`) and the MLFlow example.

#### Contribution checklist

- [X] Unit tests: adapted some tests to look up children's cache, remove some other tests.


skore/src/skore/_sklearn/_comparison/inspection_accessor.py
```diff
@@ -322,69 +321,30 @@ def permutation_importance(
...
+                    report.inspection.permutation_importance(
+                        data_source=data_source,
+                        at_step=at_step,
+                        metric=metric,
+                        n_repeats=n_repeats,
+                        max_samples=max_samples,
+                        n_jobs=n_jobs,
+                        seed=seed,
+                    )
+                    .importances.copy()
```
thomass-dev:
```suggestion
                    .importances
```

skore/src/skore/_sklearn/_cross_validation/inspection_accessor.py
```diff
@@ -249,71 +248,30 @@ def permutation_importance(
...
+                    report.inspection.permutation_importance(
+                        data_source=data_source,
+                        at_step=at_step,
+                        metric=metric,
+                        n_repeats=n_repeats,
+                        max_samples=max_samples,
+                        n_jobs=n_jobs,
+                        seed=seed,
+                    )
+                    .importances.copy()
```
thomass-dev:
```suggestion
                    .importances

Comment (thomass-dev) [reply skore/src/skore/_sklearn/_cross_validation/inspection_accessor.py]: `df.assign` returns a new dataframe.
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html

File: skore/src/skore/_sklearn/_comparison/inspection_accessor.py (hunk 1 above)
Comment (thomass-dev) [reply skore/src/skore/_sklearn/_comparison/inspection_accessor.py]: `df.assign` returns a new dataframe.
https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html

File: skore/src/skore/_sklearn/_comparison/inspection_accessor.py [hunk 3]
```diff
@@ -322,69 +321,30 @@ def permutation_importance(
         -----
         Even if pipeline components output sparse arrays, these will be made dense.
         """
-        # NOTE: to temporary improve the `project.put` UX, we always store the
-        # permutation importance into the cache dictionary even when seed is None.
```
thomass-dev, line 326: Keep in mind that calling `permutation_importance` from `skore-hub-project` with default parameters (`seed=None`) will not have any gain from the cache, as it is always re-computed.

```python
# skore/src/skore/_sklearn/_estimator/inspection_accessor.py::113

def permutation_importance(
[...]
# NOTE: avoid to fetch from the cache if the seed is None because we want      
# to trigger the computation in this case. We only have the permutation        
# stored as a workaround for the serialization for skore-hub as explained      
# earlier.                                                                     
display = None if seed is None else self._parent._cache.get(cache_key)         
if display is None:                                                            
    display = PermutationImportanceDisplay._compute_data_for_display(          
        data_source=data_source,                                               
        estimator=self._parent.estimator_,                                     
        name=self._parent.estimator_name_,                                     
        X=X_,                                                                  
        y=y_true,                                                              
        at_step=at_step,                                                       
        metric=metric,                                                         
        n_repeats=n_repeats,                                                   
        max_samples=max_samples,                                               
        n_jobs=n_jobs,                                                         
        seed=seed,                                                             
        report_type=self._parent._report_type,                                 
    )                                                                          
                                                                               
    if cache_key is not None:                                                  
        # NOTE: for the moment, we will always store the permutation importance
        self._parent._cache[cache_key] = display                               
```

thomass-dev: This is certainly why https://github.com/probabl-ai/skore/pull/2601/changes is not working well.

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
cakedev0: Yup, I didn't noticed that because I just changed the indentation of this piece of code ^^ good catch thanks!

cakedev0: (and concat rewrite all the data too)

skore/src/skore/_sklearn/_comparison/inspection_accessor.py
cakedev0: > This is certainly why https://github.com/probabl-ai/skore/pull/2601/changes is not working well.

That's indeed why, I reached this same conclusion. Hence I will close #2601 and as stated in my PR desc:

> This will break skore-hub-project cache inspection (used for permutation importance). I tried to find a fix, but it's not really possible to preserve the current behavior entirely without important changes in skore. Conclusion: let's move forward with https://github.com/probabl-ai/skore/discussions/2587 first, it will make the fix easier.
> 
> Note: It could be nice to release skore before merging this PR. This would release skore[mlflow] (for now, it can only be installed using skore-mlflow-project) and the MLFlow example.

thomass-dev: Let's draft the PR until tomorrow.

thomass-dev:
> Note: It could be nice to release skore before merging this PR. This would release skore[mlflow] (for now, it can only be installed using skore-mlflow-project) and the MLFlow example.

Yes totally agree! I will make a release tomorrow morning.

PR #2604, opened 2026-03-16
feat(skore): Add compare function as a counterpart to evaluate

### LINKED ISSUE #2603: feat(skore): Add a `compare` method

Following #2520 and especially https://github.com/probabl-ai/skore/issues/2520#issuecomment-3987627477, we can add a `compare` method that is fed built `EstimatorReports` or `CrossValidationReports` and returns a `ComparisonReport`. 
The difference with a call to `evaluate()` with multiple estimators is that with `compare` we don't create the reports.
AFAIK this would just wrap the `__init__` of `ComparisonReport` so the only use of this method would be to use natural language instead of instanciating a class.

glemaitre:
closes #2603

Follow-up to the PR introducing the `evaluate` function.

Introduce the `compare` function that allow to create a `ComparisonReport`.


sphinx/reference/index.rst
```diff
@@ -30,6 +30,8 @@ ML Assistance
 .. list-table::
    :widths: 30 70
 
+   * - :func:`compare`
```
auguste-probabl, line 35: ~`evaluate` should be here also~
i guess wait for #2573

PR #2605, opened 2026-03-17
fix(skore-hub-project): Add average precision and roc auc to binary classification performance

rouk1:
Two optional parameters were not passed when computing performance related data frames.


cakedev0:
Just out of curiosity: how did you notice that? What did it break?

thomass-dev:
> Just out of curiosity: how did you notice that? What did it break?

Missing for https://github.com/probabl-ai/skore-hub/pull/1285.

PR #2615, opened 2026-03-19
refactor(skore): Remove `._cache` from CV&Comp reports

### LINKED ISSUE #2544: chore: Remove caching logic from `CrossValidationReport` and `ComparisonReport`

cakedev0:
Depends on #2600 and #2610 (both are merged :heavy_check_mark: )

Closes #2544 

#### Change description

This PR removes `.cache` from CV&Comp reports because it's not used anymore, and adapts the `skore-*-project`s to be compatible with this removal (while being backward compatible with the current skore release, as proven by the tests).

It also prunes a bit the `Project` docstring to make the doctest pass by removing the put of a CV report. Indeed the doc test is ran with `skore:main` against `skore-local-project:latest`, which is something we don't really try to support.

#### AI usage disclosure

Used it to adapt the tests.

### DIFF
--- a/skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
+++ b/skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
@@ -44,9 +44,9 @@ class PermutationImportance(Inspection[Report], ABC):  # noqa: D101
     name: Literal["permutation_importance"] = "permutation_importance"
 
     def content_to_upload(self) -> bytes | None:  # noqa: D102
-        if not hasattr(self.report, "_cache"):
-            return None
-        for key, display in reversed(list(self.report._cache.items())):
+        cache = getattr(self.report, "_cache", {})
+
+        for key, display in reversed(list(cache.items())):
             if len(key) < 6:
                 continue
 


skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
```diff
@@ -44,6 +44,8 @@ class PermutationImportance(Inspection[Report], ABC):  # noqa: D101
     name: Literal["permutation_importance"] = "permutation_importance"
 
     def content_to_upload(self) -> bytes | None:  # noqa: D102
+        if not hasattr(self.report, "_cache"):
+            return None
         for key, display in reversed(list(self.report._cache.items())):
```
thomass-dev:
```suggestion
        cache = getattr(self.report, "_cache", {})

        for key, display in reversed(list(cache.items())):
```

skore-hub-project/src/skore_hub_project/artifact/pickle/pickle.py
```diff
@@ -48,7 +48,12 @@ def content_to_upload(self) -> Generator[bytes, None, None]:
         The report is pickled without its cache, to avoid salting the checksum.
         """
         reports = [self.report] + getattr(self.report, "estimator_reports_", [])
-        caches = [report_to_clear._cache for report_to_clear in reports]
+        reports_with_cache = [
+            report_to_clear
+            for report_to_clear in reports
+            if hasattr(report_to_clear, "_cache")
+        ]
+        caches = [report_to_clear._cache for report_to_clear in reports_with_cache]
```
thomass-dev:
```suggestion
        reports_with_cache = (
            (report, report._cache) for report in reports if hasattr(report, "_cache")
        )
```

skore-hub-project/src/skore_hub_project/artifact/pickle/pickle.py
```diff
@@ -60,5 +65,5 @@ def content_to_upload(self) -> Generator[bytes, None, None]:
 
             yield pickle_bytes
         finally:
-            for report, cache in zip(reports, caches, strict=True):
+            for report, cache in zip(reports_with_cache, caches, strict=True):
```
thomass-dev:
```suggestion
            for report, cache in reports_with_cache:
```

skore-hub-project/tests/unit/report/test_cross_validation_report.py
thomass-dev, line 1: Same comments as `pickle.py`.

skore-hub-project/tests/unit/report/test_estimator_report.py
thomass-dev, line 1: Same comments as `pickle.py`.

skore-local-project/src/skore_local_project/project.py
thomass-dev, line 1: Same comments as `pickle.py`.

skore-hub-project/src/skore_hub_project/artifact/pickle/pickle.py
cakedev0: Quite nicer indeed, applied it to every relevant places.

cakedev0: But a generator is not good here: it's lazy so it will read `report._cache` after the call to `cache_clear` below.

Personally I avoid inline generators like this unless there is a clear performance/memory gain from it, because it's much harder to reason about than lists.

thomass-dev: You're true. Sorry, i didn't want a generator here, but a tuple 😓 .
So `reports_with_cache = tuple(`.

PR #2632, opened 2026-03-19
feat(skore): First draft of repr for the reports

GaelVaroquaux:
A first draft of a meaningful display for the estimator report.

This is to be finished by the skore team:
* Add tests
* Add the same to the cross-validation report
* Update the getting started to use this
* Improve the rendering, but in a later PR

### DIFF
--- a/skore/tests/unit/reports/estimator/test_report.py
+++ b/skore/tests/unit/reports/estimator/test_report.py
@@ -327,11 +327,17 @@ def test_has_no_deep_copy():
 
 @pytest.mark.parametrize("with_train", [False, True])
 @pytest.mark.parametrize("with_test", [False, True])
-def test_report_repr_html(with_train, with_test):
+@pytest.mark.parametrize("bad_estimator", [False, True])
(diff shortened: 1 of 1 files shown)


glemaitre:
Thanks @GaelVaroquaux 

We actually discuss this earlier in the week with @jeromedockes and @GaetandeCast when tackling another issue.

I'll take over since from a the discussion with @GaelVaroquaux and @jeromedockes yesterday, it was pretty clear what we would want to have.

jeromedockes:
is this ok (the first dropdown is open and the others closed when the repr is first displayed)

screenshot_2026-03-20T10:11:51+01:00

GaelVaroquaux:
That's great. I think that longer run, we need to find a lighter way of showing the "help" hint, so that it gets less in the way.

Also, putting a summary of the estimator name (eg the title of the column used in the dataframe) in the title (visible part) of the dropdown that is folded would be great, because it helps seeing what's going on at a glance.

Thanks a lot!!

Issue #2617, opened 2026-03-19
Auto diagnosis: implement model checks


GaetandeCast:
### What would you like to say?

Implement all the checks on the model that are gathered [here](https://docs.google.com/document/d/1GaH5mXV1xcFbkwU-quO8h2Pzcce7GLrI1HipWnEUy-A/edit?tab=t.0 ), by priority order.

glemaitre:
I already had request for adding custom diagnosis. From the work that we currently do for metric I'm thinking that the higher level diagnose method could look at a diagnostics registry to who one could had a new diagnostic. A diagnostic will have a code and a function to be run. I think that what would make sense here is a function that take the report self and then it should allow the full flexibility to our user to start crafting their own. Then we can easily add an ignore them as any of our own.  

WDYT @GaetandeCast?

glemaitre:
This might a direct impact on how we want to store the diagnosis and it might be easier to already take it into account.

GaetandeCast:
Yeah I love the idea! It should not be especially hard to do since right now adding a new diagnosis is simply a matter of creating a function that reads a report and outputs a well formatted object. It would be nice to do early on, as it would make the process of adding new "built-in" checks more straightforward.

This raises the need for a consistent vocabulary within the API, so that we can document adding custom checks nicely. Here is the lingo I set up in #2714: `.diagnose` returns a _diagnostic_ of the report, composed of _issues_ found by _checks_, identified by their code and ran by `.diagnose`. I will create a dedicated PR to factor out that change and create a registry system. Then we can look at adding new diagnostics, which should become an easy process.

GaetandeCast:
Closing this as we have implemented most of the checks that were planned here.

PR #2634, opened 2026-03-20
feat(skore): Add the main components of the diagnostics API

### LINKED ISSUE #2616: Auto diagnosis: Build the API

Implement the diagnosis API:
- add `*Report.diagnose()` which runs the diagnosis. It should have a parameter to toggle expensive diagnostics on or off. It should return helpful strings with a short explanation for each diagnostic, a link to the relevant entry in the diagnostic documentation and a code/id used to mute/turn off the diagnostic.
- create a page of the documentation (in the user guide?) that lists the diagnostics and for each explains how we detected it, why it is a problem, and what to do to avoid it. Eventually it could link to a code example showing the issue on a real world dataset and the steps to take to avoid the pitfall.
- add an html representer for the `*Report` that shows the `.help()` repr and a pannel showing the number of issues detected/diagnostics made, and suggesting to run `.diagnose()` for further details. We will probably run the "fast" diagnostics by default at initialization time and suggest `.diagnose()` for further inspection.
- add a toggle system. We want the diagnostics to be opt-in, i.e. let the users choose when they want to see their diagnostics and not throw it in their face. Therefore we should have a `diagnose : bool = False` parameter on the `*Report.__init__` to optionally run `.diagnose()` at the end of the initialization and a `skore.config(diagnose=True)` to toggle that parameter on for all reports.
- We need to define how to present the collected information to the user. The type of object that is returned, and how it hierarchies and groups warnings.

GaetandeCast:
Closes #2616 and towards #2617.

Add a `diagnose()` method to all report types that detects common modeling issues and returns actionable feedback.

This PR focuses on implementing the engine and only adds two checks: overfitting and underfitting.
Overfitting (SKD001) flags when a strict majority of default metrics show a significant train/test gap, and underfitting (SKD002), flags when neither train nor test scores meaningfully outperform a dummy baseline. Thresholds are scale-aware (percentage of reference score with a floor) to avoid false positives on near-zero or large-valued metrics.

Diagnostics are computed lazily on first `diagnose()` call and cached on `_diagnostics_cache`. `.diagnose()` is called when the repr of any estimator report is generated, to include a Diagnostic panel suggesting to run `.diagnose()` for more details.

For `CrossValidationReport`, per-split diagnostics are computed at the `EstimatorReport` level and then aggregated: a code is surfaced only if a strict majority of splits triggered it. `ComparisonReport` collects diagnostics from its sub-reports and displays them together, prefixed with the estimator name.

Individual codes can be muted per-call with `diagnose(ignore=["SKD001"])` or globally with `configuration(ignore_diagnostics=["SKD001"])`. Both sources are merged and ignored codes reduce the "checks ran" counter in the output.

`DiagnosticResults` renders to Rich panels (terminal), HTML with doc links (notebooks), and plain text. 

The diagnostics are documented in a new section of the user guide and the API is showcased in a new example in the "technical details" section.


jeromedockes:
thanks @GaetandeCast ! a few high-level questions from a first glance:

- probably it's just me, but "diagnostics", "diagnose", etc. feel like a mouthful -- have we considered simpler terms like "checks"?
- why do we need the configuration? can't we just compute and return the diagnostics whenever diagnose() is called?
- I think calling the diagnostics should return an object containing all the information with good repr and html repr instead of displaying them and returning None, WDYT?
- I like that you put it directly in the report so I can do `report.diagnose` but if we plan to have several diagnostics-related methods and for consistency maybe we should consider (one more) accessor. I'm not saying I would be in favor of an accessor but just something to think about
- the diagnostic context seems quite specific to the 2 diagnostics currently implemented. I wonder if at least at first we could simplify a bit and not have a diagnostic context, and each check computes what it needs. We can allow one check to return several diagnosticresults (diagnoses I guess? :innocent: ), so that one function can setup context / collect information useful for several of them and return them all, for example the under-fitting and over-fitting checks
- rather than having separate steps to check if a diagnostic applies and then running it, I would suggest running it and having it raise a specific exception when it does not apply. this will remove the need for the pre-run-check, which I suspect for some diagnoses might end up duplicating part of the code for actually performing the check
- more generally than the last 2 suggestions above, it would be great if there are other simplifications we can make to reduce a bit the number of layers / indirections and the amount of code in the first iteration. we will want to be able to experiment and play with the api and the diagnostics themselves and that may be easier if we have less code (I know this is very vague and it maybe we actually can't reduce it further)

GaetandeCast:
Thanks for the feedback @jeromedockes, I'll try to address your questions/concerns:
> "diagnostics", "diagnose", etc. feel like a mouthful

I didn't really think about it, to me if feels okay. I do agree that we could introduce words like 'checks' and 'pitfall' in the vocabulary we use here to make it less redundant and simpler.

> why do we need the configuration? can't we just compute and return the diagnostics whenever diagnose() is called?

So that we don't have to manually call `.diagnose()` on every report. It adds a bit of complexity so if we are fine with having to explicitly run the checks on each report then I can remove it.

> I think calling the diagnostics should return an object containing all the information with good repr and html repr instead of displaying them and returning None, WDYT?

I went back and forth on that question, in the current state (bff607d25f3e23ea8f1b1080d90d9cb4bc875698) `diagnose()` has the behaviour you suggest. I rolled back a76445d4fb0d6e3119b375017396b3531cce39dc which introduced the return `None` + display behaviour. You must have reviewed there.

> if we plan to have several diagnostics-related methods

Not sure we do as of yet. If we happen to get there, then yeah it will make sense to add an accessor. Though I'd be more in favor of packing things as much as possible in `.diagnose()` and structuring its output rather than fragmenting in methods.

>  We can allow one check to return several diagnosticresults

That makes sense. I can group multiple diagnostic that use the same context into a single runner that returns a collection of `DiagnosticResult`.

> I would suggest running it and having it raise a specific exception when it does not apply.

Sure.

> it would be great if there are other simplifications we can make to reduce a bit the number of layers / indirections and the amount of code in the first iteration. 

I agree, I did my best to make things as straightforward and as modular as possible but there are probably improvements to find. The implementation of all the checks down the line will also surely dictate a lot of changes and I can't anticipate them here, so it's gonna be an iterative process.

GaetandeCast:
I made the following changes while trying to keep your comments in mind @jeromedockes:

- integrate the changes made on main to the repr of the reports
- simplify the UX of the feature: it is now limited to `report.diagnose()`, which removes the bloat around `diagnose=True` in the config and inits. Also it makes it more straightforward for the user, when I want to run a diagnosis, I do `.diagnose()` and that's it. We compute and cache results when we display the report repr though. For now it's fine because the checks are cheap but we may have to revisit this as we add more expensive checks.
- simplify the running of the diagnostics: no more shared context or diagnostic check class. Just a runner that throws and exception if test data is not available and that checks for underfitting and overfitting.
- simplify the call stack between `.diagnose()` and display, by removing the special output that was done for `ComparisonReport`. We now display a flat list of issues, prefixed by the estimator name. 

Disclaimer: I made heavy usage of claude code for these changes, to try it out with the team plan. I was careful to review everything it output.

jeromedockes:
> I made the following changes while trying to keep your comments in mind @jeromedockes

thanks @GaetandeCast ! I will look at it today

skore/src/skore/_sklearn/_diagnostics/base.py
```diff
@@ -0,0 +1,178 @@
...
+        Detailed message describing what was found
+    """
+
+    code: str
+    title: str
+    docs_anchor: str
+    explanation: str
+
+
+class DiagnosticResults:
```
jeromedockes: can we call it DiagnosticResultSet or something similar, to be more different from DiagnosticResult

skore/src/skore/_utils/repr/diagnostics_panel.html.j2
```diff
@@ -0,0 +1,4 @@
+<div class="report-diagnostics-details">{{ n_issues }} issue(s) across {{ n_checks }} check(s).</div>
+<p class="report-diagnostics-hint">
+    Run <code>.diagnose()</code> for details on detected issues.
+</p>
```
glemaitre: I would prefer to follow the style that we have for other reports, e.g.

```
        <p class="report-inline-note" role="note">
            Investigate more with
            <a
                href="{{ inspection_accessor_doc_url }}"
                target="_blank"
                rel="noopener noreferrer"
                class="report-hint-note-link"
            >{{ report_class_name }}.inspection</a>.
        </p>
```

I'm thinking that we could move this message in the estimator, cross-validation, and comparison reports template just for consistency.
The HTML repr of the diagnostic panel should only scope to the useful information that we want to display.

I think that the current way is good enough for this iteration. Later, I think that we could improve the level of information to state maybe the categories in which we have failure or that we run diagnostic. Right now, with only two checks, it make little sense.

glemaitre: Also, it allows to incorporate the documentation link to the `diagnosis` method.

skore/src/skore/_sklearn/_diagnostics/base.py
```diff
@@ -0,0 +1,178 @@
...
+            self._messages = [format_diagnostic_message(d) for d in diagnostics]
+        else:
+            self._messages = ["No issues were detected in your report!"]
+        self.header = (
+            f"Diagnostics: {len(self._diagnostics)} issue(s) detected, "
+            f"{self._checks_ran} check(s) ran, {n_ignored} ignored."
+        )
+
+    @property
+    def diagnostics(self) -> list[DiagnosticResult]:
```
jeromedockes: should this be diagnostic_results? in general, what terminology did you choose for

- the function performing the check
- the check result with a code and message

sphinx/user_guide/diagnostics.rst
```diff
@@ -0,0 +1,105 @@
+.. _diagnostics:
+
+===========
+Diagnostics
+===========
+
+`skore` diagnostics provide quick checks for common model quality pitfalls.
+Use :meth:`report.diagnose()` to get concise findings, each with:
```
glemaitre: It might be better to directly refer to `EstimatorReport.diagnose` to be able to link to the API doc. We could state that it is available for all reports.z

glemaitre: Or we could mention below when we speak specifically about those type of reports.

skore/src/skore/_sklearn/_diagnostics/base.py
```diff
@@ -0,0 +1,178 @@
...
+        """All detected diagnostic issues."""
+        return self._diagnostics
+
+    def __iter__(self):
```
jeromedockes: would it be clearer to just expose `self.diagnostics` and `self.messages`, so that we know what we are getting (the formatted messages or the full objects) and we don't need to implement this protocol here?

skore/src/skore/_sklearn/_diagnostics/base.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __eq__(self, other):
+        if isinstance(other, DiagnosticResults):
+            return self._messages == other._messages
+        if isinstance(other, (list, tuple)):
+            return list(self._messages) == list(other)
+        return NotImplemented
+
+    def __bool__(self):
+        return bool(self._messages)
```
jeromedockes: I'm not sure we need those

skore/tests/unit/utils/repr/test_base.py
```diff
@@ -243,3 +243,32 @@ def test_repr_html_mixin_repr_mimebundle():
...
+
+
+def test_report_diagnostics_html_fragment_no_issues(report_with_base_help):
+    """Fragment shows zero counts when no diagnostics are detected."""
+    html = report_with_base_help._diagnostics_html_fragment()
+    assert ".diagnose()" in html
+    assert "0 issue(s) across 0 check(s)" in html
+
+
+def test_report_diagnostics_html_fragment_with_issues(
```
glemaitre: I'm wondering if only testing the higher level HTML repr is not enough instead of importing the private object. We can test (or already test) the `DiagnosticResult` class unitely.

glemaitre: I think that we added this higher level HTML repr test as well.

skore/src/skore/_sklearn/_estimator/diagnostics.py
```diff
@@ -0,0 +1,256 @@
...
+
+    Both checks share the same pre-conditions and metric data, so they are
+    grouped in a single function.  Raises :class:`DiagnosticNotApplicable`
+    when train+test data is unavailable.
+    """
+    if not _has_train_and_test(report):
+        raise DiagnosticNotApplicable
+
+    estimator_pairs = _metric_pairs(report)
+    baseline_pairs = _baseline_metric_pairs(report)
```
jeromedockes: maybe we can join the 2 dataframes on metric name here and iterate directly on it to avoid the MetricKey and MetricPair abstractions?

skore/src/skore/_sklearn/_diagnostics/base.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+
+    def _repr_mimebundle_(self, **kwargs: object) -> dict[str, str]:
+        return {"text/plain": self._to_plain_text(), "text/html": self._repr_html_()}
+
+    def __str__(self) -> str:
+        return self._to_plain_text()
+
+    def __repr__(self) -> str:
+        return self._to_plain_text()
```
glemaitre: I would suggest a rewrap option to limit the width (and the number of characters). But I would differe to @jeromedockes that uses terminal much more than me.

skore/src/skore/_sklearn/_estimator/diagnostics.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    return max(floor, fraction * max(abs(reference) for reference in references))
+
+
+def _is_significant_gap(metric_pair: MetricPair) -> bool:
+    """Check whether the train-favored gap indicates potential overfitting.
+
+    The gap threshold is 10% of the reference score (floor 0.03).
+    """
+    if metric_pair.favorability == "(↗︎)":
```
jeromedockes: it seems a bit surprising to make decisions based on strings meant for display, do we have a more robust way to track the favorability of metrics? for example so that this doesn't break if we ever decide that `➚` is prettier than `↗` or something like that

skore/src/skore/_sklearn/_estimator/diagnostics.py
```diff
@@ -0,0 +1,256 @@
...
+                        f"than the dummy baseline for {n_positive}/{total} "
+                        "comparable metrics."
+                    ),
+                )
+            )
+
+    return results
+
+
+DiagnosticCheckFn: TypeAlias = Callable[["EstimatorReport"], list[DiagnosticResult]]
```
jeromedockes: should that be Diagnostic? DiagnosticCheck seems redundant?

skore/src/skore/_sklearn/_estimator/diagnostics.py
```diff
@@ -0,0 +1,256 @@
...
+    report: EstimatorReport,
+) -> list[DiagnosticResult]:
+    """Check for overfitting (SKD001) and underfitting (SKD002).
+
+    Both checks share the same pre-conditions and metric data, so they are
+    grouped in a single function.  Raises :class:`DiagnosticNotApplicable`
+    when train+test data is unavailable.
+    """
+    if not _has_train_and_test(report):
+        raise DiagnosticNotApplicable
```
jeromedockes: raise DiagnosticNotApplicable() (an instance not the exception type)

skore/src/skore/_sklearn/_diagnostics/base.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def __rich_console__(
```
glemaitre: I would expect to use the rich console when using the default `__repr__`. Here, I therefore have the impression that it is not used. If you want to show, then you will need to import the console singleton from skore.

skore/src/skore/_sklearn/_estimator/diagnostics.py
```diff
@@ -0,0 +1,256 @@
...
+    """Check whether the report has both train and test X/y data."""
+    return (
+        report.X_train is not None
+        and report.y_train is not None
+        and report.X_test is not None
+        and report.y_test is not None
+    )
+
+
+def check_overfitting_underfitting(
```
glemaitre: I'm thinking that we could already create a `diagnosis` module and have submodule inside it that would categorize the type of diagnostics.

skore/src/skore/_sklearn/_estimator/diagnostics.py
```diff
@@ -0,0 +1,256 @@
...
+    output: str
+
+
+@dataclass(frozen=True, slots=True)
+class MetricPair:
+    """A metric's train and test scores with its favorability direction."""
+
+    favorability: str
+    train: float
+    test: float
```
glemaitre: In the future, we should be reusing what is currently happening in the metrics registry.

skore/src/skore/_sklearn/_estimator/diagnostics.py
glemaitre: I'm sure that once we have the metric registry than we will have a `greater_is_better` that is a boolean. Here, I don't recall if we already have the info.

skore/src/skore/_sklearn/_estimator/diagnostics.py
```diff
@@ -0,0 +1,256 @@
...
+        pos_label=report.pos_label,
+    )
+    return _metric_pairs(baseline_report)
+
+
+class DiagnosticNotApplicable(Exception):
+    """Raised when a diagnostic check cannot run on the given report."""
+
+
+def _has_train_and_test(report: EstimatorReport) -> bool:
```
glemaitre: Why not only having the `if/else`?

skore/src/skore/_utils/repr/report.css
```diff
@@ -186,6 +186,28 @@
...
+    line-height: 1.5;
+    color: var(--color-text);
+    padding: 0.5em 0;
+}
+
+.report-diagnostics-details {
+    margin: 0;
+}
+
+.report-diagnostics-hint {
```
glemaitre: I think that we would need those since we would use the same style than others hints.

skore/tests/unit/reports/comparison/test_report.py
```diff
@@ -26,6 +32,105 @@ def report(request):
     return request.getfixturevalue(request.param)
 
 
+def test_diagnose_collects_component_diagnostics(report, monkeypatch):
```
glemaitre, line 35: Let's not split for the moment but we are going to quicly have several diagnostics.

skore/src/skore/_sklearn/_base.py
```diff
@@ -31,6 +38,84 @@ class _BaseReport(ReportHelpMixin):
...
+        ----------
+        ignore : list of str or tuple of str or None, default=None
+            diagnostic codes to exclude from the results, e.g.
+            `["SKD001"]`
+
+        Returns
+        -------
+        DiagnosticResults
+            an iterable of human-readable messages for every detected issue,
+            with the full :class:`~skore.DiagnosticResult` objects accessible
```
glemaitre: The `DiagnosticResult` class will not be public. We should therefore make it public.

However, I'm worried about it here: we extend skore with a new type of object with additional API. Internally, I think it is fine to define our dataclass to handle object and it is readable. But for user facing, I'm wondering if retuning a dictionary where the key are the key of the diagnostic and value are another dict with the essential information is not enough.

glemaitre: @jeromedockes do you have any thoughts on this?

GaetandeCast:
> Also, the link to the diagnostic explanation in the documentation didn't work for me in the generated sphinx gallery.

Yes it's because the user guide page does not exist yet on the documentation. It does get generated on the preview and the link should be right once the PR is deployed. But we can't test it until then unfortunately.

skore/src/skore/_sklearn/_base.py
GaetandeCast: Yeah I think we need to have a discussion on the return type for the `.diagnose` method. I understand the burden that comes with creating a public class. The advantage is that we can define a representer so that we just need to return the object and we can define how it is displayed in the different environments, with a nice html view when possible etc. Otherwise we keep it simple with a dictionary of strings but that won't be pretty. Maybe we consider that we will do something pretty in the hub and the purpose of the lib is only to expose the necessary information exhaustively. Final option I see is to have `.diagnose` directly print the info and return either None or the dictionary containing the printed info.

skore/src/skore/_sklearn/_estimator/diagnostics.py
GaetandeCast: Indeed as of now, the favorability is encoded only in that character. #2646 changes that by making it a bool.

skore/src/skore/_sklearn/_estimator/diagnostics.py
GaetandeCast: Yes, currently I'm parsing the result of the summarize which does not feel good but is the best way to get the info I need.

skore/src/skore/_sklearn/_base.py
glemaitre: I review my position. The dict will not be enough.

I'm wondering if we could not make it a display. Basically, we would not plot anything, but the `.frame` would return the current dict in the form of a table and the HTML repr of the display will be the one that we have now.

This might not be a surprising API.

sphinx/user_guide/diagnostics.rst
GaetandeCast: Went with option 2.

skore/src/skore/_sklearn/_estimator/diagnostics.py
GaetandeCast: Created a `model_checks.py` file to contain all the checks listed in #2617.

skore/src/skore/_sklearn/_estimator/diagnostics.py
GaetandeCast: Inlined the function.

skore/src/skore/_sklearn/_estimator/diagnostics.py
GaetandeCast: Done

skore/tests/unit/reports/comparison/test_report.py
GaetandeCast: Not sure what you mean by that. That we should separate the tests in new files once we have more diagnostics to test?

GaetandeCast:
Hey @jeromedockes and @glemaitre I've addressed your reviews. Here is a quick summary of the changes made:
  - get rid of unnecessary abstractions: `MetricsKey` and `MetricsPair`
  - Moved diagnostics computation to their own `model_checks.py` file
  - centralized relevant code to the `diagnostics` module.
  - added links to `*Report.diagnose` in the user guide
  - change return type to display and removed `DiagnosticResult` in favor of a dict of dict, keyed by diagnostic code.
  - align diagnostics panel in restimator repr with existing

skore/tests/unit/reports/comparison/test_report.py
glemaitre: Yes it was what my cryptic comment meant :)

examples/technical_details/plot_diagnostics_api.py
```diff
@@ -0,0 +1,112 @@
+"""
+.. _example_diagnostics_api:
+
+===================
+The diagnostics API
```
jeromedockes: can we make the title not about the thing we describe, but what it provides to the user: "Automatic detection of modelling issues" or something along those lines

GaetandeCast: done!

PR #2635, opened 2026-03-20
feat(skore): Support skrub dataops in reports

### LINKED ISSUE #2133: feat(skore): Integration with `skrubs.DataOp` plan

Recently, `skrubs` released a new type of "pipeline" called DataOps plan. More information are available here: https://skrub-data.org/stable/data_ops.html

Supporting this feature could open the door for new feature to track data and code provenance. However, there is work to do regarding the compatibility because the input are not anymore `X` and `y`.

So there is a lot to investigate to know how to integrate this feature into `skore`.

jeromedockes:
closes #2133 

Still a rough draft. The different reports can now be initialized with a SkrubLearner rather than a regular scikit-learn estimator. This could allow:

- enabling users to benefit from both skrub and skore
- capturing data wrangling / preprocessing operations that would otherwise be invisible to skore
- perhaps exploiting skrub's built-in previews and subsampling, and maybe one day caching
- maybe opportunities for reporting and diagnoses through inspection of the computation graph and intermediate results
- reporting on hyperparameter search as skrub allows easily specifying hyperparameter ranges (or other kinds of tunable choices) and can perform search either backed by scikit-learn or optuna

### DIFF
--- a/skore/src/skore/_sklearn/_cross_validation/report.py
+++ b/skore/src/skore/_sklearn/_cross_validation/report.py
@@ -185,6 +185,8 @@ class CrossValidationReport(_BaseReport, DirNamesMixin):
         super().__init__()
         self._raw_estimator = estimator
         if isinstance(estimator, skrub.DataOp):
+            if data is None:
+                data = estimator.skb.get_data()
             estimator = estimator.skb.make_learner()
         if is_clusterer(estimator):
             raise ValueError(
--- a/skore/tests/unit/reports/cross_validation/test_report.py
+++ b/skore/tests/unit/reports/cross_validation/test_report.py
@@ -1,13 +1,14 @@
 import joblib
 import numpy as np
 import pytest
+import skrub
 from sklearn.base import clone
 from sklearn.cluster import KMeans
 from sklearn.datasets import make_classification
 from sklearn.dummy import DummyClassifier
 from sklearn.ensemble import RandomForestClassifier
 from sklearn.exceptions import NotFittedError
-from sklearn.linear_model import LinearRegression
+from sklearn.linear_model import LinearRegression, LogisticRegression
 from sklearn.model_selection import train_test_split
 from sklearn.pipeline import Pipeline
 from sklearn.utils._testing import _convert_container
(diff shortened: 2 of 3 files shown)


examples/use_cases/plot_fraud_prediction.py
```diff
@@ -0,0 +1,87 @@
+"""
+Using skore with a skrub DataOp
```
glemaitre: I think that for a use case, I would formulate the example a bit differently, something along the line "tracking data preprocessing when data come from multiple table".

We could use this tiny example where "skrub" and "dataops" would appear. Right now, it would be in the technical details section but I'm thinking that could have a section about integration when we just want to show how to use skore with some specific library.

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -149,21 +152,43 @@ class CrossValidationReport(_BaseReport, DirNamesMixin):
...
+            estimator = estimator.skb.make_learner()
         self._estimator = clone(estimator)
 
-        # private storage to be able to invalidate the cache when the user alters
-        # those attributes
-        self._X = X
-        self._y = y
+        if is_skrub_learner(estimator):
+            self._initialized_with_data_op = True
+            if X is not None or y is not None:
```
glemaitre: I would isolate into a `_validate_input_data` and return X, y, data.

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -146,47 +147,85 @@ def _copy_estimator(cls, estimator: BaseEstimator) -> BaseEstimator:
 
     def __init__(
         self,
-        estimator: BaseEstimator,
+        estimator: BaseEstimator | skrub.SkrubLearner,
```
glemaitre:
```suggestion
        estimator: BaseEstimator | skrub.SkrubLearner | skrub.DataOp,
```

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -149,48 +149,86 @@ def _copy_estimator(cls, estimator: BaseEstimator) -> BaseEstimator:
...
+                split = estimator.skb.train_test_split()
+                test_data = split["test"]
+                train_data = split["train"]
+            estimator = estimator.skb.make_learner()
         if is_clusterer(estimator):
             raise ValueError(
                 "Clustering models are not supported yet. Please use a"
                 " classification or regression model instead."
             )
+        if is_skrub_learner(estimator):
```
glemaitre: I also feel that we are missing an abstraction here.

glemaitre:
I'm pretty much fine with what is going on. I'm wondering if we could a sort of abstraction to validate both the data and the estimator at the `__init__`. But it might be only that things needs to be in 2 dedicated functions and then it makes that the rest is clearer.

PR #2639, opened 2026-03-20
chore(skore): Clean-up cache keys

cakedev0:
#### Change description

Simplify cache keys (allowed because reports are immutable now), and make all cache key share the same structure:
- data source (train or test, nothing else)
- name (of the method / the display / ...)
- kwargs (can be empty)

#### AI usage: no

### DIFF
--- a/ci/requirements/skore/python-3.13/scikit-learn-1.8/sphinx-requirements.txt
+++ b/ci/requirements/skore/python-3.13/scikit-learn-1.8/sphinx-requirements.txt
@@ -103,8 +103,6 @@ graphql-core==3.2.8
     #   graphql-relay
 graphql-relay==3.2.0
     # via graphene
-greenlet==3.3.2
-    # via sqlalchemy
 gunicorn==25.1.0
     # via mlflow
 h11==0.16.0
@@ -390,7 +388,7 @@ six==1.17.0
     # via python-dateutil
 skops==0.13.0
     # via mlflow
-skore-hub-project==0.0.18
+skore-hub-project==0.0.19
     # via skore (skore/pyproject.toml)
 skore-local-project==0.0.5
     # via skore (skore/pyproject.toml)
--- a/skore-hub-project/src/skore_hub_project/authentication/login.py
+++ b/skore-hub-project/src/skore_hub_project/authentication/login.py
@@ -29,7 +29,7 @@ def login(*, timeout: int = 600) -> None:
     global credentials
 
     if credentials is not None:
-        logger.debug(f"Already logged in {URI} with {type(credentials)}.")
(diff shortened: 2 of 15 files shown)


skore-hub-project/src/skore_hub_project/report/report.py
```diff
@@ -118,10 +118,7 @@ def metrics(self) -> list[Metric[Report]]:
 
             for task in progress.track(
                 as_completed(tasks),
-                description=(
-                    f"Computing {self.report.__class__.__name__} "
-                    f"#{self.report._hash} metrics"
-                ),
+                description=f"Computing {self.report.__class__.__name__} metrics",
```
thomass-dev: The `_hash` identifier of the report is used outside the cache to reference the report (here, and in https://github.com/probabl-ai/skore/pull/2114).

Could you leave it, even if it's not used anymore in the cache?
I don't think it is needed to remove it here in this PR (a bit out-of-scope i think).
We can revisit `_hash` later, especially when we will revisit the report serialization.

cakedev0: It's out-of-scope indeed, I'll revert that.

cakedev0: (for another PR:)

But maybe using `id(report)` would be a cleaner alternative? (I'm not happy to see another access to a private thing outside of the class).

And also `_hash` is a pretty bad name. It's not a hash it's a UUID. Maybe let's change it to `self.id = uuid.uuid4()`? It would make the intend clearer.

thomass-dev: Agree with what you say, accessing a private thing is a bad idea. `uuid4` sounds good.

thomass-dev: The downside of `id(report)` is that it isn't necessarily preserved using `pickle/depickle`.
I think we should continue to have a dedicated property (public this time 😅 ).

cakedev0: Note: I re-introduced `._hash` but I still remove `._rng` in this PR (because if you see this rng attribute you will think it's used elsewhere, which is a false assumption. A code that makes you make false assumptions is bad, so let's clean it).

cakedev0: > The downside of id(report) is that it isn't necessarily preserved using pickle/depickle.

Good point!

Let's open a PR to do that (public `self.id = uuid4()`), do you want to do it? I can do it otherwise.

thomass-dev: https://github.com/probabl-ai/skore/pull/2649 It can be merged before or after the current PR, no matter.

Issue #2061, opened 2025-10-04
feat(skore): Saving custom/extra metrics to reports


divakaivan:
### Is your feature request related to a problem? Please describe.

Right now skore provides some default metrics available in a report with `report.metrics.<metric_name>` and also `report.metrics.summarize().frame()`. It'd be nice if a user could also save any extra metric they want their report to have and the report exposes access to. If I I have a `custom_metric`, in order to see it's results I need to do

```python
report.metrics.summarize(
    scoring=metric_function,
    scoring_kwargs={"response_method": "predict"},
)
```

 or 

```python
report.metrics.custom_metric(
    metric_function=metric_function,
    response_method="predict",
)
```

which requires me to have access to the `metric_function` function's definition. 

In addition, if I want to see `report.metrics.summarize().frame()`, I cannot see the default metrics + my metric by default. I'd need to pass all the metrics I want to `summarize(scoring=["rmse", "r2", metric_function])`. I think it would be nice if I could save a custom metric and whenever I do `report.metrics.summarize().frame()` all the metrics (the default ones by skore and mine) are shown. And also, if I share reports, I wouldn't necessarily expect the other user to have access to my `metric_function`, so if I could save my metric in some way into the report that'd be nice. 

### Describe the solution you'd like

If I want to save a custom metric to be part of my report, I'd like to be able to do something like

```python
report.metrics.save_custom_metric(name="my own metric", metric_function=metric_function)
```

Then I (or if I share the report, a colleague) can do:

```python
report.metrics.custom_metric("my_own_metric")
```

Also if I have a custom metric saved, I would expect to see it alongisde the provided by default metrics. (and maybe have a flag like `summarize(include_custom=True)`

```python
report.metrics.summarize().frame()
# RMSE ...
# R2 ...
# my_own_metric
```

### Additional context

from a comment on discord

glemaitre:
I think it is a good idea. We previously discussed IRL about creating a sort of register for metric such that you can plug new metric and it should be used then by default. It is really aligned with what you propose here.

PR #2646, opened 2026-03-20
refactor: Create metrics registry

auguste-probabl:
Part of #2061

#### Contribution checklist

- [ ] The code passes our style conventions (you can check this by running pre-commit on your code
requires updating tests)
- [ ] All the commits in the PR are signed (more information

#### AI usage disclosure

AI tools were involved for:


skore/tests/unit/reports/estimator/metrics/test_summarize_new.py
auguste-probabl, line 1: merge with existing tests

auguste-probabl, line 1: ensure tests follow our conventions

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +45,153 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+
+        Returns
+        -------
+        dict[str, Metric]
+        """
+        items: list[tuple[str | None, MetricLike]]
+        if metric is None or metric == []:
+            items = [(None, m) for m in self._registry]
+        elif isinstance(metric, dict):
+            items = list(metric.items())
```
auguste-probabl, line 201: supporting this case `summarize({"a": "r2", "b": "rmse"})` adds a fair amount of complexity. we could simplify things by keeping the custom names on the side until the `row` definition

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    ml_task: MLTask, estimator: BaseEstimator
+) -> dict[str, Metric]:
+    if ml_task == "binary-classification":
+        metrics = [Accuracy, Precision, Recall, RocAuc]
+        if hasattr(estimator, "predict_proba"):
+            metrics.append(Brier)
+    elif ml_task == "multiclass-classification":
+        metrics = [Accuracy, Precision, Recall]
+        if hasattr(estimator, "predict_proba"):
+            metrics += [RocAuc, LogLoss]
```
GaetandeCast: I understand it's to follow the old behaviour that you did that, but why don't we add the log loss as a built-in for binary classification?

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+    R2,
+    Rmse,
+]
+
+
+def _get_default_metrics(
+    ml_task: MLTask, estimator: BaseEstimator
+) -> dict[str, Metric]:
+    if ml_task == "binary-classification":
+        metrics = [Accuracy, Precision, Recall, RocAuc]
```
GaetandeCast: `RocAuc` requires the estimator to have a `predict_proba`. I tried creating a classifier that has only a `predict` and the `metrics.summarize()` crashed through `_check_roc_auc()`.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+        if hasattr(estimator, "predict_proba"):
+            metrics.append(Brier)
+    elif ml_task == "multiclass-classification":
+        metrics = [Accuracy, Precision, Recall]
+        if hasattr(estimator, "predict_proba"):
+            metrics += [RocAuc, LogLoss]
+    else:
```
GaetandeCast:
```suggestion
    if "classification" in ml_task:
        metrics = [Accuracy, Precision, Recall]
        if hasattr(estimator, "predict_proba"):
            metrics += [RocAuc, LogLoss, Brier]
    else:
```
So to conclude, my suggestion for defaults is this. I don't think there is a need to differentiate binary and multi-class classification here as I don't really see why we show either the log loss or the brier score, and not both always.

skore/src/skore/_sklearn/types.py
```diff
@@ -49,7 +49,7 @@ class SKLearnScorer(Protocol):
     _kwargs: dict[str, Any]
 
 
-Metric = str | Callable | SKLearnScorer
+MetricLike = str | Callable | SKLearnScorer
```
GaetandeCast, line 52: Could we find a better name than `MetricLike`? It looks like it refers to the type we expect as input argument to the public side of the metrics API. So maybe something like `MetricArgument` or `MetricInput` or something?

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -174,152 +296,32 @@ def summarize(
...
-                    )
-
-            # NOTE: we have to check specifically for `_BaseScorer` first because this
-            # is also a callable but it has a special private API that we can leverage
-            if isinstance(metric_, _BaseScorer):
-                # scorers have the advantage to have scoped defined kwargs
+        for metric_name, metric_obj in parsed_metrics.items():
+            if metric_obj.is_builtin:
+                # Built-in metric: dispatch via the accessor method by name
+                metric_fn = getattr(self, metric_obj.name)
```
```suggestion
        for metric_name, metric_object in parsed_metrics.items():
            if metric_object.is_builtin:
                # Built-in metric: dispatch via the accessor method by name
                metric_function = getattr(self, metric_object.name)
```
Nit: use full words

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +45,154 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+        metric_kwargs : dict or None
+            The kwargs to pass to each metric; each metric takes
+            only the kwargs they can accept.
+
+        Returns
+        -------
+        dict[str, Metric]
+        """
+        items: list[tuple[str | None, MetricLike]]
+        if metric is None or metric == []:
```
```suggestion
        if metric is None or metric == [] or metric == {}:
```
`.summarize(metric={})` crashes

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +45,154 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+        if not self._parent._metrics_registry:
+            self._parent._metrics_registry = _get_default_metrics(
+                self._parent._ml_task, self._parent._estimator
+            )
+
+    @property
+    def _registry(self) -> dict[str, Metric]:
+        return self._parent._metrics_registry
+
+    def _parse_metric(self, m: MetricLike, metric_kwargs: dict[str, Any]) -> Metric:
```
```suggestion
    def _parse_metric(self, metric: MetricLike, metric_kwargs: dict[str, Any]) -> Metric:
```
Nit again: `m` is not a great name IMO

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +45,154 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+            items = [(None, m) for m in self._registry]
+        elif isinstance(metric, dict):
+            items = list(metric.items())
+        elif isinstance(metric, list):
+            items = [(None, m) for m in metric]
+        else:
+            items = [(None, metric)]
+
+        result = {}
+        for display_name, m in items:
```
GaetandeCast, line 208:
```suggestion
        for display_name, metric in items:
```

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +45,154 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+            kwargs = m._kwargs.copy()
+            if "pos_label" in inspect.signature(m._score_func).parameters:
+                if (
+                    "pos_label" in kwargs
+                    and self._parent.pos_label != kwargs["pos_label"]
+                ):
+                    raise ValueError(
+                        "`pos_label` is passed both in the scorer: "
+                        f"{kwargs['pos_label']!r} and when creating "
+                        f"the report: {self._parent.pos_label!r}. "
```
```suggestion
                        "The `pos_label` passed in the scorer: "
                        f"{kwargs['pos_label']!r} and when creating "
                        f"the report: {self._parent.pos_label!r} don't match. "
```
The original formulation was always a bit weird to me.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +45,154 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+        """
+        Convert a single "metric-like" to a Metric.
+
+        `metric_kwargs` will be stripped down to only the kwargs that the metric
+        will actually accept.
+        """
+        if isinstance(m, _BaseScorer):
+            func_name = m._score_func.__name__
+            if func_name.startswith("neg_"):
+                func_name = func_name[4:]
```
GaetandeCast: I think the `neg_` prefix is not present in any scikit-learn scorer. I think it's just used in the string representation of scorers. `_BaseScorer` has a `sign` attribute to store that info.

GaetandeCast (commented):
Looks great, thanks @auguste-probabl! Here is a first round. I'll review the tests once you merged them with the suite. Cool to see that your changes fix and test the ergonomy issues that were raised previously!

skore/src/skore/_sklearn/types.py
auguste-probabl: I like it, it's similar to `ArrayLike`: something that looks like a Metric, but needs validation

skore/src/skore/_sklearn/metrics.py
auguste-probabl: I've applied this change, it's much simpler!

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: Reformulated

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: Nice catch, indeed this is dead code.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: However, it looks like until 1.7, sklearn's brier score only accepts binary classification.

auguste-probabl: I'm going to keep this change out of the PR to avoid making it bigger.

skore/src/skore/_sklearn/types.py
thomass-dev: I find it confusing that a metric-like cannot be a metric: aren't we opposing two notions, scorer (what is named `MetricLike`) and metric (what is `Metric`/accessor)?

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Can you give a reproducer? For me the following works:
```python
import skore
from sklearn.svm import SVC

r = skore.evaluate(SVC(), X, y)

print(hasattr(r.estimator_, "predict_proba"))  # False

print(r.metrics.roc_auc())  # 1.0
```

skore/src/skore/_sklearn/types.py
auguste-probabl: Not sure I understand what you mean; to me, a `Metric` is a well-defined object which I've defined in this PR; a `MetricLike` is something that may be converted to a `Metric` using `_parse_metric`, i.e. a string, a sklearn Scorer, or a callable.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Oh, I think this snippet passes because our `roc_auc` is available for binary classification models with `decision_function`, which `SVC` has. However, it is not available for multiclass models with `decision_function`.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,156 @@
...
+        by name; a callable for custom metrics.
+
+    response_method : {"predict", "predict_proba", "decision_function"} or None, \
+        default="predict"
+        The method to call to get the predicted values that will passed to
+        ``score_func``.
+
+    kwargs : dict, default={}
+        Keyword arguments to pass to ``score_func``.
+
```
thomass-dev, line 33: Nitpick: i wonder if this attribute is really useful; `score_func` can be a partial when you need specific parameter.

skore/src/skore/_sklearn/metrics.py
thomass-dev, line 1: I question the pattern/architecture, which i find tedious.

Here, you define the exhaustive list of `BUILTIN_METRICS`, but only with base attributes such as `name`, `verbose_name` and no `score_func`. So in this file, the instances of metrics remain an "empty shell".
The resolution/affectation of `score_func` is done afterward in `_parse_metrics` using inspection, even if the metric is a builtin that we control. There are bits and pieces of logic here and there. Define a new builtin metric seems harder to me than it should be.

So my question is: why not creating a dedicated subclass of `Metric` for each accessor? So, in a way, split the `metrics_accessor` into several parts: `accuracy.py`, `rmse.py`, etc, where each class defines its own logic of computing.

IMO, the code would be much more modular and readable.

thomass-dev (changes requested):
Thanks for this huge part!

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Indeed, this PR has been quite painful to write.

> Here, you define the exhaustive list of BUILTIN_METRICS

This is not supposed to be there, it's a remainder of a bad merge conflict resolution. I've removed it.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Since https://github.com/probabl-ai/skore/pull/2646 was merged my approach of declaring upfront which are the default metrics doesn't really make sense anymore; I've removed it.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: One difficulty is that the metrics often need access to information from the underlying report (e.g. the ML task or the cache) so they cannot be separated cleanly into their own objects.

skore/src/skore/_sklearn/metrics.py
GaetandeCast: Let's try and remember to do a follow-up.

skore/src/skore/_sklearn/metrics.py
GaetandeCast: Indeed I meant when there is no `predict_proba` nor `decision_function`.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,122 @@
...
+    verbose_name : str
+        Display name shown in reports (e.g. ``"Accuracy"``).
+
+    greater_is_better : bool or None, default=None
+        Whether a higher value is better (``True``), lower is better
+        (``False``), or there is no preference or information (``None``).
+
+    score_func : callable or None, default=None
+        The scoring function. ``None`` for built-in metrics that are dispatched
```
glemaitre, line 23: Do you plan to make follow-up PR to remove the `built-in` metrics. Basically, I envision that the metric accessor should define dynamically the available metric by looking at the metric registry and we don't need anymore some built-in metrics.

skore/src/skore/_sklearn/types.py
glemaitre: The definition of `ArrayLike` is the following: https://scikit-learn.org/stable//glossary.html#term-array-like

So I would be OK with `MetricLike` if what we are passing to a `as_metric()` function will return a `Metric`.

skore/src/skore/_sklearn/metrics.py
thomass-dev: > One difficulty is that the metrics often need access to information from the underlying report

Sorry i don't get it. If the report is an attribute of the metric, you have free access.

```python
class Accuracy(Metric):
    def __init__(self, report):
        self.report = report
        
        self.name = ...
        self.verbose_name = ...
```

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +57,165 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
-        "roc_auc": {"name": "ROC AUC", "icon": "(↗︎)"},
-        "log_loss": {"name": "Log loss", "icon": "(↘︎)"},
-        "r2": {"name": "R²", "icon": "(↗︎)"},
-        "rmse": {"name": "RMSE", "icon": "(↘︎)"},
-        "custom_metric": {"name": "Custom metric", "icon": ""},
-    }
-
     def __init__(self, parent: EstimatorReport) -> None:
         super().__init__(parent)
+        if not self._parent._metrics_registry:
```
glemaitre, line 63: I don't think that it should be the job of the accessor to create the registry. Basically, during the `__init__` of the report, we will have all the info to directly create it there.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +57,165 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+            self._parent._metrics_registry = self._get_default_metrics()
+
+    @property
+    def _registry(self) -> dict[str, Metric]:
+        return self._parent._metrics_registry
```
glemaitre, line 68: I think that we could access directly the registry from parent. It is one indirection less.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +57,165 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+                Recall,
+                RocAuc,
+                LogLoss,
+                Brier,
+            ]
+        else:  # regression
+            default_metric_names = [R2, Rmse]
+        metrics = [m for m in default_metric_names if hasattr(self, m.name)]
+        metrics += [FitTime, PredictTime]
+        return {m.name: m for m in metrics}
```
glemaitre, line 84: I also expect this to live at the report level as well.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -40,28 +57,165 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
...
+
+    def _parse_metric(
```
glemaitre, line 86: In some way this method is `Metric` factory. I don't think that making it part of the metric accessor is the best. I think this should belong to the a class that manage the metrics.

glemaitre (commented):
I think that in terms of design, I would prefer to go directly closer to what we had in https://github.com/probabl-ai/skore/pull/2277.

Currently, we have a pure Python dictionary as a metric registry and a couple of methods around living in the accessor to create `Metric` to add into it. Later, we will have a `register` method that is going also to leave in an accessor from what I can see in another PR.

This tell me that a registry will need a way to store `Metric` and a methods to create and manage them. And so the class design in #2277 would be best with a factory similar to you `_parse_metric` that would be equivalent to actually a `register`.

Then, in the current implementation of `Metric` in this PR adding a new method which is not built-in, will not be using the cached prediction. So it seems better to directly go the way #2277 went by adding the method that get the cached prediction and then call the `score_func`. We can simplify a bit in a first iteration by avoiding to cache the results of the metric.

auguste-probabl:
> Then, in the current implementation of `Metric` in this PR adding a new method which is not built-in, will not be using the cached prediction.

In the current impl, when you pass a custom metric, it will lead to calling the `.custom_metric()` method, which will go through `_compute_metric_scores()` which will cache the results and use the cached predictions.

PR #2651, opened 2026-03-20
feat(skore): Improve style HTML repr estimator report

### LINKED ISSUE #2655: The EstimatorReport display asks me to enable JavaScript

Playing in vscode, I find that when I want to look at the data in the display of EstimatorReport, it asks me to enable javascript:
Image
I don't understand why this is the case, given that just above I have a TableReport that works just well without asking me this.

It seems that we should try to work around this, and have a good user experience with default settings of vscode

### LINKED ISSUE #2657: Improve HTML repr styling to be compliant with the `.help()`

We introduce a better report HTML repr. We can improve it by reusing the theme style introduce by the `.help` already in place.

glemaitre:
closes #2657
closes #2655

This PR improve the style of the estimator report to be aligned with what we already have for the `.help` method.

### DIFF
--- a/skore/src/skore/_utils/repr/estimator_report-content.html.j2
+++ b/skore/src/skore/_utils/repr/estimator_report-content.html.j2
@@ -2,44 +2,47 @@
     <header class="title">EstimatorReport</header>
     <details open class="report-disclosure">
         <summary>
-            <span
-                class="method-tooltip report-disclosure-title"
-                tabindex="0"
-            >Results<span class="tooltip-text">Investigate more with <a
-                    href="{{ metrics_accessor_doc_url }}"
-                    target="_blank"
-                    rel="noopener noreferrer"
-                    class="report-tooltip-accessor-link"
-                >{{ report_class_name }}.metrics</a></span></span>
+            <span class="report-disclosure-title">Results</span>
         </summary>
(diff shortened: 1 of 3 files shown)


jeromedockes:
just a couple of display details:

- the tooltip does not disappear (screenshot below)
- can we not use a monospace font, it takes more space and is less readable

screenshot_2026-03-24T15:37:39+01:00

glemaitre:
> can we not use a monospace font, it takes more space and is less readable

Do you mean in the table or in general. I don't have a strong opinion on this. It is just to know which scope I should change.

jeromedockes:
I would say everywhere but maybe that is out-of-scope for this PR

glemaitre:
> I would say everywhere but maybe that is out-of-scope for this PR

It might be a single CSS variable so we can do here.

glemaitre:
OK so I made the following change:

- fix the font and only use monospace for the tree code structure in the help
- move the tooltip as a note text inside each section
- make that the tooltip in help is light/dark theme compliant

jeromedockes:
thanks @glemaitre !

one thing I noticed is that it does not use all the available width of the output area of jupyter notebooks, which is already quite narrow so maybe we could remove the max-width. but we can do small tweaks like this in other prs

PR #2661, opened 2026-03-23
ci: Install all skore* packages from local branch instead of PyPI in `sphinx.yml`

thomass-dev:


thomass-dev:
> Do we want to add something in the dev-doc about all packages coming from main, or we don't care?

Let's merge for now, i have another iteration on the combinations for pytest.

Issue #2642, opened 2026-03-20
Don't ask for pos_label


GaelVaroquaux:
### What would you like to say?

Using skore-hub:
```
      6 report = evaluate(model, X, y)
----> [7](vscode-notebook-cell:?execution_count=4&line=7) project.put(name, report)
...
    276         ) from exc
    278 payload: EstimatorReportPayload | CrossValidationReportPayload
    280 if isinstance(report, EstimatorReport):

ValueError: For binary classification, the positive label must be specified. You can set it using `report.pos_label = <positive_label>`.
```

I'm just a simple man, I want to get an evaluation done quickly. Don't ask me for pos-label, and in particular don't raise a traceback on me, it breaks my flow :).

On top of this, I have the feeling that if I don't use the hub, I don't need to give "pos_label", so there might be an inconsistency (but maybe it's a version of skore inconsistency)

cakedev0:
I'm working on that 😄 see issue #2592. It's the very last moment to give an opinion there, because I already started to work on it ^^

glemaitre:
@GaelVaroquaux When we have label with -1/1 and 0/1, there is not issue and we don't need `pos_label`.

With string labels, then we are having issue the same way scikit-learn has with the precision-recall. And there is to solution in there:

1. raise an error requesting as early as possible (in `evaluate`) to provide the `pos_label` for the rest of the reporting.
2. not request `pos_label` but it implication. When computing the metrics summary, we can compute the metrics for both labels. It is the same for the ROC / PR, etc. But is also means that, in particular for the curve, you need to introduce a `pos_label` to all the `plot` and other functions to allow removing the unnecessary information. It also have an implication on the hub side because we need to send and compute information for both labels and change the UI to switch from one label to another.

So my question here is, in particular for the string case, do we want to have the friction of setting the `pos_label` early (1.) or add friction points at different points, e.g. API and serialization (2.).

GaelVaroquaux:
I understand the technical reasons, but as a simple user wanting to do a simple task, you are bothering me with problems that I don't have :). I am not trying the do a PR curve, and I may actually not know what it is. I am using the number one entry point of a project (skore), that promises to simplify my life for model evaluation. And suddenly I'm first with difficult choices, while all I wanted was to compute accuracy 😀 

We have several ways forward:
* Don't compute PR is no pos_label is given
* Compute PR for all positive pos_label and have a selector to select between them? (potentially, do that only if the cardinality of y is small)

[...]

Issue #2592, opened 2026-03-13
API: inconsistent behavior of `pos_label` across metrics


cakedev0:
### What would you like to say?

While playing around with `pos_label` in various cases for different metrics, I found some behaviors a bit surprising, as mentioned here https://github.com/probabl-ai/skore/pull/2588#issuecomment-4048720725. Here is a table that tries to give a complete view of the various behaviors. Legend:

- **labels column**: labels present in `y`
- **pos_label column**: `pos_label` passed to the `EstimatorReport`
- **other columns**: metrics. Cells in those columns describe what happened when called for the given `labels` and `pos_label`.

| labels    | pos_label   | precision                       | precision(average="binary")          | roc_auc     | brier_score     | roc*                       |
|:----------|:------------|:--------------------------------|:-------------------------------------|:------------|:----------------|:---------------------------|
| [0,1]     | unset       | all classes                     | error: pos_label unset               | scalar      | scalar          | pos_label=1                |
| [0,1]     | fixed (1)   | pos_label=1                     | pos_label=1                          | scalar      | scalar          | pos_label=1                |
| ["A","B"] | unset       | all classes                     | error: pos_label unset               | scalar      | scalar          | error: pos_label required  |
| ["A","B"] | fixed (B)   | pos_label=B                     | pos_label=B                          | scalar      | scalar          | pos_label=B                |
| [0,1,2]   | unset       | all classes                     | error: multiclass + average='binary' | all classes | attribute error | all classes                |
| [0,1,2]   | fixed (2)   | all classes (warning: pos_label ignored) | error: multiclass + average='binary' | all classes | attribute error | all classes                |

\* behavior is the same for the `confusion_matrix` and `precision_recall` displays.

### What surprises me

- Setting `pos_label` seems to have no visible effect for multiclass, I think.
- `roc` / `precision_recall` / `confusion_matrix` plots silently choose a `pos_label` when it is not set and labels are `{0, 1}` or `{-1, 1}`, but crash otherwise.
- `precision(average="binary")` does not seem to bring any feature compared to simply `precision()`.
- Brier score is not supported for multiclass.

### What I propose

I'm not sure yet, this needs discussion. Some ideas:

- **For multiclass**: raise an error if a `pos_label` is set, or use it in some meaningful way?
- **For binary**:
  - enforce picking a `pos_label` (automatically picking `1` if labels are `{0, 1}` or `{-1, 1}`, and requiring it to be set explicitly otherwise)
[...]

PR #2663, opened 2026-03-25
refactor(skore): Make `pos_label` more consistent

### LINKED ISSUE #2671: Bug: predict time is not stable across calls
The cache slot for the predict time can be overwritten by calling a metric that relies on a `response_method` that was not in cache yet. For instance:

```Python
from skore import evaluate

X, y = make_classification()
report = evaluate(LogisticRegression(), X, y, splitter=0.2)
report.metrics.roc_auc()
print(report.metrics.timings())
# {'fit_time': 0.001524532970506698, 'predict_time_test': 0.0005574490060098469}
report.metrics.accuracy()
print(report.metrics.timings())
# {'fit_time': 0.001524532970506698, 'predict_time_test': 0.000334909011144191}
```

I'd say that this isn't a desired behavior. 

The expected behavior needs decision, we can either:
- only record the time when `response_method="predict"`
- only record the time for the first predictions (`response_method="predict"/"predict_proba"/"decision_function"`)
- have "predict_time_{data_source}" + optionally "predict_proba_time_{data_source}" and/or  "decision_function_time_{data_source}"

I think all options are valid because prediction speed seems to be the same for the different respond methods, for what I could try. But options 1 or 3 feels a bit more rigorous.

### LINKED ISSUE #2672: BUG: Check for decision function shape (OVR vs OVO)
### Describe the bug

When using the `decision_function` of an estimator in the multi-class case, we don't check it's an OVR decision function shape (could be OVO), which can lead us to make bad plots / return bad results.

### Steps/Code to Reproduce

```Python
from sklearn.svm import SVC
from skore import evaluate

report = evaluate(SVC(decision_function_shape="ovo", random_state=0), X, y, splitter=0.2)
report.metrics.roc().plot()
```

### Expected Behavior

Raise an error if shape doesn't match an OVR decision function shape.

### Actual Behavior

It plots incorrect curves/AUCs because it reads the `decision_function` output as if it was OVR:

(this model doesn't underperform random on this simple dataset.)

Image

### Environment

```shell
System:
    python: 3.14.3 (main, Feb 12 2026, 00:42:54) [Clang 21.1.4 ]
executable: /home/arthur/probabl/skore/skore/.venv/bin/python
   machine: Linux-6.17.0-19-generic-x86_64-with-glibc2.39

Python dependencies:
        skore: 0.0.0+unknown
          pip: 26.0.1
       jinja2: 3.1.6
       joblib: 1.5.3
   matplotlib: 3.10.8
        numpy: 2.4.3
       pandas: 2.3.3
       plotly: 6.6.0
         rich: 14.3.3
 scikit-learn: 1.8.0
      seaborn: 0.13.2
 skore[local]: None
        skrub: 0.7.2
skore-hub-project: 0.0.18
```

cakedev0:
Towards #2642 and #2592

#### Change description

Go from behavior table described in #2592 to this:

| labels    | pos_label   | precision/recall                | roc_auc     | brier_score     | roc/PR/conf. matrix                       |
|:----------|:------------|:--------------------------------|:------------|:----------------|:---------------------------|
| [0,1] / ["A","B"] | unset       | all classes                      | scalar      | scalar          | all classes             |
| [0,1] / ["A","B"] | fixed (1/"A")   | pos_label=1/"A"     | scalar      | scalar          | pos_label=1/"A"                |
| [0,1,2]   | unset       | all classes                     |  all classes | attribute error | all classes                |
| [0,1,2]   | fixed (2)   | error at init|error at init | error at init  | error at init               |

This PR also **refactors how predictions are made** to make those changes easier and too prepare the ground for moving `pos_label` from the init to the arguments of metrics/plots. Indeed, now cached predictions don't depend on the `pos_label` anymore (we adapt for `pos_label` on the fly if needed).
By doing so, it fixes #2671 (with the option 1 described in the issue)
And this refactor also fixes #2672

#### Contribution checklist

- [ ] Documentation was added or updated. TODO: check I updated everything that was needed.
- [ ] TODO? A new changelog entry was added to CHANGELOG.rst 

#### AI usage disclosure

For review and tests; not for the core changes, those were a bit too delicate* for AI.

*too delicate = not well-specified enough when I started working on it :sweat_smile:

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -51,6 +51,11 @@ Changed
 - **Breaking change:** :class:`~skore.EstimatorReport` now requires ``X_test``. If
   ``X_test`` is omitted (``None``), construction raises ``ValueError``.
 
+- **Breaking change:** binary classification metrics and curve displays no longer
+  require ``pos_label`` to be set, and skore no longer infers it implicitly. When
+  ``pos_label`` is left unset, ``precision`` and ``recall`` metrics, as well as
+  ROC and precision-recall curves, now expose both classes instead of failing.
+
 Added
 -----
 


cakedev0:
Tests for https://github.com/probabl-ai/skore/pull/2663/commits/8e22d892b510b932ba4508a9f575875e2141475a pass, but when `pos_label=None` (in binary-classif) you get really bad plots :sweat_smile: 

image

cakedev0:
Looking good now (fix: https://github.com/probabl-ai/skore/pull/2663/commits/e130809322eca002110665e96cdca2ea55e8b82f), but I had to rewrite mannnny tests :sweat_smile: (https://github.com/probabl-ai/skore/pull/2663/commits/9ccb49d22b71ba61a2e59ae92f1d1099d886dd32)

image

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -215,6 +235,7 @@ def clear_cache(self) -> None:
     def cache_predictions(
         self,
         response_methods: Literal["auto"] | str | list[str] = "auto",
```
GaetandeCast, line 246: One thing we could do to solve #2671 is to remove the `response_methods` here and cache all prediction methods in `cache_prediction`. This means we predict with either `predict_proba` or `decision_function` when available and deduce the predictions, or compute them with `predict` otherwise. This way we only predict once with the most informative function and can access any type of prediction in the cache later.

GaetandeCast (commented):
Looks great, thanks @cakedev0! One suggestion to fix #2671. Also I did not follow the whole discussion but it looks like we don't infer the `pos_label` anymore?

cakedev0:
> we don't infer the `pos_label` anymore?

Indeed. Inferring it only for the case {0, 1} and {-1, 1} and crashing otherwise is not a great behavior. Instead we choose to not crash and display/return for both labels.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Yes, I 100% love this idea. This is a very valuable optimization I think (basically 2x/3x speed-up for predictions-dominated models).

Do you think we can decide to go with that? Or should we gather more opinions first?

GaetandeCast: I think we can go with it. I don't really see any drawbacks.

cakedev0: Hum... `DummyClassifier(strategy="uniform")` the argmax of predict proba is not predict (predictions are random, predict proba is 1/n_classes everywhere), this make a bunch of tests break.

More realistic examples are: 
- `SVC(probability=True)`: but for this one, we can use the decision function
- `FixedThresholdClassifier` and `TunedThresholdClassifierCV` (those probably break assumptions made in other places too).

I keep running on such small but assumption/abstraction-breaking edge cases those days :sob: 

Anyway, what do we do?
1. Ignore that and change the tests to avoid `DummyClassifier(strategy="uniform")`, but maybe this is on purpose that we have this in the tests.
2. Implement a special treatment for `DummyClassifier` (and maybe `FixedThresholdClassifier`, `TunedThresholdClassifierCV`), and use `decision_function` first when available (this fixes the `SVC(probability=True)` case)
3. Give up on this nice optimization (which also simplifies the code quite a lot honnestly) 

I vote for 2 ^^ but I'd like to have @glemaitre opinion on that.

glemaitre: I think that we can go for the optimization. In short scikit-learn has common test to test this assumption, so most of the classifiers in scikit-learn would benefit for it. However, we cannot provide the same optimization for estimator outside of scikit-learn because we are not aware. In an ideal world, maybe the system of tags could help in this direction but we cannot bet that people implement it.

So in short, we can implement it in another PR where we will have a fallback for `*ThresholdClassifier*` and non-scikit-learn estimator.

cakedev0: Great. I opened a draft PR with the refactor/optim: #2677.

In this PR, I went with option 1 for #2671 (i.e. record only the time for `predict`, and not for other response methods).

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -341,25 +378,84 @@ def get_predictions(
...
-            if not is_cached:
-                self._cache[key] = value
-        return results[0][1]  # return the predictions only
+
+    def _get_predictions(
+        self,
+        *,
+        data_source: Literal["train", "test"],
+        response_method: str | list[str] | tuple[str, ...],
+        pos_label: PositiveLabel | None = None,
```
GaetandeCast, line 408: There should still be some amount of validation on the `pos_label`. Passing a wrong `pos_label` value currently crashes in an ugly way. For instance this snippet
```python
X, y = make_classification(random_state=0)
split_data = train_test_split(X, y, test_size=0.2, random_state=0, as_dict=True)
report = EstimatorReport(LogisticRegression(), **split_data, pos_label="a")
display = report.metrics.roc(data_source="both")
```
throws `IndexError: index 0 is out of bounds for axis 0 with size 0` when I would expect a helpful error message telling me that this label does not exist in `y` and what the available ones are.

Open for debate: passing `"1"` should be interpreted as int `1` when it is an available label. 

I don't know where that validation should occur though. Probably in the init of the report?

GaetandeCast: > passing `"1"` should be interpreted as int `1`

Motivation: currently, passing `pos_label=True` or `pos_label=1.0` has the same effect as `pos_label=1`.

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -341,25 +378,84 @@ def get_predictions(
...
+            The predictions.
+            The shape is (n_samples,) if:
+                - response_method is "predict"
+                - OR if pos_label is specified (binary-classification only)
+            Otherwise it's (n_samples, n_classes)
+        """
+        if data_source not in ("train", "test"):
+            raise ValueError(f"Invalid data source: {data_source}")
+        if pos_label is not None and self.ml_task != "binary-classification":
+            raise ValueError(f"Cannot specify a `pos_label` for task {self.ml_task}")
```
GaetandeCast, line 426: Isn't this redundant with the error in the report init?

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -417,34 +413,31 @@ def _compute_data_for_display(
...
             )
         data_source = cast(DataSource, data_source)
-        # When provided, the positive label is set in second position.
-        if ml_task == "binary-classification" and pos_label_validated is not None:
-            neg_label = next(
-                label for label in display_labels if label != pos_label_validated
-            )
-            display_labels = [str(neg_label), str(pos_label_validated)]
+
+        # FIXME? Usually, TP is the top-left cell
```
GaetandeCast: I prefer TP in the top-left too but I think we decided with @glemaitre to align ourselves with scikit-learn. For this, I'd rather we do what everybody else does.

skore/tests/unit/displays/precision_recall_curve/test_common.py
```diff
@@ -99,16 +99,15 @@ def test_relplot_kwargs(self, pyplot, fixture_prefix, task, request):
...
+        palette = ["red", "green", "blue"] if task == "multiclass" else ["red", "green"]
+        relplot_kwargs = {"palette": palette}
 
         display.set_style(relplot_kwargs=relplot_kwargs)
         fig = display.plot()
         ax = fig.axes[0]
-        assert ax.get_lines()[0].get_color() == "red"
+        actual_colors = {line.get_color() for line in ax.get_lines()}
+        expected_colors = set(palette[: 3 if task == "multiclass" else 2])
+        assert actual_colors == expected_colors
```
GaetandeCast:
```suggestion
        actual_colors = {line.get_color() for line in ax.get_lines()}
        assert actual_colors == set(palette)
```
This should be enough. Nice improvement nevertheless :wink:

skore/tests/unit/displays/roc_curve/test_common.py
```diff
@@ -96,16 +96,15 @@ def test_relplot_kwargs(self, pyplot, fixture_prefix, task, request):
...
+        palette = ["red", "green", "blue"] if task == "multiclass" else ["red", "green"]
+        relplot_kwargs = {"palette": palette}
 
         display.set_style(relplot_kwargs=relplot_kwargs)
         fig = display.plot()
         ax = fig.axes[0]
-        assert ax.get_lines()[0].get_color() == "red"
+        actual_colors = {line.get_color() for line in ax.get_lines()[:-1]}
+        expected_colors = set(palette[: 3 if task == "multiclass" else 2])
+        assert actual_colors == expected_colors
```
GaetandeCast: same

skore/tests/unit/reports/comparison/cross_validation/metrics/test_numeric.py
```diff
@@ -246,8 +246,9 @@ def test_binary_classification_pos_label(pyplot, metric):
...
     report_2 = CrossValidationReport(LogisticRegression(C=2), X, y)
     report = ComparisonReport([report_1, report_2])
-    with pytest.raises(ValueError, match="pos_label is not specified"):
-        getattr(report.metrics, metric)()
+    display = getattr(report.metrics, metric)()
+    fig = display.plot()
+    assert "Positive label" not in fig.get_suptitle()
 
     report_1 = CrossValidationReport(LogisticRegression(C=1), X, y, pos_label="A")
     report_2 = CrossValidationReport(LogisticRegression(C=2), X, y, pos_label="A")
```
GaetandeCast, line 254: I can't comment up there but the docstring of the test should be updated
when `pos_label` needs to be set.""" -> when `pos_label` is not set."""

skore/tests/unit/reports/comparison/estimator/metrics/test_numeric.py
```diff
@@ -193,9 +193,7 @@ def test_display_binary_classification_pos_label(
         y_train=y_train,
         y_test=y_test,
     )
-    report = ComparisonReport([report_1, report_2])
-    with pytest.raises(ValueError, match="pos_label is not specified"):
-        getattr(report.metrics, metric)()
+    # TODO: test when pos_label is not set
```
GaetandeCast: Yes please do :wink:

skore/tests/unit/reports/estimator/metrics/test_plot.py
```diff
@@ -37,15 +38,48 @@ def test_display_binary_classification_pos_label(pyplot, metric):
...
+    pyplot, metric
+):
+    """Check that the default binary behaviour works with 1D decision scores."""
+    X, y = make_classification(
+        n_classes=2,
+        n_features=5,
+        n_informative=3,
+        n_redundant=0,
+        random_state=0,
+    )
```
GaetandeCast:
```suggestion
    X, y = binary_classification_data

File: skore/tests/unit/reports/estimator/metrics/test_plot.py [hunk 10]
```diff
@@ -89,3 +123,19 @@ def test_display_regression_switching_data_source(
     assert report._cache != {}
     display_second_call = getattr(report.metrics, display)(data_source="train")
     assert display_first_call is not display_second_call
+
+
+def test_display_confusion_matrix_data_source_both_is_not_supported(
```
GaetandeCast: I'd rather have this test in `.../confusion_matrix/test_common.py`, outside of the parametrize block since we don't have to test it for all the kinds of report/task.
Good catch seeing it was missing though!

cakedev0:
> the tests helped you in the dev process for this change or was it pure tedium to deal with them?

The tests didn't help: I had them green without changing them and without too much efforts, but I was actually not displaying the good thing at all. Then I fixed the displays (+8/-8 change), and this made dozen of tests go red.

This is not super surprising though: I'm changing the plot behavior quite a lot...  Maybe tests on a higher level of assumptions would have been better for this case. But at the same time, plots are a lot about details.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
cakedev0: I won't change that in this PR, but I documented it.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: I'll do the same check as it was before: `pos_label in classes.tolist()` (this was done in sklearn's `_get_response_values`).

But I'll do it at init time.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: For the users yes, but for us no so I'd rather keep it (and it will become user facing soon, when pos_label is moved away from the init).

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: > passing pos_label=True or pos_label=1.0 has the same effect as pos_label=1.

This will be the case, but `"1" in [0, 1]` is false, so this will crash. We can add a `validate_and_sanitize_pos_label` in another PR.

skore/tests/unit/displays/precision_recall_curve/test_common.py
cakedev0: Thx!

skore/tests/unit/reports/comparison/cross_validation/metrics/test_numeric.py
cakedev0: Good catch, thx!

skore/tests/unit/reports/estimator/metrics/test_plot.py
cakedev0: Moved it.

PR #2664, opened 2026-03-25
feat(skore-hub-project): Change the way the report cache is inspected for permutation importances

thomass-dev:
Hotfix for weekly release.

### DIFF
--- a/skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
+++ b/skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
@@ -112,14 +112,14 @@ class PermutationImportance(Inspection[Report], ABC):  # noqa: D101
                 if len(key) != 3:
                     continue
 
-                data_source, name, kwargs = key
-                kwargs = str(kwargs)
+                key_data_source, key_name, key_kwargs = key
+                key_kwargs = str(key_kwargs)
 
                 if (
-                    data_source == data_source
-                    and name == "permutation_importance"
-                    and "('at_step', 0)" in kwargs
-                    and "('metric', None)" in kwargs
+                    key_data_source == data_source
+                    and key_name == "permutation_importance"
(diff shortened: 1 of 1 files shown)


cakedev0:
Hum, don't you want to wait for `skore` to expose a clean way of doing that before fixing it? (which is one of the main goal of this discussion https://github.com/probabl-ai/skore/discussions/2587) 

Because, we might have to rewrite a similar fix in one week or so (without accessing `_cache` this time hopefully).

thomass-dev:
> Hum, don't you want to wait for `skore` to expose a clean way of doing that before fixing it? (which is one of the main goal of this discussion #2587)
> 
> Because, we might have to rewrite a similar fix in one week or so (without accessing `_cache` this time hopefully).

I don't think it will be ready before the next release (it implies many changes in many places).
So IMO, we have to implement a temporary/dirty fix in the meantime.

skore-hub-project/src/skore_hub_project/artifact/media/inspection.py
```diff
@@ -43,31 +47,101 @@ class PermutationImportance(Inspection[Report], ABC):  # noqa: D101
     accessor: ClassVar[str] = "inspection.permutation_importance"
     name: Literal["permutation_importance"] = "permutation_importance"
 
-    def content_to_upload(self) -> bytes | None:  # noqa: D102
-        cache = getattr(self.report, "_cache", {})
+    if Version("0.0.1") < Version(metadata.version("skore")) <= Version("0.14"):
```
thomass-dev, line 50: Only for backward compatibility.
This block has to be revert if this PR is merged <ins>**after**</ins> the release of `skore==0.15`.

glemaitre:
> Hum, don't you want to wait for skore to expose a clean way of doing that before fixing it?

I would say that for being able to still release, we can have this dirty way to not have regression. At the moment, that we are going to go forward we get revert this code.

PR #2673, opened 2026-03-25
chore(skore)!: Impose to pass `X_test` to EstimatorReport

### LINKED ISSUE #2627: make test data mandatory?

ATM a report can be created without X_test and y_test. I get that training data is optional when we pass a fitted estimator, but maybe we should force a test set to be provided?

glemaitre:
closes #2627

Enforce to have some testing data when creating an `EstimatorReport`.

### DIFF
--- a/skore/tests/unit/reports/comparison/estimator/test_report.py
+++ b/skore/tests/unit/reports/comparison/estimator/test_report.py
@@ -141,4 +141,9 @@ def test_clustering():
         match="Clustering models are not supported yet. Please use a "
         "classification or regression model instead.",
     ):
-        ComparisonReport([EstimatorReport(KMeans()), EstimatorReport(KMeans())])
+        ComparisonReport(
+            [
+                EstimatorReport(KMeans(), X_test=None),
+                EstimatorReport(KMeans(), X_test=None),
+            ]
+        )
--- a/skore/tests/unit/reports/estimator/test_report.py
+++ b/skore/tests/unit/reports/estimator/test_report.py
@@ -39,7 +39,7 @@ def test_estimator_not_fitted(fit):
     estimator = LinearRegression()
     err_msg = "The training data is required to fit the estimator. "
     with pytest.raises(ValueError, match=err_msg):
-        EstimatorReport(estimator, fit=fit)
+        EstimatorReport(estimator, fit=fit, X_test=None)
 
 
 @pytest.mark.parametrize("fit", [True, "auto"])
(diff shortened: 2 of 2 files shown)


skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -390,7 +390,7 @@ def create_estimator_report(
         self,
         *,
         report_key: str,
-        X_test: ArrayLike | None = None,
+        X_test: ArrayLike,
         y_test: ArrayLike | None = None,
```
GaetandeCast, line 394: Since we dropped support for unsupervised learning, we should also expect `y_test` no?

glemaitre: I would do in another PR then because we would need to clean up at several places. I think that we still detect clustering in some way but we don't have nice support for it.

glemaitre:
I broke the `skore-mlflow-project` I think :)

skore/src/skore/_sklearn/_estimator/report.py
thomass-dev: Why not leaving this parameter as positional, without default value?

thomass-dev: It's counterintuitive that the parameter is mandatory but optional in the signature.

glemaitre: Since we are keyword only, then we can do that indeed.

(still used to the all time where we don't use `*` and you could not have "positional" argument after one specifying a default).

skore/src/skore/_sklearn/_comparison/report.py
glemaitre: I'll open an issue now and affect myself to this cleanup.

skore-mlflow-project/src/skore_mlflow_project/reports.py
```diff
@@ -199,8 +199,7 @@ def iter_cv(report: CrossValidationReport) -> Generator[NestedLogItem, None, Non
     """Yield loggable objects for a cross-validation report."""
     yield from iter_cv_metrics(report)
 
-    estimator_report = report.create_estimator_report()
-    estimator = estimator_report.estimator_
+    estimator = clone(report.estimator).fit(report.X, report.y)
```
glemaitre, line 202: @thomass-dev this change is backward compatible but we touch the `skore-mlflow-project`.

PR #2677, opened 2026-03-27
perf(skore): Compute predictions from `decision_function`/`predict_proba` when possible

cakedev0:
Context: this follows the discussion in https://github.com/probabl-ai/skore/pull/2663#discussion_r2993909658.

#### Change description

This PR optimizes `cache_predictions` by avoiding an extra `predict` call when class labels can be derived from cached `decision_function` or `predict_proba` outputs. This is a big optimization for predict-time dominated estimators, but this is also a **risky** optimization as it can lead to incorrect results see 
https://github.com/probabl-ai/skore/pull/2677#discussion_r3022131393

Changes:
- compute and cache `predict` from `decision_function` / `predict_proba` when possible
- keep a fallback to `predict` for estimators where deriving labels is not reliable (see discussion )
- simplify `cache_predictions` by removing the `response_methods` and `n_jobs` parameters (not relevant anymore): I made a change log entry for this **breaking change**.
- update examples and tests to reflect the new caching behavior
- cache `predict_log_proba` alongside `predict_proba`: this was needed to still have it available as a `respond_method` (for summarize for instance). **Alternatively**, we could choose to stop accepting `predict_log_proba` as a response method: I'm open to this solution too (would simplify the code).

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/report.py
+++ b/skore/src/skore/_sklearn/_estimator/report.py
@@ -376,19 +376,20 @@ class EstimatorReport(_BaseReport, DirNamesMixin):
         response : ndarray of shape (n_samples, n_classes)
             For binary decision_function, the returned array is reshaped to
             (n_samples, 2) so it can be aligned with classes_.
-        predictions : ndarray of shape (n_samples,)
+        predictions : ndarray of shape (n_samples,) or None
             Predicted labels derived from response
+            or None for ill-shaped decision function (OVO)
         pred_time : float
             Time spent computing ``response_method(data)`` in seconds.
         """
         with MeasureTime() as pred_time:
             response = getattr(self._estimator, response_method)(data)
-        if (
-            response_method == "decision_function"
-            and self.ml_task == "binary-classification"
-        ):
-            response = np.vstack((-response, response)).T
         classes = to_estimator(self._estimator).classes_
+        if response_method == "decision_function":
+            if self.ml_task == "binary-classification":
+                response = np.vstack((-response, response)).T
+            if response.shape[1] != len(classes):
+                return response, None, pred_time()
         predictions = classes[np.argmax(response, axis=1)]
         return response, predictions, pred_time()
 


skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -282,60 +263,81 @@ def cache_predictions(
...
+        preds_only = not (has_proba or has_decision)
+
+        # if this is True, we call .predict(...)
+        # otherwise we derive predictions from predict_proba/decision_function
+        # (which is a big optimization for models with an expansive predict function).
+        compute_preds = (
+            # FixedThresholdClassifier or TunedThresholdClassifierCV:
+            "ThresholdClassifier" in self._estimator.__class__.__name__
+            or "Dummy" in self._estimator.__class__.__name__
         )
```
cakedev0: I'm realizing that this kind of tests is not robust to pipelines / bagging / `GridSearchCV` / ... This is good to pass tests, but probably not super robust for real-life situations... 

Maybe we need to do a white-list check rather, i.e.: 

```
deduce_preds = (
    any(isinstance(estimator, cls) for cls in MODELS_WHITELIST)
    or is_pipeline_f_whitelist_models(estimator, MODELS_WHITELIST)
    or (is_bagging and internal_estimator_is_whitelisted)
    ...
)
```

This is more painful though...

GaetandeCast: As mentioned by @glemaitre [here](https://github.com/probabl-ai/skore/pull/2663#discussion_r2996584892), we should also not do it on non scikit-learn estimators: custom estimator may not deduce predictions from predicted probabilities.

GaetandeCast: Indeed it feels a bit painful to inspect the whole pipeline/gridsearch/bag of estimators etc. so maybe for now we limit the optimization to models from scikit-learn that are not meta estimators. I don't know if there is a straightforward way of checking this, maybe checking if it inherits from `MetaEstimatorMixin` covers all annoying corner cases?

GaetandeCast: so to me we could apply this optim to models that satisfy the following rule: is a sklearn estimator & is not `MetaEstimatorMixin` & is not `DummyClassifier` & is not `Pipeline`
WDYT @cakedev0 ?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -282,60 +263,81 @@ def cache_predictions(
...
 
-        # trigger the computation
-        # do not mutate directly `self._cache` during the execution of Parallel
-        keys, to_compute = zip(*keys_and_to_compute, strict=True)
-
-        for (ds, response_method), (preds, time) in zip(
-            keys,
-            track(
-                parallel(delayed(_compute_predictions)(*args) for args in to_compute),
```
GaetandeCast, line 379: `_compute_predictions` is dead code now

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -282,60 +263,81 @@ def cache_predictions(
...
-        parallel = Parallel(
-            **_validate_joblib_parallel_params(n_jobs=n_jobs, return_as="generator")
+        has_proba = hasattr(self._estimator, "predict_proba")
+        has_decision = hasattr(self._estimator, "decision_function")
+        preds_only = not (has_proba or has_decision)
+
+        # if this is True, we call .predict(...)
+        # otherwise we derive predictions from predict_proba/decision_function
+        # (which is a big optimization for models with an expansive predict function).
+        compute_preds = (
```
GaetandeCast:
```suggestion
        # This is for special cases where `predict` cannot be infered from `predict_proba` or `decision_function`
        do_predict = (
```
To (hopefully) add a bit of clarity

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -282,60 +263,81 @@ def cache_predictions(
...
+        time_key = make_cache_key(data_source, "predict_time")
+
+        if pred_key in self._cache:
             return
 
-        parallel = Parallel(
-            **_validate_joblib_parallel_params(n_jobs=n_jobs, return_as="generator")
+        has_proba = hasattr(self._estimator, "predict_proba")
+        has_decision = hasattr(self._estimator, "decision_function")
+        preds_only = not (has_proba or has_decision)
```
GaetandeCast:
```suggestion
        has_predict_only = not (has_proba or has_decision)
```
Nit for clarity

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -282,60 +263,81 @@ def cache_predictions(
...
+            elif not compute_preds:
+                self._cache[time_key] = pred_time()
+                self._cache[pred_key] = classes[np.argmax(decision_func, axis=1)]
+
+        if has_proba:
+            with MeasureTime() as pred_time:
+                predicted_proba = self._estimator.predict_proba(X)
+            proba_key = make_cache_key(data_source, "predict_proba")
+            self._cache[proba_key] = predicted_proba
+            log_key = make_cache_key(data_source, "predict_log_proba")
```
GaetandeCast, line 361: Do we ever use use predict_log_proba in skore? If not this may not be useful.

GaetandeCast (commented):
Some minor remarks, looks good otherwise! Let's merge once we converged on when to perform the optim.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Hum... I think this would make it too restrictive for the optimization to be worth it.

I have another idea though! I propose to **probe**! Something like:
```python
X_sampled = sample(X_test, min_fraction=0.02, min_samples=100)
deduce_preds = (estimator.predict(X) == np.argmax(estimator.predict_proba(X), axis=1)).all()
```

I think it's safe enough this way.

GaetandeCast: Okay I like the idea! Let's do it

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -282,60 +263,81 @@ def cache_predictions(
...
+            return
 
-        keys_and_to_compute = [
-            (key, data)
-            for key, data in keys_and_to_compute
-            if make_cache_key(*key) not in self._cache
-        ]
+        X = self._X_test if data_source == "test" else self._X_train
+        if X is None:
+            raise ValueError(f"Missing X_{data_source}")
```
auguste-probabl: Can you add a TODO to use `_get_X_y` here once it's been extracted from `_BaseAccessor`?

auguste-probabl: see https://github.com/probabl-ai/skore/pull/2713, which has now been merged

cakedev0: I don't think we want to use `_get_X_y`: we don't need y and we want to be able to give predictions to users even if y is missing.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Maybe not, but now that I rethink about it, I'd say it's cheap enough to keep it (100ms for 10M samples - 2 classes).

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: I refactored a bit when I introduced the probing, not a variable anymore :sweat_smile:

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Refactored, but added this comment, thanks.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Removed it.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Did that, which lead me to refactor a bit what I dit. I think it's pretty readable now :crossed_fingers: 

I also added a simple test with `FixedThresholdClassifier` etc. to test predictions correctness.

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -287,60 +274,100 @@ def cache_predictions(
...
+        X = self._X_train if self._X_test is None else self._X_test
+        X_sample = _sample(X, min_fraction=0.01, min_samples=100)
+        predictions = self._estimator.predict(X_sample)
+        _, deduced_predictions, _ = self._get_response_and_derived_predictions(
+            X_sample,
+            response_method=method.__name__,
+        )
+        if deduced_predictions is None:
+            return False
+        return np.array_equal(predictions, deduced_predictions)
```
GaetandeCast, line 437:
```suggestion
        return np.array_equal(predictions, deduced_predictions)
```
`np.array_equal` should be able to accept `None`

skore/tests/unit/reports/estimator/test_report.py
```diff
@@ -183,15 +182,37 @@ def test_cache_predictions(
...
+        ),
+    ],
+)
+def test_get_predictions_is_correct_for_special_classifiers(estimator):
+    X, y = make_classification(n_samples=100, random_state=42)
+    report = evaluate(estimator, X, y, splitter=0.2)
+    np.testing.assert_array_equal(
+        report.get_predictions(data_source="test"),
+        report.estimator_.predict(report.X_test),
+    )
```
GaetandeCast, line 213:
```suggestion
    report = evaluate(estimator, X, y, splitter=0.2)
    np.testing.assert_array_equal(
        report.get_predictions(data_source="test"),
        report.estimator_.predict(report.X_test),
    )
    assert not report._can_deduce_predictions
```

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -47,12 +41,17 @@
...
+    sample_size = min(
+        n_samples, max(min_samples, int(np.ceil(min_fraction * n_samples)))
+    )
+    if sample_size == n_samples:
+        return X
 
-    return predictions, predict_time()
+    rng = np.random.default_rng(0)
+    indices = np.sort(rng.choice(n_samples, size=sample_size, replace=False))
+    return _safe_indexing(X, indices, axis=0)
```
auguste-probabl: is the `min_fraction` argument necessary?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -287,60 +274,100 @@ def cache_predictions(
...
+                response = np.vstack((-response, response)).T
+            if response.shape[1] != len(classes):
+                return response, None, pred_time()
+        predictions = classes[np.argmax(response, axis=1)]
+        return response, predictions, pred_time()
+
+    @cached_property
+    def _can_deduce_predictions(self):
+        # Check if `predict` can be inferred reliably from response_method
+        # (predict_proba or decision_function)
```
auguste-probabl:
```suggestion
    def _can_deduce_predictions(self) -> bool:
        """
        Whether `predict` can be inferred reliably from
        `predict_proba` or `decision_function`.
        """
```

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -473,44 +500,27 @@ def _get_predictions(
...
-            # predict_proba
-            return np.vstack((-predictions, predictions)).T
-        else:
-            raise ValueError(f"Unexpected response_method: {method_name}")
+        # check shape if needed:
+        is_multiclass = self.ml_task == "multiclass-classification"
+        if is_multiclass and method_name == "decision_function":
+            _, d = predictions.shape
+            if d != len(self.estimator_.classes_):
+                raise ValueError(f"Unexpected decision function shape[1]: {d}")
```
auguste-probabl: Should this check be in `cache_predictions`?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -473,44 +500,27 @@ def _get_predictions(
...
-                return predictions
-            # scikit-learn returns a (n_samples,) array that corresponds to classes[-1]
-            # we normalize to a (n_samples, 2) shape with similar semantic than
-            # predict_proba
-            return np.vstack((-predictions, predictions)).T
-        else:
-            raise ValueError(f"Unexpected response_method: {method_name}")
+        # check shape if needed:
+        is_multiclass = self.ml_task == "multiclass-classification"
+        if is_multiclass and method_name == "decision_function":
```
auguste-probabl:
```suggestion
        if self.ml_task == "multiclass-classification" and method_name == "decision_function":
```

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl:
```suggestion
            if predictions.shape[1] != len(self.estimator_.classes_):
                raise ValueError(
					"Decision function output should have as many columns as there are classes; "
					f"expected {len(self.estimator_.classes_)} but got {predictions.shape[1]}."
				)
```

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: I like to be:
- safe enough with 100 samples
- the safest possible as long as the cost of this extra safety is negligible (1%)

So I like to have both.

cakedev0:
@auguste-probabl This PR should not conflict with #2701 but at the same time, I'm really not in a hurry to merge it, so any order SGTM.

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: I don't quite understand what you mean. 100 samples is very few, so it seems that it should suffice? do you mean that if the dataset is very large, 100 samples might not be enough to tell whether `argmax(predict_proba()) == predict()`?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -287,60 +274,98 @@ def cache_predictions(
...
+        if response_method == "decision_function":
+            if self.ml_task == "binary-classification":
+                response = np.vstack((-response, response)).T
+            if response.shape[1] != len(classes):
+                return response, None, pred_time()
+        predictions = classes[np.argmax(response, axis=1)]
+        return response, predictions, pred_time()
+
+    @cached_property
+    def _can_deduce_predictions(self):
```
auguste-probabl:
```suggestion
    def _can_skip_predict(self):
```

auguste-probabl: I feel like that would carry the intent better. I think the docstring should also expand on what the function does and why, as it was hard for me to grasp.

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -47,12 +41,17 @@
...
+    n_samples = _num_samples(X)
+    sample_size = min(
+        n_samples, max(min_samples, int(np.ceil(min_fraction * n_samples)))
+    )
+    if sample_size == n_samples:
+        return X
 
-    return predictions, predict_time()
+    rng = np.random.default_rng(0)
+    indices = np.sort(rng.choice(n_samples, size=sample_size, replace=False))
```
auguste-probabl: why the sorting?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -287,60 +274,98 @@ def cache_predictions(
...
+        classes = self._estimator.classes_
+        if response_method == "decision_function":
+            if self.ml_task == "binary-classification":
+                response = np.vstack((-response, response)).T
+            if response.shape[1] != len(classes):
+                return response, None, pred_time()
+        predictions = classes[np.argmax(response, axis=1)]
+        return response, predictions, pred_time()
+
+    @cached_property
```
auguste-probabl, line 396: I'd rather make this a method rather than a property, since it does non-trivial computation. Caching is fine though.

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -47,12 +41,17 @@
     from skore._sklearn._estimator.metrics_accessor import _MetricsAccessor
 
 
-def _compute_predictions(estimator, X, response_method):
-    prediction_method = _check_response_method(estimator, response_method)
-    with MeasureTime() as predict_time:
-        predictions = prediction_method(X)
+def _sample(X, *, min_fraction=0.0, min_samples=1):
```
auguste-probabl: I would be inclined to inline this function since it's used only once

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -287,60 +274,98 @@ def cache_predictions(
...
-                (self._estimator, X, response_method),
+        if data_source == "both":
+            self.cache_predictions(data_source="test")
+            if self._X_train is not None:
+                self.cache_predictions(data_source="train")
+            return
+
+        X = self._X_test if data_source == "test" else self._X_train
+        if X is None:
+            raise ValueError(
```
auguste-probabl, line 316: Could be nice to factor out the error message from _get_X_y, I'll let you decide

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -287,60 +274,98 @@ def cache_predictions(
...
+            # Most sklearn's estimator derive predict_log_proba this way
+            # except for *NB models (naive bayes) that derive predict_proba
+            # from predict_log_proba using exp:
+            with np.errstate(divide="ignore"):
+                self._cache[log_key] = np.log(response)
+            if self._can_deduce_predictions:
+                self._cache[time_key] = pred_time
+                self._cache[pred_key] = predictions
+
+    def _get_response_and_derived_predictions(self, X, response_method):
```
auguste-probabl: Add some docs, at least Returns

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: I don't think so because we don't want to raise an error until the user actually needs the `decision_function` (not needed if you just compute the accuracy for instance).

cakedev0: Nice, thx!

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Oops, I didn't noticed that

```suggestion
    indices = rng.choice(n_samples, size=sample_size, replace=False)
```

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: The message is shorter after the last PR got merged on main, so I won't do that.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: I'd rather not, it's easier to read this way IMO

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Shorter and clearer, I like that :+1:

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: I mean that if the dataset if large, it doesn't cost much to probe over 1k/10k samples, so why not. But I'm ok with settling for 100 samples if you think it's too much complexity for too little gain.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: "Python: pattern for caching a method with no argument?" to ChatGPT => use `@cached_property`... I'd rather keep this pattern, it's much lighter than alternatives.

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: it sounds like YAGNI

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: fair enough, please reconsider if you end up simplifying according to https://github.com/probabl-ai/skore/pull/2677#discussion_r3032135118, otherwise maybe add a docstring

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Not really, it might be over-carefulness though. I'll remove it.

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Done

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: inlined it.

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: ok, works for me

auguste-probabl (dismissed):
Great work!

PR #2678, opened 2026-03-27
chore(skore): Use xgboost instead of xgboost-cpu on macOS

ogrisel:
`xgboost-cpu` is not available on non-Linux/Windows platform and hence breaks install from source on macOS unless I missed something.


glemaitre:
Thanks Olivier. I'll take care to push a commit to update the lockfile.

glemaitre:
I could not push in your branch directly. I open the following PR: https://github.com/ogrisel/skore/pull/1

glemaitre:
Thanks @ogrisel

PR #2679, opened 2026-03-27
chore: Make update compliant for bash<4

glemaitre:
The current file uses `mapfiles` that is only available for bash 4 while MacOS has only bash 3.2

We can then avoid to use this function.

### DIFF
--- a/ci/pip-compile.sh
+++ b/ci/pip-compile.sh
@@ -46,8 +46,7 @@ case $1 in
         esac
 
         for PACKAGE in "${PACKAGES[@]}"; do
-            while IFS= read -r combination || [[ -n "${combination}" ]]; do
-                [[ -z "${combination}" ]] && continue
+            while IFS= read -r combination; do
                 python=$(jq -rc '.python' <<< "${combination}")
                 dependencies=$(jq -rc '.dependencies' <<< "${combination}")
 
@@ -83,9 +82,8 @@ set -eu
     # Move to `TMPDIR` to avoid absolute paths in requirements file
     cd "${TMPDIR}"
 
-    i=0
-    while [ "${i}" -lt "${#COMBINATIONS[@]}" ]; do
-        combination="${COMBINATIONS[i]}"
+    for combination in "${COMBINATIONS[@]}"
+    do
         IFS="|" read -r PACKAGE EXTRA PYTHON DEPENDENCIES <<< "${combination}"
 
         # Escape dependencies to be used in filename, both in Linux and Windows:
@@ -128,6 +126,5 @@ set -eu
            "${@:1}"
 
         let counter++
-        i=$((i + 1))
     done
 )


ci/pip-compile.sh
```diff
@@ -46,19 +46,18 @@ case $1 in
         esac
 
         for PACKAGE in "${PACKAGES[@]}"; do
-            mapfile -t combinations < <(
-                jq 'unique_by([.python, .dependencies]) | .[]' "${CWD}/../${PACKAGE}/supported-versions.json" -c
-            )
-
-            for combination in "${combinations[@]}"; do
+            # Read jq lines without `mapfile` (requires bash 4+; macOS ships bash 3.2).
```
glemaitre:
```suggestion
```

ci/pip-compile.sh
```diff
@@ -85,8 +84,10 @@ set -eu
     # Move to `TMPDIR` to avoid absolute paths in requirements file
     cd "${TMPDIR}"
 
-    for combination in "${COMBINATIONS[@]}"
-    do
+    # Index loop: `for ... in "${COMBINATIONS[@]}"` with `set -u` can error on empty arrays in bash 3.2.
```
glemaitre:
```suggestion
```

ci/pip-compile.sh
```diff
@@ -46,19 +46,17 @@ case $1 in
         esac
 
         for PACKAGE in "${PACKAGES[@]}"; do
-            mapfile -t combinations < <(
-                jq 'unique_by([.python, .dependencies]) | .[]' "${CWD}/../${PACKAGE}/supported-versions.json" -c
-            )
-
-            for combination in "${combinations[@]}"; do
+            while IFS= read -r combination || [[ -n "${combination}" ]]; do
+                [[ -z "${combination}" ]] && continue
```
thomass-dev:
```suggestion
```

LLM safeguard.

ci/pip-compile.sh
```diff
@@ -129,5 +128,6 @@ set -eu
            "${@:1}"
 
         let counter++
+        i=$((i + 1))
     done
```
thomass-dev, line 129: Please revert the block, it is not impacted by the previous change and has no additional value.

ci/pip-compile.sh
```diff
@@ -46,19 +46,17 @@ case $1 in
         esac
 
         for PACKAGE in "${PACKAGES[@]}"; do
-            mapfile -t combinations < <(
-                jq 'unique_by([.python, .dependencies]) | .[]' "${CWD}/../${PACKAGE}/supported-versions.json" -c
-            )
-
-            for combination in "${combinations[@]}"; do
+            while IFS= read -r combination || [[ -n "${combination}" ]]; do
```
thomass-dev:
```suggestion
            while IFS= read -r combination; do
```

PR #2692, opened 2026-03-30
refactor: Move `without_predict_proba` tests from `displays/` to `reports/`

auguste-probabl:
Follow up on #2684 

Tests in `unit/displays/metrics_summary/` are for testing `MetricSummaryDisplay`'s ability to display its data, e.g. with `flat_index` and `favorability`.

### DIFF
--- a/skore/tests/conftest.py
+++ b/skore/tests/conftest.py
@@ -564,7 +564,7 @@ def linear_regression_comparison_report(linear_regression_with_train_test):
     return report
 
 
-class CustomClassifierWithoutPredictProba(ClassifierMixin, BaseEstimator):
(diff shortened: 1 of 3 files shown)


skore/tests/conftest.py
```diff
@@ -558,3 +562,37 @@ def linear_regression_comparison_report(linear_regression_with_train_test):
         }
     )
     return report
+
+
+class CustomClassifierWithoutPredictProba(ClassifierMixin, BaseEstimator):
```
cakedev0: `CustomClassifierPredictOnly` would be a clearer name, because this clf has neither predict_proba nor decision_function (that's why it doesn't have a roc auc). 

`WithoutPredictProba` sound like you could have used `SVC()` instead for instance.

PR #2694, opened 2026-03-31
perf(skore-mlflow-project): Skip data.analyze() for sub-reports

cakedev0:
Skip `data.analyze()` for sub-reports of CV reports, it's redundant and can take 50% of the put time in some cases, maybe even more for datasets with many columns (and not too many samples).


cakedev0:
@thomass-dev, the hub is different so we can't apply the same optimization as easily, but you might be interested to know how expensive `report.data.analyze()` even for small `n_samples`, if `n_features` is big. On my computer: 0.04s x n_features x (n_splits + 1)

glemaitre:
I'm not sure that we should reduce the time by cutting feature. Here, a `TableReport` per split is a way to understand why your cross-validation would not work between shift. If you don't drop me the `TableReport` then I don't really have the information to inspect back when using MLflow.

cakedev0:
For now, I'm just doing `report.data.analyze()` (=> `data_source="both"` by default) for each sub report, which is not informative (100% the same as the CV report).

But maybe the good answer is to do the train & test analyses instead?

glemaitre:
> (=> data_source="both" by default) 

Ah I see, on the hub we iterate on `data_source="train"` and `data_source="test"` and we have the toggle train/test to check the both dataset in an independent manner.

cakedev0:
Should we merge this?

Or should I close this PR and instead align the behavior with the hub (iterating on `data_source="train"` and `data_source="test"` for estimator reports)?

@glemaitre wdyt? I don't have a preference personnally.

cakedev0:
Closing this one: I will wait for the "export results" feature in skore, and use that.

PR #2700, opened 2026-04-01
fix: Use unique local report IDs 

Ashish-Kumar-Dash:
#### Change description

This PR fixes local project report identity in persisted summaries. Previously, when the same report was saved multiple times in local mode, the summary exposed the shared `artifact_id` as the public report `id`. As a result, repeated identical saves were not uniquely addressable, which made report history harder to reason about and made `get(id)` less useful than expected.

```python
project.put("model", report)
project.put("model", report)

summary = project.summarize()
summary[0]["id"] == summary[1]["id"]  # previously True
```
That made report history harder to reason about, because multiple saved entries looked like the same persisted object from the summary output. It also made get(id) less useful from the summary, since the visible ID was really the artifact identifier rather than the metadata entry identifier.

This change makes summarize() expose the metadata entry ID as the public report id, while preserving artifact deduplication internally. Project.get(id) now resolves those summary IDs correctly.

#### Contribution checklist

- [ ] The code passes our style conventions (you can check this by running pre-commit on your code with `pre-commit run --all-files`)
- [ ] The documentation builds and renders properly (if it does, our bot will add a comment linking to a preview of the documentation to review it visually)
- [ ] A new changelog entry was added to CHANGELOG.rst (if necessary; typically, if your change requires updating tests)
- [x] All the commits in the PR are signed (more information [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits))
- [x] The pull request title respects the Conventional Commits convention (more information [here](https://docs.skore.probabl.ai/stable/contributing.html#pull-requests))

#### AI usage disclosure
AI tools were involved for:

Closes: N/A, I ran into these issues while working with and exploring the codebase, and put together small targeted fixes for them.

If there is any detail I may have missed or needs clarification, I would be happy to attend to it.

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -230,6 +230,10 @@ jobs:
           path: skore/venv
           key: ${{ steps.cache-python-venv.outputs.cache-primary-key }}
 
+      - name: Install graphviz
+        run: |
+          sudo apt-get install -y graphviz
+
       - name: Install all packages from local branch
         run: |
           python -m pip install --force-reinstall --no-deps skore/
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -42,6 +42,11 @@ Release highlights
 Changed
 -------
 
+- **Breaking change:** ``cache_predictions`` no longer accepts ``n_jobs`` or
+  ``response_methods`` on estimator, cross-validation, and comparison reports.
+  It now infers and caches the relevant prediction outputs automatically. See
+  :pr:`2677` by :user:`cakedev0`.
+
 Added
 -----
 
--- a/ci/requirements/skore/python-3.10/scikit-learn-1.5/test-requirements.txt
+++ b/ci/requirements/skore/python-3.10/scikit-learn-1.5/test-requirements.txt
@@ -198,7 +198,7 @@ six==1.17.0
     # via python-dateutil
 skore-local-project==0.0.5
     # via skore (skore/pyproject.toml)
-skrub==0.7.2
+skrub==0.8.0
     # via skore (skore/pyproject.toml)
 stack-data==0.6.3
     # via ipython
--- a/ci/requirements/skore/python-3.10/scikit-learn-1.7/test-requirements.txt
+++ b/ci/requirements/skore/python-3.10/scikit-learn-1.7/test-requirements.txt
@@ -198,7 +198,7 @@ six==1.17.0
     # via python-dateutil
 skore-local-project==0.0.5
     # via skore (skore/pyproject.toml)
-skrub==0.7.2
+skrub==0.8.0
     # via skore (skore/pyproject.toml)
 stack-data==0.6.3
     # via ipython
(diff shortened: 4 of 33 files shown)


skore-local-project/src/skore_local_project/project.py
auguste-probabl, line 1: Which issues does this PR address? Add `Closes: ` statements to the PR

Ashish-Kumar-Dash: Hello, I ran into these issues while working with and exploring the codebase, and put together small targeted fixes for them. I didn't find there to be a requirement that every PR must be backed by an issue, and for these small, localized bug fixes I did not think opening a separate issue was necessary.

thomass-dev:
Please split your PR in two: you are addressing two fixes in one.
It should be more easy to review and iterate, thanks!

Ashish-Kumar-Dash:
I agree, I have redacted the changes made, I will raise a separate PR for the second one once this one gets approved

skore-local-project/src/skore_local_project/project.py
auguste-probabl: While not strictly necessary, raising an issue lets us get a better understanding of what the problem actually is; I would appreciate more details about that, e.g. a reproducer, maybe in the PR description

auguste-probabl:
> I agree, I have redacted the changes made, I will raise a separate PR for the second one once this one gets approved

Can you also edit the PR title and description to make them accurate?

Ashish-Kumar-Dash:
Let me do it now and also fix the lint errors while I'm at it.

Ashish-Kumar-Dash:
Hey, I've fixed the PR title, description, added a reproducer & ran pre-commit successfully. Kindly review! @thomass-dev

Ashish-Kumar-Dash:
@auguste-probabl  can you trigger the workflow run again, I'm also attaching the screenshot of pre-commit passing locally 
Screenshot 2026-04-02 214148

Ashish-Kumar-Dash:
@thomass-dev @GaetandeCast  @auguste-probabl  Since the workflow checks have now passed can you review my changes
Thanks

thomass-dev:
Thanks for your PR, but unfortunately i don't think the change is relevant for now.
However, we will revisit the widget that displays the summary in next iterations.

> because multiple saved entries looked like the same persisted object

This is the case: when you put several times the exact same entry, it is saved only once and it is the same persisted object.

IMO, this can be useful to know that two objects putted at different moment are equivalent.

Image

thomass-dev:
Next time, i recommend opening an issue to discuss before starting development.

Ashish-Kumar-Dash:
Thanks for the clarification, that makes sense.
I had interpreted the exposed `id` in `summarize()` as the identifier of the persisted summary entry rather than the deduplicated artifact, which is why I treated the repeated ID as ambiguous. If the intended semantics are that identical saved reports should deliberately surface the same persisted artifact ID so users can see equivalence across saves, then I agree.
I appreciate the context, and I’m glad to hear the summary widget will be revisited in upcoming iterations. I’d be happy to help again with a follow up once that direction is clearer.

PR #2701, opened 2026-04-01
refactor: Create metrics registry

auguste-probabl:
#### Change description

This refactor is a precursor to a new feature allowing users to register their own metrics. It introduces **no breaking changes whatsoever**.

It consists in introducing two new concepts, `Metric` and `MetricRegistry`. `MetricRegistry` manages the metrics that can be computed on an EstimatorReport; it a dict containing `Metric`s.
`Metric` is our definition of a metric which may be computed on a report. 

It also re-organizes the `reports/estimator/metrics/test_summarize.py` file to be more systematic about which `MetricLike`s are passed (nothing, strings, Scorers, callables, mix of all of those).

Addresses part of #2061
Combines elements of #2646 and #2277, superseding both.

#### Contribution checklist

requires updating tests)

#### AI usage disclosure

AI tools were involved for:

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -1,13 +1,8 @@
-import inspect
-from collections.abc import Callable
-from functools import partial
+from collections.abc import Callable, Sized
 from typing import Any, Literal, cast
 
-import numpy as np
 import pandas as pd
-import sklearn
 from numpy.typing import ArrayLike
-from sklearn.metrics._scorer import _BaseScorer
 from sklearn.utils.metaestimators import available_if
 
 from skore._externals._pandas_accessors import DirNamesMixin
@@ -20,14 +15,22 @@ from skore._sklearn._plot import (
     PredictionErrorDisplay,
     RocCurveDisplay,
 )
+from skore._sklearn.metrics import (
+    BUILTIN_METRICS,
+    R2,
+    Accuracy,
+    Brier,
+    FitTime,
+    LogLoss,
     Metric,
-from skore._utils._accessor import (
-    _check_all_checks,
-    _check_estimator_has_method,
-    _check_roc_auc,
-    _check_supported_ml_task,
+    Precision,
+    PredictTime,
+    Recall,
+    Rmse,
+    RocAuc,
 )
 from skore._sklearn.types import DataSource, MetricLike, PositiveLabel
+from skore._utils._accessor import _check_supported_ml_task
 from skore._utils._cache_key import make_cache_key
 
 
@@ -37,22 +40,55 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
     You can access this accessor using the `metrics` attribute.
     """
 
-    _score_or_loss_info: dict[str, dict[str, str]] = {
-        "fit_time": {"name": "Fit time (s)", "icon": "(↘︎)"},
-        "predict_time": {"name": "Predict time (s)", "icon": "(↘︎)"},
-        "accuracy": {"name": "Accuracy", "icon": "(↗︎)"},
-        "precision": {"name": "Precision", "icon": "(↗︎)"},
-        "recall": {"name": "Recall", "icon": "(↗︎)"},
-        "brier_score": {"name": "Brier score", "icon": "(↘︎)"},
-        "roc_auc": {"name": "ROC AUC", "icon": "(↗︎)"},
-        "log_loss": {"name": "Log loss", "icon": "(↘︎)"},
-        "r2": {"name": "R²", "icon": "(↗︎)"},
-        "rmse": {"name": "RMSE", "icon": "(↘︎)"},
-        "custom_metric": {"name": "Custom metric", "icon": ""},
-    }
+    def __getattribute__(self, name):
+        """Hide some metric methods conditionally.
 
-    def __init__(self, parent: EstimatorReport) -> None:
-        super().__init__(parent)
+        When the registry is initialized, the report is analyzed to filter metrics
+        depending on the report's characteristics (e.g. the ML task and the estimator's
+        prediction methods).
+        """
+        if (
+            name in {m.name for m in BUILTIN_METRICS}
+            and name not in self._parent._metric_registry
+        ):
+            raise AttributeError(
+                f"'{self.__class__.__name__}' object has no attribute '{name}'"
+            )
+        return super().__getattribute__(name)
+
+    def _parse_metrics(
+        self,
+        metric: MetricLike | list[MetricLike] | dict[str, MetricLike] | None,
+        metric_kwargs: dict[str, Any],
+    ) -> dict[str, Metric]:
+        """Normalize arguments into a mapping from verbose name to Metric.
+
+        Parameters
+        ----------
+        metric : MetricLike, list of MetricLike, dict of MetricLike, or None
+            The metrics to parse.
+        metric_kwargs : dict
+            Kwargs to pass to each metric.
+        """
+        items: list[tuple[str | None, MetricLike]]
+        if metric is None or (isinstance(metric, Sized) and len(metric) == 0):
+            items = [(None, m) for m in self._parent._metric_registry]
+        elif isinstance(metric, dict):
+            items = list(metric.items())
+        elif isinstance(metric, list):
+            items = [(None, m) for m in metric]
+        else:
+            items = [(None, metric)]
+
+        result = {}
+        for display_name, m in items:
+            parsed_metric = self._parent._metric_registry.check_metric(m, metric_kwargs)
+            key = (
+                display_name if display_name is not None else parsed_metric.verbose_name
+            )
+            result[key] = parsed_metric
+
+        return result
 
     def summarize(
         self,
@@ -144,13 +180,6 @@ class _MetricsAccessor(_BaseAccessor[EstimatorReport], DirNamesMixin):
         ROC AUC                         0.99...                     0.99...          (↗︎)
         Log loss                        0.08...                     0.11...          (↘︎)
         Brier score                     0.02...                     0.03...          (↘︎)
-        >>> # Using scikit-learn metrics
-        >>> report.metrics.summarize(
-        ...     metric=["f1"],
-        ... ).frame(favorability=True)
-                                  LogisticRegression Favorability
-        Metric   Label / Average
-        F1 Score               1             0.95...          (↗︎)
         """
         if data_source == "both":
             train_summary = self.summarize(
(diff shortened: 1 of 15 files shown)


skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,508 @@
...
+        **kwargs
+            Additional keyword arguments passed to ``score_func``.
+        """
+        if self.score_func is None:
+            raise NotImplementedError(
+                f"Metric {self.name!r} has no score_func and must override __call__."
+            )
+
+        # FIXME: Should not refer to the metrics accessor
+        X, y_true = report.metrics._get_X_y(data_source=data_source)
```
auguste-probabl: Note: this should be a method of `EstimatorReport`

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -20,48 +15,94 @@
...
+        if (
+            name in _REGISTRY_GATED_METHODS
+            and name not in self._parent._metric_registry
+        ):
+            raise AttributeError(
+                f"'{self.__class__.__name__}' object has no attribute '{name}'"
+            )
+        return super().__getattribute__(name)
+
+    def _parse_metrics(
```
auguste-probabl, line 59: I'm not sure whether this method should be in MetricRegistry tbh

skore/tests/unit/reports/estimator/metrics/test_summarize.py
auguste-probabl, line 1: Would recommend looking at the new file directly for this one

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+
+        Parameters
+        ----------
+        metric : MetricLike
+            The metric to parse.
+        metric_kwargs : dict
+            Kwargs to pass; each metric takes only the kwargs it accepts.
+        """
+        if isinstance(metric, _BaseScorer):
+            func_name = metric._score_func.__name__
```
jeromedockes, line 510: not all callables have a `__name__` (for example `functools.partial(...)` doesn't)

jeromedockes, line 510: Also I'm not sure about the isinstance check because the user could pass a scoring function that has not been created with scikit-learn's make_scorer

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+            kwargs = metric._kwargs.copy()
+            if "pos_label" in inspect.signature(metric._score_func).parameters:
+                if (
+                    "pos_label" in kwargs
+                    and self._report.pos_label != kwargs["pos_label"]
+                ):
+                    raise ValueError(
+                        "The `pos_label` passed in the scorer "
+                        "and the one used when creating the report must match; "
+                        f"got {kwargs['pos_label']!r} and {self._report.pos_label!r}."
```
jeromedockes, line 521: shouldn't the pos_label passed when calling that metric override the default one set on the scorer rather than causing an error?

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+                greater_is_better=metric._sign == 1,
+                score_func=metric._score_func,
+                response_method=metric._response_method,
+                kwargs=kwargs,
+            )
+        elif metric in self:
+            parsed_metric = self[cast(str, metric)]
+
+            kwargs = (
+                _select_kwargs(parsed_metric.score_func, metric_kwargs)
```
jeromedockes, line 535: I'm not sure about this automatic filtering of kwargs. the user should pass kwargs for the metric they are asking for, and if some don't correspond to parameters in the scorer that should probably be a TypeError

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+                raise ValueError(
+                    f"Invalid metric: {metric!r}. "
+                    "Please use a valid metric from the list of supported "
+                    f"metrics: {list(self.keys())} "
+                    "or a valid scikit-learn metric string: "
+                    f"{sklearn.metrics.get_scorer_names()}."
+                ) from err
+
+            return self.parse_metric(scorer, metric_kwargs)
+        elif callable(metric):
```
jeromedockes, line 561: in scikit-learn if a value for the `scoring` parameter is a callable, it is assumed to accept (estimator, X, y) and return either a score or a dict of scores.

here IIUC, if we pass a callable, it is interpreted as in scikit-learn only when it happens to be a BaseScorer, and otherwise it is interpreted as what scikit-learn calls a "metric", a function that accepts (y_true, y_score) and returns a score.

I find that implicit switch a bit confusing, and it means users have to learn a different convention than scikit-learn's `scoring` parameter. Moreover it makes it impossible to define a custom scorer, and it causes metrics that have a `score_func` to be partly redundant with scikit-learn's `make_scorer`.

have we considered accepting exactly the same inputs as `scoring` in scikit-learn, with the same semantics? I realize this will make using the cached predictions more difficult, but I think we should investigate whether it is possible at all

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -176,155 +208,25 @@ def summarize(
...
-                if "pos_label" in metrics_params:
-                    metrics_kwargs["pos_label"] = pos_label
-            else:
-                raise ValueError(
-                    f"Invalid type of metric: {type(metric_)} for {metric_!r}"
-                )
-
-            score = metric_fn(data_source=data_source, **metrics_kwargs)
+        for metric_name, parsed_metric in parsed_metrics.items():
+            score = parsed_metric(
```
jeromedockes, line 211: note that scorers can also return a dict `{metric name: value}` rather than a single value

jeromedockes (commented):
thanks @auguste-probabl ! this is very cool. It is an intricate topic partly due to the complexity of scikit-learn metrics and scorers but I wonder if we can make fewer implicit decisions like the distinction between scorers and other callables or the distribution of kwargs based on signatures

skore/src/skore/_sklearn/metrics.py
jeromedockes: > I realize this will make using the cached predictions more difficult, but I think we should investigate whether it is possible at all

if not, an option would be to give up on caching only for this specific case but use it for all other scorers

skore/src/skore/_sklearn/metrics.py
auguste-probabl: This logic is only relevant for `summarize`, which accepts multiple metrics but just one `metric_kwargs` dict. This PR is a refactor so I didn't allow myself to change behaviour here, but once registering metrics is supported, this could be simplified:
- `summarize` may no longer have arguments (it would just give the summary for all registered metrics)
- the `metric_kwargs` argument could change to instead detail which metric gets which arguments, e.g. `{"precision": {"average": None}, "roc_auc": {"multi_class": "ovo"}}"`

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: I would consider this use-case unsupported for now; this needs to be documented.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: That's fair; the goal here is to provide first-class support for actual sklearn scorers (`_BaseScorer` instances).

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Maybe we could, although we'd have to take care to also set `prediction_pos_label` properly, otherwise we'll get the predictions for label `"A"` and then compute the metric thinking we're getting label `"B"`.

That being said, this is out-of-scope for this PR.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Yes, this was already the case before. I think we can be smarter in a later PR.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Yes indeed, I agree that 100% compatibility with sklearn is a nice goal to set. That being said, it is out-of-scope for this PR.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: see https://github.com/probabl-ai/skore/pull/2713, please merge that PR first

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -20,48 +15,94 @@
...
+            items = list(metric.items())
+        elif isinstance(metric, list):
+            items = [(None, m) for m in metric]
+        else:
+            items = [(None, metric)]
+
+        result = {}
+        for display_name, m in items:
+            parsed_metric = self._parent._metric_registry.parse_metric(m, metric_kwargs)
+            key = display_name if display_name else parsed_metric.verbose_name
```
cakedev0: I had too many bad experiences with this trick, now I avoid it:
```suggestion
            key = display_name if display_name is not None else parsed_metric.verbose_name
```

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -20,48 +15,94 @@
...
+
+        Parameters
+        ----------
+        metric : MetricLike, list of MetricLike, dict of MetricLike, or None
+            The metrics to parse.
+        metric_kwargs : dict
+            Kwargs to pass to each metric.
+        """
+        items: list[tuple[str | None, MetricLike]]
+        if metric is None or isinstance(metric, Sized) and len(metric) == 0:
```
cakedev0: Personally, I don't know the resolution order of or vs and so this is hard to read.

skore/src/skore/_sklearn/metrics.py
cakedev0: We want to remove pos_label from the __init__ / from the report's attributes. This shouldn't be a problem then I guess.

(I have started doing that, and it seems it could simplify or even remove the `prediction_pos_label` concept... Which would be nice)

So for now, this seems good.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+        if report._ml_task == "binary-classification":
+            return has_predict_proba or has_decision_function
+        elif report._ml_task == "multiclass-classification":
+            return has_predict_proba
+        return False
+
+    def _prediction_pos_label(self, report):
+        if report._ml_task == "multiclass-classification":
+            return None
+        return report.pos_label or report._estimator.classes_[-1]
```
cakedev0: `report.pos_label or report._estimator.classes_[-1]` overwrites pos_label if it's falsy (0/False/...).

```suggestion
        if report.pos_label is None:
            return report._estimator.classes_[-1]
        return report.pos_label
```

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+    @staticmethod
+    def available(report: EstimatorReport) -> bool:
+        """Whether this metric is applicable to the given report."""
+        return True
+
+    def _prediction_pos_label(self, report: EstimatorReport) -> PositiveLabel | None:
+        """Return the pos_label to use for ``_get_predictions``.
+
+        Subclasses may override this to provide a default when ``report.pos_label``
+        is ``None`` (e.g. binary classification metrics that need 1-D predictions).
```
cakedev0, line 142: Neat explanation :+1:

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+        """
+        if self.score_func is None:
+            raise NotImplementedError(
+                f"Metric {self.name!r} has no score_func and must override __call__."
+            )
+
+        X, y_true = report._get_X_y(data_source=data_source)
+
+        # Merge default kwargs with call-time kwargs
+        merged_kwargs = {**self.kwargs, **kwargs}
```
cakedev0: Personally, I'm used to `**` but it seems that since Python 3.9 the standard is to use `|`:

```suggestion
        merged_kwargs = self.kwargs | kwargs
```

(I discovered this syntax in this repo ^^)

cakedev0 (commented):
Overall, it looks like a nice refactor. I hope it'll help simplifying further `summarize` shenanigans later on. 

I would need more review time to approve it, and esp. more researches on scikit-learn scorers. But I clearly don't see any blocker.

Also, Codex had some findings you may want to address (but I'm good if you don't want to):

<details>
# PR #2701 review findings

## 1. Built-in metrics can now raise `TypeError` for `metric_kwargs` that the old code tolerated

This PR introduces a regression in how `metric_kwargs` are handled for built-in metrics.

The issue comes from selecting kwargs against the sklearn `score_func` signature, then passing them into the metric object's `__call__`, whose signature is often narrower.

I reproduced this with:

- `metric="r2", metric_kwargs={"force_finite": False}`
- `metric="precision", metric_kwargs={"zero_division": 1}`

Both fail on this branch, even though the old code tolerated those inputs.

That makes the PR's claim of "no breaking changes whatsoever" inaccurate.

Relevant locations:

- `skore/src/skore/_sklearn/metrics.py:467`
- `skore/src/skore/_sklearn/_estimator/metrics_accessor.py:218`
- `skore/src/skore/_sklearn/metrics.py:238`
- `skore/src/skore/_sklearn/metrics.py:363`

## 2. `summarize` now silently drops duplicate metrics

`_parse_metrics` now normalizes the input list into a dict keyed by display name.

That means duplicate metrics are silently deduplicated. I reproduced this with:

- `metric=["accuracy", "accuracy"]`

and only got one row back.

A list should preserve multiplicity and order. Only explicit dict input should behave like a mapping.

Relevant location:

- `skore/src/skore/_sklearn/_estimator/metrics_accessor.py:69`

</details>

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
jeromedockes: I agree parentheses wouldn't hurt here but to help you remember precedence in the future think OR is similar to addition and AND to multiplication

auguste-probabl:
Also addressed Codex's metrics_kwargs regression from https://github.com/probabl-ai/skore/pull/2701#pullrequestreview-4054795634, but not the "drops duplicates" thing.

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
cakedev0: Oh, lovely!

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
glemaitre: I don't mind to merge to keep the same API. However, I think that I would the following change in another PR:

- `_parse_metrics` should be part of the `MetricRegistry` and basically, it is only an `add`, `append`, or `register` method.
- remove the call to this function in `summarize`. Instead `summarize` will have a `metric` parameter that should take the key to fetch the `MetricRegistry` metrics that we want to use. It is only a way to filter and it should simplify the internals of `summarize`.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -0,0 +1,576 @@
...
+        self.data = {
+            metric.name: metric
+            for metric in BUILTIN_METRICS
+            if metric.available(report)
+        }
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}({list(self.data.keys())})"
+
+    def parse_metric(self, metric: MetricLike, metric_kwargs: dict[str, Any]) -> Metric:
```
glemaitre: What about the `as_metric`. For me `parsing` is more about checking text. If you recall my point on `MetricLike`, I think that it is indeed what I would expect from this function.

jeromedockes: could also be `check_metric`, to follow scikit-learn habits

glemaitre: Yep either way is fine with me.

auguste-probabl: I tend to use `parse` as inspired from ["Parse, don't validate"](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/); happy to change it. I guess `check_metric` is good.

glemaitre (approved):

glemaitre:
Let's merge this one and go for the follow-up clean-up/breaking change.

PR #2707, opened 2026-04-02
fix(skore): Eager id to avoid mutation (which breaks checksum in `skore-hub-project`)

cakedev0:
#### Change description

Define `self.id` at init instead of a cached property. The cached property breaks the checksum stability in `skore-hub-project` (try running tests on "main-main", you'll see). 

#### AI usage disclosure

Research and understanding

### DIFF
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -1,4 +1,3 @@
-from functools import cached_property
 from io import StringIO
 from typing import Generic, Literal, TypeVar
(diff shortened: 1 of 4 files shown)


cakedev0:
> It should only affect tests, isn't?

Well it breaks the workflow:

```python
project.put(key, cv_report)
# oh I forgot to compute permutation importance
# let's compute them and send them to the hub:
cv_report.inspection.feature_importance()
project.put(key, cv_report)
```

PR #2710, opened 2026-04-02
refactor(skore): Structured cache and `_get_cached_results` "private-public" API 

cakedev0:
#### Change description
This PR takes the short-term direction discussed in #2587 (make it possible for `skore-*-project` to inspect cached results through an "private-public" contract) while still moving toward a cleaner report structure.

- Predictions and prediction timings now live in a dedicated `_predictions` cache
- accessor-computed artifacts such as `metrics.*` and `inspection.*` stay in `_cache` and are accessed through `_read_cache` / `_write_cache`. The raw cache keys are now hashed implementation details, and the stored values carry the structured information `(accessor_name, method_name, data_source, kwargs, result)`.

The new "private-public" contract is `_get_cached_results(accessor_name, method_name=None, data_source=None)`, which exposes the tuples `(accessor_name, method_name, data_source, kwargs, result)` with some convenience filtering.

This PR also removes the old `make_cache_key` helper.

Follow-up PR with `get_state/from_state`: #2718


skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -616,3 +615,107 @@ def _repr_mimebundle_(self, **kwargs):
...
+        return value
+
+    def _write_cache(
+        self,
+        accessor_name: str,
+        method_name: str,
+        data_source: DataSource | None,
+        kwargs: dict[str, Any],
+        result: Any,
+    ) -> None:
```
cakedev0, line 829: Use kw-only arguments?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -616,3 +615,107 @@ def _repr_mimebundle_(self, **kwargs):
...
+        key = joblib.hash((accessor_name, method_name, data_source, kwargs))
+        value = (
+            accessor_name,
+            method_name,
+            data_source,
+            kwargs,
+            result,
+        )
```
cakedev0: This is a big tuple, maybe it should be a named tuple?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -780,3 +781,107 @@ def _repr_mimebundle_(self, **kwargs):
         output = {"text/plain": repr(self)}
         output["text/html"] = self._repr_html_()
         return output
+
+    ####################################################################################
+    # Methods related to the cache
+    ####################################################################################
```
cakedev0, line 799: Maybe we could move the cache ownership and those methods into the base accessor class (and change the pattern to a cached accessor class for persistence). This could be cleaner I think.

Also, it would be more aligned with the choice I made for `EstimatorReport.get_state` in PR #2718 to split metrics and inspection in two different keys of the state.

cakedev0: Though, I don't know how to implement `CrossValidationReport._get_cached_results` then... I think this pattern would mean re-introducing cache (and hence cache-duplication) in CV reports.

cakedev0:
Closing this PR:
- not the validated design
- not really useful after migration to mono-package (planned for soon)

PR #2716, opened 2026-04-03
refactor(skore): Remove X and y keys from report _data

jeromedockes:
in the dictionary that is created by the report when the user passed X and y directly, those keys are redundant with `_skrub_X` and `_skrub_y`

### DIFF
--- a/skore/src/skore/_utils/_skrub.py
+++ b/skore/src/skore/_utils/_skrub.py
@@ -48,7 +48,7 @@ class _LearnerAdapter(BaseEstimator):
         def learner_method(data):
             kwargs = {"X": data["_skrub_X"]}
             if name in ["fit", "score"]:
-                kwargs["y"] = data.get("_skrub_y")
+                kwargs["y"] = data["_skrub_y"]
             return estimator_method(**kwargs)
 
         return learner_method


skore/src/skore/_utils/_skrub.py
```diff
@@ -46,9 +46,9 @@ def __getattr__(self, name):
 
         @functools.wraps(estimator_method)
         def learner_method(data):
-            kwargs = {"X": data.get("_skrub_X", data["X"])}
+            kwargs = {"X": data["_skrub_X"]}
             if name in ["fit", "score"]:
-                kwargs["y"] = data.get("_skrub_y", data["y"])
+                kwargs["y"] = data.get("_skrub_y")
```
cakedev0: Why `[]` for X and `get` for y?

cakedev0 (approved):
Thanks!

skore/src/skore/_utils/_skrub.py
jeromedockes: a reflex from situations where y is allowed to be None (unsupervised transformers), you're right it does not make sense here

glemaitre:
Nice auto-merge was already on. Thanks @cakedev0

PR #2717, opened 2026-04-03
fix(skore-hub-project/splitting_strategy): Do not use  `rng.choice` when `len(target)<100`

thomass-dev:


skore-hub-project/src/skore_hub_project/report/cross_validation_report.py
```diff
@@ -222,6 +222,7 @@ def splitting_strategy(self) -> dict[str, Any]:
 
         splitter = self.report.splitter
         target = self.report.y
+        rng_size = min(len(target), SPLITTING_STRATEGY_REPR_SAMPLE_COUNT)
```
glemaitre: actually the right fix here is to completely avoid the undersampling and just get the target as-is:

```python
if len(target) < SPLITTING_STRATEGY_REPR_SAMPLE_COUNT:
    target_repr = target
else:
    # the current code.
```

thomass-dev: fixed

PR #2725, opened 2026-04-07
fix: Use function source code to compute cache key

### LINKED ISSUE #2060: docs(skore): Custom metric cache is based only on function name
## Context 

If we run the below code for a custom metric, I get my expected output

```python
# cell 1
from sklearn.datasets import load_diabetes
from sklearn.metrics import mean_absolute_error
```

```python
# cell 2
X, y = load_diabetes(return_X_y=True)
split_data = train_test_split(X=X, y=y, random_state=0, as_dict=True)
regressor = Ridge()
report = EstimatorReport(regressor, **split_data)
```

```python
# cell 3
def metric_function(y_true, y_pred):
    return {"output": mean_absolute_error(y_true, y_pred)}

report.metrics.custom_metric(
    metric_function=metric_function,
    response_method="predict",
)
# {"output" 44.922683494675425}
```

If I change my mind as to how the output looks like (or even the calculation)

```python
# cell 3 (now change metric_function's calculation)
def metric_function(y_true, y_pred):
    return 0.2 # or just mean_absolute_error(y_true, y_pred)

report.metrics.custom_metric(
    metric_function=metric_function,
    response_method="predict",
)
```

I will still get

```
{"output" 44.922683494675425}
```

## Alternatives

Atm we cache the metric function based [on the function name](https://github.com/probabl-ai/skore/blob/7251e6974c9b81757c1433edf9e1dcb2f60ce502/skore/src/skore/_sklearn/_estimator/metrics_accessor.py#L451C9-L455C10)

```python
# from skore/src/skore/_sklearn/_estimator/metrics_accessor.py
cache_key_parts: list[Any] = [
    self._parent._hash,
    metric_fn.__name__,
    data_source,
]
```

And indeed if I change my function name

```python
def metric_function_CHANGED(y_true, y_pred):
    return 0.2 

report.metrics.custom_metric(
    metric_function=metric_function_CHANGED,
    response_method="predict",
)
# now I get 0.2
```

Would some hashing based on the function contents be better? For instance:

```python
import hashlib
import inspect

def foo1(x):
    return str(x)

def hash_function(fn) -> str:
    hasher = hashlib.sha256()
    hasher.update(fn.__code__.co_code)
    return hasher.hexdigest()

hash1 = hash_function(foo1)

# change function body to return something different but keep the same name
def foo1(x):
    return x

hash2 = hash_function(foo1)
hash1 == hash2 # False
```

Or use `id(metric_function)`

❗ a user on discord (pyphi) also mentioned they faced a similar issue when trying to use custom metrics. exact msg: *Do you know if we can disable cahing? I got into issues updating the metric function, before I realise skore is chaching stuff!*

auguste-probabl:
#### Change description

Closes #2060 (again?)

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:

### DIFF
--- a/skore/src/skore/_utils/_cache_key.py
+++ b/skore/src/skore/_utils/_cache_key.py
@@ -17,7 +17,7 @@ def make_cache_key(
 ) -> tuple[Any, ...]:
     """Build a cache key.
 
-    Enforce structure (data_source, name, sanitized_kwargs).
+    Enforce structure (data_source, "predict_time", sanitized_kwargs).
     """
     if data_source not in {"train", "test"}:
         raise ValueError(f"data_source must be 'train' or 'test'; got {data_source!r}")
--- a/skore/src/skore/_utils/_cache_key.py
+++ b/skore/src/skore/_utils/_cache_key.py
@@ -13,15 +13,18 @@ def make_cache_key(
     data_source: DataSource,
     name: str,
     kwargs: Mapping[str, Any] | None = None,
+    func_id: str | None = None,
 ) -> tuple[Any, ...]:
     """Build a cache key.
 
-    Enforce structure (data_source, "predict_time", sanitized_kwargs).
+    Enforce structure (data_source, name, sanitized_kwargs).
     """
     if data_source not in {"train", "test"}:
         raise ValueError(f"data_source must be 'train' or 'test'; got {data_source!r}")
     if not isinstance(name, str):
         raise TypeError(f"name must be a string; got {type(name)}")
+    if func_id is not None:
+        name = f"{name}:{joblib.hash(func_id)}"
     return (data_source, name, deep_key_sanitize(kwargs))
 
 
--- a/skore/src/skore/_sklearn/metrics.py
+++ b/skore/src/skore/_sklearn/metrics.py
@@ -179,7 +179,12 @@ class Metric:
         # Merge default kwargs with call-time kwargs
         merged_kwargs = self.kwargs | kwargs
 
-        cache_key = make_cache_key(data_source, self.name, merged_kwargs)
+        cache_key = make_cache_key(
+            data_source,
+            self.name,
+            merged_kwargs,
+            func_id=getattr(self, "source_code", None),
+        )
         score = report._cache.get(cache_key)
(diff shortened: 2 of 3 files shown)


skore/src/skore/_utils/_cache_key.py
```diff
@@ -13,15 +13,18 @@ def make_cache_key(
     data_source: DataSource,
     name: str,
     kwargs: Mapping[str, Any] | None = None,
+    func_id: str | None = None,
 ) -> tuple[Any, ...]:
     """Build a cache key.
 
-    Enforce structure (data_source, "predict_time", sanitized_kwargs).
+    Enforce structure (data_source, name, sanitized_kwargs).
```
auguste-probabl:
```suggestion
    Enforce structure (data_source, "predict_time", sanitized_kwargs).
```

jeromedockes:
I think detecting whether the function would return the same value is not so easy (it may depend on other functions that get called or some other global or nonlocal variables etc, and as your code anticipates it may not be possible to retrieve the code). moreover, the fact that the report outputs is tied to one particular revision of the user's codebase and environment is not only the case for metrics but also for the estimator for example. so I would suggest making it clear in the documentation that the report results are only computed once, and not trying to do any kind of cache invalidation. instead, we can focus on adding tooling to skore to help with keeping track of the code and environment and helping with reproducibility. WDYT?

auguste-probabl:
> I would suggest making it clear in the documentation that the report results are only computed once, and not trying to do any kind of cache invalidation

I agree that eventually we want to track things more precisely; however, in the meantime, #2060 is a hindrance for users. Having to re-create a report because e.g. there was a typo in the custom metric is very painful, and I think we should at least provide a best-effort solution until we can solve it properly. Point taken about the docs though, I'll work on that.

jeromedockes:
from IRL discussion: in the `.metrics.add(custom_scorer)` function that will soon be added, if a scorer with an existing name is added it replaces the existing one and clears the cache. once the ongoing work around the metrics registry is stabilized we will have a clearer idea of what we want to do; it looks like `custom_metric` may disappear altogether

PR #2727, opened 2026-04-07
feat: Add check registry to diagnostic

GaetandeCast:
Adds a registry system to the diagnosis feature.

A check is now an object comprised of a callable that returns an explanation if an issue is found and a false value otherwise; along with some meta data used to run the check and construct the result: a unique code identifier, a title, a report type and an optional documentation url. This url can be a full url starting with `https` or an anchor to the diagnostic section of our userguide. 

I also improved the tests a good bit (which explains the rather large diff) in order to achieve the following goals:
- readability through reduction of duplicated code and usage of fixtures,
- simplification by not re testing logic that is independent of report type on the different report objects.
- create a dedicated file for comparison report.

I also added a new example to the technical details section of the documentation to showcase how to create and register custom checks.


skore/src/skore/_sklearn/_diagnostic/model_checks.py
```diff
@@ -18,14 +19,11 @@
     from skore._sklearn._estimator.report import EstimatorReport
 
 
-def check_overfitting_underfitting(
```
GaetandeCast, line 21: Had to split this in two to fit the new `Check` object. This means we compute the same thing twice by calling `_get_metrics_data` two times.

I think it should not be too expensive since the underlying metrics computation should be cached (?) and the dummy model should be fast to predict. We can cache the result otherwise, especially if we end up adding other checks that need the same data.

Let me know what you think.

jeromedockes: we can do it like this for now but I suspect that there will be other cases where it is easier to check several things in one go, which is why you allowed the callable to return multiple check results in your previous pr. maybe we want to keep that 🤔 

just reminding us of the context here: in this PR, we want to be able to have information about the check that a function performs, before running it or even if it finds nothing so we can report it.

If we want both of those things I guess we need structured objects both for the checker and the check result, WDYT?

maybe the checker has an attribute that is the list of check types it performs, and a method like `check(report)`. check returns a list of check results (empty if no issues), which are instances of the announced check types. each check type has a static description and code, and the instance can contain more detail about the specific issue detected.

maybe not exactly this but I guess we need something along those lines

jeromedockes: from IRL discussion: we keep returning a simple string for now; we can revisit later

examples/technical_details/plot_custom_checks.py
```diff
@@ -0,0 +1,143 @@
...
+import numpy as np
+from skore import Check, DiagnosticNotApplicable
+
+
+def check_high_feature_count(report):
+    """Flag when the number of features exceeds a threshold."""
+    if report.X_test is None:
+        raise DiagnosticNotApplicable()
+
+    n_features = np.asarray(report.X_test).shape[1]
```
jeromedockes: i know this is just a dummy example but if X_test happens to be a large dataframe copying it into an array to check the shape seems very wasteful -- as X is an array in the example you can just do X.shape[1]?

examples/technical_details/plot_custom_checks.py
```diff
@@ -0,0 +1,143 @@
...
+# any newly added checks on top of the built-in checks.
+from sklearn.linear_model import LinearRegression
+from skore import evaluate
+
+rng = np.random.default_rng(42)
+X = rng.normal(size=(200, 80))
+y = X[:, 0] + rng.normal(size=200)
+
+report = evaluate(LinearRegression(), X, y)
+report.add_checks([custom_check_1])
```
jeromedockes, line 71: maybe it could be `add_check(check_func=..., code=..., title=...)` to have one less step to add a check instead of first building the check then addign it? alternatively if we define the check ourselves it might be more natural to some users to define a class like 

```
class MyCheck:
    title = '...'
    @staticmethod
    def run(report): ...
```

🤔

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -100,7 +100,83 @@ def __repr__(self) -> str:
...
+        `""`) when there is nothing to report.
+
+    code : str
+        unique identifier for this check , used in
+        :meth:`~skore.EstimatorReport.diagnose` and `ignore` lists.
+
+    title : str
+        short label shown for the issue when one is reported.
+
+    report_type : str
```
jeromedockes, line 129: is it too cumbersome to handle this through the same NotApplicable exception mechanism?

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -100,7 +100,83 @@ def __repr__(self) -> str:
...
+    """Single diagnostic check.
+
+    Each check wraps a callable that inspects a report. If the callable returns a
+    non-empty string, that text is recorded as an issue under :attr:`code` with the
+    given :attr:`title`. Checks are scoped to a single report kind via
+    :attr:`report_type` so they only run on matching reports.
+
+    Parameters
+    ----------
+    function : callable
```
jeromedockes: maybe we can have a slightly less generic name?

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -100,7 +100,83 @@ def __repr__(self) -> str:
...
+            "comparison-cross-validation",
+        ]
+        if report_type not in report_types:
+            raise ValueError(
+                f"report_type should be one of: {', '.join(report_types)}. "
+                f"Got {report_type} instead."
+            )
+        self.report_type = report_type
+
+    def _run(self, report) -> dict[str, dict]:
```
jeromedockes: should it be run() (public), as it is meant to be called by users of this class?

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -100,7 +100,83 @@ def __repr__(self) -> str:
...
+        explanation = self.function(report)
+        if explanation:
+            return {
```
jeromedockes, line 180: I find the return format a bit strange: why a dict with a single key, rather than just the inner dict with a `code` key added? and if nothing was found maybe return None would be clearer?

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -100,7 +100,83 @@ def __repr__(self) -> str:
...
+                }
+            }
+        return {}
+
+
+def _get_issue_documentation_url(issue: dict) -> str | None:
+    docs_url = issue.get("docs_url")
+    if docs_url is None:
+        return None
+    if docs_url.startswith("https"):
```
jeromedockes: maybe startswith("http")

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -110,29 +186,43 @@ def get_issue_documentation_url(*, docs_anchor: str) -> str:
...
+    return issue.get("docs_url", "").startswith("https")
 
 
-def format_issue_message_html(code: str, issue: dict) -> str:
+def _format_issue_message(code: str, issue: dict) -> str:
+    msg = f"[{code}] {issue['title']}. {issue['explanation']}"
+    docs_url = _get_issue_documentation_url(issue)
+    if docs_url is not None:
+        if _is_external_url(issue):
+            msg += f" Read more about this here: {docs_url}."
```
jeromedockes: i think this message works for both external and internal urls

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: Yes it is. `report_type` is necessary for cross-validation and comparison reports to know whether to run a check on their component reports and aggregate the results or if the check should be run on themselves directly.

examples/technical_details/plot_custom_checks.py
GaetandeCast: yeah that was useless, switched to `X.shape[0]`

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: Changed to `check_function`. I'm still not very happy with it but it's better it think.

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: Made `run` public, added documentation

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: Heritage from previous code where one check could return multiple issues. Return type is now `dict | None`

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: done

examples/technical_details/plot_custom_checks.py
GaetandeCast: > maybe it could be add_check(check_func=..., code=..., title=...) to have one less step to add a check 

I see the point but creating a class instance means we can add the same check to different reports, like we do in the example.

>  it might be more natural to some users to define a class like

It does not feel more natural to me :sweat_smile: and it becomes a bit annoying to make sure that the user passes a class that is compatible with what we do. Either we do a base class that user checks inherit from that ensures the presence of the right attributes and method or we inspect the object we get to check that it has the required elements.

GaetandeCast: > it might be more natural to some users to define a class

I had the same feedback from @thomass-dev and we discussed a little about how to go with it. We landed on creating a `Check` protocol that has to be implemented by the user to create their check class. I'll update the PR with this change.

jeromedockes: oops sorry I was going to reply "Ok then" to your "It does not feel more natural to me" :) -- I think you're right and we can keep it like this for now. I was just wondering a bit about passing the function to Check, as some beginners are not super comfortable with the functions as first-class objects thing, but it is quite similar to what is done in make_scorer for example and having to inherit from the Check protocol might seem even more complex

let's merge without further changes on that aspect and defer that to another issue or pr if we do decide to change? also it will be good to think about this in the light of what the metrics registry looks like once the dust settles on that

GaetandeCast: Oops @thomass-dev convinced me and I did the change x) Let me know what you think about what I did here, and I can always revert the last change and come back to the idea in a second time if it's not good enough. Looking at the changes to the `plot_custom_check` example should show how the flow of creating a new check changed.

examples/technical_details/plot_custom_checks.py
```diff
@@ -0,0 +1,138 @@
...
+    report_type = "estimator"
+    docs_url = "https://scikit-learn.org/stable/modules/feature_selection.html#feature-selection"
+
+    def check_function(self, report):
+        """Flag when the number of features exceeds a threshold."""
+        if report.X_test is None:
+            raise DiagnosticNotApplicable()
+
+        n_features = X.shape[1]
+        if n_features > 50:
```
jeromedockes, line 45: maybe we can make the threshold (50 here or the 0.1) later a parameter of the check's `__init__` -- this way it seems a bit more natural that we define a check class and then pass an instance?

```
class CustomCheck1(Check):
    def __init__(self, threshold=50): ...

report.add_checks(CustomCheck1())
```

jeromedockes (approved):

examples/technical_details/plot_custom_checks.py
GaetandeCast: Yes for sure

PR #2731, opened 2026-04-08
feat(skore): Calls history inspection

cakedev0:
#### Change description

This PR adds call history tracking for `inspection.permutation_importance()` on
estimator and cross-validation reports.
**Goal:** Allow `skore-*-project`s to do call history inspection via a clean public interface instead of relying on cache inspection (because cache is not meant for that, see https://github.com/probabl-ai/skore/discussions/2587#discussioncomment-16477740) 

This was done by introducing a `record_calls` descriptor that is used a decorator
on `inspection.permutation_importance` and could be used on other accessor methods later.

This required a small update `get_public_methods()` to keep methods wrapped by custom
descriptors visible in the help tree.

**Choices:**

- `record_calls` is a descriptor and not a simple decorator. It was needed to have
  the API "`.calls` is a list" instead of "`.calls()` returns a list". And it was
  not really more complicated the descriptor.
- Call records are attached to the report via `._parent` because the accessor instance
  is not persistent (`report.metrics is not report.metrics`). Initially, I had changed
  the accessor pattern to a cached-accessor pattern, but apparently it's not a good idea:
  https://github.com/pandas-dev/pandas/issues/47667
- `record_calls` deduplicates calls: not sure this is needed? It feels nice to have but 
  it  adds some complexity.

**Testing:**
Tests for:
- estimator permutation-importance call history
- cross-validation permutation-importance call history
- accessor help discovery with descriptor-wrapped methods

#### AI usage disclosure

Used during the whole process in an interactive way.

### DIFF
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -139,8 +139,8 @@ class record_calls:
     """Descriptor that records normalized method calls on the parent report."""
 
     def __init__(self, func):
-        self.func = func
-        self.sig = inspect.signature(func)
+        self.function = func
+        self.signature = inspect.signature(func)
         self.attr_name = f"__calls_{id(func)}"
         update_wrapper(self, func)
 
@@ -156,16 +156,16 @@ class _BoundRecordedMethod:
     def __init__(self, decorator, instance):
         self._decorator = decorator
         self._instance = instance
-        update_wrapper(self, decorator.func)
+        update_wrapper(self, decorator.function)
 
     def __call__(self, *args, **kwargs):
-        dec = self._decorator
+        decorator = self._decorator
         instance = self._instance
         # Accessors are re-instantiated on each lookup, so the bound wrapper stores
         # call history on the shared parent report rather than on the accessor.
         storage_owner = instance._parent
 
-        bound_args = dec.sig.bind(instance, *args, **kwargs)
+        bound_args = decorator.signature.bind(instance, *args, **kwargs)
         bound_args.apply_defaults()
 
         args_dict = dict(bound_args.arguments)
@@ -175,9 +175,9 @@ class _BoundRecordedMethod:
         # parameter combinations rather than every invocation.
         h = joblib.hash(args_dict)
 
-        result = dec.func(instance, *args, **kwargs)
+        result = decorator.function(instance, *args, **kwargs)
 
-        calls = storage_owner.__dict__.setdefault(dec.attr_name, {})
+        calls = storage_owner.__dict__.setdefault(decorator.attr_name, {})
         calls[h] = args_dict
 
         return result
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -139,36 +139,41 @@ class record_calls:
     def __init__(self, func):
         self.func = func
         self.sig = inspect.signature(func)
-        self.attr_name = f"__calls_{func.__name__}"
+        self.attr_name = f"__calls_{id(func)}"
         update_wrapper(self, func)
 
     def __get__(self, instance, owner):
         if instance is None:
             return self
+        return _BoundRecordedMethod(self, instance)
 
-        def bound(*args, **kwargs):
-            bound_args = self.sig.bind(instance, *args, **kwargs)
-            bound_args.apply_defaults()
 
-            args_dict = dict(bound_args.arguments)
-            args_dict.pop("self")
+class _BoundRecordedMethod:
+    def __init__(self, decorator, instance):
+        self._decorator = decorator
+        self._instance = instance
+        update_wrapper(self, decorator.func)
 
-            h = joblib.hash(args_dict)
+    def __call__(self, *args, **kwargs):
+        dec = self._decorator
+        instance = self._instance
+        storage_owner = instance._parent
 
-            if not hasattr(instance, self.attr_name):
-                setattr(instance, self.attr_name, {})
+        bound_args = dec.sig.bind(instance, *args, **kwargs)
+        bound_args.apply_defaults()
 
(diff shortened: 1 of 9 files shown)


skore/src/skore/_sklearn/_base.py
```diff
@@ -130,3 +133,57 @@ def _rich_repr(self, class_name: str) -> str:
...
         return string_buffer.getvalue()
+
+
+class record_calls:
+    """Descriptor that records normalized method calls on the parent report."""
+
+    def __init__(self, func):
+        self.func = func
+        self.sig = inspect.signature(func)
+        self.attr_name = f"__calls_{id(func)}"
```
auguste-probabl:
```suggestion
        self.function = func
        self.signature = inspect.signature(func)
        self.attribute_name = f"__calls_{id(func)}"
```

skore/tests/unit/utils/repr/test_data.py
```diff
@@ -251,6 +264,16 @@ def test_get_public_methods_accessor_excludes_help(accessor_with_methods):
     assert "fetch" in names
 
 
+def test_get_public_methods_accessor_includes_recorded_methods(report_with_methods):
+    """get_public_methods includes methods wrapped by custom descriptors."""
+    accessor = _AccessorWithRecordedMethod(parent=report_with_methods)
+
+    methods = get_public_methods(accessor)
+    names = [n for n, _ in methods]
```
auguste-probabl:
```suggestion
    names = [n for n, _ in get_public_methods(accessor)]
```

skore/tests/unit/utils/repr/test_data.py
```diff
@@ -251,6 +264,16 @@ def test_get_public_methods_accessor_excludes_help(accessor_with_methods):
...
+    assert len(names) == 1
+    assert names[0] == "fetch"
```
auguste-probabl:
```suggestion
    assert names == ["fetch"]
```

auguste-probabl (commented):
I'm not quite sure what the final goal is. Here is a first review anyway.

cakedev0:
@auguste-probabl  I added a "Goal: [...]" section in the PR description.

skore/src/skore/_sklearn/_base.py
cakedev0: Renamed some attributes & variables to avoid abbreviations. Kept `attr` and `args` as they are standard ones I'd say.

jeromedockes:
I am wondering if we are not introducing too much complexity here. I still think things could be a bit simpler if the report's `get_state` that @cakedev0  is working on, or some other method along the lines of `export_whatever_you_have()` returns something that contains a list of `permutation_importances: [(data_source, step, importances)]` and consumers can read / display that . but this is a first quick impression, I will take a closer look tomorrow .

skore/src/skore/_sklearn/_base.py
```diff
@@ -130,3 +133,57 @@ def _rich_repr(self, class_name: str) -> str:
...
         return string_buffer.getvalue()
+
+
+class record_calls:
+    """Descriptor that records normalized method calls on the parent report."""
+
+    def __init__(self, func):
+        self.function = func
+        self.signature = inspect.signature(func)
+        self.attr_name = f"__calls_{id(func)}"
```
jeromedockes, line 144: I think if we pickle an object with record calls methods and load it in another process we will end up with several `__calls_012abc` attributes and `.calls` will only return the calls from the current process, is that a problem?


also not a big blocker but the decorating will make tracebacks slightly worse when the function fails or the user passes the wrong arguments, so it could be nice if we can avoid decorating a lot of the report methods with this

cakedev0:
IMO, `get_state()` or `export_whatever_you_have()` suffer from the problem described here: it does not have the same purpose and because what a report stores is different from what was called (e.g. stores a result for train & test, but was called with `data_source="both"`).

skore/src/skore/_sklearn/_base.py
cakedev0: Hum, good catch for `id`: I can easily avoid to use it so I'll do that.

We could avoid the decorator if we go for an API like `report.inspection._get_calls('permutation_importance')` , this would mean adding some code a bit everywhere in the reports (report, base accessor, inspection accessor). It would make the feature a bit less self-contained, but would avoid meta-programming.

jeromedockes:
I agree but we don't really care about what was called in this case, what we want to know is what is available to display to the user. if we have permutation importances on the train set we can store and show them, regardless of whether they are here because data_source='both' or data_source='train' was used, no?

cakedev0:
The problem is that the format in which we store things is internal and answers the needs described [here](https://github.com/probabl-ai/skore/discussions/2587#discussioncomment-16477740) (no duplication, not artificial cache misses).

And the hub/projects want to rely on the public API, not some internal format.

Unless we decide to kinda double the API surface:
- the public API for users
- the state API for projects: but I'm afraid that this API will suffer from the double purpose issue too (be a serialization format vs expose results => typically for serialization you don't need to expose any result at the CV report level)

skore/src/skore/_sklearn/_base.py
cakedev0: Note: mutating some report's properties might break / cause bad behaviors to projects (see #2742)

So instead of recording that in attribute `f"__calls_{id(func)}"` I could record that in the cache with some special cache name.

cakedev0:
Re-marking this PR as draft, as after offline discussion with @jeromedockes it appears a bit less likely we want to move forward with it. Still, we might merge it just as a temporary way to avoid cache inspection in `skore-hub-project`, but let's decide that first before reviewing this PR in details.

cakedev0:
Closing this PR:
- not the validated design
- not really useful after migration to mono-package (planned for soon)

Issue #2738, opened 2026-04-09
UI details on cross-validation reports


GaelVaroquaux:
The new displays of the reports are really nice. Two aspects that could be improved:

* Too much vertical space: on a standard laptop, having turned of most of the visual distraction of vscode, the report still takes most of my screen, though it does give me a huge density of information (by default a table of 4 x 2 numbers, with associated row and column names)

<img width="1920" height="1200" alt="Image" src="https://github.com/user-attachments/assets/9607d322-c71c-4707-84de-0a18fd0ce4c2" />

* We should add a small link that points individuals to skore hub (it could be just a logo with a hover that details what the logo means)

auguste-probabl:
Also increase the contrast on the blue color, it's barely readable

PR #2739, opened 2026-04-09
UX: more compact CrossValidationReport display

GaelVaroquaux:
Partially addresses  https://github.com/probabl-ai/skore/issues/2738

Makes a more compact display for the CrossValidationReport:

## Now

image

## Before

image

I implemented this only for the CrossValidationReport. A similar change should be done for the EstimatorReport (and maybe for other objects).

Maybe someone should take over this PR and apply the same recipe to the other objects. I won't have time to work more on it. I've implemented it as a way to show simple changes that could make the UX tighter

### DIFF
--- a/skore/src/skore/_sklearn/_cross_validation/report.py
+++ b/skore/src/skore/_sklearn/_cross_validation/report.py
@@ -526,11 +526,14 @@ class CrossValidationReport(_BaseReport, DirNamesMixin):
         Used by :meth:`_repr_html_` and by :class:`~skore.ComparisonReport` to embed
         one report's views in the comparison HTML repr.
         """
-        metrics_html = (
-            self.metrics.summarize(data_source="test")
-            .frame(aggregate=("mean", "std"), favorability=False)
-            ._repr_html_()
+        metrics = self.metrics.summarize(data_source="test").frame(
+            aggregate=("mean", "std"), favorability=False
         )
+        # Transform a bit the dataframe to make it more compact
+        columns = [col for _, col in metrics.columns.tolist()]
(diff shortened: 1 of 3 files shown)


jeromedockes:
thanks. I find it problematic that the buttons move when we click them; if we want them on one row I would suggest having tabs instead of dropdowns

auguste-probabl:
Contrast on the blue is too low, at least in dark mode

jeromedockes:
>  Contrast on the blue is too low, at least in dark mode

that is true but not introduced by this PR

glemaitre:
I think that the suggestion with tabs is great in addition to what @GaelVaroquaux is proposing.

I think that it was kind of the expected behaviour by folding the section. Knowing that we don't intend the user to unfold everything to put on section into perspective to another, then tabs are completely OK.

@GaelVaroquaux I can take over if you want because I think that the tabs will slightly change what you did (in particular for the documentation) and it would be easier to have it in a single PR.

GaelVaroquaux:
> @GaelVaroquaux I can take over if you want because I think that the tabs will slightly change what you did (in particular for the documentation) and it would be easier to have it in a single PR.

Yes please. But I beg people working on UX to take in consideration the fact that a data-scientists' screen is cramped and that UX should keep in mind screen real estate. Whether it is in a notebook or in a VScode like environment, big displays do not work well

GaelVaroquaux:
> thanks. I find it problematic that the buttons move when we click them; if we want them on one row I would suggest having tabs instead of dropdowns

Sure, I agree. But the vertical space is even more problematic

PR #2736, opened 2026-04-09
feat: Add custom metrics

auguste-probabl:
#### Change description

This PR makes it possible to register a custom metric in an `EstimatorReport`, so that it's shown along with the other default metrics in `summarize`:

```python
from sklearn.datasets import load_breast_cancer
from sklearn.linear_model import LogisticRegression
from sklearn.metrics import make_scorer, mean_absolute_error
from skore import evaluate
X, y = load_breast_cancer(return_X_y=True)
classifier = LogisticRegression(max_iter=10_000)
report = evaluate(classifier, X, y, splitter=0.2, pos_label=1)
report.metrics.register(
    make_scorer(mean_absolute_error, response_method="predict")
)
report.metrics.summarize().frame()

                     LogisticRegression
Metric
Accuracy                       0.947368
Precision                      0.984127
Recall                         0.925373
ROC AUC                        0.993649
Brier score                    0.036154
Fit time (s)                   0.324200
Predict time (s)               0.000323
Mean Absolute Error            0.052632
```

Closes https://github.com/probabl-ai/skore/issues/2061
Closes https://github.com/probabl-ai/skore/issues/2203

Related:
- https://github.com/probabl-ai/skore/issues/2120
- https://github.com/probabl-ai/skore/pull/2277
- https://github.com/probabl-ai/skore/issues/2001

#### Contribution checklist

- [x] Unit tests were added or updated (if necessary)
- [x] Documentation was added or updated (if necessary)
- [ ] The documentation builds and renders properly (if it does, our bot will add a comment linking
to a preview of the documentation to review it visually)
- [x] A new changelog entry was added to CHANGELOG.rst (if necessary; typically, if your change
requires updating tests)

#### AI usage disclosure

AI tools were involved for:
- [x] Code generation (e.g., when writing an implementation or fixing a bug)
- [x] Test/benchmark generation
- [x] Documentation (including examples)
- [x] Research and understanding

I had an LLM write down specs in docs/design, write some tests based on those specs, and some of the implementation.

PR #2741, opened 2026-04-10
feat(skore): Helpers for serde (get_state/from_state)

cakedev0:
#### Change description

This PR adds an explicit serde path for `EstimatorReport` and
`CrossValidationReport` with `report.get_state()` / `cls.from_state(state)`.

**Goals:**
- A first step towards fixing #2742 
- Make report serialization more robust across skore versions than plain pickling.
  Pickling a report depends on the exact internal object layout, while the new state-based
  API is meant to preserve the information needed to rebuild a report.
- Make the serialization smarter, e.g. CV report `get_state` pops data from sub-reports states
  and `from_state` rebuilds it using `split_indices`.
- Ease future smarter serialization on the projets side, with this kind of flow:

```python 
state = report.get_state()
data = state.pop('data')
state['data_uri'] = save_data_in_data_store(report.id, data)
# `save_data_in_data_store` ensures deduplication of datasets and returns
# some kind of ID/URI to load back the data
save_pickle(state)

# later:
state = load_pickle(state)
state['data'] = load_data_from_data_store(state.pop('data_uri'))
report = cls.from_state(state)
```

**Detailed changes:**
- adds `get_state()` / `from_state()` for estimator and CV reports.
- stores report metadata in a dedicated `_metadata` dict and uses it to expose
  the report `id`;
- moves prediction responses and prediction timings into a dedicated
  `._predictions` cache, as this is considered part of the "core state" on which
  we want to ensure BW compatibility.
- updates the predict-time metric and related tests to reflect that cache split;
- adds round-trip tests to check that `from_state()` bypasses `__init__`,
  restores the relevant state, and still allows new metrics to be computed.

#### Contribution checklist

- [x] Unit tests were added or updated. ~~TODO: test CV from state with data-ops~~ done
requires updating tests)

#### AI usage disclosure

AI tools were involved for:
- [X] Code generation (implem. of CV report's `from_state` only)
- [X] Test/benchmark generation
- [X] Documentation (including examples)


skore/src/skore/_sklearn/_estimator/report.py
thomass-dev, line 1: Please add a "version" tag to your serialization format, that can be very useful when you will modify the two methods.

Also, have you consider using JSONSchema or Pydantic in `from_state`?

cakedev0: I've added serialization version format, good idea indeed.

I don't think I need such validation tools: I don't expect `state` will be tempered with. For now, the logic feels more like skore-related business logic for which you don't want to introduce additional dependencies/complexity.

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -257,6 +260,89 @@ def __init__(
...
+        report._test_data = data["test_data"]
+        report._cache = Cache()
+        report._cache.update(state["cache"])
+        report._cache.update(
+            {
+                make_cache_key(data_source, name): val
+                for (data_source, name), val in state["predictions"].items()
+            }
+        )
+        report._metric_registry = MetricRegistry(report)
```
cakedev0, line 344: For now, metrics registry is immutable so it's not in the state and it's rebuilt here. But once #2736 is merged, I'll have include it in the state. See comment https://github.com/probabl-ai/skore/pull/2736#discussion_r3085194345

jeromedockes (approved):
very nice, thanks @cakedev0 !

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -257,6 +260,89 @@ def __init__(
...
+            "estimator": self._estimator,
+            "data": {
+                "train_data": self._train_data,
+                "test_data": self._test_data,
+            },
+            "predictions": predictions,
+            # ---------- RESULTS ------------
+            # this part is less structured and not crucial for reconstructing a report
+            # so we won't try ensuring backward compatibility.
+            "cache": cached_results,
```
cakedev0: Maybe I should nest this key in a `"optional"` key? Those that the users of this API (`skore-*-project`s for now) can just save `state.pop("optional")` in a separate pickle for which they allow `pickle.loads` to fail when loading back a saved report (as they known `from_state` don't require what's inside this)...

For now it's just the cache, but maybe we'll want to add the metrics registry here? And maybe some diagnostic-related cache-like object? ...
That seems more future-proofed, as ideally we want to avoid changing top-level keys of the state (to allow backends to split some parts and serialize them individually, like `"data"`)

(naming: `"optional"`? `"optional_state"`? `"optional_parts"`? Another word instead of "optional"?)

jeromedockes: optional seems good. could also be non-critical.

> we'll want to add the metrics registry here

i'm not sure if that would be considered optional :thinking: we have to define clearly what we mean by optional

jeromedockes (dismissed):

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Yes, indeed metrics registry is a bit in-between I'd say. We'll see later ^^

jeromedockes (approved):

Issue #2742, opened 2026-04-10
Local and hub projects make the assumption that reports are byte-immutable


cakedev0:
Both `skore-local-project` and `skore-hub-project` identify persisted reports from the bytes of a `joblib` pickle of the full report object.

Both backends try to make this stable by clearing `._cache` before pickling (which btw forces them to access this private attribute to restore the cache after pickling). However, this still assumes that a report is otherwise byte-stable once constructed. **That assumption is too strong** and not known by skore-lib developers for now, typically, we recently introduced other transient state than `._cache`:
- `report.cache_predictions()` relies on a cached property: `._can_skip_predict`.
- `report.diagnose()` populates `._issues_cache`. cc @GaetandeCast 
- the metrics registry populates (or will populate?) `._metric_registry` dynamically. cc @auguste-probabl  

### Proposed direction

The fix should happen on the `hub` / `local` project side, not by trying to make the live Python object fully byte-immutable.

I think that `skore-local-project` could rely on `report.id` instead of the pickle hash. But I haven't looked into it too much yet.

For the `skore-hub-project`, the most promising direction is to persist a stable report state based on `get_state()`/`from_state()` (introduced in PR #2741). That would let `skore-hub-project` serialize a narrower, intentional representation of the report instead of the full object pickle.

cakedev0:
cc @thomass-dev

thomass-dev:
Please note that with https://github.com/probabl-ai/skore/pull/2114, we will associate a report with its id `report.id` in `skore-hub-project`.

thomass-dev:
> I think that skore-local-project could rely on report.id instead of the pickle hash.

IMO, this is the good way 👍🏼 .

Please note that it's not related to how you serialize a report: it's two different issue (!= https://github.com/probabl-ai/skore/pull/2741).
Regardless of your serialization method, you will need to identify the report to avoid duplicating it in storage.

cakedev0:
> it's not related to how you serialize a report

If we use the serialized object checksum, it is related right? But yes, if we use the `report.id` it's not.

thomass-dev:
>  If we use the serialized object checksum, it is related right? But yes, if we use the report.id it's not.

Yes.
My meaning is that #2741 is a different issue. You will have the same need, regardless of the serialization: how to identify your report to avoid duplication. IMO, use `report.id` in all backend solves this issue perfectly.

PR #2744, opened 2026-04-10
fix(skore-local-project): Use id to identify report artifact

cakedev0:
Fixes issue #2742 on the `skore-local-project`'s side.

- update lockfiles
- use `report.id` as the persisted artifact identifier in `skore-local-project` / stop deriving local report identity from pickle bytes
- simplify pickling by keeping caches in the serialized payload

### DIFF
--- a/skore-local-project/src/skore_local_project/project.py
+++ b/skore-local-project/src/skore_local_project/project.py
@@ -170,23 +170,6 @@ class Project:
         """The workspace of the project."""
         return self.__workspace
 
-    @staticmethod
-    def pickle(report: EstimatorReport | CrossValidationReport) -> tuple[str, bytes]:
-        """
-        Pickle ``report``, return the bytes and the corresponding artifact id.
-
-        Notes
-        -----
-        Artifact identity is derived from the report id, not from the serialized bytes.
-        """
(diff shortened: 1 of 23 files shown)


skore-local-project/src/skore_local_project/project.py
```diff
@@ -235,15 +224,15 @@ def put(self, key: str, report: EstimatorReport | CrossValidationReport) -> None
                 f"ort` (found '{type(report)}')"
             )
 
-        pickle_hash, pickle_bytes = Project.pickle(report)
+        artifact_id, pickle_bytes = Project.pickle(report)
```
thomass-dev: I don't think the `Project.pickle` function is needed anymore

```python
with io.BytesIO() as stream:
    joblib.dump(report, stream)
    pickle_id, pickle_bytes = str(report.id), stream.getvalue()
```

thomass-dev (changes requested):

skore-local-project/src/skore_local_project/project.py
cakedev0: Indeed, removed it, and refactored some tests.

thomass-dev (approved):
Good improvement, thanks!

PR #2747, opened 2026-04-11
feat(skore): Support estimator/X as a dict in skore.evaluate

glemaitre:
Add support for dictionary input to be able to set the name when doing a comparison.


skore/src/skore/_sklearn/evaluate.py
```diff
@@ -14,14 +15,12 @@
...
     from skore._sklearn.types import SKLearnCrossValidator
 
 
 def evaluate(
-    estimator: BaseEstimator | list[BaseEstimator],
-    X: ArrayLike | list[ArrayLike | None] | None = None,
+    estimator: BaseEstimator | list[BaseEstimator] | dict[str, BaseEstimator],
+    X: ArrayLike | list[ArrayLike | None] | dict[str, ArrayLike | None] | None = None,
     y: ArrayLike | None = None,
     data: dict | None = None,
```
GaetandeCast, line 25: Unrelated to this PR but the newly added `data` argument is not documented in the docstring. It's actually also absent in both `EstimatorReport` and `CrossValidationReport`. Is it on purpose?

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -14,14 +15,12 @@
...
-    from numpy.typing import ArrayLike
-
     from skore._sklearn.types import SKLearnCrossValidator
 
 
 def evaluate(
-    estimator: BaseEstimator | list[BaseEstimator],
-    X: ArrayLike | list[ArrayLike | None] | None = None,
+    estimator: BaseEstimator | list[BaseEstimator] | dict[str, BaseEstimator],
+    X: ArrayLike | list[ArrayLike | None] | dict[str, ArrayLike | None] | None = None,
```
GaetandeCast:
```suggestion
    X: ArrayLike | list[ArrayLike] | dict[str, ArrayLike] | None = None,
```
No need to allow `X=[None, None]` and `X={"m1": None, "m2": None}`, it adds clutter. When comparing prefit estimators, user can just use `X=None`.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -81,7 +84,13 @@ def evaluate(
     Raises
```
GaetandeCast: Should we keep this section? I feel like we don't really have it in other places in skore and it's more burden to maintain. Plus I don't think it brings much.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -131,25 +187,26 @@ def evaluate(
             n_jobs=n_jobs,
         )
 
-    if isinstance(X, list):
-        raise TypeError("X must be a single array-like when estimator is not a list.")
-
```
GaetandeCast, line 136: Why remove this ? It protects from 
```python
report = evaluate(LogisticRegression(), X=[X, X], y=y, splitter=5)
```
which breaks.

```suggestion
    if isinstance(X, (list, dict)):
        raise TypeError("X must be a single array-like when estimator is not a list or dict.")

```

GaetandeCast (commented):
Thanks for this improvement! Some minor remarks, otherwise looks good!

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -104,13 +113,51 @@ def evaluate(
...
+
+    Compare several named estimators:
+
+    >>> report = evaluate(
+    ...     {"m1": LogisticRegression(), "m2": LogisticRegression(C=2.0)},
+    ...     X,
+    ...     y,
+    ...     splitter=0.2,
+    ... )
+    >>> sorted(report.reports_)
```
auguste-probabl: Dicts are insertion-ordered so this will always pass:
```suggestion
    >>> report.reports_
```

skore/tests/unit/dispatch/test_evaluate.py
```diff
@@ -88,6 +88,84 @@ def test_evaluate_multiple_estimators_multiple_X(regression_data):
...
+    reports."""
+    X, y = regression_data
+    report = evaluate(
+        {"a": LinearRegression(), "b": LinearRegression()},
+        X,
+        y,
+        splitter=3,
+    )
+    assert isinstance(report, ComparisonReport)
+    assert all(isinstance(r, CrossValidationReport) for r in report.reports_.values())
```
auguste-probabl, line 127: Check `report._report_type`?

skore/src/skore/_sklearn/evaluate.py
glemaitre: Nop. I think it is missing.

skore/src/skore/_sklearn/evaluate.py
glemaitre: I can remove it. Indeed, it is part of the NumPy doc standards but you usually don't document all possible way that something raise. So I think it is better to remove it.

skore/src/skore/_sklearn/evaluate.py
glemaitre: Because an `ArrayLike` can actually be a list. So I don't think that we should be the one raising the error here but let the subsequent call to mention that the size is not ok.

skore/tests/unit/dispatch/test_evaluate.py
glemaitre: This is set internally to the `CrossValidationReport` or the `ComparisonReport`. So it is kind of private and I would let the test of the specific report type to check those.

skore/src/skore/_sklearn/evaluate.py
GaetandeCast: Okay that makes sense, thx.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -105,13 +104,56 @@ def evaluate(
...
+                if set(X) != set(names):
+                    raise ValueError(
+                        "When estimator and X are both dicts, they must have the "
+                        f"same keys; got estimator keys {sorted(names)!r}"
+                        f" and X keys {sorted(X)!r}."
+                    )
+
+                Xs = [X[name] for name in names]
+            else:
+                Xs = [X] * len(estimators)  # type: ignore
```
glemaitre: The type ignore defeats the purpose of typing here.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -122,27 +164,27 @@ def evaluate(
...
-                reports,
-            ),
-            n_jobs=n_jobs,
-        )
 
-    if isinstance(X, list):
-        raise TypeError("X must be a single array-like when estimator is not a list.")
+        if names is not None:
+            reports = dict(zip(names, reports, strict=True))  # type: ignore
+        return ComparisonReport(reports, n_jobs=n_jobs)  # type: ignore
```
glemaitre: can cast instead of ignoring the type

skore/src/skore/_sklearn/evaluate.py
auguste-probabl: Working on this PR makes me think we should stop using typing.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -38,14 +37,19 @@ def evaluate(
...
-        different preprocessing pipelines.
+    estimator : estimator object, list of estimators, or dict of estimators
+        A scikit-learn compatible estimator; a list of estimators to compare; or a
+        mapping of names to estimators.
+
+    X : array-like, list of array-like, dict of array-like, or None
+        Feature matrix. When ``estimator`` is a list, ``X`` can be a list of
+        feature matrices (one per estimator) to compare models with different
+        preprocessing pipelines. When ``estimator`` is a dict, ``X`` can be a
+        dict with the **same keys** mapping each name to its feature matrix, or
```
GaetandeCast:
```suggestion
        dict with the same keys mapping each name to its feature matrix, or
```
I don't think we need the bold emphasis here.

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -105,13 +104,56 @@ def evaluate(
...
+        else:  # isinstance(estimator, list)
+            names = None
+            estimators = estimator  # type: ignore
+            if isinstance(X, dict):
+                raise TypeError(
+                    "When estimator is a list, X cannot be a dict. Pass a single "
+                    "array-like broadcast to all estimators, or a list of "
+                    "array-like with one matrix per estimator."
+                )
+            Xs = X if isinstance(X, list) else [X] * len(estimators)  # type: ignore
```
GaetandeCast: Should we check that `len(X) == len(estimators)` here when `X` is a list? Or let `zip` fail?

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -122,27 +164,27 @@ def evaluate(
...
+
+    X = cast(ArrayLike | None, X)
 
     if isinstance(splitter, str):
         if splitter != "prefit":
             raise ValueError(
-                f"Invalid string value for splitter: {splitter!r}. "
-                "The only supported string value is 'prefit'."
+                f"Invalid string value for splitter: {splitter!r}."
+                " The only supported string value is 'prefit'."
```
GaetandeCast: Ultra nit but we usually have blank spaces at the end of strings not beginning right?

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -122,27 +164,27 @@ def evaluate(
...
             )
         return EstimatorReport(
-            estimator, X_test=X, y_test=y, test_data=data, pos_label=pos_label
+            estimator,
+            X_test=X,
+            y_test=y,
+            test_data=data,
+            pos_label=pos_label,
```
GaetandeCast, line 192:
```suggestion
            estimator, X_test=X, y_test=y, test_data=data, pos_label=pos_label
```
Nit again ^^

GaetandeCast (dismissed):
Very minor remarks, otherwise LGTM!

skore/src/skore/_sklearn/evaluate.py
auguste-probabl: The current way is better for reading diffs

skore/src/skore/_sklearn/evaluate.py
auguste-probabl: Let zip fail

GaetandeCast (approved):

glemaitre:
Thanks @auguste-probabl for finishing this one.

PR #2750, opened 2026-04-13
chore(skore-hub-project): Document some side-effects

cakedev0:
#### Change description

I added comments about surprising side effects of `payload.checksum` or `payload.model_dumps()` that look like small vanilla things but under-the-hood triggers all the computations and uploads.

### DIFF
--- a/skore-hub-project/src/skore_hub_project/report/report.py
+++ b/skore-hub-project/src/skore_hub_project/report/report.py
@@ -154,6 +154,7 @@ class ReportPayload(BaseModel, ABC, Generic[Report]):
             ):
                 payload = media_cls(project=self.project, report=self.report)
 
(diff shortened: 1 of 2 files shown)


skore-hub-project/src/skore_hub_project/report/report.py
```diff
@@ -154,6 +154,7 @@ def medias(self) -> list[Media[Report]]:
             ):
                 payload = media_cls(project=self.project, report=self.report)
 
+                # NOTE: Accessing `payload.checksum` lazily uploads the artifact.
```
thomass-dev, line 157: Note: this will be less implicit with https://github.com/probabl-ai/skore/pull/2114, where we call explicitly `upload`.
In a way, just like it is done in `metrics` with the explicit call of `compute`.

thomass-dev (approved):

PR #2751, opened 2026-04-13
feat: Add OvR confusion matrices

### LINKED ISSUE #2592: API: inconsistent behavior of `pos_label` across metrics

### LINKED ISSUE #2693: feat(skore): Allow computing OvR confusion matrix in multiclass classification
Since we allow to compute the ROC and PR curves in an OvR approach, we should also compute the confusion matrix with such approach. In this case, it means that we will build 2x2 matrices with threshold corresponding to the probability estimates of the given class.

I assume that we could have a parameter in `.frame` and `.plot` to control this behaviour.

GaetandeCast:
Closes #2693, closes #2592.

Adds an OvR confusion matrices in multiclass classification that support thresholding.
Two views now coexist for the confusion matrix: 
- the `predict`-based view that is always available and displayed when `threshold=None` in `plot` and `frame`
- the `predict_proba`/`decision_function` based view that is shown when `threshold=float` and is only available for classifier that implement one of the two method.

This PR also introduces a `label` parameter to `plot` and `frame` to select which class to consider as the positive one in the thresholded view. In the default `predict` view, in binary classification, it can be used to set the labels in the order: (not `label`, `label`). Ignored in multiclass. This finishes closing #2592.

For `ComparisonReport`, we raise a warning when not all estimators implement the same prediction score function, and disable the thresholded view in that case.


skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -149,67 +157,64 @@ def plot(
...
         *,
         normalize: Literal["true", "pred", "all"] | None = None,
-        threshold_value: ThresholdValue = "default",
+        threshold_value: float | None = None,
         subplot_by: Literal["split", "estimator", "auto"] | None = "auto",
+        label: PositiveLabel,
     ) -> Figure:
-        """Matplotlib implementation of the `plot` method.
-
-        Parameters
```
GaetandeCast, line 183: removed to avoid maintaining a private duplicated docstring.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -421,83 +446,91 @@ def _compute_data_for_display(
...
+        cm_pred = np.zeros_like(cm, dtype=float)
+        np.divide(cm, col_sums, out=cm_pred, where=col_sums != 0)
+
+        total_sums = cm.sum(axis=(1, 2), keepdims=True)
+        cm_all = np.zeros_like(cm, dtype=float)
+        np.divide(cm, total_sums, out=cm_all, where=total_sums != 0)
+
+        data = {
+            "true_label": np.tile(np.repeat(labels, len(labels)), len(thresholds)),
+            "predicted_label": np.tile(np.tile(labels, len(labels)), len(thresholds)),
```
cakedev0, line 520: Reduces the memory footprint of those columns by a factor 23.5 according to my experiment ^^ (for string labels)

```suggestion
            "true_label": pd.Series(np.tile(np.repeat(labels, len(labels)), len(thresholds)), dtype="category"),
            "predicted_label": pd.Series(np.tile(np.tile(labels, len(labels)), len(thresholds)), dtype="category"),
```

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -421,83 +446,91 @@ def _compute_data_for_display(
...
-                "split": None,
-                "estimator": estimator_name,
-                "data_source": data_source,
-            }
-        )
-        disp = cls(
-            confusion_matrix=confusion_matrix,
-            display_labels=display_labels,
+        confusion_matrix_predict = cls._build_confusion_frame(
+            sklearn_confusion_matrix(y_true=y_true, y_pred=y_pred)[np.newaxis, ...],
```
cakedev0, line 450: Pass `labels=classes`, codex found this one

- [P1] Preserve absent classes when building the predict confusion matrix — skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py:449-450
  Pass `labels=classes` to `sklearn_confusion_matrix` here. Without it, any test set or CV fold that happens to contain only a subset of the fitted classes produces a smaller matrix (e.g. 1×1 for a binary fold with no positives), but `_build_confusion_frame` still expands against `estimator.classes_`, which raises `ValueError: All arrays must be of the same length` and drops missing classes from the display. This is a real regression for imbalanced datasets or non-stratified splits.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -149,67 +154,64 @@ def plot(
...
-        subplot_by: Literal["split", "estimator", "auto"] | None = "auto",
+        subplot_by : {"split", "estimator", "auto"} or None, default="auto"
             The variable to use for subplotting. If None, the confusion matrix will not
             be subplotted. If "auto", the variable will be automatically determined
             based on the report type.
 
+        label : int, float, bool, str or None, default=report pos_label
+            The class to consider as positive when using the thresholded view.
+            Required when `threshold_value` is not None. Ignored when `threshold_value`
+            is None.
```
cakedev0, line 171: It's not ignored for binary classif., it changes the plot.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -149,67 +154,64 @@ def plot(
...
+
         Returns
         -------
         matplotlib.figure.Figure
             Figure containing the confusion matrix.
         """
+        label = _check_label(self.labels, label, self.report_pos_label)
```
cakedev0, line 178: Let's also add a check:

```suggestion
        label = _check_label(self.labels, label, self.report_pos_label)
        if threshold_value == "all":
            raise ValueError("threshold_value='all' is not supported for the plot method")
```

As it's supported for frame, the error is easy to make.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
cakedev0: And for multi-class, let's actually implement OVR predict-based confusion matrix? (I think we initially said it can wait but in the end I think it's easy to do, sorry :sweat_smile: )

```python
cm = report.metrics.confusion_matrix()
df = cm.confusion_matrix_predict

label = 0

mapping = {True: label, False: f"not {label}"}

df.groupby([
    (df['true_label'] == label).map(mapping),
    (df['predicted_label'] == label).map(mapping),
])['count'].sum().reset_index()
```

But with normalized columns is a tad more complex.

... You know what, let's wait. I'm happy to do that in a follow-up PR, looks fun :stuck_out_tongue_winking_eye:

cakedev0: (not action needed here: we'll remove "Ignored when `threshold_value` is None." when implemented for the multi-class case)

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
cakedev0: Also once you do that, you can use `.cat.categories.to_list()` in the labels property (doesn't make a real difference, but aligns with other displays)

cakedev0 (changes requested):
Overall LGTM! 

Some small things, and one follow-up subject. But I think we can merge soon.

skore/tests/unit/displays/confusion_matrix/test_common.py
```diff
@@ -388,3 +400,14 @@ def test_confusion_matrix_thresholded_not_available_comparison(
...
+
+def test_missing_class_in_split(pyplot, binary_classification_data):
+    """Check that the confusion matrix is correct when a class is missing in a split."""
+    X = np.array([[0], [1], [0], [0]])
+    y = np.array([0, 1, 0, 0])
+    report = evaluate(LogisticRegression(), X, y, splitter=0.5)
+
+    display = report.metrics.confusion_matrix()
+    assert display.labels == [0, 1]
+    assert display.confusion_matrix_predict.shape == (2**2, 9)
```
cakedev0: I prefer to avoid testing things too precisely when it's not about what the test is testing:

```suggestion
    assert len(display.confusion_matrix_predict) == 2**2
```

cakedev0 (changes requested):
LGTM, one last minor thing, sorry :sweat_smile:

cakedev0 (approved):
LGTM

PR #2761, opened 2026-04-14
fix(skore): Don't pass pos_label in precision/recall when average is not "binary"

cakedev0:
#### Change description

As the title says. Without this fix, you get a warning with this snippet:

```python
from skore import evaluate
import numpy as np

X, y = make_classification()
y = np.array(['A', 'B'])[y]
report = evaluate(LogisticRegression(), X, y, splitter=0.2, pos_label="A")
report.metrics.precision(average="macro")
```

Which causes failures in `skore-hub-project` tests.

And it's also a bad user experience.


skore/src/skore/_sklearn/metrics.py
```diff
@@ -322,12 +321,11 @@ def available(report: EstimatorReport) -> bool:
...
-            average is None
-            and report._ml_task == "binary-classification"
-            and report.pos_label is not None
-        ):
-            average = "binary"
+        if report._ml_task == "binary-classification":
+            if average is None and report.pos_label is not None:
+                average = "binary"
+            else:
+                kwargs.setdefault("pos_label", None)
```
auguste-probabl: `kwargs.setdefault("pos_label")` is a shorthand for 
```
if "pos_label" not in kwargs:
    kwargs["pos_label"] = None
```
I would rather we just use that

cakedev0: Re-reading this code, I see that:
- it's buggy (`report.metrics.precision(average="binary")` raises)
- "pos_label" can't really be in kwargs already anyway.

Fixed in https://github.com/probabl-ai/skore/pull/2761/commits/2657249897efffa3d71f2b3174352a24ec1b0d0b

auguste-probabl (approved):
Thanks!

PR #2762, opened 2026-04-14
feat(skore-hub-project/artifact/pickle): Use `report.id` as checksum

thomass-dev:
Extract from https://github.com/probabl-ai/skore/pull/2114.

---

- Remove `serializer.py`
    Make `compute` and `checksum` two different functions in the `Artifact` class.
    It allows checksum to be independently calculated from the representation of the content to upload.
- Use `report.id` as checksum of the pickle artifact instead of `blake2b(joblib.dump(report))`
- Explicit and assert that an artifact is not null, computed and uploaded before being sent to the hub
- Improve readability in tests


skore-hub-project/src/skore_hub_project/artifact/pickle/pickle.py
```diff
@@ -21,45 +19,35 @@ class Pickle(Artifact):
...
-
-        Notes
-        -----
-        The report is pickled without its cache, to avoid salting the checksum.
-        """
-        reports = [self.report] + getattr(self.report, "estimator_reports_", [])
-        reports_with_cache = [
-            (report, report._cache) for report in reports if hasattr(report, "_cache")
-        ]
-        self.report.clear_cache()
```
thomass-dev, line 54: Leave this change in its own stacked PR.

thomass-dev: https://github.com/probabl-ai/skore/pull/2783

PR #2767, opened 2026-04-16
fix(skore): Inspect cache before checking score_func in metrics

cakedev0:
(not really a fix, but don't know how to call it)

#### Change description

The goal is to make this flow possible:

```Python
scorer = make_scorer(lambda y_true, y_pred: np.abs(y_true - y_pred).mean())
report.metrics.add(scorer)
report.metrics.summarize()

report3 = pickle.loads(pickle.dumps(report))
report3.metrics.summarize()  # currently raises, while all the results needed are in cache
```

For that, we just need to swap the cache check and the `if self.score_func is None` check.

#### Contribution checklist

#### AI usage: no

### DIFF
--- a/skore/tests/unit/reports/estimator/metrics/test_add.py
+++ b/skore/tests/unit/reports/estimator/metrics/test_add.py
@@ -642,7 +642,7 @@ class TestSerialization:
         """Test that if added metric is a lambda, it is lost when pickling."""
         report = binary_classification_report
 
-        scorer = make_scorer(lambda y_true, y_pred: (y_true - y_pred).abs().mean())
+        scorer = make_scorer(lambda y_true, y_pred: np.abs(y_true - y_pred).mean())
         report.metrics.add(scorer)
         assert report._metric_registry["<lambda>"].score_func is not None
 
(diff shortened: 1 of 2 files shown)


skore/tests/unit/reports/estimator/metrics/test_add.py
```diff
@@ -642,7 +642,7 @@ def test_serde_lambda(self, binary_classification_report):
         """Test that if added metric is a lambda, it is lost when pickling."""
         report = binary_classification_report
 
-        scorer = make_scorer(lambda y_true, y_pred: (y_true - y_pred).abs().mean())
+        scorer = make_scorer(lambda y_true, y_pred: np.abs(y_true - y_pred).mean())
```
auguste-probabl, line 645: Why is this necessary?

cakedev0: For my numpy version (2.4.3), `np.array([-1, 2]).abs()` crashes with `AttributeError: 'numpy.ndarray' object has no attribute 'abs'
`

auguste-probabl: Oh I see, the test passed before because the metric was never called. My bad

auguste-probabl (approved):
Thanks!

PR #2770, opened 2026-04-16
refactor(skore): Summary concatenation helper

cakedev0:
Context: I started this refactor while working on the "export results" feature.

#### Change description

Changes on `MetricsSummaryDisplay`:
- Added a `_concatenate` static method, to centralize concatenation logic.
- Change input from a dataframe to rows (will be convenient so I can reuse the summary logic in the "export results" feature). Also, it simplifies concatenation logic.
- Kept a `.data` property to avoid having to edit many places/tests. But maybe we can remove it later (let me know what you think).

+ Adapted code in metric accessors to those changes (simplifies it).


skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -1,11 +1,12 @@
-from typing import Literal, cast
+from typing import Any, Literal, cast
 
 import pandas as pd
 from matplotlib.figure import Figure
 
 from skore._sklearn._plot.base import DisplayMixin
 from skore._sklearn.types import Aggregate, ReportType
 from skore._utils._index import flatten_multi_index
+from skore._utils._metric_rows import rows_to_dataframe
```
cakedev0, line 9: Note: I think we could move functions from `_metric_rows` to here, that would make sense. But it will grow artificially the diff, so I don't do it in this PR.

auguste-probabl: sure!

skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -26,12 +39,29 @@ class MetricsSummaryDisplay(DisplayMixin):
 
     def __init__(
         self,
-        data: pd.DataFrame,
+        rows: list[dict],
```
auguste-probabl, line 42: I really like `TypedDict` to document structured dictionaries, could you use that and move the Parameters description there?

auguste-probabl (approved):
A small suggestion on typing/documentation but this looks great!

skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
cakedev0: I'll merge it first, then I'll do that in a small follow-up PR where I also move functions from `_metric_rows.py` in `metrics_summary_display.py`.

PR #2775, opened 2026-04-16
fix: Fix configuration ref in diagnose docs

GaetandeCast:
The reference to `skore.configuration` is broken in the docs of `.diagnose`
image


rouk1 (approved):
Thanks for the quick fix.

thomass-dev (approved):

PR #2778, opened 2026-04-16
fix(skore): Close plots to avoid triggering a user-facing warning

cakedev0:
#### Motivation

While playing with skore, and making plots after plots, I got this warning:

```
RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`). Consider using `matplotlib.pyplot.close()`.
  fig = plt.figure(figsize=figsize)
```

and I asked codex to fixed that.

#### AI usage: design & implementation


skore/src/skore/_sklearn/_plot/base.py
```diff
@@ -210,16 +210,23 @@ def wrapper(self, *args: Any, **kwargs: Any) -> Any:
...
+                return result
+            finally:
+                # Seaborn figure-level helpers go through pyplot and keep figures
+                # registered until they are explicitly closed. Deregister all figures
+                # created during this display call while keeping the returned Figure
+                # object usable by the caller.
+                for figure_num in set(plt.get_fignums()) - figures_before:
+                    plt.close(figure_num)
                 plt.rcParams.update(original_params)
-            return result
```
auguste-probabl, line 222: not sure about this removal

skore/src/skore/_sklearn/_plot/base.py
auguste-probabl: Update the docs

skore/tests/unit/displays/test_common.py
```diff
@@ -29,3 +29,25 @@ def test_display_protocol(pyplot, capsys, plot_func, estimator, dataset):
...
+
+    X, y = make_classification(random_state=42)
+    X_train, X_test, y_train, y_test = train_test_split(X, y, random_state=42)
+    report = EstimatorReport(
+        LogisticRegression(),
+        X_train=X_train,
+        y_train=y_train,
+        X_test=X_test,
+        y_test=y_test,
+    )
```
auguste-probabl: Use one of our fixtures

skore/tests/unit/displays/test_common.py
```diff
@@ -29,3 +29,25 @@ def test_display_protocol(pyplot, capsys, plot_func, estimator, dataset):
...
+    display = report.metrics.precision_recall()
+    figures_before = set(pyplot.get_fignums())
+
+    for idx in range(25):
+        fig = display.plot()
+        fig.savefig(tmp_path / f"precision-recall-{idx}.svg")
```
auguste-probabl: Save to a BytesIO instead

auguste-probabl: In fact, I don't think you need the savefig to test the behaviour

skore/tests/unit/displays/test_common.py
```diff
@@ -29,3 +29,25 @@ def test_display_protocol(pyplot, capsys, plot_func, estimator, dataset):
...
+
+    assert set(pyplot.get_fignums()) == figures_before
```
auguste-probabl, line 46: nit: There are no figures at the beginning so the following test is equivalent:

```suggestion
    assert pyplot.get_fignums() == []
```

auguste-probabl (changes requested):

skore/tests/unit/displays/test_common.py
cakedev0: Actually, when you run it with all the other tests, this is needed.

skore/tests/unit/displays/test_common.py
cakedev0: No indeed I don't think so, removed it.

skore/src/skore/_sklearn/_plot/base.py
cakedev0: Good catch thx, done.

skore/src/skore/_sklearn/_plot/base.py
cakedev0: It's indeed not needed, reverted that.

auguste-probabl (approved):

PR #2787, opened 2026-04-17
fix: Make diagnostics work with custom metrics

GaetandeCast:
Currently, `.diagnose`, and by extension the report html reprs, crash when a report has a custom metric. This PR introduces a non regression test and a fix.

### DIFF
--- a/skore/tests/unit/reports/estimator/test_diagnose.py
+++ b/skore/tests/unit/reports/estimator/test_diagnose.py
@@ -1,11 +1,9 @@
 from pathlib import Path
 from urllib.parse import urlparse
 
-import numpy as np
 import pytest
 from sklearn.dummy import DummyClassifier, DummyRegressor
 from sklearn.linear_model import LinearRegression
-from sklearn.metrics import confusion_matrix, make_scorer
 from sklearn.tree import DecisionTreeRegressor
 
 from skore import Check, EstimatorReport, configuration, evaluate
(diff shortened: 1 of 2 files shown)


skore/tests/unit/reports/estimator/test_diagnose.py
```diff
@@ -233,3 +235,27 @@ class InvalidCheck:
...
     with pytest.raises(ValueError, match="does not implement the Check protocol."):
         report.add_checks([InvalidCheck()])
+
+
+def test_diagnose_custom_metric(binary_classification_data):
+    """Check that diagnose works with custom metrics in the report."""
+    X, y = binary_classification_data
+    report = evaluate(DummyClassifier(), X, y)
+
+    def custom_score(y, y_pred, neg_label, pos_label):
```
cakedev0: Let's use a much simpler scorer, or even just something like `report.metrics.add("f1")`

cakedev0 (changes requested):
LGTM, but let's maybe simplify the test?

skore/tests/unit/reports/estimator/test_diagnose.py
GaetandeCast: Yes, will do.

cakedev0: note: you need to give pos_label to evaluate to use f1

GaetandeCast: Done. The string changed to "7/7" because I had to choose a `pos_label` for `"f1"`.

cakedev0 (approved):
LGTM

PR #2795, opened 2026-04-20
chore(skore): Add report_type in state metadata

cakedev0:
Everything is in the title ^^

**Motivation:** Needed/useful when you have a `state` and you need to pick between `EstimatorReport.from_state` and `CrossValidationReport.from_state` to deserialize it :sweat_smile:


thomass-dev (changes requested):
Please add the corresponding check in each `from_state`.

thomass-dev (approved):

PR #2811, opened 2026-04-21
chore(skore): Hide the cv progress bar when splitter is float in evaluate

GaetandeCast:
Title.

PR #2812, opened 2026-04-22
chore(skore): Hide the cv progress bar when splitter does a single split in evaluate

GaetandeCast:
More general version of what was done in #2811 : hide the cv bar progress when the splitter only performs a single train/test split, not only for `splitter: float`. E.g. it also hides the progress bar for a custom instance of `skore.TrainTestSplit`

### DIFF
--- a/skore/src/skore/_sklearn/evaluate.py
+++ b/skore/src/skore/_sklearn/evaluate.py
@@ -147,6 +147,8 @@ def evaluate(
 
     if isinstance(splitter, float):
         splitter = TrainTestSplit(test_size=splitter)
+
+    if hasattr(splitter, "get_n_splits") and splitter.get_n_splits(X, y) == 1:
         with configuration(show_progress=False):
             report = CrossValidationReport(
                 estimator,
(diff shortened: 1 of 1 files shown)


skore/src/skore/_sklearn/evaluate.py
```diff
@@ -149,19 +149,22 @@ def evaluate(
         # It's easier to make a 1-split CrossValidationReport
         # and extract an EstimatorReport from it,
         # than to make an EstimatorReport from scratch
+        splitter = TrainTestSplit(test_size=splitter)
+
+    if hasattr(splitter, "get_n_splits") and splitter.get_n_splits(X, y) == 1:
```
auguste-probabl:
```suggestion
        splitter = TrainTestSplit(test_size=splitter)

    if hasattr(splitter, "get_n_splits") and splitter.get_n_splits(X, y) == 1:
        # It's easier to make a 1-split CrossValidationReport
        # and extract an EstimatorReport from it,
        # than to make an EstimatorReport from scratch
```

auguste-probabl (approved):

PR #2813, opened 2026-04-22
fix(skore): Clarify metrics summary column names

cakedev0:
### Motivation

I noticed that `report.metrics.summarize().frame()` returns a column named "Label / Average" that actually never contains an average, when an average is given, it is displayed in an additional "Average" column. For instance, when you do:

```python
from skore import evaluate
from sklearn.metrics import precision_score

X, y = make_classification(random_state=0)

report = evaluate(LogisticRegression(), X, y, splitter=0.2)
report.metrics.add(precision_score, average="macro", name="Precision (Macro)")
report.metrics.summarize().frame()
```

You get:

image

After this PR you get:

image

### Change description

This PR updates `report.metrics.summarize().frame()` so metric metadata is shown in separate, explicit columns instead of being merged into a combined grouping column.

The resulting frame now uses:
- `Label` for per-class classification rows.
- `Output` for per-output regression rows.
- `Average` for averaged classification or multioutput rows.

Rows that do not use one of these metadata fields leave that level empty. This avoids ambiguous names such as `Label / Average` while keeping mixed summaries readable when per-label and averaged metrics are displayed together.

Changes are limited to the metrics summary frame formatting and the corresponding expected column names in tests.

#### AI usage: implementation and tests

Spec driven codex session + human code simplification pass.


skore/src/skore/_sklearn/_comparison/metrics_accessor.py
auguste-probabl, line 1: This predates your PR, but I would rather have one "Label" column and one "Average" column than one "Label / Average" column. If we're changing the logic I would like to take the opportunity to change this too.

auguste-probabl (changes requested):

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
glemaitre: I agree with @auguste-probabl I think that today I would prefer to have two distinct columns.

cakedev0: Ok, well that's simpler to do ^^

I did that.

auguste-probabl (approved):

PR #2814, opened 2026-04-22
feat(skore/add)!: Support scorers, not metrics (in the sklearn sense)

### LINKED ISSUE #2768: feat: Support scorer callables (`(estimator, X, y) -> float`) in `report.metrics.add`
This is the most flexible way to define a score. Unlike for `(y_true, y_pred) -> score` callables, we can't straightforwardly use the cached predictions, so let's just skip that part of the logic.

auguste-probabl:
#### Change description

In an effort to stick to the sklearn API, `EstimatorReport.metrics.add` now only supports *scorers* (`est, X, y -> score`) as opposed to *metrics* (`y_true, y_pred -> score`), following sklearn's terminology. Metrics are still supported indirectly via `sklearn.metrics.make_scorer`.

Closes #2768 

#### Contribution checklist

requires updating tests)

#### AI usage disclosure

AI tools were involved for:
- [x] Review


examples/model_evaluation/plot_custom_metrics.py
```diff
@@ -67,44 +73,19 @@ def specificity(y_true, y_pred):
 
 # %%
 # Passing extra keyword arguments
-# ================================
+# ===============================
 #
 # If your metric needs extra data at scoring time (e.g. sample-level amounts,
-# a cost matrix, ...), pass them as keyword arguments to
-# :meth:`~skore.EstimatorReport.metrics.add`. They will be forwarded to the
+# a cost matrix, ...), they can passed as keyword arguments to
```
GaetandeCast:
```suggestion
# a cost matrix, ...), they can be passed as keyword arguments to
```

examples/model_evaluation/plot_custom_metrics.py
```diff
@@ -117,34 +98,10 @@ def misclassification_cost(y_true, y_pred, cost_fp, cost_fn):
...
+report.metrics.summarize(metric=["specificity", "f2"]).frame()
 
 # %%
 # Selecting ``data_source="both"`` lets you compare train vs. test in one call:
 
 # %%
 report.metrics.summarize(metric=["specificity", "f2"], data_source="both").frame()
-
-# %%
-# Using a different response method
```
GaetandeCast, line 131: Why remove this section? I think it's interesting to show the user how to use the `response_method` of `make_scorer`.

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
```diff
@@ -142,15 +141,13 @@ def add(
         Parameters
         ----------
         metric : str, sklearn scorer, or callable
-            The metric to add.
+            The metric to add. See :meth:`EstimatorReport.metrics.add` for a
+            description of accepted inputs.
```
GaetandeCast: Can't we copy the docstring of the `EstimatorReport.metrics.add` instead ? I know it means maintaining a potentially evolving docstring in multiple places but as a IDE user, not having the info I look for with the docstring preview mouse-over feels bad.

skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -124,15 +123,13 @@ def add(
         Parameters
         ----------
         metric : str, sklearn scorer, or callable
-            The metric to add.
+            The metric to add. See :meth:`EstimatorReport.metrics.add` for a
+            description of accepted inputs.
```
GaetandeCast: Same

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -230,15 +229,20 @@ class labels to floats (e.g. ``{0: 0.9, 1: 0.85}``), in which case
...
+                    greater_is_better=greater_is_better,
+                    kwargs=kwargs,
+                )
             )
-        )
+        except MissingKwargsError as e:
+            args_msg = ", ".join(f"{kw}=..." for kw in e.missing_kwargs)
+            raise ValueError(
+                f"{e.msg} Pass those kwargs to add: add({e.metric}, {args_msg})"
+            ) from e
```
```suggestion
        except MissingKwargsError as error:
            args_msg = ", ".join(f"{kwarg}=..." for kwarg in error.missing_kwargs)
            raise ValueError(
                f"{error.msg} Pass those kwargs to add: add({error.metric}, {args_msg})"
            ) from error
```
Nit but it's easier to understand code when using full words here.

skore/src/skore/_sklearn/types.py
```diff
@@ -49,7 +50,10 @@ class SKLearnScorer(Protocol):
     _kwargs: dict[str, Any]
 
 
-MetricLike = str | Callable | SKLearnScorer
+MetricCallable = Callable[[ArrayLike, ArrayLike], float]
+ScorerCallable = Callable[[BaseEstimator, ArrayLike, ArrayLike], float]
+
+MetricLike = str | MetricCallable | ScorerCallable | SKLearnScorer
```
GaetandeCast: If `add` now only accepts scorers, why do we have `MetricCallable` in `MetricLike`? Is it for our internal metrics?

GaetandeCast: It's weird because right now the signature of `add` suggests using `MetricCallable` when not supported.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -46,17 +77,20 @@ class Metric:
     verbose_name : str
         Display name shown in reports (e.g. ``"Accuracy"``).
 
+    response_method : str, list of str, or None, default="predict"
+        Estimator method to get predictions.
+
     greater_is_better : bool or None
         Whether a higher value is better.
```
GaetandeCast, line 92: Nit af but this can stay above `response_method` to respect the order the arguments appear in, in `__init__`.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -35,6 +41,31 @@
 }
 
 
+class FunctionKind(Enum):
```
GaetandeCast, line 55: Isn't this redundant with just inspecting the type of `function`?

skore/src/skore/_sklearn/metrics.py
```diff
@@ -139,29 +178,36 @@ def __call__(
...
 
-        score = self.score_func(y_true, y_pred, **call_kwargs)
+            _, y_true = report._get_data_and_y_true(data_source=data_source)
+            y_pred = report._get_predictions(
+                data_source=data_source,
+                response_method=self.response_method,
+                pos_label=call_kwargs.get("pos_label", None),
+            )
+
+            score = self.function(y_true, y_pred, **call_kwargs)  # type: ignore
```
GaetandeCast: Is there no way to infere the type of `score` to please mypy?

skore/tests/unit/reports/estimator/metrics/test_add.py
```diff
@@ -320,31 +329,19 @@ def test_on_report_without_train_data(
...
-        report = EstimatorReport(
-            estimator,
-            X_train=X_train,
-            y_train=y_train,
-            X_test=X_test,
-            y_test=y_test,
-        )
+    def test_passing_metric_callable(self, binary_classification_report):
+        """Test that adding a metric callable ``(y_true, y_pred) -> score`` succeeds
+        (we don't introspect on the callable), but ``summarize`` will fail."""
```
GaetandeCast: So what's the reason for not failing on adding a non scorer callable?

skore/tests/unit/reports/estimator/metrics/test_add.py
```diff
@@ -320,31 +329,19 @@ def test_on_report_without_train_data(
...
+        def score(y_true, y_pred):
+            return 0
 
-        report.metrics.add(incompatible_scorer)
+        report.metrics.add(score)
 
-        err_msg = "SVC has none of the following attributes: predict_proba."
-        with pytest.raises(AttributeError, match=err_msg):
-            report.metrics.summarize(metric="accuracy_score")
+        err_msg = "takes 2 positional arguments but 3 were given"
```
GaetandeCast: This is a cryptic message when we could have failed earlier in an explicit way.

GaetandeCast (commented):
Great work! Definetly a UX improvement.
A few suggestions regarding documentation and a question on why we allow `(y_true, y_pred)` signatures if we don't support them.

skore/src/skore/_sklearn/types.py
auguste-probabl: You're right, this is wrong.

skore/tests/unit/reports/estimator/metrics/test_add.py
auguste-probabl: We would need to apply heuristics to the arguments of the callable, it seems like more trouble than to just let it fail. This is what happens when you pass a metric callable to sklearn:
```python
from sklearn import datasets, linear_model
from sklearn.model_selection import cross_validate
X = diabetes.data[:150]
y = diabetes.target[:150]
lasso = linear_model.Lasso()
scoring = lambda y_true, y_pred: 0
cv_results = cross_validate(lasso, X, y, cv=3, scoring=scoring)
# UserWarning: Scoring failed. The score on this train-test partition for these parameters will be set to nan. Details:
# Traceback (most recent call last):
#   File "/home/auguste/Desktop/work/skore/.venv/lib/python3.13/site-packages/sklearn/model_selection/_validation.py", line 916, in _score
#     scores = scorer(estimator, X_test, y_test, **score_params)
# TypeError: <lambda>() takes 2 positional arguments but 3 were given
```

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Mypy is not able to tell that self.function takes `y_true, y_pred`

skore/src/skore/_sklearn/metrics.py
auguste-probabl: To explain my strategy a bit more generally: I like to do all the validation/parsing upfront, so that the result is a well-formed input. Then, when we do actual business logic, we only deal with validated inputs. This indicator type is a way to express "We already did the validation step, and this is the result", so that when actually call the metric, we are no longer concerned with whether the inputs are well-formed.

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
auguste-probabl: Fair enough. Although long-term I would really appreciate some kind of auto-docstring system that makes maintenance less painful.

examples/model_evaluation/plot_custom_metrics.py
auguste-probabl: I removed it because it's a `make_scorer` input, and skore no longer asks for it. But sure, I can put it back.

skore/tests/unit/reports/estimator/metrics/test_add.py
auguste-probabl: If we go the heuristics route then sure. Otherwise, we could also catch the error and (again, heuristically) output a clearer error.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Found a way to make it work

skore/tests/unit/reports/estimator/metrics/test_add.py
auguste-probabl: Let's do some heuristics:
- If first argument starts with `y`
- If the callable has less than 3 positional arguments

examples/model_evaluation/plot_custom_metrics.py
```diff
@@ -1,9 +1,9 @@
 """
 .. _example_custom_metrics:
 
-===========================================================
-`EstimatorReport.metrics.add`: Adapt skore to your use-case
-===========================================================
+=======================================================
+Adapt skore to your use-case by adding your own metrics
```
glemaitre, line 5: I did a single change which was the title so I think that you can safely keep your entire example.

auguste-probabl: I changed the title because otherwise the rendering was weird, so your title works too

skore/tests/unit/reports/estimator/metrics/test_add.py
auguste-probabl: done

GaetandeCast (approved):
Thanks a lot!

PR #2815, opened 2026-04-22
feat(skore-hub-project/client/authentication): Use cookies in Jupyterlite

thomass-dev:
Part of https://github.com/probabl-ai/skore-hub/issues/1402.

In Jupyterlite, when no API key or bearer token are available in the session, use browser cookies to authenticate the hub requests.

Note that writing a test is complicated till `pyodide` cannot be installed from PyPi.

### DIFF
--- a/skore-hub-project/pyproject.toml
+++ b/skore-hub-project/pyproject.toml
@@ -121,6 +121,8 @@ module = ["skore.*"]
 ignore_missing_imports = true
 module = [
   "joblib.*",
+  "js.*",
+  "pyodide.*",
   "scipy.*",
   "sklearn.*",
 ]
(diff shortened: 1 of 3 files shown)


rouk1 (changes requested):
> [!NOTE]
> Disclaimer 🤖 review

The PR assigns `HTTPXClient.__init__` (i.e. `httpx.Client.__init__`) to a partialmethod that always injects `transport=__JupyterliteTransport()`.

Effects:

- After `skore_hub_project.client.client` is imported, any `httpx.Client(...)` in the same process (including unrelated libraries or user code) gets that transport, not only Skore’s client.
- That is surprising for a library and can break other uses of `httpx` under `Pyodide`.

Suggestion: avoid touching `httpx.Client` on the class. Override `Client.__init__` in `client.py` and pass `transport=__JupyterliteTransport()` only when `JUPYTERLITE` is true. `HUBClient` already inherits that `__init__`, so behavior stays the same for hub usage without global side effects.

rouk1 (approved):
Nice work !

Issue #2665, opened 2026-03-25
Add model checks on inspection tools


GaetandeCast:
Implement the following checks:

|Failure | Description | Typical signal |
|---|---|---|
|Coefficients of unscaled features | Linear coefficients not comparable in magnitude for a simple linear model fit directly on the data | No scaling on features used in linear model |
|MDI of high-cardinality features | Tree MDI inflated for high-cardinality features | Cardinality of feature close to n_samples |
Importance of correlated features | Correlated features give unstable importance | Feature correlation near 1 for important features
|Poor calibration | Probabilities not matching observed frequencies | Bad calibration curve, high Brier score/ECE

GaetandeCast:
Closing for now as Calibration curve is not fully implemented yet in skore. We can come back to it at a later date.

PR #2816, opened 2026-04-22
feat(skore): Add severity to diagnostic checks and show those that passed

GaetandeCast:
Towards #2665, only adds the first check in the list to limit the scope of this PR, as it requires adding a good amount of infrastructure.

This PR adds three tabs to the html repr of the `DiagnosticDisplay` : 
- `Issues` that contains all the previously added checks
- `Tips` that includes SKD006 added here, which flags linear models with unscaled input features
- `Passed` that contains all the checks that passed with no finding
- Checks that were ignored or that are not applicable are not shown.

The `Check` protocol now has an optional `severity` attribute that defaults to `"issue"` which let's us sort the checks in the right tab of the repr. 

AI tools were involved in everything except writing the new check. Generated code was carefully reviewed.


examples/technical_details/plot_custom_checks.py
```diff
@@ -57,6 +61,9 @@ def check_function(self, report):
 # :meth:`~skore.EstimatorReport.add_checks` accepts a list of ``Check`` instances,
 # and registers them. The next call to :meth:`~skore.EstimatorReport.diagnose` runs
 # any newly added checks on top of the built-in checks.
+#
+# We can then find the new check in the Tips tab of the diagnostic, along another tip
+# information us that the dataset is not standardized.
```
auguste-probabl: Typo?

skore/src/skore/_sklearn/_comparison/report.py
```diff
@@ -468,11 +468,11 @@ def pos_label(self) -> PositiveLabel | None:
 
     def _aggregate_checks(self) -> tuple[dict[str, dict], set[str]]:
```
auguste-probabl: Is there no type for diagnostic codes? It's confusing to see `str` everywhere.

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,229 @@
...
+        """All emitted tips, keyed by check code."""
+        return self._tips
+
+    @property
+    def passed(self) -> dict[str, dict]:
+        """Checks that ran, were applicable, and did not report anything.
+
+        Ignored checks are excluded.
+        """
+        return self._passed
```
auguste-probabl: Make the attributes public directly?

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,229 @@
...
-            f"{items_html}"
-            "</div>"
+        buckets = [
+            {
+                "label": "Issues",
+                "entries": [
+                    _format_finding_html(code, entry)
+                    for code, entry in self._issues.items()
+                ],
+                "empty_message": "No issues were detected in your report!",
```
auguste-probabl:
```suggestion
                "empty_message": "No issues were detected for your report.",
```
for consistency. It's also a bit weird to have a different verb for issues and tips, but less bad

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,229 @@
...
+                    f"- [{code}] {entry['title']}" for code, entry in entries.items()
+                )
+            else:
+                lines.extend(
+                    f"- {_format_finding_message(code, entry)}"
+                    for code, entry in entries.items()
+                )
+        if not (self._issues or self._tips or self._passed):
+            lines.append("- No checks were run on your report.")
+        return "\n".join(lines)
```
auguste-probabl: Shouldn't this come at the beginning?

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,229 @@
...
 class Check(Protocol):
     """Protocol for defining diagnostic checks.
 
     Each check wraps a callable that inspects a report. If the callable returns a
-    non-empty string, that text is recorded as an issue under :attr:`code` with the
-    given :attr:`title`. Checks are scoped to a single report kind via
-    :attr:`report_type` so they only run on matching reports.
+    non-empty string, that text is recorded as a finding under :attr:`code` with the
+    given :attr:`title` and :attr:`severity`. Checks are scoped to a single report
+    kind via :attr:`report_type` so they only run on matching reports.
```
auguste-probabl:
```suggestion
    type via :attr:`report_type` so they only run on matching reports.
```

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -149,7 +269,8 @@ def check_function(self, report: _BaseReport) -> str | None:
         Returns
         -------
         str or None
-            An explanation string, or None if the check did not find any issues.
+            An explanation string, or None if the check did not find any
+            finding.
```
auguste-probabl:
```suggestion
            An explanation string, or None if the check did not find anything.
```

skore/src/skore/_utils/repr/diagnostic_display.js
```diff
@@ -0,0 +1,20 @@
...
+    if (!wrapper || !host) {
+        return;
+    }
+
+    const template = document.getElementById(containerId + "-template");
+    if (!template) {
+        return;
+    }
+
+    const shadowRoot = host.attachShadow({ mode: "closed" });
```
auguste-probabl: Grr shadow DOM. This is an anti-pattern because it makes it impossible for downstream users to style the widget how they want with CSS. If the goal is to prevent downstream inadvertently overriding our styles we should instead prefix our elements with a clear convention, e.g. `skore-{element_id}`

skore/tests/unit/reports/estimator/test_diagnose.py
```diff
@@ -287,3 +296,93 @@ def test_diagnose_custom_metric(binary_classification_data):
...
+    assert passed_frame["explanation"].isna().all()
+
+    all_frame = result.frame()
+    assert set(all_frame["code"]) >= {"TST001", "TST002"}
+
+
+def test_diagnose_header_reports_all_counts(regression_data):
+    """The header string reports issue, tip, passed and ignored counts."""
+    X, y = regression_data
+    report = evaluate(LinearRegression(), X, y)
```
auguste-probabl: Use a fixture

skore/tests/unit/reports/estimator/test_diagnose.py
```diff
@@ -287,3 +296,93 @@ def test_diagnose_custom_metric(binary_classification_data):
...
+    title = "Tip check"
+    report_type = "estimator"
+    docs_url = "tst_tip"
+    severity = "tip"
+
+    def check_function(self, report):
+        return "Be careful about this."
+
+
+def test_diagnose_tip_goes_to_tips_not_issues(regression_data):
```
auguste-probabl:
```suggestion
def test_tip_goes_to_tips_not_issues(regression_data):
```
We're already in `test_diagnose`

skore/src/skore/_sklearn/_diagnostic/model_checks.py
```diff
@@ -266,10 +266,46 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+        if not (
+            isinstance(report, EstimatorReport)
+            and report.X_train is not None
+            and report.X_test is not None
+            and hasattr(report.estimator, "coef_")
+        ):
+            raise DiagnosticNotApplicable()
+
+        X = np.concatenate([report.X_train, report.X_test])
+        if not np.allclose(X.mean(axis=0), 0) and not np.allclose(X.std(axis=0), 1):
```
auguste-probabl:
```suggestion
        if not np.allclose(X.mean(axis=0), 0) or not np.allclose(X.std(axis=0), 1):
```

auguste-probabl: equivalently
```suggestion
        if not (np.allclose(X.mean(axis=0), 0) and not np.allclose(X.std(axis=0), 1)):
```

skore/src/skore/_sklearn/_base.py
```diff
@@ -36,45 +36,53 @@ def _aggregate_checks(self) -> tuple[dict[str, dict], set[str]]:
...
-        """Get the issues from the cache or compute them."""
+    def _get_findings(self) -> tuple[dict[str, dict], set[str]]:
+        """Get the findings from the cache or compute them.
+
+        Returns ``(findings, applicable_codes)`` where ``applicable_codes``
+        contains the codes of the checks that actually ran on the report,
+        i.e. those that did not raise :class:`DiagnosticNotApplicable`.
+        """
         if not hasattr(self, "_issues_cache"):
             self._issues_cache: dict[str, dict] = {}
```
auguste-probabl: Maybe
```suggestion
            self._findings_cache: dict[str, dict] = {}
```

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -172,26 +293,31 @@ def _get_issue_documentation_url(issue: dict) -> str | None:
...
 
-def _format_issue_message(code: str, issue: dict) -> str:
-    msg = f"[{code}] {issue['title']}. {issue['explanation']}"
-    docs_url = _get_issue_documentation_url(issue)
+def _format_finding_message(code: str, finding: dict) -> str:
+    msg = f"[{code}] {finding['title']}. {finding['explanation']}"
+    docs_url = _get_issue_documentation_url(finding)
     if docs_url is not None:
         msg += f" Read more about this here: {docs_url}."
     msg += f" Mute with `ignore=['{code}']`."
```
auguste-probabl: The "Mute" part of the message also shows for checks that have passed, which is a tiny bit strange, but not blocking.

auguste-probabl: Since the code is shown in this message, we should have one place where we document that checks can be skipped with `ignore=`, rather than mentioning at every single place. But that's for another issue.

skore/src/skore/_sklearn/_base.py
```diff
@@ -123,12 +132,24 @@ def diagnose(
                 for code in configuration.ignore_checks
                 if code.strip()
             )
-        issues, checked_codes = self._get_issues()
+        findings, applicable_codes = self._get_findings()
         filtered = {
-            code: issue for code, issue in issues.items() if code not in ignored
+            code: finding for code, finding in findings.items() if code not in ignored
+        }
+        checks_metadata: dict[str, dict] = {
```
auguste-probabl: Claude says that this information is already available in `issues_cache`, I haven't taken the time to verify

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -130,12 +244,18 @@ class Check(Protocol):
...
+        Severity of the finding. ``"issue"`` flags a modeling problem to fix;
+        ``"tip"`` invites caution (e.g. on the interpretation of a result)
+        without signaling a defect.
     """
 
     code: str
     title: str
     report_type: ReportType
     docs_url: str | None
+    severity: Severity = "issue"
```
auguste-probabl: Claude says, maybe set `severity` on all Checks, without relying on inheritance; I'm not fussed about it

auguste-probabl (changes requested):
Very strong work overall!

skore/src/skore/_utils/repr/diagnostic_display-content.html.j2
```diff
@@ -0,0 +1,37 @@
+<div class="container">
+    <header class="title">{{ header }}</header>
+    <div class="report-tabset">
+{% for bucket in buckets %}
```
glemaitre: Could you add a similar help symbol:

Image

It would be helpful to describe what are we going to find in each tab in one small sentence. If we have documentation later on the different category, it could redirect to the documentation.

skore/src/skore/_utils/repr/diagnostic_display-content.html.j2
```diff
@@ -0,0 +1,37 @@
...
+        <input type="radio" name="{{ container_id }}-report-tab" id="{{ container_id }}-report-tab-{{ loop.index0 }}" class="report-tab-input"{% if loop.index0 == first_non_empty %} checked{% endif %}>
```
glemaitre: I would find it more readable to have something similar as in:

https://github.com/probabl-ai/skore/blob/main/skore/src/skore/_utils/repr/estimator_report-content.html.j2#L1-L6

to define the tab. Maybe, either we can modify a bit the current code to make it generic and not specific to "accessor" or define a new template specific to the diagnostic but I don't think it is necessary.

glemaitre (commented):
I think that we need to update the reports to dissociate issues and tips as well:

Image

examples/technical_details/plot_custom_checks.py
```diff
@@ -25,6 +25,9 @@
 # with ``"https"``), it is used as-is. When it is a plain anchor string
 # it points to the skore diagnostic user guide. When omitted entirely,
 # no documentation link is shown.
+#
+# We set the severity to "tip" to indicate that this is not an issue to fix,
```
glemaitre, line 29: I would expect here, which type of severities that I can choose from.

skore/src/skore/_sklearn/_base.py
glemaitre: I agree.

skore/src/skore/_sklearn/_base.py
```diff
@@ -123,12 +132,24 @@ def diagnose(
                 for code in configuration.ignore_checks
                 if code.strip()
             )
-        issues, checked_codes = self._get_issues()
+        findings, applicable_codes = self._get_findings()
         filtered = {
```
glemaitre: I'm wondering why we don't pass `ignored` to the `_get_findings`. It could prevent computation sometimes?

skore/src/skore/_sklearn/_base.py
glemaitre: Yep I have the impression that we populate it in l.68 above

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: Yes indeed, and that was raised by @francoisgoupil as well when I showed him the feature.

examples/technical_details/plot_custom_checks.py
GaetandeCast: yes x)

skore/src/skore/_sklearn/_comparison/report.py
GaetandeCast: Oh you're right I forgot to add that type, will do.

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,229 @@
...
     ) -> None:
-        self._issues = issues
-        self._checks_ran = checks_ran
-        if issues:
-            self._messages = [
-                _format_issue_message(code, d) for code, d in issues.items()
-            ]
-        else:
-            self._messages = ["No issues were detected in your report!"]
+        self._findings = findings
```
glemaitre: I'm wondering why all those elements could be flatten in a single big dataframe of records?

glemaitre: Somehow, the `.frame` calling `_records_from` is reconstructing this dataframe. So if I'm not missing, we could indeed have everything flatten and make that `.frame` is just querying the desired row and filter only the column of interest?

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: Hmm I don't know, for me Tips and Issues are quite different so it's normal that they convey information in different ways. Having no issues is a good thing while having no tips is neutral.

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,229 @@
...
-            items_html = f"<li>{escape(self._messages[0])}</li>"
-        items_html = f'<ul style="margin:8px 0 0 18px;padding:0;">{items_html}</ul>'
-        return (
-            '<div style="margin:8px 0;padding:10px;border:1px solid #f97316;'
-            "border-radius:4px;display:inline-block;"
-            'font-family:monospace;font-size:13px;line-height:1.5;">'
-            f'<div style="font-weight:700;">{escape(self.header)}</div>'
-            f"{items_html}"
-            "</div>"
+        buckets = [
```
glemaitre: I have the impression that we do the same job whatever the type of severity but that only the `empty_message` is changing. We shoudl be able to make a nested loop. Or maybe having a small function to construct each fragment as the pattern in the report.

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,229 @@
...
+        }
         self.header = (
-            f"Diagnostic: {len(self._issues)} issue(s) detected, "
-            f"{self._checks_ran} check(s) ran, {n_ignored} ignored."
+            f"Diagnostic: {len(self._issues)} issue(s), "
+            f"{len(self._tips)} tip(s), {len(self._passed)} passed, "
+            f"{len(self._ignored_codes)} ignored."
         )
 
     @property
```
glemaitre, line 84: I would probably not bother with those public attributes and rely on `.frame` to return what we need.

skore/src/skore/_sklearn/_diagnostic/model_checks.py
```diff
@@ -266,10 +266,46 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+            return (
+                "The input features do not appear to be standardized. Be careful "
+                "when interpreting the magnitudes of a linear model's coefficients "
+                "as feature importance: they depend on the scale of each feature "
+                "and are not directly comparable across features."
```
glemaitre: I think that we need more balanced as in https://scikit-learn.org/dev/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#interpreting-coefficients-scale-matters

Basically, it is fine if they are not scale because you can intepret the coefficients has having a direct impact in the native unit but you cannot compare coefficients between them. If you want to compare the relative impact then, you need to do what you mention.

I think that the section of this documentation could be useful to be cited somewhere.

skore/src/skore/_sklearn/_diagnostic/model_checks.py
glemaitre: I think that we don't have the right test here.

I think that we should only check that the variance of the features is close. If not then we should warn that comparison should take into account the variance when comparing feature between them but that it is right if one wants to look at the magnitude of a single feature as a measure in the original unit space.

Similarly we should have most probably another check for something where all features are scaled with the same standard deviation and in this case you should not interpret the relation between a feature and the target in the original dataset unit because the scaling makes it "unitless".

sphinx/user_guide/automatic_diagnostic.rst
```diff
@@ -209,3 +214,41 @@ How to reduce the risk
 - resample the dataset (oversampling rare classes or undersampling frequent ones),
 - use class weights in the estimator,
 - collect more data for the underrepresented classes if possible.
+
+
+.. _skd006-unscaled_coefficients:
+
+SKD006 - Unscaled coefficients
+------------------------------------
```
glemaitre:
```suggestion
SKD006 - Unscaled coefficients
------------------------------
```

glemaitre: I assume that what comes below should be a bit rephrase with my comments that I put on the diagnostic.

sphinx/user_guide/automatic_diagnostic.rst
```diff
@@ -209,3 +214,41 @@ How to reduce the risk
...
+Why it matters
+^^^^^^^^^^^^^^
+
+The magnitude of a linear model's coefficients depends on the scale of each input
+feature. When features live on different scales, comparing raw coefficients as a
+measure of feature importance is misleading: a large coefficient may only reflect
+a small-scale feature, not a strong effect. Standardizing the inputs puts all
+coefficients on a common footing and makes them directly comparable.
+Read more about this in `the scikit-learn documentation
+<https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html>`_.
```
glemaitre: I would specifically drive to the scaling section.

skore/src/skore/_sklearn/_diagnostic/model_checks.py
GaetandeCast: Okay, I'll revisit the check logic

skore/src/skore/_sklearn/_base.py
GaetandeCast: In `_issues_cache` we only stored the metadata for the checks that found something and we needed to store the information of those that passed somewhere. I changed it so that we store the information of all checks in the cache, and set the explanation to `None` for those that did not find an issue/a tip. An added bonus is that we don't rerun checks that were found not applicable before.

skore/src/skore/_sklearn/_base.py
GaetandeCast: done

skore/src/skore/_sklearn/_diagnostic/base.py
GaetandeCast: So I reworked the internal storage of `DiagnosticDisplay` so that everything is in a private attribute `_check_results`. The frame method then queries the relevant information.

skore/src/skore/_sklearn/_diagnostic/base.py
auguste-probabl: At least both messages should use `for`, and neither should use an exclamation point.

GaetandeCast:
@auguste-probabl @glemaitre I addressed everything. Here is the new look for the html repr:
image
image
image

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,179 @@
...
     from skore._sklearn._base import _BaseReport
 
 
+CheckCode = str
+
+
+_TAB_SPECS: list[tuple[str, Literal["issue", "tip", "passed"], str]] = [
+    ("Issues", "issue", "No issues were detected in your report."),
+    ("Tips", "tip", "No tips were emitted for your report."),
+    ("Passed", "passed", "No checks passed without reporting anything."),
```
auguste-probabl:
```suggestion
    ("Passed", "passed", "All checks passed without reporting anything."),
```

GaetandeCast: Actually no, when the `Passed` section is empty, it means that either all checks found something or no checks were applicable. `Passed` collects the check that ran but did not flag an issue/tip.

skore/src/skore/_sklearn/_diagnostic/base.py
```diff
@@ -3,115 +3,179 @@
...
+                            ' target="_blank" rel="noopener noreferrer">here</a>.'
+                        )
+                    items.append(f"<li>{line}</li>")
+                fragments[f"{severity}_html"] = (
+                    "<ul class='report-diagnostic-list'>" + "".join(items) + "</ul>"
+                )
+            else:
+                fragments[f"{severity}_html"] = (
+                    f"<p class='report-diagnostic-empty'>{escape(empty_message)}</p>"
+                )
```
glemaitre: We need to create a jinja template for this.

GaetandeCast: Moved all the html generation in the existing jinja template, we now provide it with a dictionary containing all the information necessary for building the repr.

skore/src/skore/_sklearn/_diagnostic/base.py
auguste-probabl: Ok, since this message will appear very rarely in practice this is good enough.

auguste-probabl (approved):

PR #2719, opened 2026-04-03
docs(skore): Create an integration section in the doc

glemaitre:
Supersedes https://github.com/probabl-ai/skore/pull/2683 to be able to have the documentation preview when related to the public project on the hub.

PR #2819, opened 2026-04-23
Revert "fix(skore): Close plots to avoid triggering a user-facing warning"

glemaitre:
Reverts probabl-ai/skore#2778

I don't think that we should automatically close figure on the behalf of the user. Here, the side effect is that it is breaking the generation of the documentation because we close the figure before to save them probably.


thomass-dev (approved):

thomass-dev:
Check in https://github.com/probabl-ai/skore/pull/2719#pullrequestreview-4162665848 to see the problem.

PR #2822, opened 2026-04-24
docs(skore): Do not show matplotlib Figure repr in example

glemaitre:
@thomass-dev 

As discussed, here is a PR that catch the figure representation in the example to avoid the textual output.


thomass-dev (approved):
Great, thanks!

PR #2823, opened 2026-04-24
docs(skore): Add report methods foldable table in API ref

GaetandeCast:
Adds a `Report Methods` folded table above the `Accessor Methods` table to show methods that live directly on reports such as `diagnose` and `get_predictions`.


glemaitre (approved):
LGTM

PR #2824, opened 2026-04-24
feat(skore): Add position parameter when adding metric in registry

glemaitre:
Currently, the user metric is always shown after our predefined metric. If a user is adding a metric, I think that we should provide the possibility to show it first also.

So I'm adding a `position` argument and use an `OrderedDict` to accommodate this use-case.

### DIFF
--- a/skore/src/skore/_sklearn/metrics.py
+++ b/skore/src/skore/_sklearn/metrics.py
@@ -641,13 +641,13 @@ class MetricRegistry(UserDict[str, Metric]):
                 f"Cannot add {metric.name!r}: it is a built-in metric name."
             )
 
-        # Invalidate cached results when re-adding an existing metric
         if metric.name in self.data:
-            keys_to_delete = [k for k in self._report._cache if k[1] == metric.name]
-            for k in keys_to_delete:
-                del self._report._cache[k]
+            raise ValueError(
+                f"Cannot add {metric.name!r}: it already exists. "
+                "Remove it before adding it again."
+            )
 
         self.data[metric.name] = metric
-        
+
         if position == "first":
             self.data.move_to_end(metric.name, last=False)
--- a/skore/src/skore/_sklearn/metrics.py
+++ b/skore/src/skore/_sklearn/metrics.py
@@ -3,9 +3,9 @@ from __future__ import annotations
 import copy
 import inspect
 import pickle
-from collections import UserDict
+from collections import OrderedDict, UserDict
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Any, Literal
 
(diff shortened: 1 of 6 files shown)


skore/src/skore/_sklearn/metrics.py
```diff
@@ -633,3 +648,4 @@ def add(self, metric: Metric) -> None:
                 del self._report._cache[k]
 
         self.data[metric.name] = metric
+        self.data.move_to_end(metric.name, last=(position == "last"))
```
thomass-dev:
```suggestion
        
        if position == "first":
            self.data.move_to_end(metric.name, last=False)
```

thomass-dev (changes requested):
LGTM, only a small "optimization".

thomass-dev (dismissed):
Better now, thanks!

skore/tests/unit/reports/estimator/metrics/test_add.py
```diff
@@ -364,13 +485,17 @@ def score(y_true, y_pred):
         def score(y_true, y_pred):
             return 1
 
-        report.metrics.add(make_scorer(score, response_method="predict"))
+        err_msg = re.escape(
+            "Cannot add 'score': it already exists. Remove it before adding it again."
+        )
+        with pytest.raises(ValueError, match=err_msg):
+            report.metrics.add(make_scorer(score, response_method="predict"))
```
auguste-probabl, line 532: Can you make an issue for that and take it out of this PR?

auguste-probabl (changes requested):

skore/tests/unit/reports/estimator/metrics/test_add.py
glemaitre: I open a PR in #2836 which needs to be merged first then.

thomass-dev (dismissed):

skore/tests/unit/reports/estimator/metrics/test_registry.py
```diff
@@ -260,6 +260,124 @@ def test_summarize_with_mixed_metrics(self, binary_classification_report):
...
+        err_msg = re.escape(
+            "Cannot add 'reorder_me': it already exists. "
+            "Remove it first using the `remove` method."
+        )
+        with pytest.raises(ValueError, match=err_msg):
+            report.metrics.add(
+                make_scorer(score_v2, response_method="predict"),
+                name="reorder_me",
+                position="last",
+            )
```
thomass-dev, line 365: Already covered by `def test_duplicate_name_raises` even if it not tests explicitly `position='last'`, isn't?

thomass-dev (approved):
LGTM, just a nitpick.

auguste-probabl (approved):

PR #2825, opened 2026-04-24
feat(skore): Add remove method to remove a metric from registry

glemaitre:
It could be handy to be able to remove metrics that we might not be interested from the registry.

This PR is adding a `.remove` method.


skore/src/skore/_sklearn/metrics.py
```diff
@@ -633,3 +633,28 @@ def add(self, metric: Metric) -> None:
...
+
+        Parameters
+        ----------
+        name : str
+            The technical name of the metric to remove.
+
+        Raises
+        ------
+        KeyError
+            If *name* is not registered.
```
auguste-probabl:
```suggestion
            If `name` is not registered.
```

auguste-probabl (dismissed):
LGTM

auguste-probabl (approved):

PR #2826, opened 2026-04-25
feat(skore): Add available method to know the available metrics in registry

glemaitre:
Add an `available` to have a programmatic way to list the available key metric from the registry.

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_estimator/metrics_accessor.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from collections.abc import Iterable
 from typing import Any, Literal, cast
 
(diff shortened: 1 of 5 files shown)


skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -173,6 +175,16 @@ def _metric(
...
 
+    def available(self) -> list[str]:
+        """List available metric names in the registry.
+
+        Returns
+        -------
+        list[str]
+            The list of available metric names.
+        """
+        return list(self._parent._metric_registry)
```
thomass-dev, line 186: Nit: IMO, `metrics` should be exposed directly as a registry with only available metrics, so that someone can do `for metric in report.metrics`.

glemaitre: I think that it is an orthogonal feature. Here, it is more a feature to discover what metrics are inside the registry rather than allowing to iterate over it. Basically, it is to find the string that I can use in `summarize` or other methods accepting `metrics` to filter them.

And I find it a more explicit than `report.metrics.keys()` if we were going to expose the registry directly.

auguste-probabl (approved):

PR #2829, opened 2026-04-25
chore(skore,skore-*-project)!: Remove clear_cache and cache_predictions from public API

glemaitre:
This PR is removing the public interface for `clear_cache` and `cache_predictions`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -48,6 +48,13 @@ Added
 Removed
 -------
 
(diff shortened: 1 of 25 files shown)


cakedev0:
This is typically the kind of things that's much harder too do in a multi-package, you should probably wait for the mono-package architecture change before moving forward here (maybe that's already you're plan).

glemaitre:
I wanted to see to now, what is breaking ;)

glemaitre:
I'm closing it since @auguste-probabl is on the way to do it and it will need a complete update.

PR #2832, opened 2026-04-27
feat(skore-local-project): Support strings for `workspace`

### LINKED ISSUE #2791: enh: Make that `workspace` accept string when using local Project

When using a local project, `workspace` expect only a `Path`. However, I often only want to provide a string.

auguste-probabl:
#### Change description

Makes it possible to pass a string to `Project` for the `workspace` argument.

Closes #2791

#### Contribution checklist

requires updating tests)

#### AI usage disclosure

AI tools were not involved

### DIFF
--- a/skore-local-project/src/skore_local_project/project.py
+++ b/skore-local-project/src/skore_local_project/project.py
@@ -229,9 +229,13 @@ class Project:
     @ensure_project_is_not_deleted
     def get(self, id: str) -> EstimatorReport | CrossValidationReport:
         """Get a persisted report by its id."""
-        with io.BytesIO(self.__artifacts_storage[id]) as stream:
-            report = joblib.load(stream)
-            return cast("EstimatorReport | CrossValidationReport", report)
+        if id in self.__artifacts_storage:
+            with io.BytesIO(self.__artifacts_storage[id]) as stream:
+                return cast(
+                    "EstimatorReport | CrossValidationReport", joblib.load(stream)
+                )
+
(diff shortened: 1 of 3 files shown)


skore-local-project/src/skore_local_project/project.py
thomass-dev:
```suggestion
    def delete(name: str, *, workspace: str | Path | None = None) -> None:
```

thomass-dev:
```suggestion
        workspace : Path-like, optional
```

skore-local-project/src/skore_local_project/project.py
```diff
@@ -226,13 +229,9 @@ def put(self, key: str, report: EstimatorReport | CrossValidationReport) -> None
     @ensure_project_is_not_deleted
     def get(self, id: str) -> EstimatorReport | CrossValidationReport:
         """Get a persisted report by its id."""
-        if id in self.__artifacts_storage:
-            with io.BytesIO(self.__artifacts_storage[id]) as stream:
-                return cast(
-                    "EstimatorReport | CrossValidationReport", joblib.load(stream)
-                )
-
-        raise KeyError(id)
```
thomass-dev, line 235: I don't think we need this change: by raising fast the `KeyError` here, we avoid users to see `DiskCache` tracebacks, that are not more relevant.

thomass-dev (changes requested):

thomass-dev (approved):

PR #2833, opened 2026-04-27
chore(skore)!: Create a checks accessor

### LINKED ISSUE #2828: api: Create a namespace to interact with the diagnostic registry

Right now, we added a `add_checks` for the diagnostic registry. So the way we are going is that every methods to interact with the registry will be in the first layer of the report and we will need to scope in the naming. It is kind of the opposite of what we have for the metrics:

```python
report.metrics.add(...)
report.metric.remove(...)
report.metrics.available(...)
```

I agree that the metrics accessor is a bit different because under this namespace you don't have only metrics but also display which make it necessary.

For the diagnosis, I'm wondering if we should something like:

```python
report.diagnosis()  # run the diagnostic as now
report.diagnosis.add(...) # add checks
report.diagnosis.remove(...)
report.diagnosis.available(...)
```

Somehow, the first call is to avoid to have `report.diagnosis.run()`

GaetandeCast:
Closes #2828
Adds a `checks` accessor to concentrate all checks related methods.
Moves `Report.diagnose()` to `Report.checks.summarize()` and `Report.add_checks()` to `Report.checks.add()`.
Renames internal modules, files and objects to not use the diagnose/diagnostic vocabulary anymore.


thomass-dev:
> Adds `Report.diagnosis.remove()` and `Report.diagnosis.available()`.

You should extract these changes in their own stacked PRs.

skore/src/skore/_sklearn/_diagnosis/accessor.py
```diff
@@ -0,0 +1,136 @@
...
+        Parameters
+        ----------
+        code : str
+            The code of the check to remove.
+        """
+        code = code.strip().upper()
+        for i, check in enumerate(self._parent._checks_registry):
+            if check.code.upper() == code:
+                del self._parent._checks_registry[i]
+                break
```
auguste-probabl:
```suggestion
        registry_ids = [i for i, check in enumerate(self._parent._checks_registry)  if check.code.upper() == code]
        if registry_ids:
            del self._parent._checks_registry[registry_ids[0]]
```

skore/src/skore/_sklearn/_diagnosis/accessor.py
```diff
@@ -0,0 +1,136 @@
...
+        if hasattr(self._parent, "_check_results_cache"):
```
auguste-probabl: when would this not be the case?

skore/src/skore/_sklearn/_diagnosis/accessor.py
```diff
@@ -0,0 +1,136 @@
...
+    def available(self) -> list[str]:
+        """List available checks in the registry.
+
+        Returns
+        -------
+        list[str]
+            The list of available checks in the format "code - title".
+        """
+        return [
+            check.code + " - " + check.title for check in self._parent._checks_registry
```
auguste-probabl:
```suggestion
            f"{check.code} - {check.title}" for check in self._parent._checks_registry
```

skore/src/skore/_sklearn/_diagnosis/accessor.py
```diff
@@ -0,0 +1,106 @@
...
+    ) -> None:
+        """Register additional diagnostic checks for this report.
+
+        Checks are defined by implementing the :class:`~skore.Check` protocol.
+
+        Parameters
+        ----------
+        checks : list of Check
+            Additional checks to register
+
```
auguste-probabl:
```suggestion
```

skore/src/skore/_sklearn/_checks/accessor.py
```diff
@@ -0,0 +1,106 @@
...
+        report_types = [
+            "cross-validation",
+            "estimator",
+            "comparison-estimator",
+            "comparison-cross-validation",
+        ]
+        for check in checks:
+            if not isinstance(check, Check):
+                raise ValueError(f"{check} does not implement the Check protocol.")
+            if check.report_type not in report_types:
```
auguste-probabl, line 103: will this ever happen?

auguste-probabl (changes requested):
lots of nitpicks, almost good to go

skore/src/skore/_sklearn/_checks/accessor.py
glemaitre: Basically it should not never happen on the side of the user because developer are in charge of setting those ;) I would remove it.

jeromedockes: if it only happens in case of a bug in skore code but you still want to help detect that bug you could replace that check with an `assert check.report_type in report_types`

thomass-dev: Change it to `assert` if it is only for developers.

skore/src/skore/_sklearn/_diagnosis/accessor.py
GaetandeCast: Extracted `remove` to please @thomass-dev, I'll apply the change in #2844 and resolve the comment here.

jeromedockes: maybe

```
self._parent._checks_registry = [check for check in self._parent._checks_registry if check.code.upper() != code]
```

or something like that?

skore/src/skore/_sklearn/_diagnosis/base.py
glemaitre, line 189: We need to expose the `Check` as we do for `Display` such that in `add`, people can find the associated documentation.

skore/src/skore/_sklearn/_diagnosis/accessor.py
GaetandeCast: If you remove a check without calling `summarize` first, the cache will be empty.

GaetandeCast: The cache is not instantiated at init, which maybe is weird.

skore/src/skore/_sklearn/_checks/accessor.py
GaetandeCast: Yes for custom checks we need to enforce that the `report_type` is correct.

examples/technical_details/plot_diagnostic_api.py
```diff
@@ -40,20 +40,22 @@
...
-# ==========================================================
+# Calling :meth:`~skore.EstimatorReport.diagnosis.summarize` explicitly
+# =====================================================================
 #
-# Every report exposes a :meth:`~skore.EstimatorReport.diagnose` method.
-# Checks are computed lazily and cached, so calling
-# :meth:`~skore.EstimatorReport.diagnose` is always cheap after the first call.
+# Every report exposes a :meth:`~skore.EstimatorReport.diagnosis` accessor which provides
+# access to several methods:
+# - `.summarize()`` to run checks and get a summary of the findings
```
auguste-probabl:
```suggestion
# - `.summarize()` to run checks and get a summary of the findings
```

examples/technical_details/plot_checks_api.py
auguste-probabl, line 1: update the changelog

auguste-probabl (dismissed):

skore/src/skore/_sklearn/_checks/accessor.py
GaetandeCast: Sorry gh did not show me your messages when I wrote the previous one. This check ensures that the `report_type` chosen by the user in their own `Check` is correct.

skore/src/skore/_sklearn/_diagnosis/accessor.py
GaetandeCast: Though about that too, idk what is the most efficient.

skore/src/skore/_sklearn/_checks/accessor.py
thomass-dev: héhé @jeromedockes same idea at the same moment 10:58 🥂

skore/src/skore/_sklearn/_diagnosis/accessor.py
jeromedockes: I don't think you need to think about efficiency here

skore/src/skore/_sklearn/_checks/accessor.py
jeromedockes: :) but in the end it seems it _is_ something that can happen due to bad user input and requires the check that is currently there

skore/src/skore/_sklearn/_diagnosis/base.py
```diff
@@ -42,7 +42,7 @@
 
 
 class DiagnosticDisplay(DisplayHelpMixin):
-    """Display for the diagnostic returned by :meth:`Report.diagnose`.
+    """Display for the diagnostic returned by :meth:`Report.diagnosis.summarize`.
```
jeromedockes: I wonder if the distinction between diagnostic and diagnosis is going to be an issue for users to remember the names of objects. IIUC diagnostic is often used as an adjective, and when it is used as a noun it refers to the method used to yield a diagnosis (which is the conclusion of the diagnostic), is that correct? for example here I find it hard to get a strong intuition that the accessor should be called `diagnosis` and the display class `Diagnostic` and I wonder if this kind of thing will add friction for users

jeromedockes: also the things that do the diagnostics are called Checks ... maybe we could have something like `report.checks.summarize()` , `Checks` and `CheckResults`

GaetandeCast: Yeah when iterating on the userguide here I noticed that we should probably rename to `DiagnosisDisplay` too, so that wording is consistent: {Checks} => diagnosis.
From my understanding, a diagnostic is a medical check/exam that can yield to a diagnosis which is the "verdict" on the condition. So I guess this renaming makes sense. Hopefully afterwards we are happy and don't have to rename things again :))

GaetandeCast: once again gh was not up to date when I wrote the previous message.

examples/technical_details/plot_checks_api.py
```diff
@@ -40,20 +40,22 @@
...
-# Calling :meth:`~skore.EstimatorReport.diagnose` explicitly
-# ==========================================================
+# Calling :meth:`~skore.EstimatorReport.diagnosis.summarize` explicitly
+# =====================================================================
 #
-# Every report exposes a :meth:`~skore.EstimatorReport.diagnose` method.
-# Checks are computed lazily and cached, so calling
-# :meth:`~skore.EstimatorReport.diagnose` is always cheap after the first call.
+# Every report exposes a :meth:`~skore.EstimatorReport.diagnosis` accessor which provides
+# access to several methods:
```
jeromedockes, line 47: as we compute them anyway (to show the number of issues) I guess we could show the actual results in the repr of the report (rather than just the number of checks in each category issue/tip/passed/ignored)?

examples/technical_details/plot_checks_api.py
```diff
@@ -40,20 +40,22 @@
...
+# - `.summarize()` to run checks and get a summary of the findings
```
jeromedockes: the list doesn't seem to render properly

examples/technical_details/plot_diagnostic_api.py
```diff
@@ -40,20 +40,22 @@
...
+# Let's use `.summarize()` to see what issues can be found for the linear model.
 
 from skore import evaluate
 
 linear_report = evaluate(linear, X, y)
 linear_report
 
 # %%
-linear_report.diagnose()
+linear_report.diagnosis.summarize()
```
jeromedockes: the help() on summarize lists the `header` and `n_ignored_checks` -- maybe those should be made private attributes

examples/technical_details/plot_diagnostic_api.py
```diff
@@ -63,7 +65,7 @@
 # train and test, and not significantly better than a dummy baseline.
 
 tree_report = evaluate(deep_tree, X, y)
-tree_report.diagnose()
+tree_report.diagnosis.summarize()
```
jeromedockes: maybe worth a sentence to point out the tip on coefficient interpretation is not there because it does not apply to this report

examples/technical_details/plot_checks_api.py
jeromedockes, line 47: as we compute them anyway (to show the number of issues) I guess we could show the actual results in the repr of the report (rather than just the summary)?

GaetandeCast: Hmm do we want sub tabs in the tab ? Otherwise we can show the issues tab only ?

skore/src/skore/_sklearn/_diagnosis/base.py
GaetandeCast: We decided offline to drop the diagnostic/diagnosis vocabulary and decline the word `check` everywhere. I'll update the PR accordingly.

examples/technical_details/plot_checks_api.py
GaetandeCast: This is probably out of the scope of this PR, I think we can come back to this afterwards.

jeromedockes (dismissed):
thank you @GaetandeCast !

glemaitre (approved):

Issue #2781, opened 2026-04-17
feat: Store Git commit at `put` when possible


auguste-probabl:
If the user is working in a Git repo, store the git commit hash alongside the report that was `put`. If the working tree is clean, great; if it is dirty, maybe just warn for now?

auguste-probabl:
Closed by #2834

PR #2834, opened 2026-04-27
chore(skore): Retrieve git commit upon `Project.put`

auguste-probabl:
#### Change description

Report metadata now has a new attribute `git_commit` which can be a string or None.

Saves `<hash> (working tree dirty)` if git working tree is dirty; saves None if an error occurs (e.g. git is not available).

Note that there can be surprises: for example, if the `put` occurs in an IPython session, e.g. in a VSCode notebook, and the buffer is not backed by a file on disk, Git will not have it tracked, and the working tree will be considered clean.

Addresses #2781 

#### Contribution checklist

requires updating tests)

#### AI usage disclosure

AI tools were involved for:

### DIFF
--- a/skore/src/skore/_project/git.py
+++ b/skore/src/skore/_project/git.py
@@ -1,0 +1,94 @@
+import functools
+import shutil
+import subprocess
+
+
+def git_available() -> bool:
+    """Check whether Git is installed and usable.
+
+    Returns
+    -------
+    bool
+        True if the ``git`` executable is found on the system path.
+    """
+    return shutil.which("git") is not None
+
+
+def working_tree_clean() -> bool:
+    """Check whether the Git working tree is clean.
+
+    Assumes that ``git`` is available on the system path.
+
+    Returns
+    -------
+    bool
+        True if there are no uncommitted changes, False otherwise.
+    """
+    try:
+        result = subprocess.run(
+            ["git", "status", "--porcelain"],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        return False
+    return result.stdout.strip() == ""
+
+
+def head_commit_hash() -> str | None:
+    """Obtain the hash of the HEAD commit.
+
+    Assumes that ``git`` is available on the system path.
+
+    Returns
+    -------
+    str or None
+        The full SHA-1 hash of the current HEAD commit, if available.
+    """
+    try:
+        result = subprocess.run(
+            ["git", "rev-parse", "HEAD"],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        return None
+    return result.stdout.strip()
+
+
+@functools.cache
+def git_commit() -> str | None:
+    """Obtain the hash of the latest Git commit.
+
+    Returns
+    -------
+    str or None
+        The commit hash, if available.
+
+    Examples
+    --------
+    >>> # xdoctest: +SKIP
+
+    With git not available:
+
+    >>> git_commit()
+    None
+
+    With a clean working tree:
+
+    >>> git_commit()
+    e5de78623d1227d8b7f90f949bb7830eb9433ade
+
+    With a dirty working tree (uncommitted changes):
+
+    >>> git_commit()
+    e5de78623d1227d8b7f90f949bb7830eb9433ade (working tree dirty)
+    """
+    if not git_available():
+        return None
+    commit_hash = head_commit_hash()
+    if working_tree_clean():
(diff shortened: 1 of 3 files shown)


skore-local-project/src/skore_local_project/metadata.py
```diff
@@ -29,6 +30,39 @@ def cast_to_float(value: Any) -> float | None:
...
+    try:
+        result = subprocess.run(
+            ["git", "status", "--porcelain"], capture_output=True, text=True
+        )
+    except FileNotFoundError:
+        # Git not on PATH
+        return False
+    if result.returncode != 0:
+        # e.g. not in a Git repo
+        return False
```
thomass-dev:
```suggestion
    try:
        result = subprocess.run(
            ["git", "status", "--porcelain"],
            capture_output=True,
            text=True,
            check=True,
        )
    except FileNotFoundError, CalledProcessError:
        return False   
```

skore-local-project/tests/unit/test_git_commit.py
thomass-dev, line 1: Please rename this file to `test_metadata.py`, and prefix your tests with `test_git_commit_[...]`.

skore-local-project/tests/conftest.py
```diff
@@ -13,6 +13,11 @@ def nowstr(now):
     return now.isoformat()
 
 
+@fixture
+def mock_git_commit():
+    return "000"
```
thomass-dev: You should move it in `test_project.py`, it is only used here.

thomass-dev (changes requested):
Some minor requests, otherwise LGTM.

Also, it should be cool to find a way to really test `head_commit_hash` and `working_tree_clean`.

skore-local-project/src/skore_local_project/metadata.py
auguste-probabl, line 1: @thomass-dev skore-local-project tests are failing because of the usual compatibility issues. Should I remove the parts touching skore-local-project from the PR and keep them until after #2807?

thomass-dev: Yes, please remove it for now. Will be much easier after #2807.

skore/src/skore/_project/project.py
```diff
@@ -241,6 +242,8 @@ def put(self, key: str, report: EstimatorReport | CrossValidationReport):
         else:
             self.ml_task = report.ml_task
 
+        report._metadata["git_commit"] = git_commit()
```
thomass-dev: This should be done directly in `_BaseReport::__init__`, where all metadata are computed.

thomass-dev (changes requested):

skore/src/skore/_project/project.py
auguste-probabl: This is a design consideration: my vision is that the commit hash is set upon `put`, not at report creation.

thomass-dev: IRL, we discussed that it should be integrated at the report creation, no matter the report is put or not 😅 .

With the current implementation (git system call at put), strange behavior can occur: i run my script with a long-running task (commit A) before putting a report. Meanwhile, i change the code locally on a different file and commit (commit B). The report will be tagged with commit B even though it was run with commit A.

Same, i can have a strange behavior even by putting twice the same report.

IMO, it's not a good idea to set this metadata at `put`. I'm convinced we need to do it at report creation.
What's your opinion @jeromedockes ?

thomass-dev: I'm even thinking if we should not retrieve the commit hash in the `skore` init and set it as a constant.

jeromedockes: I agree, during import is probably the most reliable or otherwise at report creation time is probably also ok if we do not want to slow down import

auguste-probabl: Done, moved to report creation time

skore/src/skore/_project/git.py
```diff
@@ -0,0 +1,92 @@
...
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        return None
+    return result.stdout.strip()
+
+
+def git_commit() -> str | None:
```
thomass-dev, line 62: Coud you please cache this call, in a global or using `@functools.cache`?
In a single session, you shouldn't have to run "git" multiple times.

thomass-dev (changes requested):

skore/src/skore/_project/git.py
auguste-probabl: Done

thomass-dev (approved):

PR #2838, opened 2026-04-27
feat(skore): Better grouping in the help menu of the metrics accessor

### LINKED ISSUE #2746: Simplify API by removing single metrics in favor of `report.metrics.summarize`

Currently, we have access to individual metric using `report.metrics.my_metric()`. Also, with the current work on the metric registry, all those metrics will definitely be available using `report.metrics.summarize()`.

Most of the time, I'm always finding myself calling the `.summary` when I'm interested about those metrics and never the individual metric which feel a normal usage from a "report".

My latest thought would be to completely remove `.accuracy` & co since once we will have easily access using:

```python
report.metrics.summarize().frame(metric="accuracy")
```

if the interest is to only observed this metric in particular.

I feel that it would simplify a lot the documentation page which is crowded and not dynamic.
We would need to introduce a helper to get which metric are currently available in the internal registry:

```python
>>> report.metrics.available_metrics
The following metric are available for the reporting:
- "accuracy"
- "precision"
- "recall"
You can add a new metric using the method `metrics.add` and it will be used by default
in the report.
```

glemaitre:
closes https://github.com/probabl-ai/skore/issues/2746

This PR improves the help of the metrics accessor by grouping the registry management methods, the metrics to be called, and the displays in to sub groups.

For the moment, we are hard coding them but if the metrics are dynamically generated, we could easily leverage the registry to populate the the registry methods and the metrics.

My initial thought is to have the following design:

image

### DIFF
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -211,12 +211,18 @@ def _summarize_report_metrics(
 class BaseMetricsAccessor(_BaseAccessor, Generic[ParentT]):
     """Base class for metrics accessor."""
 
-    # Help tree subgroups under ``.metrics``. ``Displays`` is a catch-all for
-    # methods that are neither registry management nor metric callables.
-    _HELP_METHOD_GROUPS: ClassVar[dict[str, tuple[str, ...] | None]] = {
+    # Help tree subgroups under ``.metrics``. Registry callables are injected
+    # into Metrics by ``_build_help_data``.
+    _HELP_METHOD_GROUPS: ClassVar[dict[str, tuple[str, ...]]] = {
         "Registry": ("available", "add", "remove", "get"),
         "Metrics": ("fit_time", "predict_time", "score", "timings"),
-        "Displays": None,
+        "Displays": (
+            "summarize",
+            "roc",
+            "precision_recall",
+            "prediction_error",
+            "confusion_matrix",
+        ),
     }
 
     def _callable_metric_names(self) -> list[str]:
@@ -343,9 +349,9 @@ class BaseMetricsAccessor(_BaseAccessor, Generic[ParentT]):
         Names that are not valid identifiers (e.g. containing spaces) are excluded,
         since they cannot be called as ``report.metrics.<name>(...)``.
 
-        Methods are then partitioned into Registry / Metrics / Displays groups.
-        Metrics combines registry callables with static score helpers; Displays
-        collects remaining methods that are neither registry management nor metrics.
+        Methods are then partitioned into Registry / Metrics / Displays groups
+        using the ordered names in ``_HELP_METHOD_GROUPS``, with registry
+        callables injected into Metrics.
         """
         help_data = super()._build_help_data()
         known_names = {method.name for method in help_data.methods}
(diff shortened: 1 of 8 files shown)


GaetandeCast:
>we could easily leverage the registry to populate the the registry methods and the metrics.

Then we can show the added custom metrics in the help tree which would be really cool imo

auguste-probabl:
How does this close #2746?

glemaitre:
Because we are not going to remove the individual metrics but instead try to improve the documentation as an alternative.

skore/src/skore/_sklearn/_base.py
```diff
@@ -211,6 +211,21 @@ def _summarize_report_metrics(
 class BaseMetricsAccessor(_BaseAccessor, Generic[ParentT]):
     """Base class for metrics accessor."""
 
+    # Help tree subgroups under ``.metrics``. Explicit tuples list fixed methods;
+    # ``None`` is a catch-all for remaining methods (registry callables such as
+    # ``accuracy`` / ``r2``, plus static score helpers like ``timings``).
+    _HELP_METHOD_GROUPS: ClassVar[dict[str, tuple[str, ...] | None]] = {
+        "Registry": ("available", "add", "remove", "get"),
+        "Metrics": None,
+        "Displays": (
```
glemaitre, line 237: Can we make that this list are metrics that are not available either is Metrics or Registry.

skore/src/skore/_sklearn/_base.py
```diff
@@ -347,7 +367,24 @@ def _build_help_data(self) -> AccessorHelpData:
             for name in self._callable_metric_names()
             if name not in known_names
         )
-        help_data.methods.sort(key=lambda method: method.name)
+        # Rebuild groups after injecting dynamic registry metrics.
```
glemaitre: You can remove this comment

skore/src/skore/_sklearn/_base.py
```diff
@@ -347,7 +367,24 @@ def _build_help_data(self) -> AccessorHelpData:
...
+        help_data.groups = _build_method_groups(self, help_data.methods)
+        # Within Metrics, show registry callables first (registry order), then
```
glemaitre: Remove also this comment.

skore/src/skore/_sklearn/_base.py
```diff
@@ -347,7 +367,24 @@ def _build_help_data(self) -> AccessorHelpData:
...
+                if group.name != "Metrics":
+                    continue
+                by_name = {method.name: method for method in group.methods}
+                group.methods = [
+                    by_name[name] for name in registry_order if name in by_name
+                ] + [
+                    method
+                    for method in group.methods
+                    if method.name not in registry_set
+                ]
```
glemaitre: It would be easier to read if we would have the three group with an `if` statement. It would be more explicit

glemaitre (commented):
By default, I would like the group to be unfolded to avoid an extra click to see them to have a similar experience as before.

GaetandeCast (approved):
LGTM

glemaitre:
Thanks @GaetandeCast

PR #2839, opened 2026-04-28
feat(skore)!: Plot multi-class predict-based OvR confusion matrix

cakedev0:
#### Change description

This PR adds predict-based one-vs-rest confusion matrices for multiclass classification, so `.plot(label=...)` can show the same 2x2 OvR view without requiring a `threshold_value`. 

For that, `ConfusionMatrixDisplay` now stores a new multiclass-only predict-based OvR cache in `confusion_matrix_ovr`.

Because I started to refactor, I went down the refactor rabbit hole :sweat_smile: and also ended-up refactoring/fixing those things:
- `.frame()` doesn't return irrelevant columns anymore (`data_source` is not included, and `split`/`estimator` are only included when meaningful)
- Dataframes of `ConfusionMatrixDisplay` also don't store those columns anymore when not needed.
- Introduced `_get_ovr_labels` because I had a couple of bugs due to different logic in different places around that.

#### Breaking changes:

- `.frame()` no longer always returns `data_source`, `split`, and `estimator`; it only returns the relevant metadata columns for the report.

#### AI usage disclosure

AI tools were involved for: Test/benchmark generation and review-fix iteration.

Because I iterated quite a bit (by hand and with AI), there are a couple of changes that could maybe have been avoided. Sorry for that, I tried to clean-up the biggest ones.


skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -418,13 +428,13 @@ def _compute_data_for_display(
             The estimator.
 
         estimator_name : str
-            The estimator name to attach to the display data.
+            The estimator name. Accepted for compatibility with display builders.
```
GaetandeCast: You can actually remove this parameter now, since it's not used.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -474,29 +499,20 @@ def _compute_data_for_display(
...
-                            label,
-                        ],
-                        # metadata:
-                        label=label,
-                        split=None,
-                        estimator=estimator_name,
-                        data_source=data_source,
-                    )
+                cm = np.column_stack([tns, fps, fns, tps]).reshape(-1, 2, 2).astype(int)
+                df = cls._build_confusion_frame(
```
GaetandeCast, line 498: @auguste-probabl would be mad at the amount of one-time-use local variables that could be inlined x)

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
```diff
@@ -645,25 +662,20 @@ def select_threshold_and_format(group):
...
-        groupby_cols = []
-        if "cross-validation" in self.report_type:
-            groupby_cols.append("split")
-        if "comparison" in self.report_type:
-            groupby_cols.append("estimator")
-        if label is None:
-            groupby_cols.append("label")
+            return group.query(f"threshold == {closest_threshold_value}")
 
+        groupby_cols = df.columns.intersection(
```
GaetandeCast, line 662: cool shortcut :thumbsup:

skore/tests/unit/displays/confusion_matrix/test_common.py
```diff
@@ -331,16 +377,30 @@ def test_data_source_both_is_not_supported(binary_classification_data):
         report.metrics.confusion_matrix(data_source="both")
 
 
-def test_plot_threshold_requires_label(pyplot, binary_classification_data):
-    """Check that plot raises when threshold_value is set but label is None."""
+def test_frame_threshold_label_none_returns_all_labels(binary_classification_data):
```
GaetandeCast, line 392: I don't really understand the changes to this test. Coverage is going to go down without the old version.

skore/tests/unit/displays/confusion_matrix/test_common.py
```diff
@@ -331,16 +377,30 @@ def test_data_source_both_is_not_supported(binary_classification_data):
...
+):
+    """Check that numeric labels can select OvR heatmaps."""
+    X, y = multiclass_classification_data
+    report = evaluate(LogisticRegression(), X, y, splitter=0.2)
+
+    display = report.metrics.confusion_matrix()
+    fig = display.plot(threshold_value=0.5, label=1)
+
+    heatmap_values = fig.axes[0].collections[0].get_array()
+    assert not np.ma.is_masked(heatmap_values)
```
GaetandeCast: I'm not sure it's very relevant to test this.

skore/tests/unit/displays/confusion_matrix/test_common.py
```diff
@@ -331,16 +377,30 @@ def test_data_source_both_is_not_supported(binary_classification_data):
...
+
+def test_multiclass_thresholded_plot_with_numeric_label(
+    pyplot, multiclass_classification_data
+):
+    """Check that numeric labels can select OvR heatmaps."""
+    X, y = multiclass_classification_data
+    report = evaluate(LogisticRegression(), X, y, splitter=0.2)
+
+    display = report.metrics.confusion_matrix()
+    fig = display.plot(threshold_value=0.5, label=1)
```
GaetandeCast: maybe we should test with and without a threshold value, to also cover what you added with this PR.

GaetandeCast (commented):
Thanks a lot! Just a few remarks otherwise it's okay for me.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
cakedev0: Indeed thx.

skore/src/skore/_sklearn/_plot/metrics/confusion_matrix.py
cakedev0: :sweat_smile: I can inline more if you want.

I personally think that this way is more aligned with how human brain works, though it's indeed a bit less elegant.

skore/tests/unit/displays/confusion_matrix/test_common.py
cakedev0: Yes sorry, I trusted codex too much with the tests.

skore/tests/unit/displays/confusion_matrix/test_common.py
cakedev0: Added back the old test (and kept the new one, as I think it does cover some code paths that aren't covered otherwise)

skore/tests/unit/displays/confusion_matrix/test_common.py
cakedev0: Good idea, done.

skore/tests/unit/displays/confusion_matrix/test_common.py
cakedev0: Coverage back at 100% for confusion_matrix.py

GaetandeCast (approved):
LGTM, thanks!

PR #2842, opened 2026-04-29
fix(skore): Correct unbalanced div tags returned by scikit-learn

### LINKED ISSUE #2801: html repr of reports breaks when model is a meta estimator on a Pipeline
### Describe the bug

The html representation of reports breaks (the contents of the Diagnostic, Estimator and Data tabs all appear in the Estimator tab) when the report was built for a meta estimator on a `Pipeline`.

It is very much a corner case but it happened to me when using skore in a quite natural way, see code below. It is probably bad practice to feed the whole `Pipeline` to the meta estimator (here `TunedThresholdClassifierCV`) but with skrub's `tabular_pipeline` it happened without me realizing it. The correct way is to do `tabular_pipeline(TunedThresholdClassifierCV(HistGradientBoostingClassifier()))` in which case the repr is not broken.

Note: I tried with other meta estimators, `FixedThresholdClassifier` and `GridSearchCV` and it also broke.
 
### Steps/Code to Reproduce

```python
import skore
import skrub
from sklearn.model_selection import TunedThresholdClassifierCV

X, y = load_breast_cancer(return_X_y=True, as_frame=True)

first_report = skore.evaluate(
    skrub.tabular_pipeline("classification").set_params(
        histgradientboostingclassifier__random_state=0
    ),
    X,
    y,
    splitter=5,
)

second_report = skore.evaluate(
    TunedThresholdClassifierCV(
        first_report.estimator,
        scoring="neg_brier_score",
        store_cv_results=True,
        random_state=0,
    ),
    X,
    y,
    splitter=0.2,
)
second_report
```

### Expected Behavior

The usual html repr.

### Actual Behavior

https://github.com/user-attachments/assets/dc342fad-b798-48ee-b136-3c25893b94e2

### Environment

```shell
System:
    python: 3.14.0 (main, Nov 19 2025, 22:48:15) [Clang 21.1.4 ]
executable: /home/gdc/dev/skore/.venv/bin/python3
   machine: Linux-6.17.0-20-generic-x86_64-with-glibc2.39

Python dependencies:
        skore: 0.0.0+unknown
          pip: None
       jinja2: 3.1.6
       joblib: 1.5.3
   matplotlib: 3.10.8
        numpy: 2.4.4
       pandas: 3.0.2
       plotly: 6.6.0
         rich: 14.3.3
 scikit-learn: 1.8.0
      seaborn: 0.13.2
 skore[local]: None
        skrub: 0.8.0
skore-hub-project: 0.0.0+unknown
xgboost-cpu; (platform_system: None
xgboost; (platform_system: None
```

glemaitre:
closes #2801 

This PR tries to get around the bug in scikit-learn for which the HTML repr as unalanced `<div>` tag.

For the minimal reproducer in #2801, you get:

image

### DIFF
--- a/skore/src/skore/_utils/repr/utils.py
+++ b/skore/src/skore/_utils/repr/utils.py
@@ -18,7 +18,7 @@ def repair_estimator_html_for_slotted_host(html: str) -> str:
     following slotted elements inside the estimator subtree, so they no longer
     receive slot assignment and can appear under the wrong tab.
 
-    This function compares counts of ``<div`` vs ``</div>`` on a script-stripped copy
+    This function compares counts of ``<div>`` vs ``</div>`` on a script-stripped copy
     (so ``</div>`` inside ``<script>`` is ignored) and appends the deficit of closing
     ``</div>`` to the original string.
     """
--- a/skore/src/skore/_utils/repr/utils.py
+++ b/skore/src/skore/_utils/repr/utils.py
@@ -1,0 +1,31 @@
+"""Miscellaneous helpers for HTML repr and embedding (slots, fragments)."""
+
+from __future__ import annotations
+
+import re
+
+# Strip script bodies when counting <div> so strings like "</div>" in JS do not skew.
+_HTML_SCRIPT_RE = re.compile(r"<script\b[^>]*>.*?</script>", re.DOTALL | re.IGNORECASE)
+
+
+def repair_estimator_html_for_slotted_host(html: str) -> str:
+    """Append missing closing ``</div>`` tags so sibling slotted nodes stay valid.
+
+    Estimator HTML from ``estimator_._repr_html_()`` is concatenated in the light DOM
+    next to ``<div slot="table-report">`` and ``<div slot="diagnostic">`` on the
+    same shadow host. Named slots only consider **direct children** of that host; if
+    the sklearn fragment leaves ``<div>`` elements unclosed, the HTML parser nests the
+    following slotted elements inside the estimator subtree, so they no longer
+    receive slot assignment and can appear under the wrong tab.
+
+    This function compares counts of ``<div`` vs ``</div>`` on a script-stripped copy
+    (so ``</div>`` inside ``<script>`` is ignored) and appends the deficit of closing
+    ``</div>`` to the original string.
+    """
+    without_scripts = _HTML_SCRIPT_RE.sub("", html)
+    n_open = len(re.findall(r"<div\b", without_scripts))
+    n_close = len(re.findall(r"</div>", without_scripts))
+    deficit = n_open - n_close
+    if deficit > 0:
+        return f"{html}{'</div>' * deficit}"
+    return html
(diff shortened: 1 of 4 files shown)


skore/src/skore/_utils/repr/utils.py
auguste-probabl, line 1: Can you reference https://github.com/scikit-learn/scikit-learn/issues/33887 so that we can drop this code once it gets closed?

skore/src/skore/_utils/repr/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    same shadow host. Named slots only consider **direct children** of that host; if
+    the sklearn fragment leaves ``<div>`` elements unclosed, the HTML parser nests the
+    following slotted elements inside the estimator subtree, so they no longer
+    receive slot assignment and can appear under the wrong tab.
+
+    This function compares counts of ``<div`` vs ``</div>`` on a script-stripped copy
+    (so ``</div>`` inside ``<script>`` is ignored) and appends the deficit of closing
+    ``</div>`` to the original string.
+    """
+    without_scripts = _HTML_SCRIPT_RE.sub("", html)
```
auguste-probabl:
```suggestion
    without_scripts = re.sub(_HTML_SCRIPT_RE, "", html)
```

skore/src/skore/_utils/repr/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    """Append missing closing ``</div>`` tags so sibling slotted nodes stay valid.
+
+    Estimator HTML from ``estimator_._repr_html_()`` is concatenated in the light DOM
+    next to ``<div slot="table-report">`` and ``<div slot="diagnostic">`` on the
+    same shadow host. Named slots only consider **direct children** of that host; if
+    the sklearn fragment leaves ``<div>`` elements unclosed, the HTML parser nests the
+    following slotted elements inside the estimator subtree, so they no longer
+    receive slot assignment and can appear under the wrong tab.
+
+    This function compares counts of ``<div`` vs ``</div>`` on a script-stripped copy
```
auguste-probabl:
```suggestion
    This function compares counts of ``<div>`` vs ``</div>`` on a script-stripped copy
```

auguste-probabl (changes requested):
I don't love that add so much logic + tests just for something that should be fixed in sklearn, so please take care to reference the upstream issue as much as possible

glemaitre:
> I don't love that add so much logic + tests just for something that should be fixed in sklearn, so please take care to reference the upstream issue as much as possible

It will be fixed upstream but since it will not be backported for the previous version of scikit-learn, we need to fix here.

skore/src/skore/_utils/repr/utils.py
glemaitre: It is not the same bug. It is another one that has not been released yet.

For the current one, I don't think that we have a specific PR in scikit-learn for it. It could have been handled while working in some independent feature as a side-effect.

Basically, I need that https://github.com/scikit-learn/scikit-learn/issues/33887 is fixed to be sure that the bug is not existing anymore.

auguste-probabl (approved):

Issue #2820, opened 2026-04-24
feat(skore,skore-*-project): Use `tqdm` as progress bar as default in WASM


glemaitre:
### What would you like to say?

As part to support WASM, we need to rely on TQDM (https://github.com/tqdm/tqdm) since rich is creating threads when doing progress bar.

thomass-dev:
Note that we don't need to use `tqdm` in replacement of `rich`: we need to disable the `auto_refresh` feature of `rich` when threading is not available.

PR #2847, opened 2026-04-29
fix(skore/progressbar): Disable `auto_refresh` mode in Jupyterlite where threading is not available

thomass-dev:
Partially addresses https://github.com/probabl-ai/skore/issues/2820.

### DIFF
--- a/skore/src/skore/__init__.py
+++ b/skore/src/skore/__init__.py
@@ -73,6 +73,7 @@ __all__ = [
     "PredictionErrorDisplay",
     "Project",
     "RocCurveDisplay",
+    "THREADABLE",
     "TableReportDisplay",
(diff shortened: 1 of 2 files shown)


skore/src/skore/__init__.py
```diff
@@ -98,3 +99,15 @@
...
+try:
+    from threading import Thread
+
+    thread = Thread()
+    thread.start()
+    thread.join()
+except Exception:
+    THREADABLE = False
+else:
+    THREADABLE = True
```
glemaitre, line 113: Do you think it would be better just to have a small utils function `_is_threadable` that returns the boolean instead to have the global?

glemaitre (approved):
LGTM.

PR #2869, opened 2026-05-04
feat(skore): Add `verbose_name` argument to `add`

### LINKED ISSUE #2784: Don't modify the name I give to my custom metric
When logging a custom metric with a custom name, e.g. :
```python
second_report.metrics.add(custom_scorer, name="Score with FP penalty")
```
it appears as "Score With Fp Penalty" in the `metrics.summarize`. I would like if it was not capitalized automatically.

Image

auguste-probabl:
#### Change description

See title. Closes #2784 

#### Contribution checklist

requires updating tests)

#### AI usage disclosure

AI tools were not involved.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -72,6 +72,11 @@ Added
 - :class:`Project` now accepts strings for the `workspace` argument. See :pr:`2832` by
   :user:`auguste-probabl`.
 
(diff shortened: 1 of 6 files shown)


GaetandeCast (approved):
LGTM. I think that it is the correct fix for the moment, we might revisit it when we converge on how to expose registered custom metrics outside summarize.

PR #2876, opened 2026-05-06
feat(skore-hub-project): Allow users to upload binary classification without `pos_label`

### LINKED ISSUE #2870: skore-hub-project: Send plots for both positive labels
After the merge of #2663 user are not forced to set there positive label. Hence hub project must now compute ROC/PR/CM for both labels and send it to the hub.

rouk1:
~Also contains a fix to make mypy happy.~

Closes #2870


skore-hub-project/pyproject.toml
```diff
@@ -112,6 +112,9 @@ convention = "numpy"
 [tool.mypy]
 exclude = ["hatch/", "tests/"]
 strict = true
+# Resolve `skore` from the monorepo so `from skore import THREADABLE` matches this tree
+# (the PyPI wheel has no `py.typed` and pre-commit installs it outside this repo).
+mypy_path = ["../skore/src"]
```
thomass-dev: This should not be done, please revert.
You are introducing an error: you are testing the code of `skore-hub-project` with a version of `skore` that is not released, while `skore-hub-project` can be released before/independently of `skore`.

`THREADABLE` is available in `skore>=0.17.0`, and `skore` is in the test dependencies.

It should be an issue with your environment, maybe play with `pre-commit clean` (the CI is green on main).

skore/src/skore/__init__.py
```diff
@@ -101,13 +101,15 @@
 )
 
 
+# Whether Rich live/progress can safely auto-refresh while worker threads run.
+# Initialized before the probe so type checkers see a definite binding (strict mypy).
+THREADABLE: bool = False
```
thomass-dev: mypy is not in strict mode for `skore/`.

You should not force strict mode for your entire IDE, and take in account the `pyproject.toml` for each directory. You will have plenty of errors otherwise.

skore/src/skore/__init__.py
thomass-dev, line 1: Let's do the opposite please: by default `THREADABLE = True`, in the except block `THREADABLE = False`.

skore/src/skore/__init__.py
```diff
@@ -101,13 +101,15 @@
 )
 
 
+# Whether Rich live/progress can safely auto-refresh while worker threads run.
+# Initialized before the probe so type checkers see a definite binding (strict mypy).
```
thomass-dev:
```suggestion
# Whether threading is available or not.
```

The comment brings no more value than the name of the variable.

skore-hub-project/pyproject.toml
thomass-dev: Note that, this question of "what to test with what" will no longer exist with mono-package.

thomass-dev (changes requested):

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: With your PR, this exception (or at least its message) has no more sense: `pos_label` is never none.

thomass-dev: Is the whole `if` block still pertinent? cc @glemaitre

thomass-dev:
For sphinx 🔴 , please rebase with main.

skore-hub-project/src/skore_hub_project/project/project.py
thomass-dev: As discussed with @glemaitre , we drop the entire `if` block.

thomass-dev (approved):

PR #2877, opened 2026-05-06
feat(skore)!: Force arguments to be passed by keyword

### LINKED ISSUE #1991: api(skore): Use keyword only argument for most of the cases

direkkakkar319-ops:
Fixes #1991 

#### Change description
added `*` to the parameters to the public methods to make consistency throughout the codebase 
this pr also makes consistency across the code base as asked by @glemaitre in the review of PR #2326 (merged)

from this 
```python 
    def available(self, report_name: str | None = None) -> list[str]:
        """List available metric names in the registry
```
to 
```python
    def available(self, *, report_name: str | None = None) -> list[str]:
        """List available metric names in the registry
```

#### Contribution checklist
- [ ] Documentation was added or updated
- [x] The code passes our style conventions 
- [x] All the tests pass
- [x] The documentation builds and renders properly
- [x] A new changelog entry was added to CHANGELOG.rst
- [x] The pull request title respects the Conventional Commits convention


direkkakkar319-ops:
Hi there are few things that i need to ask 
1. `get_n_splits` and split in `TrainTestSplit` are not touched as they follow the sklearn cross-validatro is it fine ??
2. Should internal utilities like logger_context as i only done for the public methods only
3. do we need to add it to the _externals/ as it has some signatures where we can

direkkakkar319-ops:
do we need to run a deprecation cycle before it ??

auguste-probabl:
> 1. `get_n_splits` and split in `TrainTestSplit` are not touched as they follow the sklearn cross-validatro is it fine ??

That's correct

> 2. Should internal utilities like logger_context as i only done for the public methods only

It's a bit of a case-by-case thing, with a bias for making everything keyword arguments.

> 3. do we need to add it to the _externals/ as it has some signatures where we can

Code in `_external` is generally copied verbatim from other libraries, so we should not touch it.

auguste-probabl (approved):
Looks great, thanks!

direkkakkar319-ops:
Hi @auguste-probabl the pytest had timed out so this was removed from the queue

auguste-probabl (approved):

PR #2879, opened 2026-05-07
chore(skore/_sklearn_compat): Remove sections about `scikit-learn<1.5`

thomass-dev:
Simplify `_sklearn_compat.py` since we don't support anymore `scikit-learn<1.5` and `python<3.11`.

### DIFF
--- a/skore/src/skore/_externals/_sklearn_compat.py
+++ b/skore/src/skore/_externals/_sklearn_compat.py
@@ -1,5 +1,6 @@
 # ruff: noqa
 # mypy: ignore-errors
+
 """Ease developer experience to support multiple versions of scikit-learn.
 
 This file is intended to be vendored in your project if you do not want to depend on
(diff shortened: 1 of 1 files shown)


skore/src/skore/_externals/_sklearn_compat.py
glemaitre, line 1: I think that I would not bother changing the code here because it is a vendoring from https://github.com/sklearn-compat/sklearn-compat

So if we need to change something, we mainly copy-paste the new version and we should not include any coverage for it.

glemaitre: It is to avoid to depend on the package itself.

thomass-dev: Supporting "dead-code" is not the way we should go. Feel free to close this PR!

PR #2880, opened 2026-05-07
fix(skore): Fix estimator report for skrub learners

jeromedockes:
do not rely on to_estimator when the report was initialized with a skrub learner to avoid an error (see test)

we do not need the conversion here as the learner exposes the methods we check and the classes_ attribute


cakedev0 (approved):
Nice thx!

I had skrub 0.7.2 in my venv, that's probably why I used `to_estimator(...)` (otherwise .classes_ breaks)

PR #2883, opened 2026-05-07
feat(skore): Add high card mdi and high correlation checks

GaetandeCast:
Towards #2665.

Adds 
- SKD007 which raises a tip when at least one feature has cardinality higher than `0.5*n_sample` for a tree model (model that exposes `feature_importances_`)
- SKD008 which raises an issue when at least one pair of features has a Spearman correlation greater than 0.9.


skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -165,10 +170,7 @@ class CheckMetricsConsistencyAcrossSplits(Check):
     severity = "issue"
 
     def check_function(self, report: _BaseReport) -> str | None:
-        from skore._sklearn._cross_validation.report import CrossValidationReport
-
-        if not isinstance(report, CrossValidationReport):
```
jeromedockes, line 170: IIUC we do not need this check because `report_type` is inspected before invoking the check anyway?

skore/src/skore/_sklearn/_checks/utils.py
```diff
@@ -77,3 +78,66 @@ def detect_outliers_modified_zscore(scores, threshold=3):
 
 class CheckNotApplicable(Exception):
     """Raised when a check cannot run on the given report."""
+
+
+def _unwrap_estimator(estimator):
```
jeromedockes: I would recommend a less generic name, because there could be many sorts of wrapping going on (other meta-estimators, dataops, skore reports, ...) so maybe something like `split_pipeline` would make it clearer that it only works for Pipelines and what it does?

jeromedockes: also I guess this is a matter of personal habit but for me this function should not start with an underscore, as it is meant to be used by other modules. `_unwrap_estimator` makes it look like it is meant only for internal use by `utils.py`
conversely, if `utils.py` is internal to `_checks` and not meant to be used outside of  `_checks`, it could be renamed `_utils` to make that clear

skore/src/skore/_sklearn/_checks/utils.py
```diff
@@ -77,3 +78,66 @@ def detect_outliers_modified_zscore(scores, threshold=3):
...
+        when true and both train and test are available, return their
+        concatenation. Otherwise return train if available, else test.
+
+    Returns
+    -------
+    np.ndarray or None
+    """
+    from skore._sklearn._estimator.report import EstimatorReport
+
+    if not isinstance(report, EstimatorReport):
```
jeromedockes: would it make more sense to raise a typeerror here?

skore/src/skore/_sklearn/_checks/utils.py
```diff
@@ -77,3 +78,66 @@ def detect_outliers_modified_zscore(scores, threshold=3):
...
+
+    Splits sklearn :class:`~sklearn.pipeline.Pipeline` into its preprocessing
+    steps and final predictor.
+    """
+    if isinstance(estimator, Pipeline):
+        return estimator[:-1], estimator[-1]
+    return None, estimator
+
+
+def _get_data(report, *, target="X", concatenate=False) -> np.ndarray | None:
```
jeromedockes: get_preprocessed_data?

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -77,3 +78,66 @@ def detect_outliers_modified_zscore(scores, threshold=3):
...
+        data = train
+    elif test is not None:
+        data = test
+    else:
+        return None
+
+    if target == "X":
+        preprocessor, _ = _unwrap_estimator(report.estimator_)
+        if preprocessor is not None and len(preprocessor.steps) > 0:
+            data = preprocessor.transform(data)
```
jeromedockes, line 152: some of the preprocessors can require y

skore/src/skore/_sklearn/_checks/utils.py
```diff
@@ -77,3 +78,66 @@ def detect_outliers_modified_zscore(scores, threshold=3):
...
+    elif test is not None:
+        data = test
+    else:
+        return None
+
+    if target == "X":
+        preprocessor, _ = _unwrap_estimator(report.estimator_)
+        if preprocessor is not None and len(preprocessor.steps) > 0:
+            data = preprocessor.transform(data)
+    return np.asarray(data)
```
jeromedockes: do we need to convert to an array? for many dataframes it will be an extra copy. for dataframes that have non-numeric values it will be an object array btw

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -307,11 +290,94 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+
+        n_samples, n_features = X.shape
+        high_cardinality_features = [
+            idx
+            for idx in range(n_features)
+            if len(np.unique(X[:, idx])) > 0.5 * n_samples
+        ]
+
+        if high_cardinality_features:
+            names = ", ".join(str(s) for s in high_cardinality_features[:3])
```
jeromedockes, line 331: if X is a dataframe column names might be more informative than indices

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -307,11 +290,94 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+    severity = "issue"
+
+    def check_function(self, report: _BaseReport) -> str | None:
+        report = cast("EstimatorReport", report)
+        X = _get_data(report, target="X")
+
+        if X is None or X.shape[1] < 2:
+            raise CheckNotApplicable()
+
+        corr = np.abs(spearmanr(X).statistic)
```
jeromedockes, line 370: should we return None when we cannot compute correlations, eg input is not numeric?

sphinx/user_guide/automated_checks.rst
```diff
@@ -264,3 +264,83 @@ How to reduce the risk
...
+Highly correlated features carry largely redundant information. This redundancy
+can cause two problems:
+
+- **Unstable coefficients and importances**: in linear models, correlated
+  predictors share the explanatory burden, making individual coefficients
+  erratic across small data perturbations. Feature-importance methods such as
+  MDI and permutation importance inherit the same instability.
+- **Numerical issues**: near-perfect collinearity can inflate the variance of
+  coefficient estimates and, in extreme cases, make the design matrix
+  ill-conditioned.
```
jeromedockes: the check is done regardless of the model so maybe we should point out issues for non-linear models, if any?

jeromedockes (commented):
do you think it would be possible to restrict the situations where those checks are made, eg only check cardinality if user actually looks at mdi?

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: Yes indeed, [here](https://github.com/probabl-ai/skore/blob/58155b71ac020034118c2a3afa7559f281af27dd/skore/src/skore/_sklearn/_base.py#L66-L72) we make sure that the report's type matches the expected check report type.

skore/src/skore/_sklearn/_checks/utils.py
GaetandeCast: I looked at it and skore is pretty inconsistent when it comes to that convention. All helper files are called `utils.py` without the underscore and helper functions are sometimes preceded or not by an underscore, even when used outside their file. Anyways, I'm going with what you suggest.

skore/src/skore/_sklearn/_checks/utils.py
GaetandeCast: Actually I will just add typing to the signature to expect an `EstimatorReport` and remove this.

skore/src/skore/_sklearn/_checks/_utils.py
GaetandeCast: Hmm at `fit` yes but not at `transform` time afaik. The `Pipeline.transform` signature does not let me pass y.

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: Added filtering of only numeric types and raise `CheckNotApplicable` where there are less than 2.

skore/src/skore/_sklearn/_checks/utils.py
GaetandeCast: No we don't, I removed in favor of returning `np.ndarray | pd.DataFrame | pd.Series | None` and raising `CheckNotApplicable` when `None`. This let's us use operations such as `X.std` or `X.shape` safely later, without copying the data. However this means that checks that use the helper are not applicable for sparse data or `np.matrix` which is probably fine.

sphinx/user_guide/automated_checks.rst
GaetandeCast: So for any model it's going to make any feature importance measure unstable. Maybe I should split the check in an issue for linear models, where there is something to fix (high rank feature matrix) and a tip for all models to warn of the impact on feature importance?

GaetandeCast:
> do you think it would be possible to restrict the situations where those checks are made, eg only check cardinality if user actually looks at mdi

I thought about it a little and the main reason I did not do it yet is that it would require adapting the caching logic as it is currently independent of the report state, which we would change by let's say adding a flag that some accessor was called. So I think it's better to reconsider it later.
For now, since the mdi check is a tip and only applicable when the model has `feature_importances_` I think it is fine to always show it.

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -307,11 +292,96 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+
+    def check_function(self, report: _BaseReport) -> str | None:
+        report = cast("EstimatorReport", report)
+        _, predictor = split_preprocessor_estimator(report.estimator)
+        X = get_preprocessed_data(report, target="X")
+
+        if X is None or not hasattr(predictor, "feature_importances_"):
+            raise CheckNotApplicable()
+
+        X_df = X if isinstance(X, pd.DataFrame) else pd.DataFrame(X)
```
jeromedockes: if X is a numpy array this could make a copy, so maybe instead we should keep the indices if it is an array and use column names if it is a dataframe

sphinx/user_guide/automated_checks.rst
jeromedockes: I would keep the check as it is but in the first paragraph make it clearer that the feature importance issue is for all models not only linear. also maybe mention the fact that dropping redundant features could reduce memory usage and computation time for most models

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: done

sphinx/user_guide/automated_checks.rst
GaetandeCast: updated it to reflect this.

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -307,11 +292,102 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+
+    def check_function(self, report: _BaseReport) -> str | None:
+        report = cast("EstimatorReport", report)
+        _, predictor = split_preprocessor_estimator(report.estimator)
+        X = get_preprocessed_data(report, target="X")
+
+        if X is None or not hasattr(predictor, "feature_importances_"):
+            raise CheckNotApplicable()
+
+        if isinstance(X, (pd.DataFrame, pd.Series)):
```
jeromedockes:
```suggestion
        if isinstance(X, pd.DataFrame):
```

jeromedockes (approved):
thank you @GaetandeCast !

PR #2893, opened 2026-05-11
chore(skore): Type using `pandas-stubs`

thomass-dev:


auguste-probabl (commented):
This PR adds a (test) dependency and makes the code more complicated. Unless we can find a concrete problem that this PR fixes, I'd rather not merge it. I'm generally growing less fond of typing in skore in general, because we keep having to introduce hacks like `cast` to make mypy happy.

Issue #2884, opened 2026-05-07
Use the cv splitter and scoring of skrub learners


jeromedockes:
it is possible in skrub DataOps to specify a cross-validation splitter and a scoring metric (and any additional arguments they need):

```python
import skore
import skrub
from sklearn.dummy import DummyClassifier
from sklearn.model_selection import LeaveOneGroupOut

df = skrub.datasets.toy_products()
data = skrub.var("df", df)
groups = data["seller"]
# Configure the cv splitter
X = data[["description", "price"]].skb.mark_as_X(
    cv=LeaveOneGroupOut(), split_kwargs={"groups": groups}
)
sample_weight = X["price"]
y = data["category"].skb.mark_as_y()
pred = (
    X.skb.apply(DummyClassifier(), y=y)
    # Configure scoring
    .skb.with_scoring("accuracy")
    .skb.with_scoring(
        "accuracy",
        kwargs={"sample_weight": sample_weight},
        name="weighted_accuracy",
    )
)

cv = pred.skb.cross_validate() # splits grouping by seller, scores with and without sample weights
print(cv)
```

```
   fit_time  score_time  test_accuracy  test_weighted_accuracy
0  0.001563    0.003420       0.666667                0.594595
1  0.001237    0.002599       0.666667                0.909091
```

At the moment skore ignores this configuration (it overrides it with its own defaults).

```python
report = skore.evaluate(pred.skb.make_learner(), data={"df": df})
summary = report.metrics.summarize().frame()
print(summary)
```

```
                             SkrubLearner
                                     mean
Metric           Label
Accuracy                         0.500000
Precision        electronics     0.500000
                 tools           0.000000
Recall           electronics     1.000000
                 tools           0.000000
ROC AUC                          0.500000
Log loss                         0.836988
Brier score                      0.312500
Fit time (s)                     0.002137
Predict time (s)                 0.000943
```

It would be helpful if it could use the configured splitter by default when it exists, and either use the configured scorers or add them to the default ones in .metrics.summarize() (e.g. weighted_accuracy would show up in the summary above, and splitting would use the groups)

jeromedockes:
more generally we should probably also include the estimator's `score()` in the displayed metrics

jeromedockes:
[...]

PR #2898, opened 2026-05-12
feat: Compute `estimator.score` in `summarize`

auguste-probabl:
Create a new `Score` Metric to compute `estimator.score(X, y)`. This metric also takes into account the case where the estimator is a `skrub.DataOp`.

Note: For now, if the metric returns a dict, then there is no error but the metric names show up in the Label column of `report.metrics.summarize`. This will be addressed later.

Addresses part of #2884

#### Contribution checklist

requires updating tests)

#### AI usage disclosure

AI tools were involved for:

### DIFF
--- a/skore/src/skore/_sklearn/metrics.py
+++ b/skore/src/skore/_sklearn/metrics.py
@@ -189,35 +189,7 @@ class Metric:
         if score is not None:
             return score
 
-        if self.function is None:
-            raise ValueError(f"Metric {self.name!r} has no scoring function.")
-
-        metric_params = inspect.signature(self.function).parameters
-        call_kwargs = merged_kwargs.copy()
-        if "pos_label" in metric_params and "pos_label" not in call_kwargs:
-            call_kwargs["pos_label"] = report.pos_label
-
-        if self.function_kind == FunctionKind.METRIC:
-            assert self.response_method is not None
-
-            _, y_true = report._get_data_and_y_true(data_source=data_source)
-            y_pred = report._get_predictions(
-                data_source=data_source,
-                response_method=self.response_method,
-                pos_label=call_kwargs.get("pos_label", None),
-            )
-
-            score = cast(MetricCallable, self.function)(y_true, y_pred, **call_kwargs)
-        elif self.function_kind == FunctionKind.SCORER:
-            data, y_true = report._get_data_and_y_true(data_source=data_source)
-            X = data["_skrub_X"]
-
-            score = cast(ScorerCallable, self.function)(
-                report.estimator_,
-                X,
-                y_true,
-                **call_kwargs,
-            )
+        score = self._compute(report=report, data_source=data_source, **merged_kwargs)
 
         if isinstance(score, np.ndarray):
             score = cast(NDArray, score).tolist()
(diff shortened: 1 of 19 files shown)


skore/src/skore/_sklearn/metrics.py
```diff
@@ -668,6 +689,7 @@ def __call__(
     Mape(),
     FitTime(),
     PredictTime(),
+    Score(),
```
auguste-probabl: Not sure if it should be first or last

skore/src/skore/_sklearn/metrics.py
```diff
@@ -654,6 +654,27 @@ def __call__(
...
+    @staticmethod
+    def available(report: EstimatorReport) -> bool:
+        return hasattr(report.estimator_, "score")
+
+    @staticmethod
+    def function(estimator, X, y, *args, **kwargs):
+        if hasattr(estimator, "__skrub_to_Xy_pipeline__"):
+            # SkrubLearner.score expects an env dict, not (X, y); swap to the
+            # sklearn-style pipeline (empty env: X and y are passed directly).
+            estimator = estimator.__skrub_to_Xy_pipeline__({})
```
auguste-probabl: @jeromedockes I like this impl because it fits cleanly in the `Metric` interface (no need to override `Metric.__call__`, but I'm a bit out of my depth with regards to the skrub API.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: @jeromedockes i'll wait for your opinion before I fix the tests, it's tedious

skore/src/skore/_sklearn/metrics.py
jeromedockes: this will not work because there can be more than X and y in the environment. rather than relying on function here we should override `__call__` with something along the lines of 

```
def __call__(self, report):
    return report._estimator.score(report.test_data)
```

(plus handling of data source etc)

skore/src/skore/_sklearn/metrics.py
jeromedockes: i guess maybe first? if it is the default scikit-learn one it will be just 1 column so will not shift the rest by much. if it is a scorer set by the user eg on a dataop it is likely more relevant than the default ones to the user

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Done

skore/src/skore/_sklearn/metrics.py
auguste-probabl: Done

auguste-probabl:
Tests are failing because we rely on `DataOp.skb.with_scoring` which just came out in skrub 0.9, while CI is running skrub 0.8.

skore/tests/unit/reports/estimator/metrics/test_summarize.py
```diff
@@ -402,3 +414,98 @@ def test_data_source_both(forest_binary_classification_data):
...
+    assert set(score_labels) == {"accuracy", "weighted_accuracy"}
+
+
+def test_score_skrub_learner_with_extra_env_vars():
+    """``score`` works when the DataOp env has variables beyond X and y.
+
+    Non-regression test: the previous implementation invoked
+    ``__skrub_to_Xy_pipeline__({})`` with an empty environment, which dropped
+    any additional variables referenced by the DataOp (e.g. a parent table
+    from which X and y are derived via ``mark_as_X``/``mark_as_y``).
```
jeromedockes: the docstring is correct but in the specific example below it would still work, because the 'df' var is only used to compute X and y so when X and y are provided it is not required. but it is probably not worth the effort to make a more complex example

jeromedockes (dismissed):
LGTM, follow-up is to have the different scores as separte columns rather than showing up as labels

jeromedockes:
> Tests are failing because we rely on DataOp.skb.with_scoring which just came out in skrub 0.9, while CI is running skrub 0.8.

to clarify: it is required for the added tests to run because they check the behavior when with_scoring is used, but skore itself still works with 0.8 . still the easiest thing to do is probably bump the skrub dependency

auguste-probabl:
Waiting for #2913

skore/tests/unit/reports/estimator/metrics/test_summarize.py
auguste-probabl: thanks, i changed the test a bit to be more thorough

auguste-probabl:
I didn't think to add a `score` method to MetricsAccessor.

auguste-probabl:
@jeromedockes Added `report.metrics.score()`.

jeromedockes (dismissed):

jeromedockes (approved):

PR #2905, opened 2026-05-13
feat(skore): Inspectable local export

jeromedockes:
see discussion #2730  
```python
import skore
from skore import _export as e

e.init_data_dir()

X, y = make_regression(random_state=0)
X_train, X_test, y_train, y_test = train_test_split(X, y, shuffle=False)
report = skore.EstimatorReport(
    DummyRegressor(), X_train=X_train, X_test=X_test, y_train=y_train, y_test=y_test
)
path = e.export(report, name="dummy_regressor")
loaded = e.load(path)
print(loaded.metrics.r2())
```

```
❯ tree skore_data
skore_data
├── datasets
│   ├── 7d5bff15810e66661885935a7fa32c5c.joblib
│   ├── d21d0b5340551abab2f4a637ef777027.joblib
│   ├── d9a8e18d72b8483fd6befc69f72ec620.joblib
│   └── fbe12b2ec185cc6b86a1bda843ed967e.joblib
└── projects
    └── my_project
        └── reports
            ├── 2026-06-26T09-12-16.363332+00-00__id_a06b5f3812f64f6392bf23048084d2cc__estimator__dummy
            │   ├── checks
            │   │   ├── _check_results_cache.json
            │   │   └── summarize.csv
            │   ├── data
            │   │   ├── test_data.json
            │   │   └── train_data.json
            │   ├── estimator.pickle
            │   ├── estimator_.pickle
            │   ├── inspection
            │   │   └── permutation_importance
            │   │       └── permutation_importance__test__0__None
            │   │           ├── cache_key.json
            │   │           ├── importances.csv
            │   │           └── kwargs.json
            │   ├── learner_.pickle
            │   ├── metadata.json
            │   ├── metrics
            │   │   ├── cache.pickle
            │   │   ├── registry.pickle
            │   │   └── summarize.csv
            │   ├── predictions
            │   │   └── test__predict.joblib
            │   ├── state.json
            │   └── user
            │       └── README
            └── latest__dummy -> /home/jerome/workspace/scratch/2026/06/skore_data/projects/my_project/reports/2026-06-26T09-12-16.363332+00-00__id_a06b5f3812f64f6392bf23048084d2cc__estimator__dummy

15 directories, 21 files
```


thomass-dev:
I would like to review this first POC, can you please wait for my return? Thanks!

jeromedockes:
> I would like to review this first POC, can you please wait for my return? Thanks!

sure! as I mentioned it is mostly intended to give a more concrete basis to the discussion, we will not commit to any decisions before you are back

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
+                display.importances.to_csv(display_dir / "importances.csv", index=False)
+                (display_dir / "kwargs.json").write_text(json.dumps(kwargs), "UTF-8")
+                (display_dir / "cache_key.json").write_text(
+                    json.dumps((data_source, "permutation_importance", kwargs))
+                )
+            case _:
+                pass
+
+
+def _dump_estimator_report(
```
auguste-probabl:
```suggestion
def _write_estimator_report(
```

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
+    predictions_dir = output_dir / "predictions"
+    predictions_dir.mkdir(exist_ok=True)
+    for (subset_name, meth_name), val in report.to_dict()["predictions"].items():
+        with open(predictions_dir / f"{subset_name}__{meth_name}.joblib", "wb") as f:
+            joblib.dump(val, f)
+
+    return output_dir
+
+
+def load_report_metadata(report_dir: Path) -> dict[str, Any]:
```
auguste-probabl:
```suggestion
def read_report_metadata(report_dir: Path) -> dict[str, Any]:
```

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
+        reports_path = self.path / "reports"
+        result = []
+        for p in sorted(reports_path.glob("*__id_*")):
+            if p.is_symlink():
+                continue
+            try:
+                result.append(load_report_metadata(p))
+            except Exception as e:
+                warnings.warn(f"Failed to load report at {p}: {e!r}", stacklevel=2)
+                raise
```
auguste-probabl:
```suggestion
                raise Exception(f"Failed to load report at {p}") from e
```
with the right choice of Exception class

skore/tests/unit/plugins/local/test_project.py
```diff
@@ -63,380 +72,87 @@ def cv_binary_classification() -> CrossValidationReport:
...
+        d.mkdir(parents=True)
+    monkeypatch.setattr(Path, "home", lambda: home)
+    monkeypatch.chdir(pwd)
+    assert Project("regression").path == home / "skore_data" / "projects" / "regression"
+    Project("abc", workspace=local / "skore_data")
+    assert (
+        Project("regression").path == local / "skore_data" / "projects" / "regression"
+    )
+    monkeypatch.setenv("SKORE_WORKSPACE", str(env))
+    assert Project("regression").path == env / "projects" / "regression"
```
auguste-probabl, line 177: these should be several tests, it's hard to understand what is tested

skore/pyproject.toml
```diff
@@ -101,6 +100,9 @@ local = "skore._plugins.local:Project"
 hub = "skore._plugins.hub.project.project:Project"
 mlflow = "skore._plugins.mlflow:Project"
 
+[project.scripts]
+skore-init-workspace = "skore._plugins.local.project:init_workspace"
+
```
auguste-probabl: do we want this? if yes, when/how is it called? can we have some docs on it?

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
+
+def find_workspace() -> Path:
+    if env_workspace := os.environ.get("SKORE_WORKSPACE"):
+        return init_workspace(Path(env_workspace))
+    start = Path(".").resolve()
+    for candidate in [start, *start.parents[::-1]]:
+        workspace = candidate / "skore_data"
+        if workspace.is_dir() and (workspace / ".SKORE_WORKSPACE").exists():
+            return workspace
+    return init_workspace(Path.home() / "skore_data")
```
auguste-probabl: I'd rather keep the current default location, it follows the XDG base directory spec:

> A ``skore/`` directory in the user data directory:
>
> - on Windows, usually ``C:\Users\%USER%\AppData\Local\skore``,
> - on Linux, usually ``${HOME}/.local/share/skore``,

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
-        | The workspace can be shared between all the projects.
-        | The workspace can be set using kwargs or the environment variable
-          ``SKORE_WORKSPACE``.
-        | If not, it will be by default set to a ``skore/`` directory in the user
-          cache directory:
-
-        - on Windows, usually ``C:\Users\%USER%\AppData\Local\skore``,
-        - on Linux, usually ``${HOME}/.local/share/skore``,
-        - on macOS, usually ``${HOME}/Library/Application Support/skore``.
-    """
```
auguste-probabl, line 106: we need docs, especially to document where the workspace is found/created.

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
-        # Prune artifacts not related to a project
-        for artifact in artifacts:
-            if artifact not in remaining_artifacts:
-                del artifacts[artifact]
-
-        # Delete the project
-        del projects[name]
+    def delete(*, name: str, workspace: str | Path) -> None:
+        workspace = Path(workspace)
+        if not (workspace / ".SKORE_WORKSPACE").exists():
```
auguste-probabl, line 501: kind of unrelated: i find it weird that this takes a `workspace` argument

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
-            | The workspace can be shared between all the projects.
-            | The workspace can be set using kwargs or the environment variable
-              ``SKORE_WORKSPACE``.
-            | If not, it will be by default set to a ``skore/`` directory in the USER
-              cache directory:
-
-            - on Windows, usually ``C:\Users\%USER%\AppData\Local\skore``,
-            - on Linux, usually ``${HOME}/.cache/skore``,
-            - on macOS, usually ``${HOME}/Library/Caches/skore``.
-        """
```
auguste-probabl, line 303: don't delete the docs completely

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
-
-        # Prune artifacts not related to a project
-        for artifact in artifacts:
-            if artifact not in remaining_artifacts:
-                del artifacts[artifact]
-
-        # Delete the project
-        del projects[name]
+    def delete(*, name: str, workspace: str | Path) -> None:
+        workspace = Path(workspace)
```
auguste-probabl: would be nice for it to work with `~`. use `expanduser`. we can deal with this later.

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
-
-        if project.name not in project._Project__projects_storage:
-            raise RuntimeError(
-                f"Skore could not proceed because {project!r} does not exist anymore."
+import numpy as np
+import pandas as pd
+
+from ... import CrossValidationReport, EstimatorReport
+from ..._utils._cache_key import deep_key_sanitize
+
```
auguste-probabl: i would prefer more qualified paths

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
+
+def _check_name(name: Any) -> str:
+    return re.sub(r"[^-_.a-zA-Z0-9]", "_", str(name)) or "_"
```
auguste-probabl, line 21: this can return `"_"`, is this on purpose?

auguste-probabl (changes requested):
First round of reviews. Looks great so far, although Project methods need docstrings

skore/src/skore/_plugins/local/project.py
glemaitre: Here, we are dumping file that are user documents and it is really weird that they are inside an hidden folder `.local` that does not make it easy to find them. I think that the user data directory is more intended for our application data like token or whatever private thing we can have.

The current report make more sense to be in the user documents directory as mentioned by `platformdirs`.

skore/src/skore/_plugins/local/project.py
jeromedockes: I don't think we want to turn the whole project unusable when there is one corrupted report in it so it is probably best to have a warning rather than an exception here, until we have more robust error recovery and utils for managing the contents of the workspace

skore/pyproject.toml
jeromedockes: I think it can be useful so you can do `skore-init-workspace` in your repository and it will create a workspace there, then by default skore local projects will use this workspace when python is running from inside the repo -- it is a bit the skore equivalent of git init

however I will change the command to `skore local init` according to IRL discussion with @glemaitre to be more consistent with the skore-cli commands

skore/src/skore/_plugins/local/project.py
jeromedockes: that is the interface imposed by skore.Project to the plugins

skore/src/skore/_plugins/local/project.py
jeromedockes: yes to ensure it is never empty

skore/tests/unit/plugins/local/test_project.py
jeromedockes: I added comments to explain each step but each time we check that the new workspace takes priority over the previous one so I think there would be a lot of repetition between the tests

skore/src/skore/_plugins/local/project.py
auguste-probabl: I'm confused; the code already has a `raise` here that will terminate the program

skore/tests/unit/plugins/local/test_project.py
auguste-probabl: comments make it a lot better already, thanks

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
+    project_dir = _init_project_dir(workspace, project_name)
+    reports_dir = project_dir / "reports"
+    date_str = str(report._metadata["creation-date"]).replace(":", "-")
+    name_str = "" if name is None else f"__{_check_name(name)}"
+    output_dir = (
+        reports_dir / f"{date_str}__id_{report._metadata['id']:x}__"
+        f"{report._metadata['report_type']}{name_str}"
+    )
+    output_dir.mkdir(parents=True, exist_ok=True)
+    symlink = reports_dir / f"latest{name_str}"
```
auguste-probabl:
```suggestion
    date_str = str(report._metadata["creation-date"]).replace(":", "-")
    name_str = "" if name is None else _check_name(name)
    filename = "__".join([
        date_str,
        f"id_{report._metadata['id']:x}",
        report._metadata['report_type'],
        name_str,
    ])
    output_dir = reports_dir / filename
    output_dir.mkdir(parents=True, exist_ok=True)
    symlink = reports_dir / ("__".join("latest", name_str))
```

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,327 +1,397 @@
...
+            case (
+                data_source,
+                "permutation_importance",
+                ("mapping", kwarg_items),
+            ), display:
+                kwargs = dict(kwarg_items)
+                display_dir = (
+                    importances_dir / "permutation_importance__"
+                    f"{data_source}__{_check_name(kwargs['at_step'])}"
+                    f"__{_check_name(kwargs['metric'])}"
```
auguste-probabl:
```suggestion
                    importances_dir / ("__".join([
						"permutation_importance",
	                    data_source,
						_check_name(kwargs['at_step']),
						_check_name(kwargs['metric']),
					]))
```

skore/pyproject.toml
auguste-probabl: will you change it in this PR?

jeromedockes: @glemaitre I added the skore local init but kept the other one for now so yo can try both. having 2 levels of subcommands when only 1 command is available feels a bit weird for now, can you lmk what you think?

skore/src/skore/_cli.py
```diff
@@ -0,0 +1,31 @@
+import argparse
+
+from skore._plugins.local.project import init_workspace
+
+
+def init_workspace_command(args):
+    workspace = init_workspace().resolve()
+    print(f"Skore workspace created: {workspace}")
```
auguste-probabl: use console.print instead

skore/pyproject.toml
```diff
@@ -176,7 +177,7 @@ select = [
     "UP",
     "RUF010", "RUF015",
 ]
-ignore = ["PERF203"]
+ignore = ["PERF203", "T201"]
```
auguste-probabl: revert once `print` was replaced with `console.print`; catching stray `print` is useful because they are used for debugging

skore/src/skore/_plugins/local/project.py
jeromedockes: :) that was here for debugging and I forgot to remove it : thanks

jeromedockes: I'll add a test

skore/src/skore/_plugins/local/project.py
auguste-probabl: > it is really weird that they are inside an hidden folder `.local` that does not make it easy to find them.

If we really want to make the files easy to find, we should put them in the project directory.

If we don't want to do that, the way I see it we have 2 cases:
- either the user uses the terminal, and in this case they know about the `.local/share` thing which every app uses, and they expect stuff to be there and will be surprised if it's not (that's me ^^),
- or the user doesn't use the terminal, and in that case looking in the home directory is as strange as looking in `.local/share`.

Regardless of where we put the files, we will have to document clearly where the user needs to look, and if we do that, I would rather "just follow the spec".

> I think that the user data directory is more intended for our application data like token or whatever private thing we can have.

You're thinking of `XDG_STATE_HOME`:

> The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
> - actions history (logs, history, recently used files, …)
> - current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

https://specifications.freedesktop.org/basedir/

skore/src/skore/_plugins/local/project.py
jeromedockes: added them. but note those are not visible to the user

skore/src/skore/_project/project.py
jeromedockes, line 52: TODO update this once a decision is made on default location

skore/src/skore/_plugins/local/project.py
```diff
@@ -353,6 +381,33 @@ def _get_data_ref(value: Any, workspace: Path) -> dict[str, str]:
...
+        resides. The project is in <workspace>/projects/<project.name>.
+
+        If not provided, a workspace is found or created according to those
+        rules (see find_workspace() in this module):
+
+        - if the SKORE_WORKSPACE environment variable is set, use that.
+        - otherwise look in the current working directory and its parent for a
+          skore workspace: a directory named 'skore_data' containing a file named
+          '.SKORE_WORKSPACE'. If found, use that.
+        - otherwise use a global default location: ~/skore_data/
```
auguste-probabl: may not be idiomatic for windows

skore/src/skore/_plugins/local/project.py
```diff
@@ -361,11 +416,25 @@ def __init__(self, name: str, workspace: str | Path | None = None):
...
+
+        If the key already exists, its last report is modified to point to this new
+        report, while keeping track of the report history.
+
+        Parameters
+        ----------
+        key : str
+            The key to associate with ``report`` in the local project.
+        report : skore.EstimatorReport | skore.CrossValidationReport
+            The report to associate with ``key`` in the local project.
```
auguste-probabl, line 455: add Returns

skore/src/skore/_plugins/local/project.py
```diff
@@ -192,136 +428,49 @@ def put(self, key: str, report: EstimatorReport | CrossValidationReport) -> None
...
-                key=key,
-            )
+        return _write_report(
+            report, workspace=self.workspace, project_name=self.name, name=key
         )
 
-    @ensure_project_is_not_deleted
-    def get(self, id: str) -> EstimatorReport | CrossValidationReport:
+    def get(self, report_id: str) -> EstimatorReport | CrossValidationReport:
         """Get a persisted report by its id."""
```
auguste-probabl:
```suggestion
        """Get a persisted report by its ID."""
```

skore/src/skore/_plugins/local/project.py
jeromedockes: yeah that we need to decide if we do it here or in another pr; currently in theory it should return None but I think all put should return some kind of location string like a path or URI

also I would add an attribute to reports to store the local path if they have one, so that after getting a report I can interact directly with its data for example to add stuff to that directory

skore/src/skore/_project/project.py
auguste-probabl: discussion at https://github.com/probabl-ai/skore/pull/2905#discussion_r3474818697

skore/src/skore/_plugins/local/project.py
jeromedockes: discussion at https://github.com/probabl-ai/skore/pull/2905#discussion_r3474818697

auguste-probabl: i meant the `~` and the forward-slashes may not mean anything on windows

jeromedockes: ah I see. I'll keep it in mind when updating this after the default location is decided

skore/src/skore/_plugins/local/project.py
glemaitre: > You're thinking of XDG_STATE_HOME:

Basically, my point is that I don't want to think about those specifications because some other people already thought about it and wrote a library for me for different operating system.

And here the winner is:

https://platformdirs.readthedocs.io/en/latest/platforms.html#user-documents-dir

Here is a detailed search for the different operating system and the reasoning:

<details>

I now have all three official sources. Here they are:

---

Here are the **official platform guidelines** directly linking to the primary sources:

### 🍎 macOS — Apple Developer Documentation

**[File System Programming Guide — "File System Overview"](https://developer.apple.com/library/archive/documentation/FileManagement/Conceptual/FileSystemProgrammingGuide/FileSystemOverview/FileSystemOverview.html)**

This is the canonical Apple guide. The relevant quote, straight from the page:

> *"Put user data in `Documents/`. User data generally includes any files you might want to expose to the user — anything you might want the user to create, import, delete or edit."*

And the warning:

> *"The files in the user's Documents and Desktop directories should reflect only the documents that the user created and works with directly. […] Those directories must never be used to store data files that your app creates and manages automatically."*

---

### 🪟 Windows — Microsoft Learn

**[Windows app restore: maximize the value of your app — Microsoft Learn](https://learn.microsoft.com/en-us/windows/apps/develop/windows-app-restore)**

The directly relevant quote:

> *"Keep it separate from app state — if your app produces user-generated content (files, sound clips, videos, etc.), then you should write that content to the Windows Known Folders (Documents, Pictures, Music, Videos, etc.). You should store data that's not machine-specific in the `FOLDERID_Documents` location; typically in a sub-folder."* [^1]

And for the complementary "don't use AppData for user data" side:

**[Store and retrieve settings and other app data — Microsoft Learn](https://learn.microsoft.com/en-us/windows/apps/develop/data/store-and-retrieve-app-data)**

> *"Don't use app data to store user data or anything that users might perceive as valuable and irreplaceable. We recommend that the user's libraries and Microsoft OneDrive be used to store this sort of information."* [^2]

---

### 🐧 Linux — freedesktop.org (XDG)

**[XDG Base Directory Specification — specifications.freedesktop.org](https://specifications.freedesktop.org/basedir/latest/)**

This is the official spec maintained by freedesktop.org. It defines `$XDG_DATA_HOME` for app-internal data and, via the companion **XDG User Directories** spec, `$XDG_DOCUMENTS_DIR` for user-visible documents. [^3]

The companion user directories spec (which defines `$XDG_DOCUMENTS_DIR`) is at:
**[XDG User Directories — ArchWiki (canonical reference)](https://wiki.archlinux.org/title/XDG_user_directories)** [^4]

**References**

[1]: [Windows app restore: maximize the value of your... | Microsoft Learn](https://learn.microsoft.com/en-us/windows/apps/develop/windows-app-restore?view=winui-2.1) (39%)
[2]: [Store and retrieve settings and other app data - Windows apps](https://learn.microsoft.com/en-us/windows/apps/develop/data/store-and-retrieve-app-data) (25%)
[3]: [XDG Base Directory Specification - specifications.freedesktop.org](https://specifications.freedesktop.org/basedir/latest/) (24%)
[4]: [XDG user directories - ArchWiki](https://wiki.archlinux.org/title/XDG_user_directories) (13%)

</details>

skore/pyproject.toml
thomass-dev:
```suggestion
```

skore/pyproject.toml
thomass-dev: I don't think it is a good idea to maintain two CLI from two different packages for the same purpose.

In state, the `skore` entrypoint will surely conflict with https://github.com/probabl-ai/skore-cli.
Can we choose the one to use/support and remove the other?

skore/src/skore/_plugins/local/project.py
thomass-dev: The `platformdirs.user_data_dir` directory was used to store "hidden" files, like the database itself.
These files were not intended to be browsed extensively, but it was important to store them in a central place so that you can retrieve all your XP wherever you are on your disk.

Since we will be providing explicit files, `platformdirs.user_document_dir` sounds good to me for lambda users.

thomass-dev (changes requested):
Review in progress.

skore/src/skore/_plugins/local/project.py
thomass-dev: > we will have to document clearly where the user needs to look

Definitely agree. We should keep something as before:

```
    workspace : Path
        The directory where the project (metadata and artifacts) are persisted.

        | The workspace can be shared between all the projects.
        | The workspace can be set using kwargs or the environment variable
          ``SKORE_WORKSPACE``.
        | If not, it will be by default set to a ``skore/`` directory in the user
          cache directory:

        - on Windows, usually ``C:\Users\%USER%\AppData\Local\skore``,
        - on Linux, usually ``${HOME}/.local/share/skore``,
        - on macOS, usually ``${HOME}/Library/Application Support/skore``.
```

auguste-probabl: > Basically, my point is that I don't want to think about those specifications because some other people already thought about it and wrote a library for me for different operating system.

Just to be clear, this is also what I believe I'm doing! `platformdirs` literally implements the spec that I'm pushing.

As for the choice of `user_documents_dir`, I note that the macOS docs cited in https://github.com/probabl-ai/skore/pull/2905#discussion_r3496688896 say

> Those directories [e.g. ~/Documents] must never be used to store data files that your app creates and manages automatically.

So we're pretty much ignoring that warning.

I still think `user_data_dir` is the more correct choice, and that the best choice would be to store files in the user's project directory (e.g. the one with a `.git`). That being said, `~/Documents` is an okay compromise.

thomass-dev: > and that the best choice would be to store files in the user's project directory

I don't think it should be the default because we would lose the ability to retrieve (by default) any XP wherever you are running your script on your disk. But that's another issue, and we've never agreed on it ^^.

thomass-dev: On the strict specification, i agree with @auguste-probabl that `user_data_dir` is the way to go, and that's what advanced users expect.

And for my personal use, if we change the default from `user_data_dir` to `user_document_dir`, i will always use `skore` with `user_data_dir`. I don't want my `~/Documents` to be polluted with something i don't control and might even delete or conflict with my personal files (this is not the common pattern).

But for lambda user, `~/Documents` sounds better.

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,185 +1,431 @@
...
+def _check_name(name: Any) -> str:
+    return re.sub(r"[^-_.a-zA-Z0-9]", "_", str(name)) or "_"
 
 
-class Project:
-    r"""
-    API to manage a collection of key-report pairs persisted in a local storage.
+def init_workspace(workspace_dir: str | Path | None = None) -> Path:
+    workspace_dir = (
+        (Path(".") / "skore_data" if workspace_dir is None else Path(workspace_dir))
```
auguste-probabl:
```suggestion
        (Path(".") / "skore" if workspace_dir is None else Path(workspace_dir))
```

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,185 +1,431 @@
...
+                not in (
+                    "estimator",
+                    "learner",
+                    "data",
+                    "predictions",
+                    "metric_registry",
+                    "optional",
+                    "split_indices",
+                    "estimator_reports",
+                )
```
auguste-probabl, line 160: i'd rather this be an allowlist than a block list

skore/src/skore/_plugins/local/project.py
auguste-probabl: do this in a later PR

skore/src/skore/_project/project.py
```diff
@@ -180,12 +182,16 @@ def __init__(self, name: str, *, mode: ProjectMode = "local", **kwargs):
...
+                  named ``skore_data`` containing a special marker file
+                  ``.SKORE_WORKSPACE``).
+                | If none is found a global default location in the user
+                  Documents directory is used:
 
-                - on Windows, usually ``C:\Users\%USER%\AppData\Local\skore``,
-                - on Linux, usually ``${HOME}/.local/share/skore``,
-                - on macOS, usually ``${HOME}/Library/Application Support/skore``.
+                    - on Linux, usually ``~/Documents/skore_data``,
+                    - on macOS, usually ``~/Documents/skore_data``,
```
rouk1: Note that since macOS Mojave/Catalina, Apple added TCC (Transparency, Consent, and Control) protections around a few user folders — Documents, Desktop, Downloads, iCloud Drive — under the "Files and Folders" privacy category. It's not tied to the Python binary itself; it's tied to the parent app actually doing the file I/O from the OS's point of view. So the first time your terminal app tries to write into ~/Documents, macOS pops up a dialog like:

"Terminal" would like to access files in your Documents folder.

Once you click Allow, it's remembered — you can check/manage it under System Settings → Privacy & Security → Files and Folders, where you'd see Terminal listed with a Documents toggle.
A few practical notes:

It's granted per terminal app, not per script. So once Terminal has permission, any Python script run from Terminal can write there without a new prompt.
If you deny it (or dismiss it), subsequent writes will just fail silently or throw a PermissionError/OSError in Python rather than re-prompting — macOS doesn't nag repeatedly.
If you ever run the same script from a different terminal app (say, VS Code's integrated terminal or a cron job), you'll get a separate prompt for that app.
Writing to ~/Documents/some_subfolder still falls under the same Documents grant — no extra prompts for subdirectories.

Also note that this folder is often back up to user's iCloud drive, writing large file here will be painful for users.
I recommend to fallback to a place with write access to avoid permission error handling.

jeromedockes: thanks for this information @rouk1 

about the fallback, I wonder if it could create a situation where a user saves reports in one location when they use say Terminal and in another one when they use VSCode and then they get confused. maybe we should not use documents then and look for some location where we are likely able to write and raise an error if we cannot :thinking: 

then i guess there is some tension between saving reports in a location where users will find them easily, and the risk that they get included in backups unintentionally. maybe we can do some things like adding a CACHEDIR.TAG file and setting appropriate xattr so some backup tools will ignore it. but also if we make storing the datasets optional the workspace contains mostly results that might require some effort to recreate if needed, so some users will want them backed up

thomass-dev: @rouk1 didn't you already have the issue with `${HOME}/Library/Application Support/skore`?

rouk1: User's Library folder is not among the list of protected folders (Documents, Desktop, Downloads, iCloud Drive).

thomass-dev: Isn't an argument in favor of `user_data_dir` over `user_document_dir`?

skore/src/skore/_plugins/local/project.py
Aljutor: Currently, skore cli is using .skore folder in the project dir, so I support this change to unify the layout

skore/src/skore/_project/project.py
rouk1: I do not express an opinion here, just saying that there will be a popup and we need to have a retry mechanism and to listen to folder permission changes if you chose `Documents`.

I think that apple users never navigate to the `Library` folder it is a for machine folder.

thomass-dev: Being the only one thinking that storing all the projects in a central place (a la PostgreSQL) is a better option than storing in the current directory (a la git), you should go for the second option.

Let's go with the majority's view.

skore/tests/unit/plugins/local/test_project.py
```diff
@@ -63,380 +78,120 @@ def cv_binary_classification() -> CrossValidationReport:
...
+def test_find_workspace(tmp_path, monkeypatch):
+    """Check the priority order of workspace lookup"""
+    env = tmp_path / "env"
+    local = tmp_path / "repo"
+    pwd = local / "a" / "b"
+
+    pwd.mkdir(parents=True)
+    monkeypatch.chdir(pwd)
+
+    # TODO: check git repo (if we decide to keep that part)
```
jeromedockes: This test should be updated / completed once we settle on the exact behavior for  choosing the storage location

skore/src/skore/_plugins/local/project.py
jeromedockes: it is a list of things we manually write in a different way

skore/src/skore/_project/git.py
```diff
@@ -92,3 +92,19 @@ def git_commit() -> str | None:
...
+        result = subprocess.run(
+            ["git", "rev-parse", "--show-toplevel"],
+            capture_output=True,
+            text=True,
+            check=True,
+        )
+    except subprocess.CalledProcessError:
+        return None
+    path = result.stdout.strip()
+    return path or None
```
thomass-dev:
```python
from pathlib import Path

[...]

def git_repo_root() -> Path | None:
    """Get the root of the repository if we are in one, otherwise None."""
    cmd = subprocess.run(
        ["git", "rev-parse", "--show-toplevel"],
        capture_output=True,
        text=True,
    )

    return ((not cmd.returncode) and Path(cmd.stdout.strip())) or None
```

skore/pyproject.toml
```diff
@@ -83,7 +83,7 @@ sphinx-base = [
 sphinx = ["mlflow", "polars", "skore[sphinx-base, hub, mlflow]"]
 sphinx-lts-cpu = ["mlflow", "polars-lts-cpu", "skore[sphinx-base, hub, mlflow]"]
 dev = ["ipykernel", "nbformat"]
-local = ["diskcache", "platformdirs"]
+local = []
```
thomass-dev:
```suggestion
```

skore/src/skore/_plugins/local/project.py
```diff
@@ -192,136 +448,49 @@ def put(self, key: str, report: EstimatorReport | CrossValidationReport) -> None
...
-                "log_loss_std": value.get("log_loss_std"),
-                "roc_auc_std": value.get("roc_auc_std"),
-                "fit_time_std": value.get("fit_time_std"),
-                "predict_time_std": value.get("predict_time_std"),
-            }
-            for value in self.__metadata_storage.values()
-            if value["project_name"] == self.name
-        ]
-
-    def __repr__(self) -> str:  # noqa: D105
```
thomass-dev, line 277: Please revive and adapt, so that we have a uniform behavior between modes.

skore/src/skore/_project/project.py
```diff
@@ -207,6 +214,7 @@ def __init__(self, name: str, *, mode: ProjectMode = "local", **kwargs):
 
         self.__mode = mode
         self.__project = plugin(**parameters)
+        self.local_path = self.__project.path if self.__mode == "local" else None
```
thomass-dev: Now i see the `tracking_uri`, i find strange that we have properties here that are available only on one mode.

Couldn't we overload `__getattr__` and `__dir__` such as:

```python
def __getattr__(self, name):
    # return name content from self.__project if it's a public property

def __dir__(self):
    # return dir(self) + public properties from `self.__project`    
```

---

At least, could you please stick to the "property layout", as it is done for `tracking_uri` ?
As the property is user-faced, that way, the user discovery is better.

skore/pyproject.toml
thomass-dev: If so, address simultaneously https://github.com/probabl-ai/skore/pull/2905/changes#r3497231778.

thomass-dev (changes requested):

skore/src/skore/_project/project.py
jeromedockes: ok I just removed it for now we can deal with that in another PR. I think it could be a unified attribute like "location" that contains a URL for hub, tracking uri for mlflow, and local path for local

auguste-probabl (approved):
Great work!

skore/src/skore/_plugins/local/project.py
```diff
@@ -1,185 +1,446 @@
...
+        .resolve()
+    )
+    new_workspace = not workspace_dir.exists()
+    workspace_dir.mkdir(parents=True, exist_ok=True)
+    if new_workspace:
+        (workspace_dir / ".gitignore").write_text("*\n")
+    (workspace_dir / ".SKORE_WORKSPACE").touch()
+    (workspace_dir / "projects").mkdir(exist_ok=True)
+    (workspace_dir / "datasets").mkdir(exist_ok=True)
+    return workspace_dir
```
thomass-dev, line 37: I have an issue with this behavior: as a user, and especially as a contributor, creating a project should not overlap with my `skore/` or `skore/skore/` sources.

If a `skore/` directory exists at cwd but `.SKORE_WORKSPACE` is missing, we should ask for an explicit validation (or something else). Otherwise, this could lead to conflicts or corruption of the user's directory. We don't manage the user's environment and what might be inside it compared to `.local/share` & co.

thomass-dev (changes requested):
Let's merge in state and iterate.

glemaitre:
Great work everyone. This is nice milestone.

PR #2908, opened 2026-05-13
chore(skore/project/plugin): Assume all native plugins are installed

thomass-dev:
Since we are now in monopackage, where all plugins are natively available, we don't need anymore to ensure that external packages are installed.


glemaitre (approved):
LGTM.

PR #2910, opened 2026-05-13
perf(hub): Send fewer requests per put

auguste-probabl:
#### Change description

Aims to make `put`s faster by reducing the number of HTTP requests per `put`.

This is done for the most part by collecting payloads locally first, and only then sending them, rather than making one request every time an Artifact's `checksum` attribute is accessed.

#### Contribution checklist

- [ ] The code passes our style conventions (you can check this by running pre-commit on your code
requires updating tests)
- [ ] All the commits in the PR are signed (more information

#### AI usage disclosure

~This PR is pretty vibe-coded at this stage.~ Review and editing in progress.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


thomass-dev:
Please consider there is already an opened PR https://github.com/probabl-ai/skore/pull/2114 before working on that.

There is already an open work to parallelize the uploads as soon are they are available (which is the real goal, not to reduce the number of calls afterward) with many issues around `matplotlib`.

> This PR is pretty vibe-coded at this stage.

This is a sensitive part. Please explain the ins and outs first and consider there is already an opened work.

thomass-dev:
You are making concurrent changes with #2114, while it was voluntary postponed to start by handling SVGs directly on the frontend and avoid matplotlib issues.

I don't think it's the way to go. As said previously, asking for presigned URLs in batch is not where we will have significant gain.

auguste-probabl:
> You are making concurrent changes with #2114

That PR is not finished or testable so I don't think the comparison is warranted.

I definitely agree that no longer sending the plots will be way more impactful, and now that the work hub-side is getting good progress I'll most likely be closing this.

I also wanted to note that this part of the code could be made more procedural and simpler (for example, doing network calls in properties is awkward). I hope that the upcoming performance improvements can cover that as well.

PR #2918, opened 2026-05-19
feat(hub): Make skore.login a noop when run in hub's jupyterlite

rouk1:
This PR introduces an env variable `SKORE_HUB_JUPYTERLITE`. If set to a true(ish) value this var makes `skore.login` a no op. 

This is useful to lower the friction when putting an first estimator within Hub's Jupyterlite playground.


glemaitre (approved):
LGTM.

PR #2926, opened 2026-05-21
fix(skore): In `Project`, check dependencies before trying to load those dependencies

cakedev0:
#### Bug description

We call `assert_optional_dependencies_installed(mode)` after actually important the `skore._plugins.[mode]` module, which will import dependencies (like httpx or mlflow). So users don't get a nice error message but get a:
`ModuleNotFoundError: No module named 'httpx'`/`'mlflow'` error.

#### Change description

Move the call to `assert_optional_dependencies_installed` earlier.

#### AI usage disclosure

AI tools were involved for:
- Research and understanding
- Test generation

### DIFF
--- a/skore/src/skore/_project/project.py
+++ b/skore/src/skore/_project/project.py
@@ -142,6 +142,8 @@ class Project:
                 f'`mode` must be "hub", "local" or "mlflow" (found {mode})'
             )
 
+        assert_optional_dependencies_installed(mode)
+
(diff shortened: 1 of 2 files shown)


jeromedockes (approved):
LGTM, thanks. I'm always a bit on the fence about these kind of checks, the default error when we don't catch it is pretty clear -- if I get a no module httpx i install httpx . but as we have them anyway, best use them in the right place :)

PR #2929, opened 2026-05-21
chore(skore): Using pixi to manage the lock files and CI

### LINKED ISSUE #1659: repo: Add pixi

### LINKED ISSUE #2809: ci: move from multi-package to mono-package
Both pytest and release have to be changed.

glemaitre:
closes https://github.com/probabl-ai/skore/issues/1659
closes https://github.com/probabl-ai/skore/issues/2809

Now that we are in monorepo, I would rather use a tool like pixi to orchestrate task and the CI jobs and update the lock file.

This PR is going towards this direction.

However, it does not force any user to absolutely use pixi as a development tools (the same way that it was not necessary before hand to use `uv` for instance).

What this CI proposes:

- [x] create a cross-platform lock file
- [x] use the lock file for the documentation and code testing
- [x] add a `codecov` integration instead of add the report in a message.


auguste-probabl:
Yes, please!

scripts/activate_dot.sh
auguste-probabl, line 1: What's this? Do we need it?

jeromedockes: it is a command that needs to run on macos to be able to use graphviz, used to display dataop graphs

glemaitre: In MacOS, when installing from source, you need to run `dot -c` to initialize otherwise skrub will have warning.

auguste-probabl, line 1: could you move this to a `scripts` dir

pixi.toml
auguste-probabl, line 1: Could this go in pyproject.toml?

auguste-probabl:
Is there something specific keeping this in draft @glemaitre ? I'm very excited to have a coherent, reproducible dev environment.

pixi.toml
```diff
@@ -0,0 +1,76 @@
...
+py311-sklearn18 = ["py311", "sklearn18", "test"]
+py312-sklearn15 = ["py312", "sklearn15", "test"]
+py312-sklearn18 = ["py312", "sklearn18", "test"]
+py313-sklearn15 = ["py313", "sklearn15", "test"]
+py313-sklearn18 = ["py313", "sklearn18", "test"]
+py314-sklearn17 = ["py314", "sklearn17", "test"]
+py314-sklearn18 = ["py314", "sklearn18", "test"]
+py314-sklearn18-pandas2 = ["py314", "sklearn18", "pandas2", "test"]
+py314-sklearn18-mlflow3 = ["py314", "sklearn18", "mlflow3", "test"]
+sphinx = ["py314", "sklearn18", "sphinx"]
```
auguste-probabl: update to sklearn 1.9

docs/how-to/bump-minimal-python-version.md
auguste-probabl, line 1: I feel like this is still relevant?

.github/workflows/pr-display-code-coverage.yml
rouk1, line 1: Who cares about coverage after all ^^

rouk1 (commented):
IMHO this needs to be documented in contribution guidelines.
How to install pixi and dot and how to spawn a virtualenv.

.github/workflows/pr-display-code-coverage.yml
glemaitre: :). It is replace by a codecov integration that is going to scream at you. You're not going to like it :)

glemaitre:
> IMHO this needs to be documented in contribution guidelines.

We can indeed document the way of using `pixi` if one is interested in.

thomass-dev:
> > IMHO this needs to be documented in contribution guidelines.
> 
> We can indeed document the way of using `pixi` if one is interested in.

It's not only "if one is interested in".

The lockfiles used in the CI are now only in pixi "language".
If someone wants to debug and reproduce the environment, they cannot use a simple requirements file anymore.
He has to use pixi, where `requirements.txt` was more "cross-package-manager".

glemaitre:
> If someone wants to debug and reproduce the environment, they cannot use a simple requirements file anymore.

OK I see. I was not considering this use case.

thomass-dev:
Also, "how to update lockfiles" should be documented, both when adding a new dependency or when dependencies have to be upgraded.

In a way, this PR forces maintainers to adopt pixi.

.github/dependabot.yml
```diff
@@ -8,19 +8,7 @@ updates:
...
-    groups:
-      test-dependencies:
-        patterns:
-          - "*"
 
-    # It uses `ci/requirements/**/.python-version` to specify the python version used to
-    # update `ci/requirements/**/test-requirements.txt`
+# Python deps are now resolved by pixi (see `pixi.toml` / `pixi.lock`).
+# Dependabot has no native pixi ecosystem; refresh the lockfile with `pixi update`
+# (manually or via a scheduled workflow).
```
thomass-dev: This is really not cool :(.

.github/workflows/pytest.yml
```diff
@@ -73,216 +78,96 @@ jobs:
         with:
           filters: ${{ runner.temp }}/filters.yaml
 
-  pytest-lockfiles:
```
thomass-dev, line 76: How to ensure pixi lockfile is coherent?

.github/dependabot.yml
glemaitre: We can create a bot as we do in `skrub` that open a PR weekly.

glemaitre: e.g. https://github.com/skrub-data/skrub/blob/main/.github/workflows/update_pixi_lock_files.yml

pixi.toml
glemaitre: We have a choice between having it in `pixi.toml` and `pyproject.toml`. Up to us.

.github/workflows/pytest.yml
glemaitre: It is the default behaviour of pixi: I think that there is a `pixi lock --check` that is run. In short, if the `pixi.toml` or `pyproject.toml` change and the lock files are not coherent, it will raise.

Here, the only think is that we are in `--frozen` mode, meaning that we don't check if there is a new version available for the library that are not pin. We can use a bot to update.

CONTRIBUTING.rst
glemaitre, line 1: So I added the information in this file @rouk1. Would you rather add it somewhere else?

glemaitre:
> It's not only "if one is interested in".

> The lockfiles used in the CI are now only in pixi "language".
If someone wants to debug and reproduce the environment, they cannot use a simple requirements file anymore.
He has to use pixi, where requirements.txt was more "cross-package-manager".

I added a `pixi` task that could help at exporting a pure `requirements.txt` file. Maybe we can revisit when `pylock.toml` is supported.

CONTRIBUTING.rst
```diff
@@ -85,22 +85,116 @@ You will find below some code you can use in your terminal, using HTTPS connecti
...
-Once your environment is set up, install the development dependencies and setup pre-commit with:
+- with `pixi <https://pixi.sh>`_, which is the workflow used by our CI and the
+  recommended way to reproduce it exactly (see `Using pixi`_);
+- with a plain virtual environment and ``pip`` (see `Using a virtual environment and pip`_).
 
-    .. code-block:: bash
+Using pixi
+""""""""""
 
-        make install-skore
```
rouk1, line 94: Do we still need the makefile now ?

.github/workflows/sphinx.yml
```diff
@@ -253,7 +178,7 @@ jobs:
           SPHINX_EXAMPLE_BASE_URL: ${{ vars.DOCUMENTATION_EXAMPLE_BASE_URL }}
 
       - name: Upload documentation
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        uses: actions/upload-artifact@bbbca2ddaa5d8feaa63e36b76fdaad77386f024f # v7.0.0
```
rouk1: Is this downgrade on purpose ?

.github/workflows/sphinx.yml
```diff
@@ -418,7 +343,7 @@ jobs:
     needs: [sphinx-deploy-html, sphinx-deploy-root-files, sphinx-deploy-stable]
     steps:
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@d979d5b3a71173a29b74b5b88418bfda9437d885 # v6.1.1
+        uses: aws-actions/configure-aws-credentials@ec61189d14ec14c8efccab744f656cffd0e33f37 # v6.1.0
```
rouk1: Is this downgrade on purpose ?

rouk1 (commented):
Image

🤣 

Side note: cursor/vscode never succeed in auto enabling the correct pixi venv.

jeromedockes:
FWIW the pixi workflow proposed by @glemaitre is more or less what he set up for skrub like 1.5 year ago and it has worked well for skrub. personally I don't use pixi for my day-to-day work, only when there is an error in one of the CI jobs that I don't reproduce in my local env I will run `pixi run -e ci-env-name pytest test_.py::test_it`. the `pixi run -e <env>` part I copy-paste from the ci log. so I would say I use pixi less than monthly. Usually when I do it is for the env using the nightly builds of scientific packages which are not on PyPI or conda-forge but on other indexes so just doing `pixi run -e ci-nightly-deps` is convenient (I'm sure there are other ways besides pixi to have that convenience)

thomass-dev:
This message is meant to be factual, and i'm looking for arguments in favor of pixi (i hope this isn't taken as passive-aggressive).

> FWIW the pixi workflow proposed by @glemaitre is more or less what he set up for skrub like 1.5 year ago and it has worked well for skrub. personally I don't use pixi for my day-to-day work, only when there is an error in one of the CI jobs that I don't reproduce in my local env I will run `pixi run -e ci-env-name pytest test_.py::test_it`. the `pixi run -e <env>` part I copy-paste from the ci log. so I would say I use pixi less than monthly. Usually when I do it is for the env using the nightly builds of scientific packages which are not on PyPI or conda-forge but on other indexes so just doing `pixi run -e ci-nightly-deps` is convenient (I'm sure there are other ways besides pixi to have that convenience)

So what is the benefit of changing the CI from uv+bash+gh to pixi+gh? The argument of "bash" doesn't convince me, pixi is another language to learn, and the diff +30K -5K neither. The same goes for reproducibility; that's what lockfiles are for, whether you're using pixi or not.

I'm not convinced.

(by the way, i'm starting my review)

.github/workflows/sphinx.yml
```diff
@@ -253,7 +178,7 @@ jobs:
           SPHINX_EXAMPLE_BASE_URL: ${{ vars.DOCUMENTATION_EXAMPLE_BASE_URL }}
 
       - name: Upload documentation
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
```
thomass-dev, line 254: please revert

.github/workflows/sphinx.yml
```diff
@@ -418,7 +343,7 @@ jobs:
     needs: [sphinx-deploy-html, sphinx-deploy-root-files, sphinx-deploy-stable]
     steps:
       - name: Configure AWS Credentials
-        uses: aws-actions/configure-aws-credentials@d979d5b3a71173a29b74b5b88418bfda9437d885 # v6.1.1
```
thomass-dev: please revert

CONTRIBUTING.rst
```diff
@@ -85,22 +85,116 @@ You will find below some code you can use in your terminal, using HTTPS connecti
...
+    curl -fsSL https://pixi.sh/install.sh | sh
+
+pixi creates an isolated environment for you (you do not need to create a virtual
+environment yourself) and downloads the dependencies on first use. The environments are
+named in the ``[environments]`` table of ``pixi.toml``. The most common commands are:
+
+.. code-block:: bash
+
+    # Run the test suite in a given environment
+    pixi run -e py311-sklearn19 tests
```
auguste-probabl: this line will need to be maintained; maybe there is a way to have a `main` environment? otherwise add it to the how-to

CONTRIBUTING.rst
```diff
@@ -85,22 +85,116 @@ You will find below some code you can use in your terminal, using HTTPS connecti
...
+
+    # Activate it (Windows, PowerShell)
+    .venv\Scripts\Activate.ps1
+
+Once your environment is activated, install the development dependencies and setup
+pre-commit with:
+
+.. code-block:: bash
+
+    make install-skore
```
auguste-probabl: so we keep the makefile? maybe we can inline it

.github/dependabot.yml
auguste-probabl: yes, we should

CONTRIBUTING.rst
auguste-probabl: same q

.github/workflows/sphinx.yml
auguste-probabl: duplicate

.github/workflows/sphinx.yml
auguste-probabl: duplicate

CONTRIBUTING.rst
auguste-probabl: duplicate

.github/workflows/sphinx.yml
thomass-dev: please revert

.github/workflows/sphinx.yml
thomass-dev: please revert

jeromedockes:
> and i'm looking for arguments in favor of pixi

I wasn't arguing for or against pixi, just saying what was my experience with this setup in another project. I don't have an opinion either way.

> (i hope this isn't taken as passive-aggressive).

don't worry :) it did not come across like that

.github/dependabot.yml
thomass-dev: Maintaining our own dependabot 😅 .

auguste-probabl: we are maintaining a whole in-house dependency tracking system right now, so I think it is a net improvement tbh

thomass-dev: > whole in-house dependency tracking system

That's not true, and it's really unfair. 😉

Before, we used:
- `pyproject.toml` to define dependencies,
- a `txt` file to override dependencies (matrix),
- a `bash` script to read the `txt` file and generate `requirements.txt` files using `uv`.

With this PR, you need:
- `pyproject.toml` to define dependencies,
- a `pixi.toml` to define dependencies,
- a pixi command to read the `pixi.toml` file and generate the `pixi.lock`.

Also, the `txt` file was used in the CI, so with one file we were able to generate lockfiles and to control matrix in the CI. With this PR, we have to redefine matrix both in the CI and in the `pixi.toml`.

The underlying system is the same, in a different language `pixi` vs `bash+uv`: defining lockfiles.

The only advantage is the invisible `venv` management in the CI + the automatic lockfiles check.
This latest advantage is mitigated by the fact that a replacement for dependabot needs to be created.
Otherwise, the complexity in the CI is strictly the same, and i'm still not convinced.

thomass-dev:
> This message is meant to be factual, and i'm looking for arguments in favor of pixi (i hope this isn't taken as passive-aggressive).
> 
> > FWIW the pixi workflow proposed by @glemaitre is more or less what he set up for skrub like 1.5 year ago and it has worked well for skrub. personally I don't use pixi for my day-to-day work, only when there is an error in one of the CI jobs that I don't reproduce in my local env I will run `pixi run -e ci-env-name pytest test_.py::test_it`. the `pixi run -e <env>` part I copy-paste from the ci log. so I would say I use pixi less than monthly. Usually when I do it is for the env using the nightly builds of scientific packages which are not on PyPI or conda-forge but on other indexes so just doing `pixi run -e ci-nightly-deps` is convenient (I'm sure there are other ways besides pixi to have that convenience)
> 
> So what is the benefit of changing the CI from uv+bash+gh to pixi+gh? The argument of "bash" doesn't convince me, pixi is another language to learn, and the diff +30K -5K neither. The same goes for reproducibility; that's what lockfiles are for, whether you're using pixi or not.
> 
> I'm not convinced.
> 
> (by the way, i'm starting my review)

Sorry @jeromedockes , in fact my message was implicitly addressed to @glemaitre .

.github/workflows/pytest.yml
```diff
@@ -73,216 +78,96 @@ jobs:
...
-          path: ${{ matrix.package }}/venv
-          key: ${{ steps.cache-python-venv.outputs.cache-primary-key }}
-
-      - name: Build
-        working-directory: ${{ matrix.package }}/
-        run: python -m build
-
-      - name: Install
-        working-directory: ${{ matrix.package }}/dist/
-        run: wheel=(*.whl); python -m pip install --force-reinstall --no-deps "${wheel}"
```
thomass-dev, line 253: Please revert: testing with the wheels and not the `src/` is important, otherwise we are not in a user-like environment and we could miss some failures.

How you launch tests:

```
run: pixi run -e ${{ matrix.environment }} tests
run: pixi run -e ${{ matrix.environment }} tests-coverage
```

```toml
[feature.test.tasks]
tests = { cmd = "pytest src/ tests/ --no-cov", cwd = "skore" }
tests-coverage = { cmd = "pytest src/ tests/ --showlocals --durations=20 --junitxml=coverage/pytest.xml -o junit_family=legacy --cov-config=pyproject.toml --cov=skore --cov-report=", cwd = "skore" }
```

thomass-dev (changes requested):

.github/dependabot.yml
auguste-probabl: > > whole in-house dependency tracking system
>
> That's not true, and it's really unfair. 😉

`pip-compile.sh` and `supported-versions.json` *are* an attempt at a home-grown dependency tracking system, and it works fine, don't get me wrong!

But it's less powerful (e.g granularity of lock files), less standard, and less well-documented than a bespoke tool like pixi or uv

CONTRIBUTING.rst
glemaitre: If you are using `pixi` then the tasks are replacing the `Makefile` and the writing cross-platform tasks make that it will work for all user if needed.

.github/workflows/pytest.yml
glemaitre: I see. I added a way that we test the installed wheels and not the editable in the source.

glemaitre:
> Side note: cursor/vscode never succeed in auto enabling the correct pixi venv.

It is weird, it works normally with the Python extension out of the box.

PR #2930, opened 2026-05-21
perf(mlflow): Enable data de-duplication for MLflow backend

cakedev0:
#### Change description

This PR updates the MLflow project backend to serialize reports through `get_state` / `from_state`, and stores the heavy `state["data"]` entries as separate artifacts.

Those data artifacts are keyed by their `joblib.hash`, so identical data can be de-duplicated in the hidden MLflow storage experiment instead of being logged again for each report.

I also added small serde helper tests using real report states, plus a MLflow test checking that writing the same data artifact twice only creates one storage run.

#### AI usage disclosure

AI tools were involved for:
- Test generation
- Review-fix iteration

### DIFF
--- a/skore/src/skore/_plugins/serde.py
+++ b/skore/src/skore/_plugins/serde.py
@@ -39,12 +39,12 @@ def externalize(
 ) -> dict[str, Any]:
     """Extract large artifacts in ``state`` to external storage.
 
-    Replace large artifacts in ``state`` with IDs.
+    Replace large artifacts in ``state`` with pointers.
 
     Parameters
     ----------
     state : dict[str, Any]
-        Report state whose ``"data"`` entries should be externalized.
+        Report state.
 
     artifact_writer : callable
         Called as ``artifact_writer(id, bytes)`` for each artifact extracted
--- a/skore/src/skore/_plugins/serde.py
+++ b/skore/src/skore/_plugins/serde.py
@@ -1,69 +1,85 @@
 import io
-from collections.abc import Callable, Generator
+from collections.abc import Callable
 from dataclasses import dataclass
-from typing import Any
+from typing import Any, TypeAlias
 
 import joblib
+
+ArtifactId: TypeAlias = str
 
 
 @dataclass
 class ArtifactPointer:
-    artifact_id: str
+    """A lazy reference to a serialized artifact.
 
-    @staticmethod
-    def load(artifact_bytes: bytes) -> Any:
-        raise NotImplementedError()
-
-    @staticmethod
-    def dump(artifact: Any) -> bytes:
-        raise NotImplementedError()
-
-
-class JoblibPointer(ArtifactPointer):
-    @staticmethod
-    def load(artifact_bytes: bytes) -> Any:
-        with io.BytesIO(artifact_bytes) as stream:
-            return joblib.load(stream)
-
-    @staticmethod
-    def dump(artifact: Any) -> bytes:
-        with io.BytesIO() as stream:
-            joblib.dump(artifact, stream)
-            return stream.getvalue()
-
-
-def split_data_from_state(
-    state: dict[str, Any],
-) -> Generator[tuple[str, bytes], None, None]:
-    """Extract data artifacts from a report state.
-
-    The input ``state`` is updated in place: each entry in ``state["data"]`` is
-    replaced by an artifact pointer, and the serialized artifact bytes are yielded.
+    A pointer replaces a heavy payload inside a report state so the bytes can
+    live elsewhere (typically content-addressed storage that de-duplicates
+    identical payloads across reports).
     """
-    data = {}
+
+    artifact_id: ArtifactId
+
+
+def joblib_dump(artifact: Any) -> bytes:
+    """Serialize the artifact into bytes suitable for storage."""
+    with io.BytesIO() as stream:
+        joblib.dump(artifact, stream)
+        return stream.getvalue()
+
+
+def joblib_load(artifact_bytes: bytes) -> Any:
+    """Deserialize the artifact from its stored byte representation."""
+    with io.BytesIO(artifact_bytes) as stream:
+        return joblib.load(stream)
+
+
+def externalize(
+    state: dict[str, Any],
+    artifact_writer: Callable[[ArtifactId, bytes], None],
+) -> dict[str, Any]:
+    """Extract large artifacts in ``state`` to external storage.
+
+    Replace large artifacts in ``state`` with IDs.
+
+    Parameters
+    ----------
+    state : dict[str, Any]
+        Report state whose ``"data"`` entries should be externalized.
+
+    artifact_writer : callable
+        Called as ``artifact_writer(id, bytes)`` for each artifact extracted
+        from ``state``; responsible for persisting ``bytes`` under ``id``.
+    """
+    new_data: dict[str, ArtifactPointer] = {}
     for key, obj in state["data"].items():
-        obj_bytes = JoblibPointer.dump(obj)
-        bytes_id = joblib.hash(obj_bytes)
-        yield (bytes_id, obj_bytes)
-        data[key] = JoblibPointer(bytes_id)
-    state |= {"data": data}
+        obj_bytes = joblib_dump(obj)
+        artifact_id = joblib.hash(obj_bytes)
+        artifact_writer(artifact_id, obj_bytes)
+        new_data[key] = ArtifactPointer(artifact_id)
+    return state | {"data": new_data}
 
 
-def restore_data_in_state(
+def internalize(
     state: Any,
-    artifact_loader: Callable[[str], bytes],
+    artifact_reader: Callable[[ArtifactId], bytes],
 ) -> Any:
-    """Restore artifact pointers in a report state.
+    """Replace artifact pointers in ``state`` with the data they point to.
 
-    Pointers are loaded with ``artifact_loader`` and deserialized; dictionaries
-    and lists are traversed recursively while other values are returned unchanged.
+    Parameters
+    ----------
+    state : Any
+        Possibly-nested structure (dict, list, or scalar) that may contain
+        :class:`ArtifactPointer` instances; traversed recursively.
+
+    artifact_loader : callable
+        Called as ``artifact_loader(artifact_id)`` for each pointer encountered.
     """
     if isinstance(state, ArtifactPointer):
-        artifact_bytes = artifact_loader(state.artifact_id)
-        return state.load(artifact_bytes)
+        artifact_bytes = artifact_reader(state.artifact_id)
+        return joblib_load(artifact_bytes)
     elif isinstance(state, dict):
-        return {k: restore_data_in_state(v, artifact_loader) for k, v in state.items()}
+        return {k: internalize(v, artifact_reader) for k, v in state.items()}
     elif isinstance(state, list):
-        return [restore_data_in_state(v, artifact_loader) for v in state]
+        return [internalize(v, artifact_reader) for v in state]
     else:
         return state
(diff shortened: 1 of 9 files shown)


skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,69 @@
...
+def split_data_from_state(
+    state: dict[str, Any],
+) -> Generator[tuple[str, bytes], None, None]:
+    """Extract data artifacts from a report state.
+
+    The input ``state`` is updated in place: each entry in ``state["data"]`` is
+    replaced by an artifact pointer, and the serialized artifact bytes are yielded.
+    """
+    data = {}
+    for key, obj in list(state["data"].items()):
```
glemaitre: Why do you need to make it a list instead of directly iterate on the dict items?

cakedev0: I don't, I needed it when I was mutating it, which was not a good idea.

cakedev0: removed

auguste-probabl:
Can you explain how this change can be tested manually?

cakedev0:
> Can you explain how this change can be tested manually?

I did a small manual test but my venv still contained the old `skore-mlflow-project` package, so it was not using the new code...

I encountered a few bugs (in this PR and in skore), I pushed all the fixes in this PR, let me know if I should rather do  skore fixes in a separate PR.

Manual test: for instance you can run this (after launching `uv tool run mlflow server`):

```Python
from skore import evaluate
from skore import Project

project = Project(mode="mlflow", name="smart-serde-test", tracking_uri="http://127.0.0.1:5000")
# my venv was still polluted with the skore-mlflow-project package, which was used instead of
# `skore._plugins.mlflow`, let's make sure you don't loose time with the same problem:
assert str(project._Project__project.__class__) == "<class 'skore._plugins.mlflow.project.Project'>"

X, y = make_regression(
    n_samples=250_000,
    random_state=0,
)
print(X.nbytes / 1e6)

for max_leaf_nodes in [10, 15, 20]:
    hgb = HistGradientBoostingRegressor(max_leaf_nodes=max_leaf_nodes, max_features=0.1)
    report = evaluate(hgb, X, y, splitter=0.2)
    project.put("key", report)
```

And watch `du -sh mlartifacts/` (on this branch vs on main).

skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,69 @@
...
+            joblib.dump(artifact, stream)
+            return stream.getvalue()
+
+
+def split_data_from_state(
+    state: dict[str, Any],
+) -> Generator[tuple[str, bytes], None, None]:
+    """Extract data artifacts from a report state.
+
+    The input ``state`` is updated in place: each entry in ``state["data"]`` is
```
auguste-probabl: can we clone instead of mutating?

skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,69 @@
+import io
+from collections.abc import Callable, Generator
+from dataclasses import dataclass
+from typing import Any
+
+import joblib
+
+
+@dataclass
+class ArtifactPointer:
```
auguste-probabl: Is this base class useful?

skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,69 @@
...
+    @staticmethod
+    def dump(artifact: Any) -> bytes:
+        with io.BytesIO() as stream:
+            joblib.dump(artifact, stream)
+            return stream.getvalue()
+
+
+def split_data_from_state(
+    state: dict[str, Any],
+) -> Generator[tuple[str, bytes], None, None]:
```
auguste-probabl:
```suggestion
) -> Iterator[tuple[str, bytes]]:
```

skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,69 @@
...
+    The input ``state`` is updated in place: each entry in ``state["data"]`` is
+    replaced by an artifact pointer, and the serialized artifact bytes are yielded.
+    """
+    data = {}
+    for key, obj in state["data"].items():
+        obj_bytes = JoblibPointer.dump(obj)
+        bytes_id = joblib.hash(obj_bytes)
+        yield (bytes_id, obj_bytes)
+        data[key] = JoblibPointer(bytes_id)
+    state |= {"data": data}
```
auguste-probabl:
```suggestion
    state["data"] = data
```

skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def restore_data_in_state(
+    state: Any,
+    artifact_loader: Callable[[str], bytes],
```
auguste-probabl:
```suggestion
    artifact_loader: Callable[[ArtifactId], bytes],
```

skore/tests/unit/plugins/test_serde.py
```diff
@@ -0,0 +1,40 @@
...
+        (EstimatorReport, "estimator_reports_regression"),
+        (CrossValidationReport, "cross_validation_reports_regression"),
+    ],
+    ids=["estimator", "cross-validation"],
+)
+def test_report_rebuilds_after_smart_serde(
+    request,
+    report_cls: type[EstimatorReport] | type[CrossValidationReport],
+    report_fixture: str,
+) -> None:
```
auguste-probabl:
```suggestion

def test_report_rebuilds_after_smart_serde(
    request,
    report
) -> None:
```
and then get the class of `report` with `report.__class__`

skore/tests/unit/reports/cross_validation/test_report.py
```diff
@@ -392,3 +400,24 @@ def test_from_state_rejects_unknown_version(logistic_binary_classification_data)
 
     with pytest.raises(ValueError, match="Unexpected state version"):
         CrossValidationReport.from_state(state)
+
+
+def test_state_has_no_unexpected_data_copy():
```
auguste-probabl: not super clear what this tests, can you add a docstring?

auguste-probabl (dismissed):
looks great overall. I have a few comments. Also, I'm a bit surprised that MLFlow doesn't provide a data dedup mechanism that we could rely on. Have you checked if such a thing exists?

skore/tests/unit/reports/cross_validation/test_report.py
cakedev0: (This test is based on a heuristic "X is big")

auguste-probabl:
I will take over from @cakedev0 as he is busy on other things. Thanks!

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -400,7 +401,9 @@ def from_state(cls, state: dict[str, Any]) -> EstimatorReport:
                 for (data_source, name), val in state["predictions"].items()
             }
         )
-        report._metric_registry = state["metric_registry"]
+        report._metric_registry = MetricRegistry(report)
```
auguste-probabl: we need to do this because MetricRegistry holds `report`, and hence all of its members, including the data

skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,85 @@
...
+    state : Any
+        Possibly-nested structure (dict, list, or scalar) that may contain
+        :class:`ArtifactPointer` instances; traversed recursively.
+
+    artifact_loader : callable
+        Called as ``artifact_loader(artifact_id)`` for each pointer encountered.
+    """
+    if isinstance(state, ArtifactPointer):
+        artifact_bytes = artifact_reader(state.artifact_id)
+        return joblib_load(artifact_bytes)
```
jeromedockes: i think dump and load could be methods of the Pointer class, so that it could be possible to have different kinds of serialization and the right loader will be used

skore/src/skore/_sklearn/_estimator/report.py
jeromedockes: AFAICT metricsregistry.remove is only used by the metrics accessor, which has a reference to the report right? so maybe we could avoid the circular reference between the report and registry by passing the report to MetricsRegistry.remove

skore/src/skore/_plugins/serde.py
auguste-probabl: That's how it was originally, but i'd rather keep it simple for now

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: done

skore/src/skore/_plugins/serde.py
jeromedockes: ok that seems reasonable WDYT @cakedev0 ?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -401,6 +402,7 @@ def from_state(cls, state: dict[str, Any]) -> EstimatorReport:
             }
         )
         report._metric_registry = state["metric_registry"]
+        report._checks_registry = list(_BUILTIN_CHECKS)
```
jeromedockes, line 412: doing get and set state resets the checks registry to the default, let's open a separate issue about that

jeromedockes (dismissed):
LGTM but if you have the time it would be great to have a review from @cakedev0 as you started the PR and know the mlflow part much better than me

skore/src/skore/_plugins/serde.py
cakedev0: Hum... That works too, it does allow additions like npz serde with easy/no-effort backward compat'.

Let's maybe just rename `ArtifactPointer` to `JoblibPicklePointer`?

skore/src/skore/_sklearn/_estimator/report.py
cakedev0: Ah yes thanks, I wanted to do it but forgot.

skore/src/skore/_plugins/serde.py
```diff
@@ -0,0 +1,85 @@
...
+    Replace large artifacts in ``state`` with pointers.
+
+    Parameters
+    ----------
+    state : dict[str, Any]
+        Report state.
+
+    artifact_writer : callable
+        Called as ``artifact_writer(id, bytes)`` for each artifact extracted
+        from ``state``; responsible for persisting ``bytes`` under ``id``.
```
cakedev0, line 51: So much better than my initial pattern! Nice!

cakedev0, line 51: So much better than my initial pattern! Nice!

cakedev0 (commented):
LGTM! (I can't select "approve" as I opened it) 

Overall, this looks quite better than what I wrote initially!

skore/src/skore/_plugins/serde.py
auguste-probabl: I was a bit puzzled working on this code initially because it seemed that having the Pointer both hold data and know how to serde itself (with static methods) seemed to mix many things. It feels a bit more modular to me to have the pointers just hold the data and maybe a tag like "joblib", and then upon serde we read the tag and call the right serde. But then I thought that there was no need to consider that yet since there is only joblib.

cakedev0: > maybe a tag like "joblib

A tag or the class name. Maybe a tag is indeed more adequate for this purpose.

cakedev0: (anyway, this is not blocking, I'm completely ok with merging the current code as it is)

auguste-probabl: renamed

jeromedockes: a tag stored in `__class__`? calling the appropriate implementation based on the type of object is part of why we have classes. but anyway let's revisit if / when we add a different serialization format

jeromedockes (approved):
thanks @auguste-probabl and @cakedev0 !

auguste-probabl:
I found a gnarly bug because of these test failures. Investigating

EDIT: Hopefully fixed by https://github.com/probabl-ai/skore/pull/2950

PR #2931, opened 2026-05-22
fix(skore/checks): Catch exceptions in baseline report creation

GaetandeCast:
cc @jeromedockes

### DIFF
--- a/skore/tests/unit/reports/estimator/test_checks.py
+++ b/skore/tests/unit/reports/estimator/test_checks.py
@@ -239,6 +239,24 @@ def test_exception_when_train_data_missing(regression_train_test_split):
                 check.check_function(report)
 
 
+def test_exception_when_baseline_report_creation_fails(
+    regression_train_test_split, monkeypatch
+):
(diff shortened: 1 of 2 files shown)


skore/tests/unit/reports/estimator/test_checks.py
```diff
@@ -239,6 +239,21 @@ def test_exception_when_train_data_missing(regression_train_test_split):
...
 
 
+def test_exception_when_baseline_report_creation_fails(regression_data, monkeypatch):
+    """Check that an exception is raised when the baseline report creation fails."""
+    X, y = regression_data
+    report = evaluate(LinearRegression(), X, y)
+    monkeypatch.setattr(
+        EstimatorReport,
+        "_fit_estimator",
+        lambda self, **kwargs: RuntimeError("Test error"),
```
jeromedockes: this function returns an exception, i think you meant to raise it? (which you cannot do in an expression so you will have to define a function rather than lambda expression). but it seems to cause a failure anyway so the test still works

jeromedockes (approved):

PR #2943, opened 2026-05-24
fix #2941: add text and HTML reprs for displays and reports

### LINKED ISSUE #2941: Implement text and interactive reprs for all `*Display` and `*Report` objects
Currently 
- only `EstimatorReport` has a nice html repr, with a text repr that could be improved: it's just a text view of the metrics summary display
- only `ChecksSummaryDisplay` has an html and text repr.
- every other display and report types are lacking representations.

For displays, we thought with @glemaitre that it would make sense to have the output of `.plot()` in the interactive repr, if the method is implemented. Otherwise we can use the pandas repr of the `.frame()` output. For their text repr, we should show the text version of a nice table, i.e. in a wide format, which means we need to iterate on that. 

For the text repr of reports, we thought of writing a summary of the report in a markdown format.

One of the goals motivating this change is to have a more agent-friendly interface for skore: nice text repr and markdown.
The other goal is having something meaning full displayed when displaying `report.accessor().method()`, when not explicitly calling `.frame()` or `.plot()`. When the interactive repr shows a `.plot()` output, it should also prompt the user to use `.plot()` explicitly to control what's shown.

utsab345:
## Summary
This PR adds text and interactive HTML reprs for the relevant display and report objects, improving notebook and terminal ergonomics.

## What changed
- Added repr support for display and report objects
- Fixed invalid repr method placement in the ROC and precision-recall display modules
- Added a markdown-style text repr for `EstimatorReport`
- Aligned the HTML title for the report repr
- Fixed NumPy compatibility in the checks path
- Added repr-focused tests for displays and reports

## Verification
- `pytest -o addopts='' --import-mode=importlib tests/unit/displays/test_common.py` → **4 passed**
- `pytest -o addopts='' --import-mode=importlib tests/unit/reports/test_common.py` → **25 passed**

## Notes
- There are still two warnings during the targeted test run:
  - a `RuntimeWarning` from plotting utilities
  - deprecation warnings from `skrub.TableReport`

Fixes #2941

### DIFF
--- a/skore/src/skore/_sklearn/_checks/model_checks.py
+++ b/skore/src/skore/_sklearn/_checks/model_checks.py
@@ -257,7 +257,7 @@ class CheckHighClassImbalance(Check):
         if report.ml_task != "binary-classification" or y is None:
             raise CheckNotApplicable()
 
-        values, counts = np.unique_counts(y)
+        values, counts = np.unique(y, return_counts=True)
         overrepresented_class = values[counts >= 0.8 * counts.sum()]
 
         if overrepresented_class.size > 0:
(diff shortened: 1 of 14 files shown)


glemaitre:
This PR is of too low quality to be reviewed.

PR #2945, opened 2026-05-25
chore(skore): Improve consistency of backslashes in docs

glemaitre:
Make the use of `\` consistent in the docstrings:

- `\` only used to break type where there is a need for an extra indent on next line


auguste-probabl:
I wonder why the sphinx deployment comment is not showing up

thomass-dev (approved):

PR #2952, opened 2026-05-28
fix: Ensure that `report.metrics.summarize` and `report.metrics.<metric>` use the same cache keys

auguste-probabl:
#### Change description

Even if the two share mostly the same code, the way they are called in `_raw_cached` causes one to use cache key `(data_source, "r2", ('mapping', ()))` while the other uses `(data_source, "r2", ('mapping', ('multioutput', 'raw_values'),))`.

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:
Nothing

### DIFF
--- a/skore/src/skore/_sklearn/metrics.py
+++ b/skore/src/skore/_sklearn/metrics.py
@@ -147,6 +147,8 @@ class Metric:
       differ from what the base metric returned.
     """
 
+    kwargs: dict[str, Any] = {}
+
     def __init__(
         self,
         *,
@@ -166,7 +168,7 @@ class Metric:
         # When name is None, the metric is being instantiated from a subclass
         # (e.g. Accuracy()) whose fields are defined as class attributes.
         # Only `kwargs` needs to be set as an instance attribute.
-        self.kwargs = kwargs or {}
+        self.kwargs = kwargs or self.kwargs
 
         if name is None:
             return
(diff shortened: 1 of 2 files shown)


skore/tests/unit/reports/estimator/metrics/test_summarize.py
```diff
@@ -392,6 +392,19 @@ def test_cache(forest_binary_classification_with_test):
     assert_frame_equal(result.data, result_from_cache.data)
 
 
+def test_cache_interaction(linear_regression_with_test):
```
glemaitre: Could you parametrize taking into account the following metrics:

- precision / recall (average=None)
- roc_auc (multi_class="ovr")
- fit_time / predict_time (cast=True)

Since they would be affected the same way

skore/src/skore/_sklearn/metrics.py
```diff
@@ -574,6 +577,7 @@ class PredictTime(Metric):
     greater_is_better = False
     function = None
     function_kind = None
+    kwargs = {"cast": True}
 
     @staticmethod
     def available(report: EstimatorReport) -> bool:
```
glemaitre, line 583: Maybe change the signature of the function `_raw` such that the signature match with the `FitTime._raw` metric.

glemaitre (commented):
In summarize we have:

```python
            metric_rows = parsed_metric.rows(
                report=self._parent,
                data_source=data_source,
                **parsed_metric.kwargs,
            )
```

we don't need the line `**parsed_metric.kwargs` because the `parsed_metric.rows` already have `self.kwargs | kwargs` internally.

glemaitre (commented):
Otherwise, it looks good.

skore/src/skore/_sklearn/metrics.py
auguste-probabl: nice catch

skore/tests/unit/reports/estimator/metrics/test_summarize.py
auguste-probabl: done

jeromedockes (approved):

thomass-dev (approved):

PR #2955, opened 2026-05-28
feat(skore/EstimatorReport): Add to_markdown and mention it in repr

GaetandeCast:
Towards #2941

Keeps the `__repr__` concise and refers the reader to the `to_markdown()` method which constructs a comprehensive mardown summary of the report's contents.

Only for `EstimatorReport` to keep the scope controlled.

cc @glemaitre

<details>
<summary>Example output</summary> 

# EstimatorReport: TabICLRegressor

## Estimator

- name: `TabICLRegressor`
- kind: tabicl estimator
- ml task: regression
- fit time: 0.212 s

```text
TabICLRegressor()
```

Full estimator info: `report.estimator_`.

## Metrics

```text
                  TabICLRegressor (train)  TabICLRegressor (test)
Metric                                                           
Score                            0.999985                0.995280
R²                               0.999985                0.995280
RMSE                             0.781705                9.442092
MAE                              0.569714                6.772416
MAPE                             0.012629                0.160790
Fit time (s)                     0.211974                0.211974
Predict time (s)                 0.530039                0.407629
```

Full metrics: `report.metrics.summarize().frame()`.

## Checks

```text
Checks summary: 2 issue(s), 1 tip(s), 3 passed, 1 ignored.
Issues:
- [SKD001] Potential overfitting. Significant train/test gaps were found for 3/5 default predictive metrics. Read more about this here: https://docs.skore.probabl.ai/dev/user_guide/automated_checks.html#skd001-overfitting.
- [SKD010] Model slower than baseline. Fit is ~13.3x slower than a fast linear baseline without significantly better test scores (5/5 default predictive metrics). Read more about this here: https://docs.skore.probabl.ai/dev/user_guide/automated_checks.html#skd010-slower-than-baseline.
Tips:
- [SKD012] Useless features. Feature(s) ['Feature #5', 'Feature #6', 'Feature #7', 'Feature #8', 'Feature #9'] have permutation importance overlapping with zero and could likely be dropped without degrading performance. Read more about this here: https://docs.skore.probabl.ai/dev/user_guide/automated_checks.html#skd012-useless-features.
Passed:
- [SKD002] Potential underfitting
- [SKD008] Highly correlated input features
- [SKD009] Model worse than baseline
Mute a check with .checks.summarize(ignore=['<code>']).
```

Full checks: `report.checks.summarize()`.

## Data

Computed on the train+test dataset (n_rows=100, n_columns=11, n_constant_columns=0).

| column | dtype | null_count | n_unique |
| ------ | ----- | ---------- | -------- |
| Feature 0 | Float64DType | 0 | 100 |
| Feature 1 | Float64DType | 0 | 100 |
| Feature 2 | Float64DType | 0 | 100 |
| Feature 3 | Float64DType | 0 | 100 |
| Feature 4 | Float64DType | 0 | 100 |
| Feature 5 | Float64DType | 0 | 100 |
| Feature 6 | Float64DType | 0 | 100 |
| Feature 7 | Float64DType | 0 | 100 |
| Feature 8 | Float64DType | 0 | 100 |
| Feature 9 | Float64DType | 0 | 100 |
| Target | Float64DType | 0 | 100 |

Full analysis: `report.data.analyze()`.
</details>


skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -815,25 +816,33 @@ def _get_help_legend(self) -> str:
...
         """Return a string representation."""
         return f"""{self.__class__.__name__}:
-        {self.estimator_!r}
+        {self.estimator_name_!r}
 
-        {self.metrics.summarize().frame()}"""
+        {
+            "No data provided."
+            if (data_source := self._repr_data_source()) is None
+            else self.metrics.summarize(data_source=data_source).frame()
```
auguste-probabl, line 826: I don't love running `summarize` in `__repr__`; `__repr__` is supposed to be:
- very fast
- stateless (summarize touches the cache)
- if possible, sufficient to re-create a new object (so closer to `EstimatorReport({self.estimator}, ...)`)

Since `to_markdown` now does it, could we remove it from here?

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -926,3 +933,128 @@ def _repr_mimebundle_(self, **kwargs):
         output = {"text/plain": repr(self)}
         output["text/html"] = self._repr_html_()
         return output
+
+    def to_markdown(self) -> str:
+        """Return a markdown summary of the report, suitable for agent consumption.
```
auguste-probabl:
```suggestion
        """Return a markdown summary of the report.
```

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -926,3 +933,128 @@ def _repr_mimebundle_(self, **kwargs):
...
+                f"# {self.__class__.__name__}: {self.estimator_name_}",
+                self._markdown_estimator_section(),
+                self._markdown_metrics_section(),
+                self._markdown_checks_section(),
+                self._markdown_data_section(),
+            )
+        )
+
+    def _markdown_estimator_kind(self) -> str:
+        est = self.estimator
```
auguste-probabl: inline

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -926,3 +933,128 @@ def _repr_mimebundle_(self, **kwargs):
...
+        )
+
+    def _markdown_data_section(self) -> str:
+        lines = ["## Data", ""]
+        data_source = self._repr_data_source()
+        if data_source is None:
+            lines.append("No data provided.")
+            lines += ["", "Full analysis: `report.data.analyze()`."]
+            return "\n".join(lines)
+        summary = summarize_dataframe(
```
auguste-probabl: Maybe it could make sense for each accessor to own its markdown repr? I'm okay with the current code but I see that the data accessor part uses a lot of logic from the data accessor code

auguste-probabl (commented):
I quite like this, especially if it means `__repr__` no longer runs `summarize`.

skore/src/skore/_sklearn/_estimator/report.py
glemaitre: Since we are going to compute the `prediction` at `__init__` at some point, this call will be free and will not change the state.

So I would keep this info to not have to pay the cost of making a copy.

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -926,3 +933,128 @@ def _repr_mimebundle_(self, **kwargs):
...
+                    f"meta-estimator {type(est).__name__} "
+                    f"wrapping {type(inner).__name__}"
+                )
+            return f"meta-estimator {type(est).__name__}"
+        if type(est).__module__.startswith("sklearn."):
+            return "scikit-learn estimator"
+        return f"{type(est).__module__.split('.')[0]} estimator"
+
+    def _markdown_estimator_section(self) -> str:
+        lines = [
```
glemaitre: Let's use a `jinja` template. It works whatever the output language :)

glemaitre: Then, it makes me think that a bunch of thing need to be moved in the `_utils/repr`. We could now, organize the module their in subfolder: html, mardkown, text.

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: Can we remove `summarize` from `__repr__` and put it back when we compute predictions at init time then?

GaetandeCast: I'm +1 for removing it and adding it back once we compute predictions at init. I can add a `FIXME` comment as reminder.

skore/src/skore/_sklearn/_estimator/report.py
GaetandeCast: Discussed IRL: as we plan to add reprs to `Display` objects, the text repr of the relevant displays can be the markdown fragments that we generate in the report for now. Aka follow what is already the case for the checks summary: the display has a text repr that is used in the checks section of `to_markdown`.

skore/src/skore/_sklearn/_estimator/report.py
GaetandeCast: Could you expand a bit on why create a jinja template ? I'm not sure I understand the benefits here.

Like what would the scope of the template be ? Creating the whole markdown file or just the estimator section that you commented on ? And what does creating a template bring compared to the current python logic ?

glemaitre: It looks the same pattern than trying to write HTML into Python strings which become not maintainable rapidly.

The point of the template is to have static test and inject the data which is exactly what we do here.

And I mean creating the full markdown file. Then each method can create data object to be injected, I'm fine with this.

skore/src/skore/_sklearn/_estimator/report.py
glemaitre: We already have the pattern for the HTML repr and I'm just more worried that we are not going to think to put it back.

skore/src/skore/_sklearn/_estimator/report.py
GaetandeCast: Okay done, lmk what you think, I'm not super familiar with jinja templates. Also ideally later down the line we can drop the hand made data section in favor of a text repr of the `TableReportDisplay`.

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: > I'm just more worried that we are not going to think to put it back

We can write an issue to remind ourselves

GaetandeCast: Okay to unblock this I'm picking the side of @glemaitre : let's tackle the issue of when and where we want expensive computations to happen in an other place and keep the current behavior as is.

skore/src/skore/_utils/repr/estimator_report_markdown.j2
```diff
@@ -0,0 +1,61 @@
...
+# {{ report_class_name }}: {{ estimator_name }}
+
+## Estimator
+
+- name: `{{ estimator_name }}`
+- kind: {{ estimator_kind }}
+- ml task: {{ ml_task }}
+{%- if fit_time is not none %}
+- fit time: {{ "%.3g"|format(fit_time) }} s
+{%- endif %}
```
auguste-probabl, line 14: this will be in the metrics

skore/src/skore/_utils/repr/estimator_report_markdown.j2
```diff
@@ -0,0 +1,61 @@
...
+{%- if data_source is none %}
+No data provided.
+{%- else %}
+
+```text
+{{ metrics_text }}
+```
+{%- endif %}
+
+Full metrics: `report.metrics.summarize().frame()`.
```
auguste-probabl:
```suggestion
Get the metrics results with: `report.metrics.summarize().frame()`.
```

skore/src/skore/_utils/repr/estimator_report_markdown.j2
```diff
@@ -0,0 +1,61 @@
...
+
+## Checks (fast mode)
+
+```text
+{{ checks_text }}
+```
+
+Full checks: `report.checks.summarize()`.
```
auguste-probabl:
```suggestion
Get the checks results with: `report.checks.summarize()`.
```

skore/src/skore/_utils/repr/estimator_report_markdown.j2
```diff
@@ -0,0 +1,61 @@
...
+
+## Data
+
+{%- if data_source is none %}
+No data provided.
+
+Full analysis: `report.data.analyze()`.
```
auguste-probabl:
```suggestion
```

skore/src/skore/_utils/repr/estimator_report_markdown.j2
```diff
@@ -0,0 +1,61 @@
...
+
+Computed on the {{ data_label }} dataset (n_rows={{ data_n_rows }}, n_columns={{ data_n_columns }}, n_constant_columns={{ data_n_constant_columns }}).
+
+| column | dtype | null_count | n_unique |
+| ------ | ----- | ---------- | -------- |
+{%- for col in data_columns %}
+| {{ col.name }} | {{ col.dtype }} | {{ col.null_count }} | {{ col.n_unique }} |
+{%- endfor %}
+
+Full analysis: `report.data.analyze()`.
```
auguste-probabl:
```suggestion
Get the data analysis with: `report.data.analyze()`.
```

auguste-probabl: Move it after `endif`

skore/src/skore/_utils/repr/estimator_report_markdown.j2
GaetandeCast: I think that it makes sense to have fit time in this section. What do you think about adding predict time too, and dropping both from the metrics dataframe ?

skore/src/skore/_utils/repr/estimator_report_markdown.j2
GaetandeCast: Actually it makes no sense to redirect to `analyze` when there is no data provided, so it's best kept only in this branch. I applied the same idea for metrics.summarize.

skore/src/skore/_utils/repr/estimator_report_markdown.j2
auguste-probabl: Even if their implementation is a bit different, I really think fit time and predict time should be treated like all the other metrics; e.g. they allow comparing models and querying them from a project.

GaetandeCast: I agree but in the context of a single model summary, I like reading the fit and predict time with the components of the pipeline, the type of estimator etc. I'm trying to quickly grasp the contents of my model and I think execution time is helpful.

GaetandeCast: I'm not suggesting removing timing metrics from skore's metrics summary, only changing the order of how information is presented in the markdown summary of a report.

jeromedockes: if they are useful for different reasons in both contexts there is not much harm in showing them twice

auguste-probabl: > I'm not suggesting removing timing metrics from skore's metrics summary

but earlier you wrote

> dropping both from the metrics dataframe ?

so I'm a bit confused. My main point was to keep the timing metrics in `summarize`, but I guess I'm okay to show the information twice; feel free to add the predict time if you feel it's relevant

skore/src/skore/_utils/repr/estimator_report_markdown.j2
```diff
@@ -8,6 +8,7 @@
 - ml task: {{ ml_task }}
 {%- if fit_time is not none %}
 - fit time: {{ "%.3g"|format(fit_time) }} s
+- predict time: {{ "%.3g"|format(predict_time) }} s
```
auguste-probabl: why is it in the if block? should we note that it's the predict time on the test set?

GaetandeCast: Sorry I went too fast. I put it in it's own if/else and added which set it was computed on.

auguste-probabl (approved):
nice job @GaetandeCast !

PR #2960, opened 2026-05-30
feat: Implement Project.Summary as a proper Display and improve component

### LINKED ISSUE #2654: feat(skore): Replace ipywidget by simplified HTML component for `project.summarize`
Right now, the UX of the `ipywidget` is not ideal. We should revisit the UX and UI of this component to simplify its use.

glemaitre:
closes #2654 

First iteration to have a better components to go through the estimator stored in a project:

- a table with the information
- a plot view with the parallel coordinate plot
- another plot to have a time vs. score view

We don't try to modify the state, we use what the `TableReport` provide: a query-string that can be used with the display:

```python
summary = project.summarize()
# interact and get the string
report = summary.query("the query-string").compare(return_as="report")
```

The point here is to use the plot only for the selection that they are useful and reduce any dependencies around `anywidgets` and `ipywidgets` as well.

The documentation has been updated and we have an updated landing page that show off the web components.


jeromedockes:
from a functional point of view it seems very complete and nice. I haven't reviewed the code yet

glemaitre:
@rouk1 Could you have a look at the javascript components to see if you see some gotcha that I did not observe before.

I'll solve the merge conflicts but they will not have any impact on the design.

glemaitre:
As seen with @jeromedockes let's merge and iterate.

PR #2965, opened 2026-06-02
docs(skore): Use the california housing dataset on landing page

### LINKED ISSUE #2963: Docs: Landing page is different accross versions
Every time we regenerate the landing page, the results are slightly different. As far as I can tell, it's because the `random_state` of the `StringEncoder` in the `TableVectorizer` created by `tabular_pipeline` is `None`.

Do we care ?

GaetandeCast:
Closes #2963

This PR changes the landing page to use the `california_housing` dataset from skrub. This avoids usage of `StringEncoder` in `tabular_pipeline` so the outputs are reproducible. 

As a bonus side effect, the prediction error plot is a little nicer I think, in the sense that it provides insight: The house values are capped in the dataset and we can distinguish some groups in the splits.

### DIFF
--- a/sphinx/sphinxext/landing_page.py
+++ b/sphinx/sphinxext/landing_page.py
@@ -15,9 +15,10 @@ import matplotlib.pyplot as plt
 from jinja2 import Environment, FileSystemLoader
 from sphinx.application import Sphinx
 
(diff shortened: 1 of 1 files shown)


GaetandeCast:
It's strange, the plot does not get updated on the preview

jeromedockes (approved):

PR #2966, opened 2026-06-02
fix: Set `Score.greater_is_better` to None

auguste-probabl:


auguste-probabl:
Actually, sklearn assumes that `.score()` is `greater_is_better`.

PR #2974, opened 2026-06-03
ci: Drop support of scikit-learn `1.5` in favor of `1.9`

thomass-dev:
Following https://github.com/probabl-ai/skore/pull/2967.

### DIFF
--- a/.github/workflows/sphinx.yml
+++ b/.github/workflows/sphinx.yml
@@ -68,8 +68,8 @@ defaults:
 
 env:
   PYTHON: '3.14'
-  SCIKIT_LEARN: '1.8'
-  LOCKFILE: 'ci/requirements/skore/python-3.14/scikit-learn-1.8/sphinx-requirements.txt'
+  SCIKIT_LEARN: '1.9'
+  LOCKFILE: 'ci/requirements/skore/python-3.14/scikit-learn-1.9/sphinx-requirements.txt'
 
 jobs:
   sphinx-changes:
--- a/README.md
+++ b/README.md
@@ -115,17 +115,18 @@ Join our mission to promote open-source and make machine learning development mo
 
 Skore is tested on Linux and Windows, for at most 4 versions of Python, and at most 4 versions of scikit-learn:
 - Python 3.11
-  - scikit-learn 1.5
-  - scikit-learn 1.8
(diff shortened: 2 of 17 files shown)


.github/workflows/sphinx.yml
```diff
@@ -68,8 +68,8 @@ defaults:
 
 env:
   PYTHON: '3.14'
-  SCIKIT_LEARN: '1.8'
-  LOCKFILE: 'ci/requirements/skore/python-3.14/scikit-learn-1.8/sphinx-requirements.txt'
+  SCIKIT_LEARN: '1.9'
+  LOCKFILE: 'ci/requirements/skore/python-3.14/scikit-learn-1.9/sphinx-requirements.txt'
```
auguste-probabl, line 72: do we need to mention the versions here?

GaetandeCast:
I realize we might want to clean up https://github.com/probabl-ai/skore/blob/main/skore/src/skore/_externals/_sklearn_compat.py to remove compatibility with versions we don't support anymore.

glemaitre:
> I realize we might want to clean up https://github.com/probabl-ai/skore/blob/main/skore/src/skore/_externals/_sklearn_compat.py to remove compatibility with versions we don't support anymore.

I don't think we care because it is vendoring of https://github.com/sklearn-compat/sklearn-compat/

So the idea would be to go there pick it up and replace the entire file.

glemaitre:
The bug might actually be an underlying bug in the data where we have twice the same column that polars would not support while pandas was. Now narwhals is a dependency of scikit-learn as a layer of abstraction and it is the reason why it is the package screaming.

We might only need to rename the columns most probably. To be checked.

thomass-dev:
> The bug might actually be an underlying bug in the data where we have twice the same column that polars would not support while pandas was. Now narwhals is a dependency of scikit-learn as a layer of abstraction and it is the reason why it is the package screaming.
> 
> We might only need to rename the columns most probably. To be checked.

Fixed by #2979 .

auguste-probabl (dismissed):
Let's go

.github/workflows/sphinx.yml
thomass-dev: needed for sphinx

auguste-probabl (approved):

PR #2975, opened 2026-06-03
docs(skore): Change dataset used in getting_started

jeromedockes:
to not rely on openml

to make it more interesting, we could change the preprocessing to replace StringEncoder by TextEncoder as this improves the scores noticeably, but it requires pytorch

### DIFF
--- a/examples/getting_started/plot_getting_started.py
+++ b/examples/getting_started/plot_getting_started.py
@@ -70,8 +70,8 @@ X_experiment, X_holdout, y_experiment, y_holdout = train_test_split(
 # We will investigate two different families of models using cross-validation.
 #
 # 1. A :class:`~sklearn.linear_model.LogisticRegression` which is a linear model
-# 2. A :class:`~sklearn.ensemble.HistGradientBoostingClassifier` which is a more
-#    powerful model (gradient boosting).
+# 2. A :class:`~sklearn.ensemble.RandomForestClassifier` which is a more
+#    powerful model.
 #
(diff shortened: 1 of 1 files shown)


examples/getting_started/plot_getting_started.py
```diff
@@ -90,8 +70,8 @@
 # We will investigate two different families of models using cross-validation.
 #
 # 1. A :class:`~sklearn.linear_model.LogisticRegression` which is a linear model
-# 2. A :class:`~sklearn.ensemble.RandomForestClassifier` which is an ensemble of
-#    decision trees.
+# 2. A :class:`~sklearn.ensemble.HistGradientBoostingClassifier` which is a more
+#    powerful model (gradient boosting).
```
GaetandeCast: You keep using `RandomForest` afterwards, so it's inconsistent.

GaetandeCast (changes requested):
> but it requires pytorch

FYI we used to depend on pytorch for the docs and removed it to reduce the size of our dependencies: #2270

auguste-probabl:
sanity check: do we rely on openml in any other place in the docs?

glemaitre:
> sanity check: do we rely on openml in any other place in the docs?

I would say no: https://github.com/search?q=repo%3Aprobabl-ai%2Fskore%20fetch_openml&type=code

examples/getting_started/plot_getting_started.py
jeromedockes: thanks

GaetandeCast (approved):
Thanks!

PR #2976, opened 2026-06-03
feat(skore): Introduce CLI for installing skills

glemaitre:
Implement an alternative to `npx skills add` with Python as first citizen.

Creating interactive app using `textual` and making that the cli uses `click`, `rich` and `rich-click`.

### Interactive app

The following two commands are instantiating a textual app:

- `skore skills install`
- `skore skills find`

To guide the user to find the right information.

Those commands can be used in non-interactive mode when passing flags that corresponds to the choices in the interactive app.

### Other commands

- `skore skills remove`
- `skore skills update`
- `skore skills list`

are three other command to manage the skills.

### Things to consider

The commands are a bit slow and I assume that it is related to the import. I think that the lazy loading of module would be a good thing to try to improve the state.

https://github.com/user-attachments/assets/1b4d7c0a-b0f0-4493-9da0-6237b7c3e9f3


rouk1 (changes requested):
I tested that a bit and I have a few feedbacks.

- launching the CLI is a bit slow
- launching with no parameter does not trigger an interactive option UX (so you need to relaunch the script and it is slow so it is a bit frustrating)
- when selection skills pressing tab is hard to discover (bold text is not readable enough to me)
- when selecting agent and scope you must press two keys to select what you want (using the arrow should select the highlighted choice no ?) 


IMO all this could have been done with a multistep non fullscreen questionary like experience.
1. chose action (list, find, ... plus exit)
2. chose action first choice (if only one skip) (for install select among workflows + cancel)
3. continue through list of questions or go back
4. final step may have a progress bar

glemaitre:
> IMO all this could have been done with a multistep non fullscreen questionary like experience.

I actually try `questionary`, `beaupy`and `InquirierPy` and I was not happy with some styling issues (selection or background) for each of them and I ending up doing the textual app because it was where I can control the flow.

@rouk1 I think that I can correct all your point easily apart of the speed to launch. For me this one could be done in a later step with lazy import or doing a separate package with only the cli (not sure that it is a good idea).

glemaitre:
The following PR (https://github.com/probabl-ai/skore/pull/2988) will decrease the import time from 2.5s to 0.15s with the lazy loading mechanism.

jeromedockes:
- the skills do not seem to take a huge amount of disk space, so is there a downside to just downloading everything / do we need a TUI vs just a command like `probabl-skills path/to/project` or something similar?
- it seems quite independent from the rest of skore so should it be in the same package?
- as the cli does not seem to import the rest of skore maybe the lazy-loading thing could be done only in the top-level init and not subpackages

thomass-dev:
> * it seems quite independent from the rest of skore so should it be in the same package?

Discussed IRL with @glemaitre , and given its independence from skore, i also think it's a good idea to host the CLI in its own repository.

By the way, great job 👏🏼 .

glemaitre:
> the skills do not seem to take a huge amount of disk space, so is there a downside to just downloading everything / do we need a TUI vs just a command like probabl-skills path/to/project or something similar?

I would say there is two points that bring me towards this complex proposal:

- others are doing this UX
- the location of the skills (and later whatever will come) is agent dependent sometimes

> it seems quite independent from the rest of skore so should it be in the same package?
> as the cli does not seem to import the rest of skore maybe the lazy-loading thing could be done only in the top-level init and not subpackages

Iterating with several persons then I think that the subpackage is actually the way to go. It postpone the complexity of lazy loading and we have a way to keep the entry point depending on `skore` to live in the skore package, and the others orthogonal thing to live in a separate package with the dependencies that are only required for the CLI.

Aljutor:
I expect that skills installation could be triggered by the agent itself in non interactive terminal. 

Maybe we can have simpler and faster CLI and only trigger TUI when it's directly needed?  
Also I haven't found a way to list skills to install without TUI.

glemaitre:
> I expect that skills installation could be triggered by the agent itself in non interactive terminal.

You can do that by passing the option directly.

glemaitre:
> Also I haven't found a way to list skills to install without TUI.

Yep I need to revisit the command that I added the TUI yesterday to be sure that we have the non-interactive one.

glemaitre:
Starting a new repo: https://github.com/probabl-ai/skore-cli

glemaitre:
OK so one remaining question that I would have is the name of the CLI where I don't know if I need to use `skore` or `skore-cli`.

If using `skore`, we will have:

```bash
skore skills install
uvx --from skore-cli skore skills install
pixi exec --spec skore-cli skore skills install
```

or with `skore-cli`:

```bash
skore-cli skills install
uvx skore-cli skills install
pixi exec skore-cli skills install
```

Aljutor:
I think it should be simple “skore skills install”

glemaitre:
This can be close now.

PR #2978, opened 2026-06-03
feat(hub): Send all metrics including custom

auguste-probabl:
Closes https://github.com/probabl-ai/skore-hub/issues/928

Replace the current class-based architecture with delegating to `summarize()` so that custom metrics are included.
Since skore-lib users can order metrics by passing `metrics.add(metric, position="last")`, derive the position attribute in the payload from the order of metrics in `summarize`.

#### Performance results

TL;DR: Metrics computation/serialization for the hub is 1.5-4.5x faster. My hypothesis: the metric computations tend to be very fast (predictions are cached) so the overhead introduced by multithreading and indirection is significant in comparison. 

<details>

Benchmark script:

```py
from __future__ import annotations
import skore

import gc
import statistics
import sys
import time
from contextlib import contextmanager

from sklearn.datasets import make_classification, make_regression
from sklearn.linear_model import LogisticRegression, Ridge

from skore import CrossValidationReport, EstimatorReport
from skore._plugins.hub.report import (
    CrossValidationReportPayload,
    EstimatorReportPayload,
)


@contextmanager
def timer(label):
    gc.collect()
    t0 = time.perf_counter()
    yield
    dt = time.perf_counter() - t0
    print(f"  {label}: {dt:.3f}s")


def _new_payload(cls, report):
    """Construct a payload without Pydantic validation, no Project needed."""
    p = cls.model_construct(project=None, report=report, key="bench")
    # ``model_post_init`` is needed for CrossValidationReportPayload.
    try:
        p.model_post_init(None)
    except Exception:
        pass
    return p


def _fresh_payload_and_time(cls, report):
    payload = _new_payload(cls, report)
    gc.collect()
    t0 = time.perf_counter()
    _ = payload.metrics
    return time.perf_counter() - t0


def bench(name, builder, n_runs=3):
    print(f"\n[{name}]")
    # warmup
    report = builder()
    _ = _fresh_payload_and_time(*report)
    times = []
    for i in range(n_runs):
        report = builder()
        dt = _fresh_payload_and_time(*report)
        times.append(dt)
        print(f"  run {i + 1}: {dt:.3f}s")
    print(f"  median: {statistics.median(times):.3f}s  mean: {statistics.mean(times):.3f}s")
    return times


# --------- builders ---------


def build_estimator_classif():
    X, y = make_classification(n_samples=10_000, n_features=20, n_informative=10, random_state=0)
    X_tr, X_te, y_tr, y_te = train_test_split(X, y, random_state=0)
    est = LogisticRegression(max_iter=1000)
    report = EstimatorReport(
        est, X_train=X_tr, y_train=y_tr, X_test=X_te, y_test=y_te
    )
    return EstimatorReportPayload, report


def build_estimator_regress():
    X, y = make_regression(n_samples=10_000, n_features=20, n_informative=10, random_state=0)
    X_tr, X_te, y_tr, y_te = train_test_split(X, y, random_state=0)
    est = Ridge()
    report = EstimatorReport(
        est, X_train=X_tr, y_train=y_tr, X_test=X_te, y_test=y_te
    )
    return EstimatorReportPayload, report


def build_cv_classif():
    X, y = make_classification(n_samples=5_000, n_features=20, n_informative=10, random_state=0)
    est = LogisticRegression(max_iter=1000)
    report = CrossValidationReport(est, X, y, splitter=5)
    return CrossValidationReportPayload, report


def build_cv_regress():
    X, y = make_regression(n_samples=5_000, n_features=20, n_informative=10, random_state=0)
    est = Ridge()
    report = CrossValidationReport(est, X, y, splitter=5)
    return CrossValidationReportPayload, report


def main():
    print(f"Python {sys.version.split()[0]}")
    print(f"skore from: {skore.__file__}")
    bench("estimator classif", build_estimator_classif)
    bench("estimator regress", build_estimator_regress)
    bench("cv classif", build_cv_classif)
    bench("cv regress", build_cv_regress)


if __name__ == "__main__":
    main()
```

Before:
```
Python 3.13.0
skore from: /home/auguste/Desktop/work/skore/skore/src/skore/__init__.py

[estimator classif]
  run 1: 0.033s
  run 2: 0.030s
  run 3: 0.042s
  median: 0.033s  mean: 0.035s

[estimator regress]
  run 1: 0.018s
  run 2: 0.018s
  run 3: 0.019s
  median: 0.018s  mean: 0.018s

[cv classif]
  run 1: 0.237s
  run 2: 0.222s
  run 3: 0.221s
  median: 0.222s  mean: 0.227s

[cv regress]
  run 1: 0.142s
  run 2: 0.137s
  run 3: 0.136s
  median: 0.137s  mean: 0.138s
```

After:

```
Python 3.13.0
skore from: /home/auguste/Desktop/work/skore/skore/src/skore/__init__.py

[estimator classif]
  run 1: 0.025s
  run 2: 0.030s
  run 3: 0.018s
  median: 0.025s  mean: 0.024s

[estimator regress]
  run 1: 0.005s
  run 2: 0.005s
  run 3: 0.006s
  median: 0.005s  mean: 0.005s

[cv classif]
  run 1: 0.088s
  run 2: 0.090s
  run 3: 0.091s
  median: 0.090s  mean: 0.089s

[cv regress]
  run 1: 0.030s
  run 2: 0.034s
  run 3: 0.031s
  median: 0.031s  mean: 0.032s
```

Speedup based on means:

estimator classif: 1.45x
estimator regression: 3.6x
cv classif: 2.55x
cv regress: 4.31x

</details>


thomass-dev (changes requested):
To ensure i can review.

skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -372,3 +296,62 @@ def estimators(self) -> list[EstimatorReportPayload]:
...
+        The list of scalar metrics that have been computed from the report.
+
+        Notes
+        -----
+        All metrics whose value is not a scalar are currently ignored:
+        - ignore ``list[float]`` for multi-output ML task,
+        - ignore ``dict[str: float]`` for multi-classes ML task.
+
+        The position field is used to drive the ``hub``'s parallel coordinates plot:
+        - int [0, inf[, to be displayed at the position,
```
rouk1: This cant be set a report level. It must be at project level because we can't fetch all reports of a project before knowing in which order metrics needs to be displayed.

rouk1 (commented):
I wont make any comment on the implementation ◝(ᵔᗜᵔ)◜

I tested it locally and the seed script has sent a score metrics as intended 👍🏻

skore/src/skore/_plugins/hub/report/cross_validation_report.py
auguste-probabl: Moreover, the Hub already has `position` defined for every builtin metric, so sending it from lib is useless

auguste-probabl: > we can't fetch all reports of a project before knowing in which order metrics needs to be displayed.

I'm a bit confused by this; do you mean "we [the Hub frontend] can't fetch [from the Hub backend]"?

rouk1: > Moreover, the Hub already has position defined for every builtin metric, so sending it from lib is useless

Iirc the plan was to be able to upsert metrics so the lib can update the positions.

> I'm a bit confused by this; do you mean "we [the Hub frontend] can't fetch [from the Hub backend]"?

I'll try to clarify: I'm concerned that if a project contains lot of reports with various metrics (especially custom ones), the API will have to fetch all of them to build the ranges of the parallel coordinates plot. But this is in fact already happening and we might want to rethink that search API anyway.

That being said this is probably not the right place to discuss this, even if my gut feeling says "we should completely remove that position fiel in favor of an heuristic".

skore/src/skore/_plugins/hub/report/estimator_report.py
```diff
@@ -29,29 +32,17 @@
...
-    RocAucTest,
-    RocAucTrain,
-)
-from skore._plugins.hub.metric.metric import Metric
+from skore._plugins.hub.metric import EstimatorReportMetric
 from skore._plugins.hub.report.report import ReportPayload
+from skore._sklearn._plot.metrics.metrics_summary_display import MetricsSummaryRow
+
+
+def _filter(row: MetricsSummaryRow) -> bool:
```
jeromedockes: can you give it a more explicit name?

skore/src/skore/_plugins/hub/metric.py
```diff
@@ -0,0 +1,83 @@
...
+
+Report = TypeVar("Report", bound=(EstimatorReport | CrossValidationReport))
+
+
+class Metric(BaseModel, Generic[Report]):
+    """Payload used to send a metric to ``hub``."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    report: Report = Field(repr=False, exclude=True)
```
jeromedockes: why isit needed?

skore/src/skore/_plugins/hub/report/cross_validation_report.py
jeromedockes: I have the impression the lib should not deal with the notion of position, and in the future, only if it is useful, maybe the user can select and reorder stuff in the UI, and the hub remembers that display setting. but indeed I would wait for the project overview / search thing to stabilize before thinking about that and remove the position field here if we can (it is arbitrary anyway, and places user-defined metrics behind the default ones AFAICT)

skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -372,3 +296,62 @@ def estimators(self) -> list[EstimatorReportPayload]:
...
+            for row in self.report.metrics.summarize(data_source="both").rows
+            if _filter(row)
+        ]
+
+        first_report = self.report.reports_[0]
+        positions = {name: i for i, name in enumerate(first_report._metric_registry)}
+
+        result = []
+        key = itemgetter("metric_name", "data_source")
+        for (name, data_source), group_ in groupby(sorted(rows, key=key), key=key):
```
jeromedockes: why not put the rows in a dataframe to do these operations?

auguste-probabl: i like to use the standard library when possible

skore/src/skore/_plugins/hub/metric.py
auguste-probabl: Before this PR it was needed so that each `Metric` subclass could compute its own value. With this PR `report` indeed no longer needs to be an attribute

skore/src/skore/_plugins/hub/report/cross_validation_report.py
jeromedockes: I can relate to that but in a datascience context where pandas and numpy available it is very unusual to use itertools groupby or the statistics module

auguste-probabl: fair enough, i'll take a look

skore/src/skore/_plugins/hub/metric.py
auguste-probabl: Removed

skore/src/skore/_plugins/hub/report/cross_validation_report.py
auguste-probabl: set position to `None` all the time, didn't remove the attribute completely since the hub expects it

skore/src/skore/_plugins/hub/report/cross_validation_report.py
auguste-probabl: done

jeromedockes (dismissed):
LGTM but @thomass-dev will review as well

skore/src/skore/_plugins/hub/report/report.py
```diff
@@ -72,59 +67,19 @@ def estimator_class_name(self) -> str:
...
-                self.report.y_test
-                if isinstance(self.report, EstimatorReport)
-                else self.report.y
-            ),
+        fingerprint: str = joblib.hash(
+            self.report.y_test
+            if isinstance(self.report, EstimatorReport)
+            else self.report.y
         )
+        return fingerprint
```
thomass-dev, line 76: Please revert, the cast function is a noop at runtime.

skore/src/skore/_plugins/hub/report/report.py
```diff
@@ -72,59 +67,19 @@ def estimator_class_name(self) -> str:
...
 
     @computed_field  # type: ignore[prop-decorator]
     @property
     def ml_task(self) -> str:
         """The type of ML task covered by the report."""
         return self.report.ml_task
 
-    @computed_field  # type: ignore[prop-decorator]
-    @cached_property
-    def metrics(self) -> list[Metric[Report]]:
```
thomass-dev, line 94: Please set the property as abstact: it is part of the `Report` contract and must be implemented by the sub-classes.

Something like:

```python
    @computed_field    # type: ignore[prop-decorator]
    @property
    @abstractmethod
    def metrics(self) -> list[Metric[Report]]:
    """"
    doctstr
    """"
    ...
```

skore/src/skore/_plugins/hub/metric.py
thomass-dev, line 1: In your implementation, compared to the previous one, there is no fundamental difference between `EstimatorReportMetric` and `CrossValidationReportMetric`. `Metric` should be used everywhere in place of the two mentioned classes.

Also, the docstring should stay high-level: you don't need here the details about what the name is in such and such cases, or about the future version.

skore/src/skore/_plugins/hub/metric.py
```diff
@@ -0,0 +1,76 @@
...
+class Metric(BaseModel, Generic[Report]):
+    """Payload used to send a metric to ``hub``."""
+
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    name: str
+    verbose_name: str
+    data_source: Literal["train", "test"] | None
+    greater_is_better: bool | None
+    # TODO: Remove the attribute if Hub no longer requires it
```
thomass-dev: Please remove and prefer a dedicated issue in GH.

skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -101,7 +61,7 @@ class CrossValidationReportPayload(ReportPayload[CrossValidationReport]):
 
     Attributes
     ----------
-    METRICS : ClassVar[tuple[Metric, ...]]
+    metrics : list[CrossValidationMetric]
         The metric classes that have to be computed from the report.
```
thomass-dev:
```suggestion
```

Other properties are not listed in the attribute section, you can just drop it.

skore/src/skore/_plugins/hub/report/estimator_report.py
```diff
@@ -54,8 +36,8 @@ class EstimatorReportPayload(ReportPayload[EstimatorReport]):
 
     Attributes
     ----------
-    METRICS : ClassVar[tuple[Metric, ...]]
-        The metric classes that have to be computed from the report.
+    metrics : list[EstimatorReportMetric]
+        The metrics that have to be computed from the report.
```
thomass-dev:
```suggestion
```

Other properties are not listed in the attribute section, you can just drop it.

skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -361,3 +280,50 @@ def estimators(self) -> list[EstimatorReportPayload]:
...
+        )
+
+        return [
+            CrossValidationReportMetric(
+                name=f"{row['metric_name']}_{suffix}",
+                verbose_name=f"{row['metric_verbose_name']} - {suffix.upper()}",
+                data_source=row["data_source"],
+                greater_is_better=row["greater_is_better"]
+                if suffix == "mean"
+                else False,
```
thomass-dev, line 325: Nitpick:

```suggestion
                greater_is_better=((suffix == "mean") and row["greater_is_better"]),
```

skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -361,3 +280,50 @@ def estimators(self) -> list[EstimatorReportPayload]:
...
+        """
+        The list of scalar metrics that have been computed from the report.
+
+        Notes
+        -----
+        All metrics whose value is not a scalar are currently ignored:
+        - ignore ``list[float]`` for multi-output ML task,
+        - ignore ``dict[str: float]`` for multi-classes ML task.
+        """
+        data = self.report.metrics.summarize(data_source="both").data
```
thomass-dev, line 296: Could you confirm whether "unavailable" metrics are filtered-out in the summary?

skore/src/skore/_plugins/hub/report/estimator_report.py
```diff
@@ -107,3 +67,33 @@ class EstimatorReportPayload(ReportPayload[EstimatorReport]):
...
+            data["label"].isna() & data["output"].isna() & data["average"].isna()
+        ]
+
+        return [
+            EstimatorReportMetric(
+                value=row["score"],
+                name=row["metric_name"],
+                verbose_name=row["metric_verbose_name"],
+                data_source=row["data_source"],
+                greater_is_better=row["greater_is_better"],
```
thomass-dev: Nitpick: to be sorted as in `cross_validation_report.py`

```suggestion
                name=row["metric_name"],
                verbose_name=row["metric_verbose_name"],
                data_source=row["data_source"],
                greater_is_better=row["greater_is_better"],
                value=row["score"],
```

skore/src/skore/_plugins/hub/metric.py
thomass-dev: Please update all files of report payloads.

skore/tests/unit/plugins/hub/test_metric.py
thomass-dev, line 1: There is a regression in the tests. As it was done here in `test_metrics.py`, could you please test exhaustively each payload in both `test_cross_validation_report.py:test_metrics` and `test_estimator_report.py:test_metrics` ?

```python

# test_metrics.py

...

        assert_almost_equal(metric_payload.pop("value"), value)
        assert metric_payload == {
            "name": name,
            "verbose_name": verbose_name,
            "greater_is_better": greater_is_better,
            "data_source": data_source,
            "position": position,
        }

...
```

thomass-dev (changes requested):
In terms of code readability, it is a great improvement to use `report.metrics.summarize`, thanks.

Could you please benchmark the "time" of the compute of the metrics payload before and after your PR (without upload)? I'm curious to see the impact, because before the compute was threaded, and now (maybe i'm wrong), i don't see any "parallelization" in the `report.metrics.summarize`.

skore/src/skore/_plugins/hub/report/report.py
auguste-probabl: I find this a bit cleaner, since it doesn't require an extra import

thomass-dev: It's purely coding-style appreciation, and the difference is not mandatory till ruff doesn't complain.

skore/src/skore/_plugins/hub/metric.py
auguste-probabl: https://github.com/probabl-ai/skore/issues/3025

skore/src/skore/_plugins/hub/report/cross_validation_report.py
auguste-probabl: even if they are equivalent, I think the existing formulation is clearer

skore/src/skore/_plugins/hub/report/cross_validation_report.py
thomass-dev: Discussed IRL: for some reasons, `None` values can flow here.

So this PR introduces a regression: null value can't be send to the hub, and were filtered-out in previous implementation.

So can you fix it and remove all metrics that are None, NaN etc.?

skore/tests/unit/plugins/hub/test_metric.py
auguste-probabl: I removed the timings because they are non-deterministic, let me know if that's ok

skore/src/skore/_plugins/hub/report/cross_validation_report.py
auguste-probabl: Made the `value` attribute of `Metric` a `float`, rather than a `float | None`. `report.metrics.summarize` indeed should never contain metrics that are in fact not available (e.g. a classification metric on a regression report). If the metric itself returns None for another reason (e.g. precision with `zero_division=np.nan`), Pydantic will cast it to `float('nan')`, which is a float so the Hub will accept it.

auguste-probabl:
> Could you please benchmark 

Added a benchmark in the PR description

auguste-probabl:
I don't know how but my attempt to merge `main` in put some of the `main` commits in the PR. I have to clean up with a rebase. Sorry reviewers; you can still see my changes since last review by navigating the individual commits (first one is "address review")

skore/tests/unit/plugins/hub/test_metric.py
auguste-probabl: Added them back by relying on pytest.approx

skore/src/skore/_plugins/hub/report/cross_validation_report.py
thomass-dev: FYI, `NaN` are not allowed in the hub, but since we use orjson, `float('nan')` is converted to `b'null'` and not `b'NaN'`.

So in the current implementation, `float('nan')` is equivalent to `None` and should likewise a `None` be filtered out/not sent to the hub. The hub will fail otherwise. Could you fix it please?

skore/src/skore/_plugins/hub/report/report.py
thomass-dev: I'm still not convinced that this change is necessary, otherwise id make plenty of styling changes to the code myself in the places i don't like.

I'm not going to argue over and over on a minor difference. Just try to hold back in future, it makes the reviews longer.

skore/tests/unit/plugins/hub/test_metric.py
thomass-dev: I didn't know about `approx`, that's cool!

thomass-dev (changes requested):
One last comment to address https://github.com/probabl-ai/skore/pull/2978#discussion_r3426623611.

Good job otherwise!

skore/src/skore/_plugins/hub/report/cross_validation_report.py
auguste-probabl: Skipped `NaN` completely.

thomass-dev (approved):

PR #2986, opened 2026-06-04
feat: Add `metrics.get` to compute custom metrics without `summarize`

auguste-probabl:
Closes https://github.com/probabl-ai/skore-hub/issues/1539.

### DIFF
--- a/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_comparison/metrics_accessor.py
@@ -255,6 +255,55 @@ class _MetricsAccessor(_BaseAccessor[ComparisonReport], DirNamesMixin):
         for report in self._parent.reports_.values():
             report.metrics.remove(name)
 
+    def get(
+        self,
+        name: str,
+        data_source: DataSource = "test",
+        aggregate: Aggregate | None = ("mean", "std"),
+        flat_index: bool = False,
+        **kwargs,
+    ) -> pd.DataFrame | None:
+        """Get a metric value.
+
+        Parameters
+        ----------
+        data_source : {"test", "train"}, default="test"
+            The data source to use.
+
+            - "test" : use the test set provided when creating the report.
+            - "train" : use the train set provided when creating the report.
+
+        aggregate : {"mean", "std"}, list of such str or None, default=("mean", "std")
+            Function to aggregate the scores across the cross-validation splits.
+            None will return the scores for each split.
+
+        flat_index : bool, default=True
+            Whether to return a flat index or a multi-index.
+
+        Returns
+        -------
+        pd.DataFrame
+            The metric values, or None if the metric is not available.
+
+        Examples
+        --------
+        >>> from sklearn.datasets import load_breast_cancer
+        >>> from sklearn.linear_model import LogisticRegression
+        >>> from skore import evaluate
+        >>> X, y = load_breast_cancer(return_X_y=True)
+        >>> estimator_1 = LogisticRegression(max_iter=10000, random_state=42)
+        >>> estimator_2 = LogisticRegression(max_iter=10000, random_state=43)
+        >>> comparison_report = evaluate([estimator_1, estimator_2], X, y, splitter=0.2)
+        >>> comparison_report.metrics.get("precision")
+        Estimator        LogisticRegression_1  LogisticRegression_2
+        Metric    Label
+        Precision 0                  0.901961              0.901961
+                  1                  0.984127              0.984127
+        """
+        return self._metric(metric_name=name, data_source=data_source, **kwargs).frame(
+            aggregate=aggregate, flat_index=flat_index
+        )
+
     def timings(
         self,
         *,
--- a/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
@@ -215,6 +215,55 @@ class _MetricsAccessor(_BaseAccessor[CrossValidationReport], DirNamesMixin):
         for report in self._parent.reports_:
             report.metrics.remove(name)
 
+    def get(
+        self,
+        name: str,
+        data_source: DataSource = "test",
+        aggregate: Aggregate | None = ("mean", "std"),
+        flat_index: bool = False,
+        **kwargs,
+    ) -> pd.DataFrame | None:
+        """Get a metric value.
+
+        Parameters
+        ----------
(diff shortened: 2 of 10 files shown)


thomass-dev:
What is the state of the metric accessors with this PR?

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -288,6 +288,42 @@ def remove(self, name: str) -> None:
...
+        >>> X, y = load_breast_cancer(return_X_y=True)
+        >>> classifier = LogisticRegression(max_iter=10_000)
+        >>> report = evaluate(classifier, X, y, splitter=0.2)
+        >>> report.metrics.get("precision")
+        {0: 0.90..., 1: 0.98...}
+        """
+        metric = self._parent._metric_registry.get(name)
+        if metric is None:
+            return None
+        return metric.pretty(report=self._parent, data_source=data_source, **kwargs)
```
jeromedockes, line 325: this is not introduced by this PR (.accuracy accepted kwargs before) but I wonder about the role of kwargs outside of `add` (which is where the metric is created and given a name). It would simplify things a bit if we considered that a metric is a name associated with a number (or list of numbers) -- pinball loss with alpha=.1 and alpha=.5 are just 2 different metrics, each with a name, an entry in the summarize, and a line in the hub. it would also simplify the code to have less messing around and combining dicts of kwargs. WDYT?

skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -215,6 +215,55 @@ def remove(self, name: str) -> None:
         for report in self._parent.reports_:
             report.metrics.remove(name)
 
+    def get(
+        self,
+        name: str,
+        data_source: DataSource = "test",
+        aggregate: Aggregate | None = ("mean", "std"),
+        flat_index: bool = False,
+        **kwargs,
```
jeromedockes, line 224: kwargs seems to be ignored

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
```diff
@@ -255,6 +255,55 @@ def remove(self, name: str) -> None:
         for report in self._parent.reports_.values():
             report.metrics.remove(name)
 
+    def get(
+        self,
+        name: str,
+        data_source: DataSource = "test",
+        aggregate: Aggregate | None = ("mean", "std"),
+        flat_index: bool = False,
+        **kwargs,
```
jeromedockes, line 264: kwargs seems to be ignored

thomass-dev:
Also, i would like to re-open a discussion about user API @glemaitre @jeromedockes 

```
EstimatorReport                                                                   
└── .metrics                                                                      
    ├── .add(...)                  -                                              
    ├── .available(...)            - List available metric names in the registry. 
    ├── .fit_time(...)             - Get time to fit the estimator.               
    ├── .get(...)                  - Get a metric value.                          
    ├── .mae(...)                  - Compute the mean absolute error.             
    ├── .mape(...)                 - Compute the mean absolute percentage error.  
    ├── .predict_time(...)         - Get prediction time if it has been already   
    │   measured.                                                                 
    ├── .prediction_error(...)     - Plot the prediction error of a regression    
    │   model.                                                                    
    ├── .r2(...)                   - Compute the R² score.                        
    ├── .remove(...)               - Remove a metric from the registry.           
    ├── .rmse(...)                 - Compute the root mean squared error.         
    ├── .score(...)                - Compute the estimator's default score.       
    ├── .summarize(...)            - Report a set of metrics for our estimator.   
    └── .timings(...)              - Get all measured processing times related to 
        the estimator.                                                            
```

It is really confusing in a user point of view that "scores" and "plots" are mixed under the same accessor `.metrics`.
Especially, the `summarize` and `get` functions can only address "scores" and not "plots", but i can call myself other functions that return plots.

As a user, the term "metric" is really unclear if all functions under the `.metrics` accessor are not "metrics".

`.get(...)                  - Get a metric value.`
`.summarize(...)            - Report a set of metrics for our estimator. `

jeromedockes:
> It is really confusing in a user point of view that "scores" and "plots" are mixed under the same accessor .metrics.

i agree, and plots cannot be custom or added to the registry. also when thinking of a metric I think it is a number or possibly a few numbers. maybe they should go in a `plots` accessor?

jeromedockes:
> What is the state of the metric accessors with this PR?

WDYM?

thomass-dev:
> i agree, and plots cannot be custom or added to the registry. also when thinking of a metric I think it is a number or possibly a few numbers. maybe they should go in a `plots` accessor?

For my part, I was thinking about `.metrics` and `.displays`.

thomass-dev:
> > What is the state of the metric accessors with this PR?
> 
> WDYM?

Sorry, i mean: will the related functions (`accuracy` etc.) in the metric accessors be deleted?

jeromedockes:
> will the related functions (accuracy etc.) in the metric accessors be deleted?

currently that is the plan, or at least they will become dynamic attributes that do not appear in the api documentation and are only there as shorthands for a call to get():

https://github.com/probabl-ai/skore-hub/issues/941#issuecomment-4603885788

(of course that is up for debate, and the original issue is probably the right place to discuss this plan)

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: It's a convenience to let users quickly try a different kwarg value without having to add a whole new metric. This feature is historically there to let users call skore metrics just like they would sklearn metrics, e.g. `report.metrics.precision(average="micro")`. If these explicit methods disappear, maybe the kwargs are less relevant, but I think having to add new metrics for every value of a kwarg would be cumbersome. On the other hand, there is a proposal for adding while computing at the same time; maybe that would solve that problem

jeromedockes: I see what you mean but calling a metric is not just about getting the immediate result but also adding that result to the report and its local and skore-hub exports, and that is more usable if it has a name / there are not several results associated with a single metric name.

but we can revisit later

jeromedockes (approved):
thanks!

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
auguste-probabl: i would tend to agree. I think the "compute and add" feature would be a good compromise then

PR #2988, opened 2026-06-04
feat: Lazy load modules at import

glemaitre:
This change will reduce the time for using the CLI from 2.5s to 0.15s thanks to the lazy import mechanism.

`_lazy_loader` is vendored to avoid a new dependency.

We follow the design proposed in https://scientific-python.org/specs/spec-0001/ and add `.pyi` file to not break static typing tools.


thomass-dev:
Discussed IRL with @glemaitre: this PR should be closed if https://github.com/probabl-ai/skore/pull/2976 is moved in its own repository (no urge; keep the code simple; wait for python improvements).

glemaitre:
Closing this one because we will not need right now.

PR #2989, opened 2026-06-05
feat(skore): Add repr_mimebundle to MetricsSummaryDisplay

GaetandeCast:
Towards #2941.

Adds reprs to the output of `report.metrics.summarize()` which is the repr of the dataframe obtained after `.frame()` with default parameters + a suffix suggesting to use `.frame` to control the format of the ouput.

image

image


auguste-probabl (dismissed):
LGTM Thanks!

auguste-probabl (approved):

PR #2993, opened 2026-06-05
fix: Make `EstimatorReport.metrics.get` raise when metric is not known

auguste-probabl:
To align with the other reports


skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -321,7 +321,7 @@ def get(
         """
         metric = self._parent._metric_registry.get(name)
         if metric is None:
-            return None
+            raise KeyError()
```
jeromedockes:
```suggestion
            raise KeyError(name)
```

ATM we can only get() metrics that correspond to an entry in the registry. this is not necessarily the same thing as the metrics shown in summarize. I think this is fine but we should mention it in the doc

auguste-probabl: added some docs

jeromedockes (approved):

PR #2998, opened 2026-06-08
feat(skore): Add repr to accessors

GaetandeCast:
Towards #2941

Adds repr to all 4 accessors:

- `.data` and `.inspection` show their respective help trees.
- `.checks` shows the repr of the `ChecksSummaryDisplay`
- `.metrics` shows the metrics summary frame.

### DIFF
--- a/skore/src/skore/_sklearn/_checks/accessor.py
+++ b/skore/src/skore/_sklearn/_checks/accessor.py
@@ -140,5 +140,5 @@ class _ChecksAccessor(_BaseAccessor[_BaseReport], DirNamesMixin):
     def __repr__(self) -> str:
         return repr(self.summarize())
 
-    def _html_repr(self) -> str:
+    def _repr_html_(self) -> str:
         return self.summarize()._repr_html_()
--- a/skore/src/skore/_sklearn/_checks/accessor.py
+++ b/skore/src/skore/_sklearn/_checks/accessor.py
@@ -138,7 +138,11 @@ class _ChecksAccessor(_BaseAccessor[_BaseReport], DirNamesMixin):
             self._parent._applicable_codes.discard(code)
 
     def __repr__(self) -> str:
-        return repr(self.summarize(fast_mode=True))
+        return (
+            "Checks summary: (fast mode)\n"
+            f"{self.summarize(fast_mode=True)!r}\n"
+            "Explore available methods with .help()."
+        )
 
     def _repr_html_(self) -> str:
         return self.summarize(fast_mode=True)._repr_html_()
--- a/skore/src/skore/_sklearn/_checks/accessor.py
+++ b/skore/src/skore/_sklearn/_checks/accessor.py
@@ -138,7 +138,7 @@ class _ChecksAccessor(_BaseAccessor[_BaseReport], DirNamesMixin):
             self._parent._applicable_codes.discard(code)
 
     def __repr__(self) -> str:
-        return repr(self.summarize())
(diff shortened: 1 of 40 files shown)


skore/src/skore/_sklearn/_checks/accessor.py
```diff
@@ -142,3 +136,9 @@ def remove(self, code: CheckCode) -> None:
             self._parent._check_results_cache.pop(code, None)
         if hasattr(self._parent, "_applicable_codes"):
             self._parent._applicable_codes.discard(code)
+
+    def __repr__(self) -> str:
+        return repr(self.summarize())
```
jeromedockes: summarize can take some time as it runs the baseline right? i'm not sure if that's ok for a `__repr__` which is usually supposed to be lightweight and very unlikely to fail. however maybe it's ok here because it is on the accessor so not an object that is likely to be manipulated on its own, stored in a variable, printed in a log or by a debugger etc 🤔

GaetandeCast: Yes it can be expensive, and actually I should use `fast_mode=True` to filter out some of the checks. I meant to do it and forgot.

IMO since the "only" method behind `.checks` is summarize, it's ok to run the fast checks in accessor repr as you'll want to run `summarize` next and the results will be cached.

We can also just have a short text saying to run `summarize` instead.

GaetandeCast: Makes me realize that the non-dummy baseline checks should be marked as slow so they are not run in `fast_mode`.

skore/src/skore/_sklearn/_checks/accessor.py
```diff
@@ -142,3 +136,9 @@ def remove(self, code: CheckCode) -> None:
             self._parent._check_results_cache.pop(code, None)
         if hasattr(self._parent, "_applicable_codes"):
             self._parent._applicable_codes.discard(code)
+
+    def __repr__(self) -> str:
+        return repr(self.summarize(fast_mode=True))
```
auguste-probabl: `checks` doesn't have a `.help` right?

auguste-probabl (dismissed):
Looks great!

skore/src/skore/_sklearn/_checks/accessor.py
GaetandeCast: Actually it does, the accessor inherits from `_BaseAccessor` which does.

auguste-probabl: Then mention it in the repr for symmetry with the other accessors

skore/src/skore/_sklearn/_checks/accessor.py
jeromedockes: will you do that in a separate pr?

jeromedockes: also I think checks that are skipped due to fast mode should be shown in a skipped section, or there should be a message like "3 more checks were skipped because they are more expensive, run summarize() to display"

GaetandeCast: > will you do that in a separate pr?

Yep. I can bundle it with a skipped section. I also want to add a progress bar in the loop on checks.

skore/src/skore/_sklearn/_checks/accessor.py
GaetandeCast: done

auguste-probabl (approved):

PR #3001, opened 2026-06-08
feat(skore): Add .plot method to MetricSummaryDisplay

glemaitre:
This PR add a `.plot` method to the `MetricSummaryDisplay`.

The use case for the `MetricSummaryDisplay` is to filter a single metric and show a bar or box+swarmplot


glemaitre:
Looking at the current implementation, I think that we should modify the look of the frame provided by `.frame` by not allowing anything else that a flat index. It would then be aligned with the other display.

We can use the HTML repr to reformat to a multi index if we want an improved visual.

jeromedockes:
nice addition! fyi ATM subplot_by="data_source" was giving an error for me

```
ValueError: No columns to group by. `subplot_by` is expected to be None or 'auto'.
```

glemaitre:
Yep I wanted to know what we need to change on the data frame. Most probably it will be the first thing to do.

skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -561,7 +575,354 @@ def __repr__(self) -> str:
...
+        -------
+        list of str
+            Unique metric names from the ``name`` column of :attr:`summary`,
+            in the order they first appear. These are the same names accepted
+            by :meth:`plot`.
+        """
+        if self.summary.empty or "name" not in self.summary.columns:
+            return []
+        return list(dict.fromkeys(self.summary["name"].tolist()))
+
```
glemaitre: Actually, we should not need this public method. We should instead inline on the unique value of the column of `summary`. The user will not need it.

skore/tests/unit/displays/conftest.py
```diff
@@ -8,6 +8,15 @@
...
 
+@pytest.fixture
+def pyplot():
+    """Setup and teardown fixture for matplotlib plotting tests."""
+    pyplot = pytest.importorskip("matplotlib.pyplot")
+    pyplot.close("all")
+    yield pyplot
+    pyplot.close("all")
+
+
```
glemaitre: I think that we already have something similar that makes that the test might not need this fixture anymore. You should check the history for the change that we did by removing this fixture.

PR #3005, opened 2026-06-09
feat(skore/checks): Add a progress bar

GaetandeCast:
Adds a progress bar on the checks loop and marks SKD009 and SKD010 as slow. 

cc @jeromedockes

I'm looking at how to show checks that were skipped in a new tab, maybe show checks that were ignored there too. Also looking into showing checks that raised `CheckNotApplicable` in a separate tab.

I'm going to tackle all that separately.

### DIFF
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -13,6 +13,7 @@ from skore._project.git import git_commit
 from skore._sklearn._checks._utils import CheckNotApplicable
 from skore._sklearn._checks.base import Check, CheckCode
 from skore._sklearn._checks.model_checks import _BUILTIN_CHECKS
+from skore._utils._progress_bar import track
 from skore._utils.repr.base import AccessorHelpMixin, ReportHelpMixin
 
 
@@ -74,15 +75,19 @@ class _BaseReport(ReportHelpMixin):
         if not hasattr(self, "_applicable_codes"):
             self._applicable_codes: set[CheckCode] = set()
 
-        for check in self._checks_registry:
-            if (
-                check.report_type != self._report_type
-                or check.code in self._check_results_cache
-                or check.code in ignored_codes
-            ):
-                continue
-            if fast_mode and getattr(check, "slow", False):
-                continue
+        checks_to_run = [
+            check
+            for check in self._checks_registry
+            if check.report_type == self._report_type
+            and check.code not in self._check_results_cache
+            and check.code not in ignored_codes
+            and not (fast_mode and getattr(check, "slow", False))
+        ]
+        for check in track(
+            checks_to_run,
+            description="Running checks",
+            total=len(checks_to_run),
+        ):
             try:
                 explanation = check.check_function(self)
                 self._applicable_codes.add(check.code)
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -87,6 +87,7 @@ class _BaseReport(ReportHelpMixin):
             checks_to_run,
             description="Running checks",
             total=len(checks_to_run),
+            disable=fast_mode,
         ):
             try:
                 explanation = check.check_function(self)
--- a/skore/src/skore/_sklearn/_checks/base.py
+++ b/skore/src/skore/_sklearn/_checks/base.py
@@ -69,6 +69,7 @@ class ChecksSummaryDisplay(DisplayHelpMixin):
         self,
         check_results: dict[CheckCode, dict],
         n_ignored_codes: int,
+        fast_mode: bool,
     ) -> None:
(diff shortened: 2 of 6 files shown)


skore/src/skore/_sklearn/_base.py
```diff
@@ -74,15 +75,19 @@ def _get_results(
...
-                continue
-            if fast_mode and getattr(check, "slow", False):
-                continue
+        checks_to_run = [
+            check
+            for check in self._checks_registry
+            if check.report_type == self._report_type
+            and check.code not in self._check_results_cache
+            and check.code not in ignored_codes
+            and not (fast_mode and getattr(check, "slow", False))
```
jeromedockes, line 84: as summarize is called in repr, we should not show the progress bar in fast mode no?

jeromedockes: it would be weird to have a progress bar flash before the repr appears if you just do 

```
>>> report.checks
```

skore/src/skore/_sklearn/_base.py
```diff
@@ -74,15 +75,19 @@ def _get_results(
...
+        ]
+        for check in track(
+            checks_to_run,
+            description="Running checks",
+            total=len(checks_to_run),
+        ):
```
thomass-dev, line 91: Two solutions to disable the progress-bar:

- add a disable parameter to the track function (currently there is not), and set the value here,
- use the configuration contextmanager directly in the root `__repr__`

```python
with skore.configuration(show_progress=(skore.configuration["show_progress"] and <predicate>)):
    ....
```

GaetandeCast: Thanks :pray: 
I'll go with option 1 to as it is more robust: we don't have to hide the progress bar everywhere we run only the fast checks.

skore/src/skore/_sklearn/_checks/base.py
```diff
@@ -84,11 +85,13 @@ def __init__(
...
         )
         self._n_ignored_codes = n_ignored_codes
+        self._fast_mode = fast_mode
 
     @property
     def _header(self) -> str:
         return (
-            f"Checks summary: {len(self.frame(severity='issue'))} issue(s), "
+            f"Checks summary {'(fast mode)' if self._fast_mode else ''}: "
+            f"{len(self.frame(severity='issue'))} issue(s), "
```
GaetandeCast: Kind of unrelated but I added a mention to the fast mode here and removed the header on the accessor repr to avoid duplication: 

Before
Image

After

Image

skore/src/skore/_sklearn/_base.py
GaetandeCast: Done

skore/src/skore/_utils/_progress_bar.py
```diff
@@ -70,12 +70,20 @@ def advance(self):
...
+    disable: bool = False,
 ) -> Iterable[Any]:
     """Track progress by iterating over a sequence."""
     if total is None:
         total = float(length_hint(sequence)) or None
 
+    if disable:
+        for value in sequence:
+            yield value
+        return
```
thomass-dev: Let the `ProgressBar` do it for you (just flow the value):

```
class ProgressBar:
    """Simplified progress bar based on ``rich.Progress``."""

    def __init__(self, description: str, total: float | None, disable: bool = False):
        from skore import THREADABLE, console
        from skore._config import configuration

        progress = Progress(
            SpinnerColumn(),
            TextColumn("[bold cyan]{task.description}"),
            BarColumn(
                complete_style="dark_orange",
                finished_style="dark_orange",
                pulse_style="orange1",
            ),
            TextColumn("[orange1]{task.percentage:>3.0f}%"),
            console=console,
            expand=False,
            transient=True,
            auto_refresh=THREADABLE,
            disable=(disable or (not configuration.show_progress)),  # <--
        )
```

GaetandeCast: Ooh, I stopped at our `ProgressBar`, didn't see that rich's has the option. Doing it

skore/src/skore/_utils/_progress_bar.py
```diff
@@ -27,7 +27,7 @@ def __init__(self, description: str, total: float | None):
             expand=False,
             transient=True,
             auto_refresh=THREADABLE,
-            disable=(not configuration.show_progress),
+            disable=(not configuration.show_progress or disable),
```
thomass-dev: Just for a question of readibility

```suggestion
            disable=((not configuration.show_progress) or disable),
```

thomass-dev (dismissed):
Small nitpick, otherwise LGTM!

thomass-dev (approved):

jeromedockes (approved):
thanks!

PR #2904, opened 2026-05-13
feat: Implement `data_source="both"` for CrossValidationReport

direkkakkar319-ops:
#### Change description
Fixes #2549

Implements `data_source="both"` for `CrossValidationReport.metrics.summarize()`                                                                       
Previously, passing data_source="both" raised NotImplementedError. Now it returns train and test metrics side by side

Is "EstimatorName (train)" / "EstimatorName (test)" the right format or do they prefer something different

PR #3006, opened 2026-06-09
fix(skore): Support `data_source="both"` for `CrossValidationReport` in PR and ROC Displays

### LINKED ISSUE #2925: Fix `data_source="both"` in metric displays for cv reports
Follow up to #2549 which was only for the `MetricsSummaryDisplay`.

These are the displays to update: 
- [x] RocCurveDisplay #3006
- [x] PrecisionRecallCurveDisplay #3006
- [ ] PredictionErrorDisplay

direkkakkar319-ops:
#### Change description
Fixes #2925 (PART 1)
this pr is focused fix the `data_source=both` for metrics displays cv reports that are 
1. RocCurveDisplay
2. PrecisionRecallCurveDisplay

as mentioned in comment https://github.com/probabl-ai/skore/issues/2925#issuecomment-4660258434 these bugs are also fixed in this pr 

##### Bugs were
1. Inverted dashes 
2. Broken in comparison plots 
3. Wrong legend colors 
4. Wrong legend order

main problem found was that appeareace order was diverging from category order but seaborn used the category order to draw the graph
there the solution to the problem is to remove the divergence at the source by reordering the categories to match appearance order, both orderings become identical, so it no longer matters which one seaborn picked legend and curves always agree was done by adding a helper function `_reorder_categoricals_by_appearance`
 
code used to reproduce the bug was as mentioned in comment https://github.com/probabl-ai/skore/issues/2925#issuecomment-4669147563
```python
import matplotlib
matplotlib.use('Agg')
from sklearn.ensemble import HistGradientBoostingClassifier
from skore import compare, evaluate

report = compare([
    evaluate(HistGradientBoostingClassifier(), X, y, splitter=2),
    evaluate(DecisionTreeClassifier(max_depth=1), X, y, splitter=2),
])
display = report.metrics.roc(data_source="both")
fig = display.plot(subplot_by=None, label=0)
fig.savefig(r'F:\skore_05\fiel.png', dpi=80, bbox_inches='tight')
```
###### before
image

###### after 
image

> [!NOTE]
The issue is for new version of `pandas>=3` or more but there are not checks for it 
Test coverage: both the `roc` and `precision-recall` regression tests assert the same three things - legend color matches the actual curves, legend order follows the order estimators were passed, and train/test linestyles.    The two are intentional twins; the color assertion is the one that fails on the old code under pandas 3.

#### other comments
first i was not able to reproduce the bug but the bug was found to be specific for new version of pandas(3 or above)
tried to keep the fix and docstring similar https://github.com/probabl-ai/skore/pull/2904

code already worked with "both" but we now have just any official support of it as in the following code the `data_source` was just passed 

###### `PrecisionRecallCurveDisplay`
```python
        child_displays = [
            report.metrics.precision_recall(data_source=data_source)
            for report in self._parent.reports_
        ]
```
###### `RocCurveDisplay`
```python
        child_displays = [
            report.metrics.roc(data_source=data_source)
            for report in self._parent.reports_
        ]
        split_indices = range(len(self._parent.reports_))
```

### Ai usage
ai was used for solving the bugs in ther pandas 3.x mentioned above

### DIFF
--- a/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
@@ -1114,7 +1114,7 @@ class _MetricsAccessor(_BaseAccessor[CrossValidationReport], DirNamesMixin):
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
-            - "both" : use both the train and test and show themside-by-side.
+            - "both" : use both the train and test and show them side-by-side.
 
         Returns
         -------
--- a/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
+++ b/skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
@@ -1046,17 +1046,18 @@ class _MetricsAccessor(_BaseAccessor[CrossValidationReport], DirNamesMixin):
     def roc(
         self,
         *,
-        data_source: DataSource = "test",
+        data_source: DataSource | Literal["both"] = "test",
     ) -> RocCurveDisplay:
         """Plot the ROC curve.
 
         Parameters
         ----------
-        data_source : {"test", "train"}, default="test"
+        data_source : {"test", "train", "both"}, default="test"
             The data source to use.
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
+            - "both" : use both the train and test and show them side-by-side.
 
         Returns
         -------
@@ -1102,17 +1103,18 @@ class _MetricsAccessor(_BaseAccessor[CrossValidationReport], DirNamesMixin):
     def precision_recall(
         self,
         *,
-        data_source: DataSource = "test",
+        data_source: DataSource | Literal["both"] = "test",
     ) -> PrecisionRecallCurveDisplay:
         """Plot the precision-recall curve.
 
         Parameters
         ----------
-        data_source : {"test", "train"}, default="test"
+        data_source : {"test", "train", "both"}, default="test"
             The data source to use.
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
+            - "both" : use both the train and test and show themside-by-side.
 
         Returns
         -------
--- a/skore/src/skore/_sklearn/_plot/utils.py
+++ b/skore/src/skore/_sklearn/_plot/utils.py
@@ -255,6 +255,29 @@ def sample_mpl_colormap(
     return [cmap(i) for i in indices]
 
 
+def _reorder_categoricals_by_appearance(
+    plot_data: DataFrame, columns: Sequence[str | None]
+) -> DataFrame:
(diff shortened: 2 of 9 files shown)


direkkakkar319-ops:
hi @GaetandeCast can you please review the pr once so we ca then move to making test cases

skore/src/skore/_sklearn/_plot/utils.py
```diff
@@ -255,6 +255,33 @@ def sample_mpl_colormap(
...
+    """Reorder categorical levels of ``columns`` to their order of appearance.
+
+    Seaborn draws and colors categorical levels by the categorical's *category*
+    order, which (e.g. on pandas >= 3) can differ from the order of appearance used
+    to build the manually constructed legend. That mismatch pairs labels with the
+    wrong color/linestyle and lists estimators out of the order they were passed.
+    Reordering the categories to the order of appearance makes seaborn draw in that
+    same order, keeping curves and legend in sync.
+    """
+    plot_data = plot_data.copy()
```
GaetandeCast: This makes a deep copy by default, which means uncessary memory usage. Let's remove it. Alternatively we could do `deep=False` but I don't think we need to be this defensive here.

skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -1102,17 +1103,18 @@ def roc(
...
 
         Parameters
         ----------
-        data_source : {"test", "train"}, default="test"
+        data_source : {"test", "train", "both"}, default="test"
             The data source to use.
 
             - "test" : use the test set provided when creating the report.
             - "train" : use the train set provided when creating the report.
+            - "both" : use both the train and test and show themside-by-side.
```
GaetandeCast:
```suggestion
            - "both" : use both the train and test and show them side-by-side.
```

skore/tests/unit/displays/test_utils.py
```diff
@@ -7,11 +8,40 @@
...
+        }
+    )
+    assert list(df["estimator"].cat.categories) == ["a", "b"]
+    assert list(df["data_source"].cat.categories) == ["test", "train"]
+
+    result = _reorder_categoricals_by_appearance(df, ["estimator", "data_source", None])
+
+    assert list(result["estimator"].cat.categories) == ["b", "a"]
+    assert list(result["data_source"].cat.categories) == ["train", "test"]
+    assert list(df["estimator"].cat.categories) == ["a", "b"]
```
GaetandeCast: This will fail after removing the copy because of lost immutability which is fine.

GaetandeCast (commented):
Looks great! Just a few nitpicks. Also I'm making changes to the `test_data_source_both_legend_matches_curves` tests to make them a little more straightforward. I'm going to push directly on the branch, I hope you don't mind.

GaetandeCast (approved):
Thanks @direkkakkar319-ops !

direkkakkar319-ops:
Thanks @GaetandeCast !
happy to contribute :)

PR #3007, opened 2026-06-10
chore(login): Forward to hub login transparently

### LINKED ISSUE #3003: link about api key creation in login() docstring outdated
the link that is in the docstring of login seems outdated (at least I get "this resource does not exist")

also when navigating to https://login.probabl.ai/ui/settings i do not see where I can create an api key?

auguste-probabl:
#### Change description

Makes `login` specific to the Hub; previously it had a `mode` argument that did nothing for the other modes.

Closes #3003


thomass-dev (changes requested):
This is not the good way: we want snippets to be compatible whatever the backend is, simply by changing the `mode` parameter in `Project` and `login`. This is why `login` was design as is.

Please revert, and change only the outdated link in the docstring.

auguste-probabl:
Okay thinking a bit deeper about this, I think we can achieve this "snippets work regardless of the backend" by removing `login`, and instead have `Project`/`HUBClient` do the login step when it communicates with the Hub. I opened https://github.com/probabl-ai/skore/issues/3008

PR #3009, opened 2026-06-10
docs(login): Reword

### LINKED ISSUE #3003: link about api key creation in login() docstring outdated

auguste-probabl:
#### Change description

Closes https://github.com/probabl-ai/skore/issues/3003

### DIFF
--- a/skore/src/skore/_project/login.py
+++ b/skore/src/skore/_project/login.py
@@ -11,47 +11,31 @@ logger = getLogger(__name__)
 
 def login(*, mode: ProjectMode = "hub", **kwargs):
     """
-    Login to the storage backend.
+    Log in to Skore Hub for the duration of the session (e.g. script).
 
-    It configures the credentials used to communicate with the desired storage backend.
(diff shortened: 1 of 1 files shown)


jeromedockes (approved):
thanks @auguste-probabl

skore/src/skore/_project/login.py
```diff
@@ -11,47 +11,31 @@
 
 def login(*, mode: ProjectMode = "hub", **kwargs):
     """
-    Login to the storage backend.
+    Log in to Skore Hub for the duration of the session (e.g. script).
```
thomass-dev, line 14: This specific rephrase should be changed, as it goes against the signature: 
```suggestion
    Log in for the duration of the session (e.g. script).
```

auguste-probabl: You mean because the default mode is "hub"? If so this is mentioned in the Parameters section

PR #3010, opened 2026-06-10
feat(checks): Refine hyperparams to tune

GaetandeCast:
Following review with @jeromedockes and @ogrisel


jeromedockes (approved):
thanks @GaetandeCast

PR #3017, opened 2026-06-10
fix(plugin/hub/cross_validation_report/splitting_strategy): Deal with `len(split) == 1`

### LINKED ISSUE #3004: `splitting_strategy` fails with a custom splitter
the cross-validation report `splitting_strategy` field assumes that get_n_splits can be called without inputs, and that data passed to split() can be replaced by randomly-generated numbers. both of which are usually not true. Here is a toy example trying to split a time series by date:

```python
import skore
import pandas as pd
import numpy as np

df = pd.DataFrame(
    {
        "when": pd.to_datetime(
            [
                "2020-01-01",
                "2020-01-02",
                "2020-01-03",
                "2020-01-04",
                "2020-01-05",
                "2020-01-06",
                "2020-01-07",
                "2020-01-08",
            ]
        ),
        "what": ["A", "B", "A", "B", "A", "B", "A", "B"],
        "result": [1, 10, 1, 10, 1, 10, 1, 10],
    }
)


class Splitter:
    def split(self, X, y=None, groups=None):
        df = X.assign(row_index=np.arange(X.shape[0]))
        for split_date in pd.date_range(X["when"].min(), X["when"].max(), freq="48h"):
            train = df[df["when"] <= split_date]["row_index"].to_numpy()
            test = df[df["when"] > split_date]["row_index"].to_numpy()
            if len(train) > 2 and len(test):
                yield train, test

    def get_n_splits(self, X, y=None, groups=None):
        return len(list(self.split(X, y=y, groups=groups)))


report = skore.evaluate(
    DummyRegressor(), X=df.drop(columns="result"), y=df["result"], splitter=Splitter()
)
print(report.metrics.summarize().frame())

# ok: prints
#
#                  DummyRegressor
#                            mean       std
# Metric
# Score                 -0.308148  0.016761
# R²                    -0.308148  0.016761
# RMSE                   5.012413  0.134376
# MAE                    4.914286  0.197949
# MAPE                   1.211429  0.603743
# Fit time (s)           0.000157  0.000074
# Predict time (s)       0.000023  0.000003


skore.login()
project = skore.Project("jerome-workspace-1/project-1", mode="hub")
project.put("CrossValidationReport", report)

# TypeError: Splitter.get_n_splits() missing 1 required positional argument: 'X'
```

also when we start using the splitter configured in a dataop we will also see that the `groups` (or any other metadata) are not passed to `split()` nor `get_n_splits`

thomass-dev:
Deal with splits not compatible with `gaussian_kde`, with only one element `len(y) == 1`.

```python-traceback
File skore/skore/src/skore/_plugins/hub/report/cross_validation_report.py:318, in CrossValidationReportPayload.splitting_strategy(self)
    316     train_kernel = gaussian_kde(train_y)
    317     train_target_distribution = [float(x) for x in train_kernel(linspace)]
--> 318     test_kernel = gaussian_kde(test_y)
    319     test_target_distribution = [float(x) for x in test_kernel(linspace)]
    321 train_target_distributions.append(train_target_distribution)

File scipy/stats/_kde.py:212, in gaussian_kde.__init__(self, dataset, bw_method, weights)
    210 self.dataset = atleast_2d(asarray(dataset))
    211 if not self.dataset.size > 1:
--> 212     raise ValueError("`dataset` input should have multiple elements.")
    214 self.d, self.n = self.dataset.shape
    216 if weights is not None:

ValueError: `dataset` input should have multiple elements.

```

Following https://github.com/probabl-ai/skore/pull/3018, and closes https://github.com/probabl-ai/skore/issues/3004.

---

Third and last iteration of a bunch of bug-fixes all related with the provided reproducer in https://github.com/probabl-ai/skore/issues/3004 .

### DIFF
--- a/skore/src/skore/_plugins/hub/report/cross_validation_report.py
+++ b/skore/src/skore/_plugins/hub/report/cross_validation_report.py
@@ -61,23 +61,28 @@ def _regression_target_kdes(
     y: np.ndarray,
 ) -> tuple[list[list[float]], list[tuple[float, float]]]:
     """Estimate per-target density curves for regression splitting strategy."""
-    sample_count = TARGET_DISTRIBUTION_REPR_SAMPLE_COUNT
-    y = np.asarray(y)
-    if y.ndim == 1:
-        lo, hi = float(y.min()), float(y.max())
-        linspace = np.linspace(lo, hi, num=sample_count)
-        kde = gaussian_kde(y)
-        return [[float(x) for x in kde(linspace)]], [(lo, hi)]
 
-    curves: list[list[float]] = []
-    ranges: list[tuple[float, float]] = []
-    for col in y.T:
-        col_arr = np.asarray(col).ravel()
+    def _target_distribution(
+        col_arr: np.ndarray,
+    ) -> tuple[list[float], tuple[float, float]]:
+        col_arr = np.asarray(col_arr).ravel()
         lo, hi = float(col_arr.min()), float(col_arr.max())
-        linspace = np.linspace(lo, hi, num=sample_count)
-        curves.append([float(x) for x in gaussian_kde(col_arr).evaluate(linspace)])
-        ranges.append((lo, hi))
-    return curves, ranges
+        linspace = np.linspace(lo, hi, num=TARGET_DISTRIBUTION_REPR_SAMPLE_COUNT)
+
+        if len(col_arr) > 1:
+            curve = [float(x) for x in gaussian_kde(col_arr).evaluate(linspace)]
+        else:
+            # Fallback when `y` is not compatible with `gaussian_kde`
+            distribution = np.zeros(TARGET_DISTRIBUTION_REPR_SAMPLE_COUNT, dtype=float)
+            distribution[np.abs(linspace - col_arr[0]).argmin()] = 1.0
+            curve = [float(x) for x in distribution]
+
+        return curve, (lo, hi)
+
+    y = np.asarray(y)
+    columns: Iterable[np.ndarray] = [y] if y.ndim == 1 else y.T
+    curves, ranges = zip(*map(_target_distribution, columns), strict=True)
+    return list(curves), list(ranges)
 
 
 def _flatten_regression_target_kdes(curves: list[list[float]]) -> list[float]:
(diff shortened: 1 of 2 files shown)


skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -313,10 +327,12 @@ def splitting_strategy(self) -> dict[str, Any]:
                     float(train_y.max()),
                     num=TARGET_DISTRIBUTION_REPR_SAMPLE_COUNT,
                 )
-                train_kernel = gaussian_kde(train_y)
-                train_target_distribution = [float(x) for x in train_kernel(linspace)]
-                test_kernel = gaussian_kde(test_y)
-                test_target_distribution = [float(x) for x in test_kernel(linspace)]
+                train_target_distribution = _regression_target_distribution(
```
jeromedockes: it is not introduced by this PR but IIUC this will plot the density of the test data over the support of the training distribution, so it will hide one of the most interesting thing we could discover by looking at such plots which is test data points outside of that support


also, as we are discretizing with the linspace anyway why do the kde instead of just a histogram?

jeromedockes (dismissed):
thanks @thomass-dev !

skore/src/skore/_plugins/hub/report/cross_validation_report.py
thomass-dev: Can we move this (interesting) discussion in its own issue?

cc @rouk1 @glemaitre

jeromedockes: yes you are right i'll open another one

skore/tests/unit/plugins/hub/report/test_cross_validation_report.py
```diff
@@ -470,6 +470,36 @@ def split(X, y=None, groups=None):
...
         }
 
+    def test_splitting_strategy_with_singleton_split(self, project):
+        # non-regression test for https://github.com/probabl-ai/skore/pull/3017
+        X = array([0, 1, 2, 3, 4])
+        y = array([5, 6, 7, 8, 9])
+
+        class Splitter:
+            def split(self, X, y=None, groups=None):
+                yield array([0]), array([1, 2, 3, 4])
```
jeromedockes: maybe we can try also with only one test sample? that is more likely to happen than 1 train sample. or for example use LeaveOneOut from scikit-learn

thomass-dev: Fixed in 70545b35.

jeromedockes (approved):
nice, thanks a lot @thomass-dev !

PR #3011, opened 2026-06-10
fix(plugin/hub/cross_validation_report/splitting_strategy): Use `len(split_indices)`

thomass-dev:
Use `len(report.split_indices)` instead of `splitter.get_n_splits()`, where `report.split_indices` is already computed and always available.
 
We can't control the way `splitter.get_n_splits()` must be called: with or without parameters?

```python-traceback
File skore/skore/src/skore/_plugins/hub/report/cross_validation_report.py:219, in CrossValidationReportPayload.splitting_strategy(self)
    216 is_classifier = "classification" in self.ml_task
    218 n_repeats = getattr(splitter, "n_repeats", None)
--> 219 n_splits = splitter.get_n_splits() // (n_repeats or 1)
    220 splitter_metadata = {
    221     "type": splitter.__class__.__name__,
    222     "n_splits": n_splits,
   (...)    228     "random_state": getattr(splitter, "random_state", None),
    229 }
    231 rng = np.random.default_rng(0)

TypeError: Splitter.get_n_splits() missing 1 required positional argument: 'X'
```

---

First iteration of a bunch of bug-fixes all related with the provided reproducer in #3004 .

PR #3018, opened 2026-06-11
fix(plugin/hub/cross_validation_report/splitting_strategy): Deal with custom splitter

thomass-dev:
Do not try to simplify/recompute splits when splitter is custom.

It addresses especially a bug when `X = rng.normal(size=(rng_size, 1))` changes the type of the parameters passed to the splitter.

```python-traceback
File skore/skore/src/skore/_plugins/hub/report/cross_validation_report.py:278, in CrossValidationReportPayload.splitting_strategy(self)
    275 splits: list[list[int]] = []
    276 X = rng.normal(size=(rng_size, 1))
--> 278 for train_idx, test_idx in simplified_splitter.split(X, target_repr):
    279     split_flags = np.full(rng_size, -1, dtype=int)
    280     split_flags[train_idx] = 0

Cell In[1], line 28, in Splitter.split(self, X, y, groups)
     27 def split(self, X, y=None, groups=None):
---> 28     df = X.assign(row_index=np.arange(X.shape[0]))
     29     for split_date in pd.date_range(X["when"].min(), X["when"].max(), freq="48h"):
     30         train = df[df["when"] <= split_date]["row_index"].to_numpy()

AttributeError: 'numpy.ndarray' object has no attribute 'assign'
```

Following https://github.com/probabl-ai/skore/pull/3011.

---

Second iteration of a bunch of bug-fixes all related with the provided reproducer in https://github.com/probabl-ai/skore/issues/3004 .

### DIFF
--- a/skore/src/skore/_plugins/hub/report/cross_validation_report.py
+++ b/skore/src/skore/_plugins/hub/report/cross_validation_report.py
@@ -4,6 +4,7 @@ from collections import Counter, defaultdict
 from collections.abc import Iterable, Sized
 from functools import cached_property
 from inspect import signature
+from itertools import chain
 from typing import Any, ClassVar, cast
 
 import numpy as np
@@ -42,6 +43,7 @@ from skore._plugins.hub.metric import Metric
 from skore._plugins.hub.report.estimator_report import EstimatorReportPayload
 from skore._plugins.hub.report.report import ReportPayload
 
+SPLITTING_STRATEGY_MAX_INDEX_COUNT = 10_000
 SPLITTING_STRATEGY_REPR_SAMPLE_COUNT = 100
 TARGET_DISTRIBUTION_REPR_SAMPLE_COUNT = 100
 SPLITTERS = {
@@ -175,6 +177,7 @@ class CrossValidationReportPayload(ReportPayload[CrossValidationReport]):
         if self.report.y is None:
             return {}
 
+        splits: list[list[int] | None] = []
         splitter = self.report.splitter
         target = cast(Sized, self.report.y)
         is_classifier = "classification" in self.ml_task
@@ -239,12 +242,19 @@ class CrossValidationReportPayload(ReportPayload[CrossValidationReport]):
                 target_repr,
             )
         else:
-            split_generator = self.report.split_indices
+            total_index_count = sum(
+                len(cast(Sized, indices))
+                for indices in chain.from_iterable(self.report.split_indices)
+            )
+
+            if total_index_count > SPLITTING_STRATEGY_MAX_INDEX_COUNT:
+                splits = [None] * len(self.report.split_indices)
+                split_generator = []
+            else:
+                split_generator = self.report.split_indices
 
         # Per split: one list of length N_SAMPLES_REPR, ordered by sample index,
         # with 0 = train fold and 1 = test fold for that split.
-        splits: list[list[int]] = []
-
         for train_idx, test_idx in split_generator:
             split_flags = np.full(len(target), -1, dtype=int)
             split_flags[train_idx] = 0
--- a/skore/src/skore/_plugins/hub/report/cross_validation_report.py
+++ b/skore/src/skore/_plugins/hub/report/cross_validation_report.py
@@ -193,54 +193,60 @@ class CrossValidationReportPayload(ReportPayload[CrossValidationReport]):
             "random_state": getattr(splitter, "random_state", None),
         }
 
-        rng = np.random.default_rng(0)
-        rng_size = min(len(target), SPLITTING_STRATEGY_REPR_SAMPLE_COUNT)
+        # create a simplified splitter without shuffling and repetitions when possible
+        if simplified_cls := SPLITTERS.get(splitter.__class__.__name__):
+            rng = np.random.default_rng(0)
+            rng_size = min(len(target), SPLITTING_STRATEGY_REPR_SAMPLE_COUNT)
 
-        if len(target) < SPLITTING_STRATEGY_REPR_SAMPLE_COUNT:
-            target_repr = target
-        elif is_classifier:
-            # create an undersampled target to create a simplify representation
-            if not isinstance(target, pd.Series):
-                target = pd.Series(target)
+            if len(target) < SPLITTING_STRATEGY_REPR_SAMPLE_COUNT:
+                target_repr = target
+            elif is_classifier:
+                # create an undersampled target to create a simplify representation
+                if not isinstance(target, pd.Series):
+                    target = pd.Series(target)
 
-            probs = target.value_counts(normalize=True)
-            target_repr = rng.choice(
-                probs.index.to_numpy(),  # classes
-                size=rng_size,
-                p=probs.to_numpy(),  # probabilities
-                replace=True,
+                probs = target.value_counts(normalize=True)
+                target_repr = rng.choice(
+                    probs.index.to_numpy(),  # classes
+                    size=rng_size,
+                    p=probs.to_numpy(),  # probabilities
+                    replace=True,
+                )
+                target_repr.sort()
+            else:  # regression
+                # uniformly sample the target because it will have no impact on the
+                # representation
+                target_repr = rng.choice(
+                    cast(np.ndarray, target), size=rng_size, replace=False
+                )
+
+            simplified_cls_parameters = {}
+
+            for key in signature(simplified_cls.__init__).parameters:
+                if key in splitter_metadata:
+                    simplified_cls_parameters[key] = splitter_metadata[key]
+                elif hasattr(splitter, key):
+                    simplified_cls_parameters[key] = getattr(splitter, key)
+
+            if "shuffle" in simplified_cls_parameters:
+                simplified_cls_parameters["shuffle"] = False
+                simplified_cls_parameters["random_state"] = None
+
+            target = target_repr
+            simplified_splitter = simplified_cls(**simplified_cls_parameters)
+            split_generator = simplified_splitter.split(
+                rng.normal(size=(rng_size, 1)),
+                target_repr,
             )
-            target_repr.sort()
-        else:  # regression
-            # uniformly sample the target because it will have no impact on the
-            # representation
-            target_repr = rng.choice(
-                cast(np.ndarray, target), size=rng_size, replace=False
-            )
-
-        # create a simplified splitter without shuffling and repetitions
-        simplified_cls = SPLITTERS.get(splitter.__class__.__name__, splitter.__class__)
-        simplified_cls_parameters = {}
-
-        for key in signature(simplified_cls.__init__).parameters:
-            if key in splitter_metadata:
-                simplified_cls_parameters[key] = splitter_metadata[key]
-            elif hasattr(splitter, key):
-                simplified_cls_parameters[key] = getattr(splitter, key)
-
-        if "shuffle" in simplified_cls_parameters:
-            simplified_cls_parameters["shuffle"] = False
-            simplified_cls_parameters["random_state"] = None
-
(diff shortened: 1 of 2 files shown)


skore/tests/unit/plugins/hub/report/test_cross_validation_report.py
```diff
@@ -355,6 +356,36 @@ def get_n_splits(self, X, y=None, groups=None):
...
+        X = array([0, 1, 2, 3, 4])
+        y = array([5, 6, 7, 8, 9])
+
+        class Splitter:
+            def __init__(self, arg):
+                # raise TypeError: missing 1 required positional argument: 'arg'
+                pass
+
+            def split(self, X, y=None, groups=None):
+                yield array([0, 1]), array([2, 3, 4])
```
jeromedockes, line 388: maybe we can strengthen the test a bit by making X something with different dtypes like `X = pd.DataFrame({'a': pd.date_range('2020-01-01', freq='1d', periods=5), 'b': ['one', 'two', 'three', 'four', 'five'])` and in split check we are getting the right X (if split were to be called at all) with `pd.testing.assert_frame_equal(X, report.X)` or at least `X.columns == ['a', 'b'] and X.shape == (5, 2)` or something of the sort?

skore/tests/unit/plugins/hub/report/test_cross_validation_report.py
```diff
@@ -355,6 +356,36 @@ def get_n_splits(self, X, y=None, groups=None):
...
+
+            def get_n_splits(self, X, y=None, groups=None):
+                raise Exception
```
jeromedockes:
```suggestion
                raise Exception()
```

skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -229,54 +229,60 @@ def splitting_strategy(self) -> dict[str, Any]:
...
 
         # Per split: one list of length N_SAMPLES_REPR, ordered by sample index,
         # with 0 = train fold and 1 = test fold for that split.
         splits: list[list[int]] = []
-        X = rng.normal(size=(rng_size, 1))
 
-        for train_idx, test_idx in simplified_splitter.split(X, target_repr):
-            split_flags = np.full(rng_size, -1, dtype=int)
+        for train_idx, test_idx in split_generator:
+            split_flags = np.full(len(target), -1, dtype=int)
```
jeromedockes, line 259: not sure how it gets serialized later or if the size matters but I guess dtype could be np.int8 here

jeromedockes (commented):
thanks @thomass-dev ! 

I guess the original motivation for doing the split on fake data is so the index array would not be too big, did you give the pr a try uploading to the hub and see what happens?

skore/src/skore/_plugins/hub/report/cross_validation_report.py
thomass-dev: The size doesn't matter here, because it will be serialized in int.

thomass-dev:
> I guess the original motivation for doing the split on fake data is so the index array would not be too big, did you give the pr a try uploading to the hub and see what happens?

Yes it works!

skore/tests/unit/plugins/hub/report/test_cross_validation_report.py
thomass-dev: With your proposition, i have to test the call done by the `CrossValidationReport` itself.
I will do something different: i will ensure that `split` is only called by the report itself before being pushed.

jeromedockes: but the crossvalidationreport call should pass right? basically we want to assert that all calls made to split are the same (and the reason that is the case is that only one call is made) but you are right checking that it is only called once is probably more straightforward

thomass-dev: > but the crossvalidationreport call should pass right?
 
Yes.

At least with this PR: actually, the only call of `splitter.split` is made in the init of the `CrossValidationReport`. The plugin doesn't call `splitter.split` anymore when the splitter is custom.

So i don't think we need to ensure other thing that the plugin doesn't call `splitter.split`.

jeromedockes:
@thomass-dev due to sending the larger indices array, if i run the toy example from the original issue with more samples now when I try to open the report in the hub it crashes my firefox tab, can you reproduce?

```python
import skore
import pandas as pd
import numpy as np

when = pd.date_range("2020-01-01", freq="60s", periods=100_000)
df = pd.DataFrame(
    {
        "when": when,
        "what": when.astype(str),
        "result": when.day_of_week,
    }
)

# %%


class Splitter:
    def split(self, X, y=None, groups=None):
        df = X.assign(row_index=np.arange(X.shape[0]))
        for split_date in pd.date_range(X["when"].min(), X["when"].max(), periods=20):
            train = df[df["when"] <= split_date]["row_index"].to_numpy()
            test = df[df["when"] > split_date]["row_index"].to_numpy()
            if len(train) > 2 and len(test):
                yield train, test

    def get_n_splits(self, X, y=None, groups=None):
        return len(list(self.split(X, y=y, groups=groups)))


report = skore.evaluate(
    DummyRegressor(), X=df.drop(columns="result"), y=df["result"], splitter=Splitter()
)
print(report.metrics.summarize().frame())

skore.login()
project = skore.Project("jerome-workspace-1/project-1", mode="hub")
project.put("CrossValidationReport", report)
```

pidefrem:
This PR also resolves a crash we hit independently: pushing a `CrossValidationReport` cross-validated with a **group-aware splitter** (`GroupKFold` / `StratifiedGroupKFold`) raises `ValueError: The 'groups' parameter should not be None.` inside `splitting_strategy` — the same root cause noted in #3004 (`groups` is never threaded into the re-instantiated splitter's `.split()`).

On current `main` the lookup `SPLITTERS.get(name, splitter.__class__)` falls back to the group-aware class itself and then calls `.split(X, target_repr)` with no groups → `ValueError`. With this PR, those splitters aren't in `SPLITTERS`, so they take the `else: split_generator = self.report.split_indices` branch and render the **real** group folds. That's strictly better than registering them as plain `KFold` (which we'd prototyped locally — it shows a fake non-grouped partition rather than the actual folds).

One small gap: there doesn't seem to be an explicit regression test covering group-aware splitters. Offering one in case it's useful. The plain constructor splits without `groups`, so the report is built with a normal splitter and the group-produced indices are reassigned:

```python
@mark.parametrize(
    ("group_splitter", "splitter_type"),
    [
        param(GroupKFold(n_splits=2), "GroupKFold", id="GroupKFold"),
        param(
            StratifiedGroupKFold(n_splits=2),
            "StratifiedGroupKFold",
            id="StratifiedGroupKFold",
        ),
    ],
)
def test_group_aware_splitting_strategy(self, project, group_splitter, splitter_type):
    # non-regression for group-aware splitters (see #3004): they are not in
    # SPLITTERS, so splitting_strategy must fall back to report.split_indices
    # instead of re-splitting without `groups`.
    X, y = make_classification(random_state=42, n_samples=8, n_classes=2)
    groups = array([0, 0, 1, 1, 2, 2, 3, 3])

    report = CrossValidationReport(
        LogisticRegression(random_state=42), X, y, splitter=KFold(n_splits=2)
    )
    report._splitter = group_splitter
    report._split_indices = tuple(group_splitter.split(X, y, groups=groups))

    payload = CrossValidationReportPayload(project=project, report=report, key="-")

    strategy = payload.splitting_strategy
    assert strategy["splitter"]["type"] == splitter_type
    assert len(strategy["splits"]) == 2
    # full-length real folds (no synthetic 100-sample representation, no -1 gaps)
    for split in strategy["splits"]:
        assert len(split) == len(y)
        assert set(split) <= {0, 1}
```

Happy to open a separate PR with just this test against your branch if that's easier — whatever fits best.

thomass-dev:
> Happy to open a separate PR with just this test against your branch if that's easier — whatever fits best.

Thanks.
I added a dedicated test, but with `skrub` and not by modifying internal/private attributes.

thomass-dev:
@jeromedockes ready for another review.

On custom/unsupported splitter, if we have less than 10_000 samples counted on all splits, the detail is sent. Otherwise, we sent null splits.

This PR should be reviewed with https://github.com/probabl-ai/skore-hub/pull/1611.

skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -193,54 +196,67 @@ def splitting_strategy(self) -> dict[str, Any]:
...
+            target = target_repr
+            simplified_splitter = simplified_cls(**simplified_cls_parameters)
+            split_generator = simplified_splitter.split(
+                rng.normal(size=(rng_size, 1)),
+                target_repr,
+            )
+        else:
+            total_index_count = sum(
+                len(cast(Sized, indices))
+                for indices in chain.from_iterable(self.report.split_indices)
```
jeromedockes, line 247: ~I think this will give you the number of splits * 2?~

split indices is a list of train_idx, test_idx pairs, where train_idx is the array of train indices and test_idx of test indices

jeromedockes:
thanks @thomass-dev  it works :rocket:  here is an example with a custom splitter

screenshot_2026-06-30T10:41:39+02:00

btw in this splitter I know there is a gap of unused samples between the train and test but I don't see it in the diagram, maybe it is too few rows for the resolution of the image

skore/src/skore/_plugins/hub/report/cross_validation_report.py
thomass-dev: This will give the sum of the total number of idx in each split (train_idx + test_idx).
The rule is: in total, we can't send more than 10_000 idx (arbitrary limit where i have observed UI latency).

thomass-dev: To summarize, this is equivalent to `len(self.report.split_indices) * len(target)` but i wanted us to be resilient against splitters that don't select all the samples.

thomass-dev:
> btw in this splitter I know there is a gap of unused samples between the train and test but I don't see it in the diagram, maybe it is too few rows for the resolution of the image

Let me check what is sent: do you have a reproducer?

```python
        for train_idx, test_idx in split_generator:
            split_flags = np.full(len(target), -1, dtype=int)
            split_flags[train_idx] = 0
            split_flags[test_idx] = 1
```

In such case we should have `-1` between couples of 0/1, not only at the end.
So it can be an UI issue or as you said, only an issue with the resolution of the image.

skore/src/skore/_plugins/hub/report/cross_validation_report.py
jeromedockes: sorry I read this part too fast indeed it looks fine but my question came from seeing a report with more than 10k samples

Image

jeromedockes: wait maybe I am on an old commit when pushing this report

jeromedockes: yes sorry that was me being stupid. the report above was from when I tried an earlier iteration of this PR that sent indices for all datasets. now I get None for this one and if I subsample the expected plot

Image

(where we do see the gap between train and test btw)

jeromedockes: > i wanted us to be resilient against splitters that don't select all the samples.

yes you are right that is often the case

jeromedockes (approved):
thanks @thomass-dev !

jeromedockes:
> only an issue with the resolution of the image.

I would imagine that because with fewer samples we can indeed see it: https://github.com/probabl-ai/skore/pull/3018#discussion_r3497621002

PR #3024, opened 2026-06-12
fix(skore): Update matplotlib rcparams type hint

jeromedockes:


skore/src/skore/_sklearn/_plot/base.py
```diff
@@ -3,6 +3,11 @@
 from typing import Any, Literal, Protocol, runtime_checkable
 
 import matplotlib.pyplot as plt
+
+try:
+    from matplotlib.typing import RcKeyType as MatplotlibRcKeyType
+except ImportError:
+    MatplotlibRcKeyType = str  # type: ignore[misc]
```
jeromedockes: I could not find a way to avoid the ignore here, suggestions welcome

thomass-dev: Could you please use a typing block:

```python
from typing import TYPE_CHECKING

...

IF TYPE_CHECKING:
    try:
        from matplotlib.typing import RcKeyType as MatplotlibRcKeyType
    except ImportError:
        MatplotlibRcKeyType = str  # type: ignore[misc]
```

thomass-dev (changes requested):
LGTM otherwise.

skore/src/skore/_sklearn/_plot/base.py
```diff
@@ -77,7 +78,7 @@ def _plot_plotly(self, **kwargs):
         )
 
 
-DEFAULT_STYLE: dict[MatplotlibRcKeyType, Any] = {
+DEFAULT_STYLE: dict["MatplotlibRcKeyType", Any] = {
```
thomass-dev: You can remove quote by importing `from __future__ import annotations`.

thomass-dev (approved):

PR #3026, opened 2026-06-12
fix(ci/pre-commit): Use user environment and lockfile when running `pre-commit mypy`

thomass-dev:
Prior to this PR, pre-commit managed its own virtual environment and we had to declare dependencies in both `pyproject.toml` and `.pre-commit-config.yaml`.

With this PR:
- for the user, `pre-commit mypy` runs now entirely in the user environment.
- for the CI, `pre-commit mypy` uses now lockfiles (and enjoy the resulting benefits) and the dedicated virtual environments.

So there will be no more difference between running mypy by hand or via pre-commit.

Also:
- add `build` and `mypy` directly in test dependencies, so that they are included in the lockfiles,
- upgrade `matplotlib` to comply linting.


rouk1 (approved):
LGTM !

PR #3032, opened 2026-06-16
feat: Add reprs to all Displays

### LINKED ISSUE #2941: Implement text and interactive reprs for all `*Display` and `*Report` objects

GaetandeCast:
Closes #2941

Implements default `__repr__` and `_repr_html_` in `DisplayMixin` so that all displays get reprs. `MetricsSummaryDisplay`, `ChecksSummaryDisplay` and `TableReportDisplay` override them to define their custom ones. 

The text repr is the default ouput of the `.frame` with a hint to use `.plot` to plot and `.frame` for access to full data. 
The html repr is the default output of `.plot` with a hint to use `.plot` to control the view and `.frame` to access the raw data.

### DIFF
--- a/skore/src/skore/_sklearn/_checks/base.py
+++ b/skore/src/skore/_sklearn/_checks/base.py
@@ -8,8 +8,8 @@ from uuid import uuid4
 import pandas as pd
 
 from skore._externals._sklearn_compat import parse_version
+from skore._sklearn._plot.base import DisplayMixin
 from skore._sklearn.types import ReportType
-from skore._utils.repr.base import DisplayHelpMixin
 from skore._utils.repr.html_repr import render_template
 
 if TYPE_CHECKING:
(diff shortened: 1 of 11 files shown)


glemaitre (approved):
LGTM

PR #3043, opened 2026-06-17
feat: Add support for polars input dataframe using Narwhals

### LINKED ISSUE #2916: polars support
This is not necessarily super high priority but at the same time the sooner we do it the easier it will be.

Given that both scikit-learn and skrub support polars dataframes as inputs it would be great if skore did, too. I don't think much needs to be done, beyond:

- I think there are 1 or 2 places where it might cause errors such as when putting on the hub, I'll try to remember and make a reproducer
- having some tests with polars inputs
- maybe deciding on an explicit strategy for handling this, for now it seems we use the private skrub utilities in a couple of places we may want to discuss if we should use narwhals

This is only about user input data being stored in polars dataframes. all data produced by skore itself such as metrics dataframes would still be in pandas dataframes

glemaitre:
closes #2916

PoC to see how to use Narwhals to add support for polars dataframe as input.


docs/design/0002-narwhals-user-input.md
GaetandeCast, line 1: I had no knowledge of those ADRs lol. I don't think they are in use anymore (some are even obsolete). I guess we should get rid of them, maybe separately.

glemaitre: Yep let's do that in another PR.

skore/src/skore/_plugins/hub/artifact/media/data.py
```diff
@@ -27,9 +29,13 @@ def content_to_upload(self) -> bytes:  # noqa: D102
         table_report = display.summary
 
         # Replace full dataset by its head/tail
-        dataframe = table_report.pop("dataframe")
-        table_report["extract_head"] = dataframe.head(3).to_dict(orient="split")
-        table_report["extract_tail"] = dataframe.tail(3).to_dict(orient="split")
+        dataframe_nw = nw.from_native(table_report.pop("dataframe"))
+        table_report["extract_head"] = (
```
jeromedockes: why to_pandas rather than convert to dict directly?

skore/src/skore/_utils/_dataframe.py
```diff
@@ -1,57 +1,72 @@
...
+def _normalize_X_as_dataframe(X: ArrayLike) -> UserDataFrame:
+    """Normalize feature data as a DataFrame with string column names."""
     if sp.issparse(X):
         raise NotImplementedError(
             "Data analysis via skrub is currently not supported for sparse matrices. "
             "Please use dense data."
         )
 
-    if not sbd.is_dataframe(X):
+    if not nw.dependencies.is_into_dataframe(X):
```
jeromedockes, line 37: don't we need dtype=None in check_array?
or skrub tablereport also accepts np arrays so maybe just check if it is a 2d numpy array is enough

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -423,11 +438,11 @@ def check_function(self, report: _BaseReport) -> str | None:
 
         if X is None:
             raise CheckNotApplicable()
-        X = X.select_dtypes(include="number")
+        X = nw.from_native(X).select(nw.selectors.numeric())
         if X.shape[1] < 2 or X.shape[1] > 1000:
             raise CheckNotApplicable()
 
-        corr = np.abs(spearmanr(X).statistic)
+        corr = np.abs(spearmanr(X.to_numpy()).statistic)
```
jeromedockes, line 445: not related to this pr but we could handle nans here

skore/src/skore/_sklearn/_plot/data/table_report.py
```diff
@@ -375,36 +376,42 @@ def _plot_distribution_1d(
...
+            # `duration_to_numeric` is a skrub helper operating on pandas.
+            plot_column, duration_unit = duration_to_numeric(column.to_pandas())
+        elif is_categorical := not (dtype.is_numeric() or dtype.is_temporal()):
+            top_k = (
+                column.drop_nulls().value_counts(sort=True).head(k)[column.name]
+            ).to_list()
+            # the ordered categorical encoding required by seaborn is pandas-only.
+            plot_column = (
+                column.filter(column.is_in(top_k))
+                .to_pandas()
```
jeromedockes: could we use numpy instead? to_pandas is sometimes problematic because of pyarrow

jeromedockes:
I ran the skore checks and hub upload on a couple of polars examples and it worked properly

GaetandeCast (commented):
I read the part on checks and it looks good. Not sure how we should proceed with testing checks with non pandas/numpy input though. We can figure it out later.

skore/src/skore/_plugins/hub/artifact/media/data.py
glemaitre: because of the `orient="split"` option. Narwhals only the option default from what I understand. Since it is required by the hub and it is a small part of the dataframe I thought it was not worth to try avoiding the conversion.

skore/src/skore/_utils/_dataframe.py
glemaitre: Yep true that we should try to preserve the `dtype`. I'll add a non-regression test where this would have fail.

glemaitre:
I will have a deeper look at the testing. I'm thinking that we could parametrize in a couple of places. Basically, a smoke test for polars should be enough hoping that `narwhals` will test the correspondence between the type of dataframe..

skore/src/skore/_plugins/hub/artifact/media/data.py
jeromedockes: ok it was not about the size but more limiting the use of to_pandas as it can sometimes require pyarrows to be installed but we can definitely revisit later

otherwise the same as dict(orient='split') can be acheived on a nw or polars dataframe with something like

```
{'index': list(range(df.shape[0])), 'columns': df.columns, 'data': df.rows()}
```

glemaitre: Good point. I made the change as you propose.

jeromedockes (approved):
LGTM @glemaitre ! we can think about gradually introducing polars more systematically into the test suite in other prs

PR #3046, opened 2026-06-17
chore(EstimatorReport)!: Cache test predictions at initialization

### LINKED ISSUE #2624: Refactor proposal: Predictions as (cached) property

I propose to:
- compute test predictions as estimator report's property at init time: `.pred_test`  (plus `predicted_proba_test` and `decision_function_test` if relevant) => naming to be discussed (for instance, do we want a `_` at the end? I'd say we don't because a report is not an estimator, it doesn't have a `.fit`)
- expose train predictions as cached property: `.pred_train` etc. Same naming as for the test predictions.
- fix a non-`None` `pos_label` at init, this will help make things cleaner/easier. See https://github.com/probabl-ai/skore/issues/2592
- remove method `cache_predictions()` => becomes useless

Goes in the same direction as:
- #2556 
- #2587
- #2612

auguste-probabl:
Addresses part of https://github.com/probabl-ai/skore/issues/2624.

This is a first step towards removing `cache_predictions` and `clear_cache` from the public API.

One additional benefit for this change is computing predictions is one less thing for `repr` to do.

Since predictions are computed at the end of `EstimatortReport.__init__`, the report now needs to be fitted when this happens, so the `fit` argument has been removed: now, either the estimator is already fitted, or it is fitted in `EstimatorReport.__init__` (the equivalent of `fit="auto"`).


skore/tests/unit/reports/estimator/metrics/test_registry.py
```diff
@@ -401,24 +401,6 @@ def my_metric(y_true, y_pred):
         # At least the metric value and the model predictions
         assert len(report._cache) >= 2
 
-    def test_callable_predictions_not_cached(self, binary_classification_report):
```
auguste-probabl, line 404: the point of this test was to show that the report wouldn't "useless-ly" compute predictions if the custom metric takes care of that, but now it does it anyway.

skore/tests/unit/reports/estimator/metrics/test_timings.py
```diff
@@ -25,18 +25,8 @@ def test_only_fit(estimator_data):
 
     result = report.metrics.timings()
     assert isinstance(result, dict)
-    assert len(result) == 1
-    assert isinstance(result.get("fit_time"), float)
-
-
-def test_only_fit_unfitted(estimator_data):
```
auguste-probabl, line 32: prefitted estimator case is covered by the next test

skore/tests/unit/reports/estimator/metrics/test_timings.py
auguste-probabl:
```suggestion
"""If the estimator is prefitted, and some predictions are computed,
   then `timings` is filled up accordingly."""
```

skore/tests/unit/reports/estimator/test_report.py
```diff
@@ -424,21 +385,21 @@ def test_report_repr_html_sklearn_estimator_bad_html_repr(with_train):
...
-    kwargs = {}
-    if with_train:
-        kwargs.update(X_train=X_train, y_train=y_train)
-    kwargs.update(X_test=X_test, y_test=y_test)
-    report = EstimatorReport(estimator, fit=False, **kwargs)
+    estimator = _DummyClassifierBadRepr(strategy="uniform", random_state=0).fit(
+        X_train, y_train
+    )
+    kwargs = {"X_test": X_test, "y_test": y_test} | (
+        {"X_test": X_test, "y_test": y_test} if with_train else {}
```
GaetandeCast:
```suggestion
        {"X_train": X_train, "y_train": y_train} if with_train else {}
```

skore/src/skore/_sklearn/_estimator/report.py
```diff
@@ -107,18 +107,17 @@ class EstimatorReport(_BaseReport, DirNamesMixin):
...
         - a skrub :class:`~skrub.SkrubLearner` extracted from a :class:`~skrub.DataOp`
           by calling :meth:`~skrub.DataOp.skb.make_learner`.
 
-    fit : {"auto", True, False}, default="auto"
-        Whether to fit the estimator on the training data. If "auto", the estimator
-        is fitted only if the training data is provided.
+        The estimator passed will not be modified in-place:
+
+        - If it is fitted, it will be cloned;
+        - If it is not fitted, it will be deep-copied, then fitted on the training data.
```
GaetandeCast: Actually we always deep copy and we only clone before fitting not fitted estimators.

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
```diff
@@ -358,8 +358,9 @@ def timings(
...
         >>> report = evaluate([estimator_1, estimator_2], X, y, splitter=0.2)
         >>> report.metrics.timings()
-                        LogisticRegression_1    LogisticRegression_2
-        Fit time (s)                     ...                     ...
+                                LogisticRegression_1    LogisticRegression_2
+        Fit time (s)                             ...                     ...
+        Predict time test (s)                    ...                     ...
         >>> report.cache_predictions()
         >>> report.metrics.timings()
                                 LogisticRegression_1    LogisticRegression_2
```
GaetandeCast, line 366: Maybe we can remove the second part that uses `cache_predictions()` now, it is quite redundant.

skore/src/skore/_sklearn/_cross_validation/metrics_accessor.py
```diff
@@ -298,8 +298,9 @@ def timings(
         >>> from skore import evaluate
         >>> report = evaluate(estimator, X, y, splitter=2)
         >>> report.metrics.timings()
-                          mean       std
-        Fit time (s)       ...       ...
+                                    mean       std
+        Fit time (s)                 ...       ...
+        Predict time test (s)        ...       ...
         >>> report.cache_predictions()
```
GaetandeCast, line 304: same

skore/src/skore/_sklearn/_estimator/metrics_accessor.py
```diff
@@ -410,10 +410,10 @@ def timings(self) -> dict:
         >>> estimator = LogisticRegression()
         >>> report = evaluate(estimator, X, y, splitter=0.2)
         >>> report.metrics.timings()
-        {'fit_time': ...}
+        {'fit_time': ..., 'predict_time_test': ...}
         >>> report.cache_predictions()
         >>> report.metrics.timings()
-        {'fit_time': ..., 'predict_time_test': ...}
+        {'fit_time': ..., 'predict_time_train': ..., 'predict_time_test': ...}
```
GaetandeCast, line 416:
```suggestion
        {'fit_time': ..., 'predict_time_train': ..., 'predict_time_test': ...}
```
Same

GaetandeCast (commented):
A few nits, otherwise looks good !

skore/src/skore/_sklearn/_estimator/report.py
jeromedockes: probably what we want is

- clone in the normal case (not fitted)
- in the slightly more unusual case of an already-fitted estimator, deep-copy to get a new object while also copying the fitted state (which would be discarded by clone)

actually i am not sure we need to do that second part, we could also say that we use the object directly. for most estimators making predictions should not modify their state, and if a user wants to be 100% sure it is not modified for whatever reason they can always deep-copy themselves. some estimators are very large objects so avoiding a deep copy could be beneficial

jeromedockes:
> the report now needs to be fitted when this happens, so the fit argument has been removed

i would imagine that `fit` was there to force refitting the estimator, more than prevent fitting an unfitted one (we can't do much if we never fit the estimator anyway).

it's fine to remove it IMO as the same result is easily acheived by cloning the estimator before making the report; maybe that can be mentioned in the docstring

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: Should we do this in this PR?

GaetandeCast: Probably worth doing separately. The docstring inconsistency predates this PR so I'd adjust it here and implement the behaviour we want in a second time.

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
auguste-probabl: the point is, if you just create the report, you get the fit time and the test predict time; if you `cache_predictions` (default is `data_source="both"`) then you also get train predict time

glemaitre:
> i would imagine that fit was there to force refitting the estimator, more than prevent fitting an unfitted one. Particularly useful with pre-fitted neural net.

It is the third case: it here to avoid fitting an already fitting estimator as what we try to have with `skore.evaluate(est, ..., cv="prefit")`.

I did not look what is the concrete impact but since `EstimatorReport` is not anymore the entry point, we could advise for using `FrozenEstimator` (https://scikit-learn.org/stable/modules/generated/sklearn.frozen.FrozenEstimator.html#sklearn.frozen.FrozenEstimator) and if needed we `cv="prefit"` in evaluate can wrap the user estimator in such `FrozenEstimator`.

Edit: I see

> now, either the estimator is already fitted, or it is fitted in EstimatorReport.__init__

which is completely fine to me.

skore/src/skore/_sklearn/_comparison/metrics_accessor.py
GaetandeCast: okay that makes sense.

skore/src/skore/_sklearn/_estimator/report.py
auguste-probabl: opened https://github.com/probabl-ai/skore/issues/3075

GaetandeCast (approved):
LGTM, thanks!

PR #3065, opened 2026-06-18
feat: Disambiguate metrics with the same name

### LINKED ISSUE #3050: Identify custom metrics (including `score`) more precisely
Identify custom metrics (including `score`) more precisely so that they can be differentiated, e.g. when comparing estimators. We can do this in a heuristic way with computing a hash of `inspect.getsource`.

auguste-probabl:
Closes #3050

`Metric` now has an optional field `fingerprint` to identify it more precisely than just by its name.

Custom metrics and `Score` now make use of this fingerprint field, whose implementation is a simple `joblib.hash` of the callable's source code. This is brittle, but more sophisticated approaches can be investigated later.

## Example
```py
from skore import evaluate, compare

X, y = make_regression(random_state=0)

class A(HistGradientBoostingRegressor):
    def score(self, X, y):
        return 1

class B(HistGradientBoostingRegressor):
    def score(self, X, y):
        return 2

def my_metric(est, X, y):
    return 1000

report_A1 = evaluate(A(random_state=0), X, y, splitter=2)
report_A1.metrics.add(my_metric)

def my_metric(est, X, y):
    return 500

report_A2 = evaluate(A(random_state=1), X, y, splitter=2)
report_A2.metrics.add(my_metric)

report_B = evaluate(B(random_state=0), X, y, splitter=2)
report_B.metrics.add(my_metric)

report = compare([report_A1, report_A2, report_B])

report.metrics.summarize().frame()
#                          mean                                std
# Estimator                 A_1         A_2           B        A_1        A_2          B
# Metric
# My Metric_1       1000.000000         NaN         NaN   0.000000        NaN        NaN
# Score_1              1.000000    1.000000         NaN   0.000000   0.000000        NaN
# R²                   0.056581    0.056581    0.056581   0.018712   0.018712   0.018712
# RMSE               126.515178  126.515178  126.515178  16.383589  16.383589  16.383589
# MAE                103.014721  103.014721  103.014721  14.201073  14.201073  14.201073
# MAPE                 3.996748    3.996748    3.996748   0.199124   0.199124   0.199124
# Fit time (s)         0.178121    0.096624    0.101581   0.093845   0.013827   0.022251
# Predict time (s)     0.001963    0.001920    0.001824   0.000130   0.000173   0.000023
# My Metric_2               NaN  500.000000  500.000000        NaN   0.000000   0.000000
# Score_2                   NaN         NaN    2.000000        NaN        NaN   0.000000
```

Note that the additional metrics appear at the end of the dataframe, as this required less logic to implement; happy to reconsider this.


skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -31,6 +34,9 @@ class MetricsSummaryRow(TypedDict):
         Dataset split used to compute the metric.
     greater_is_better : bool or None
         Whether higher or lower values are better.
+    fingerprint : str or None
+        Identifier disambiguating distinct custom metrics that share
+        ``metric_verbose_name``. ``None`` for built-in metrics.
```
jeromedockes, line 37: why is the fingerprint None for built-in metric? we also want to disambiguate if a custom metric tries to take their name no? or can yo explain what is the consequence when the fingerprint is none?

skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
```diff
@@ -111,6 +118,36 @@ def data(self):
...
 
+        data = MetricsSummaryDisplay._resolve_fingerprints(data)
+        return data.drop(columns="fingerprint")
+
+    @staticmethod
+    def _resolve_fingerprints(data: pd.DataFrame) -> pd.DataFrame:
+        """Rename ``metric_verbose_name`` per unique fingerprint by appending ``_i``."""
+        data = data.copy()
+
+        # {"Score": [[0, 3], [4]], ...}
```
jeromedockes: I find that part a bit tricky to follow and I think we could run into trouble if there are some of the verbose names that already have a _i suffix (unlikely, but it could happen in theory for example because that is also how skrub will make names unique if the dataop was configured with several scorers and there are  name clashes)

maybe it makes things slightly clearer if we maintain an explicit set of the names that are already taken, maybe along the lines of something like this

```python
already_used = set() # or set of the built-in metric names, not sure I understand the None fingerprint part
renaming = {} # (old_name, fingerprint) -> new_name

# for each unique (name, fingerprint) pair come up with a name that is not already taken
for name, fingerprint in sorted(
    df[["metric_verbose_name", "fingerprint"]].drop_duplicates().itertuples(index=False)
):
    new_name = name
    i = 0
    while new_name in already_used:
        i += 1
        new_name = f"{name}_{i}"
    renaming[(name, fingerprint)] = new_name
    already_used.add(new_name)

# apply the renaming
for (name, fingerprint), new_name in renaming.items():
    df.loc[
        (df["metric_verbose_name"] == name) & (df["fingerprint"] == fingerprint),
        "metric_verbose_name",
    ] = new_name
```

skore/src/skore/_utils/_callable.py
```diff
@@ -19,3 +22,15 @@ def _callable_name(func: Callable) -> str:
...
 
     return type(func).__name__
+
+
+def _callable_hash(func: Callable) -> str:
+    """Fingerprint a callable based on its source code."""
+    if isinstance(func, functools.partial):
+        return _callable_hash(func.func)
+
+    if hasattr(func, "__call__") and inspect.ismethod(func.__call__):  # noqa: B004
```
jeromedockes, line 40: the difference between method and method-wrapper is subtle so this makes the reader pause and wonder if this will go into an infinite recursion `f.__call__.__call__.__call__` we never want to go down more than one level anyway so I think we can avoid the recursive calls in this function eg by calling a `_hash_source` helper that does no unwrapping

skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
auguste-probabl: It's currently an error to register a metric with the same name as a pre-existing one, so there is no ambiguity issue

jeromedockes: but you could have a scorer returning a dict whose keys match the built-in metric names

basically we have 2 sets of names, the "scorer names" which are the keys in the registry and can be passed to summarize and get and can be checked for duplicated ahead of time, and the "score names", which are the names associated with scorer outputs. for some scorer there will be only one output and the names will match, for others not

skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
auguste-probabl: updated with more tests and a more robust impl addressing these points.

skore/src/skore/_utils/_callable.py
auguste-probabl: done

skore/src/skore/_sklearn/_plot/metrics/metrics_summary_display.py
auguste-probabl: you're right, i've now addressed this -- fingerprints are still `None` for builtins, but the builtins can be overridden safely now, including by multimetric scorers.

jeromedockes (approved):
thanks @auguste-probabl !

PR #3069, opened 2026-06-19
feat(checks): Filter parameters at space edge in search edge check

GaetandeCast:
So that we don't raise an issue when e.g. the found best parameter is 0 for a parameter that takes values in [0, inf).

Uses scikit-learn's `_parameter_constraints` to check if the best parameter is at the parameter space edge.


skore/tests/unit/reports/estimator/test_checks.py
```diff
@@ -387,6 +393,37 @@ def test_skd010_not_detected_for_fast_model(regression_data):
...
+            "ridge__alpha",
+            "left",
+            0.0,
+        ),
+        # Pipeline: unknown step name
+        (
+            Pipeline([("scaler", StandardScaler()), ("ridge", Ridge())]),
+            "unknown__alpha",
+            "left",
+            None,
```
auguste-probabl: No error in this case, so `None` is ambiguous -- not sure if it's really a problem

auguste-probabl: Why not return `inf` if there is no finite right bound?

skore/src/skore/_sklearn/_checks/model_checks.py
auguste-probabl, line 1: just a note on organization: eventually it would make sense for each check to have its file, so that helpers that only serve for one check can be moved out of utils

auguste-probabl: or i guess `get_space_bound` could be a method of the check class in this case

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -152,6 +153,39 @@ class CheckNotApplicable(Exception):
...
+        nested_params = owner.get_params(deep=True)
+        if step_name not in nested_params:
+            return None
+        owner = nested_params[step_name]
+    if not hasattr(owner, "_parameter_constraints"):
+        return None
+
+    # Find the Interval constraint for the leaf_param
+    closed_sides_for_bound = {"left": ("left", "both"), "right": ("right", "both")}
+    for constraint in owner._parameter_constraints.get(leaf_param, []):
```
auguste-probabl:
```suggestion
    if leaf_param not in owner._parameter_constraints:
	    return None
    for constraint in owner._parameter_constraints[leaf_param]:
```

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -152,6 +153,39 @@ class CheckNotApplicable(Exception):
     """
 
 
+def get_space_bound(estimator, param_name: str, side: str) -> float | None:
+    """Return the closed parameter-space boundary for side ('left' or 'right').
```
auguste-probabl:
```suggestion
def get_space_bound(estimator, param_name: str, side: Literal["left", "right"]) -> float | None:
    """Obtain the closed parameter-space boundary for ``side`` if it exists.
```

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: In general i find this function a bit too terse but it's not blocking.

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -152,6 +153,39 @@ class CheckNotApplicable(Exception):
     """
 
 
+def get_space_bound(estimator, param_name: str, side: str) -> float | None:
```
auguste-probabl:
```suggestion
def get_space_bound(estimator, *, param_name: str, side: str) -> float | None:
```

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: I prefer to avoid "Return" when i can ^^

auguste-probabl (commented):
Looking good, just a few comments

skore/tests/unit/reports/estimator/test_checks.py
GaetandeCast: I removed this test case as it does not really make sense in practice. We are inspecting the steps of a scikit-learn `BaseSearchCV` that was successfully fitted so we can assume that most things are well defined, e.g. there is not parameter in the `best_params_` that are unknown to the estimator. I simplified the helper to reflect that. I also tried to make it clearer and moved it to the class.

So now the outcome are straightforward: `float` for a closed bound, `None` for an open bound, including +/- inf, which is also how it's coded in scikit-learn's `Interval` and for when the estimator does not have `_parameter_constraints`, e.g. non sklearn estimators.

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: made it a static class method

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -775,17 +792,19 @@ def check_function(self, report: _BaseReport) -> str | None:
                 best_value, bool | np.bool_
             ):
                 continue
-            if np.isclose(
-                float(best_value), float(search_low), rtol=0.0, atol=0.0, equal_nan=True
-            ):
```
GaetandeCast, line 780: replaced with `==` since we already had no tol.

auguste-probabl (approved):
Looks great

PR #3070, opened 2026-06-20
feat: Support split_kwargs from DataOp for advanced cross-validation

### LINKED ISSUE #2884: Use the cv splitter and scoring of skrub learners

glemaitre:
closes #2884 

Support more advanced cross-validation by leveraging `split_kwargs` from skrub `DataOp`.

### DIFF
--- a/examples/technical_details/plot_skrub_data_op_cv.py
+++ b/examples/technical_details/plot_skrub_data_op_cv.py
@@ -71,5 +71,5 @@ holdout_report
 
 # %%
 # Explicitly passing a ``splitter`` always overrides the DataOp configuration.
-override_report = evaluate(learner, data={"df": df}, splitter=3)
+override_report = evaluate(learner, data={"df": df}, splitter=2)
 override_report
--- a/examples/technical_details/plot_skrub_data_op_cv.py
+++ b/examples/technical_details/plot_skrub_data_op_cv.py
@@ -26,7 +26,7 @@ from sklearn.dummy import DummyClassifier
 from sklearn.model_selection import LeaveOneGroupOut
 
 df = skrub.datasets.toy_products()
-data = skrub.var("df", df)
+data = skrub.var("df")
 groups = data["seller"]
 X = data[["description", "price"]].skb.mark_as_X(
     cv=LeaveOneGroupOut(), split_kwargs={"groups": groups}
--- a/examples/technical_details/plot_skrub_data_op_cv.py
+++ b/examples/technical_details/plot_skrub_data_op_cv.py
@@ -1,0 +1,75 @@
+"""
+.. _example_skrub_data_op_cv:
+
+===================================
+Using skrub DataOp cross-validation
+===================================
+
+When a skrub :class:`~skrub.DataOp` defines a cross-validation splitter on
+:meth:`~skrub.DataOp.skb.mark_as_X`, :func:`~skore.evaluate` can reuse that
+configuration — including ``split_kwargs`` such as ``groups`` — instead of
+skore's default 80/20 holdout.
+
+This example builds a small grouped cross-validation setup with skrub and
+evaluates it with skore.
+"""
+
+# %%
+# Configure cross-validation on the DataOp
+# ========================================
+#
+# We use the toy products dataset and group products by seller. The goal is to
+# assess generalization to new sellers with
+# :class:`~sklearn.model_selection.LeaveOneGroupOut`.
+import skrub
+from sklearn.dummy import DummyClassifier
+from sklearn.model_selection import LeaveOneGroupOut
+
+df = skrub.datasets.toy_products()
+data = skrub.var("df", df)
+groups = data["seller"]
+X = data[["description", "price"]].skb.mark_as_X(
+    cv=LeaveOneGroupOut(), split_kwargs={"groups": groups}
+)
+y = data["category"].skb.mark_as_y()
+pred = X.skb.apply(DummyClassifier(), y=y)
+learner = pred.skb.make_learner()
+
+# %%
+# Evaluate with skore (no explicit splitter)
+# ==========================================
+#
+# Because ``mark_as_X`` was called with an explicit ``cv`` argument, calling
+# :func:`~skore.evaluate` without a ``splitter`` returns a
+# :class:`~skore.CrossValidationReport` that respects the DataOp grouping.
+from skore import evaluate
+
+report = evaluate(learner, data={"df": df})
+report
+
+# %%
+# There are two sellers, so cross-validation runs in two folds:
+len(report.reports_)
+
+# %%
+# Inspect aggregated metrics with the same API as other skore reports:
+report.metrics.summarize().frame()
+
+# %%
+# Default behavior without an explicit DataOp cv
+# ==============================================
+#
+# If ``mark_as_X`` is called without an explicit ``cv`` argument,
+# :func:`~skore.evaluate` still defaults to a single 80/20 holdout and returns
+# an :class:`~skore.EstimatorReport`.
+simple_learner = skrub.X().skb.apply(DummyClassifier(), y=skrub.y()).skb.make_learner()
+holdout_report = evaluate(
+    simple_learner,
+    data={"X": df[["description", "price"]], "y": df["category"]},
+)
+holdout_report
+
+# %%
+# Explicitly passing a ``splitter`` always overrides the DataOp configuration.
+override_report = evaluate(learner, data={"df": df}, splitter=3)
+override_report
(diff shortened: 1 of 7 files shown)


examples/technical_details/plot_skrub_data_op_cv.py
```diff
@@ -0,0 +1,75 @@
...
+#
+# We use the toy products dataset and group products by seller. The goal is to
+# assess generalization to new sellers with
+# :class:`~sklearn.model_selection.LeaveOneGroupOut`.
+import skrub
+from sklearn.dummy import DummyClassifier
+from sklearn.model_selection import LeaveOneGroupOut
+
+df = skrub.datasets.toy_products()
+data = skrub.var("df", df)
```
jeromedockes: if we don't show it we don't really need the df here
```suggestion
data = skrub.var("df")
```

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -21,19 +21,28 @@
...
+    PositiveLabel,
+    SKLearnCrossValidator,
+    _DefaultType,
+)
 from skore._utils._fixes import _validate_joblib_parallel_params
 from skore._utils._parallel import delayed
 from skore._utils._progress_bar import track
-from skore._utils._skrub import eval_X_y, is_skrub_learner, to_estimator, to_learner
+from skore._utils._skrub import (
+    eval_X_y,
```
jeromedockes: i think instead of eval_X_y and then getting the indices separately we should use `.skb.find_X_y` which will return X, y, the splitter and split_kwargs and evaluate all in one go

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -243,8 +252,26 @@ def __init__(
             _check_estimator_and_data(clone(estimator), X, y, data)
         )
         self._pos_label = pos_label
-        self._splitter = check_cv(splitter, y, classifier=is_classifier(estimator))
-        self._split_indices = tuple(self._splitter.split(self.X, self.y))
+        if self._initialized_with_data_op:
+            cv_for_indices = splitter if splitter is not None else _DEFAULT
+            self._split_indices = resolve_data_op_split_indices(
```
jeromedockes: we could get that at the same time as the data

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -225,7 +234,7 @@ def __init__(
         y: ArrayLike | None = None,
         data: dict | None = None,
         pos_label: PositiveLabel | None = None,
-        splitter: int | SKLearnCrossValidator | Generator | None = None,
+        splitter: int | SKLearnCrossValidator | Generator | _DefaultType | None = None,
```
jeromedockes: i'm not sure why we need to accept DefaultTYpe here instead of just None?

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -659,11 +686,11 @@ def input_data(self) -> dict:
         return self._data.copy()
 
     @property
-    def splitter(self) -> SKLearnCrossValidator:
+    def splitter(self) -> SKLearnCrossValidator | None:
```
jeromedockes: if we use find_X_y here we can return the splitter that was produced by evaluating the result

skore/src/skore/_sklearn/_cross_validation/report.py
```diff
@@ -739,7 +766,11 @@ def to_markdown(self) -> str:
                 "splitter_repr": (
                     repr(self.splitter)
                     if self.splitter is not None
-                    else f"{len(self.split_indices)} folds"
+                    else (
+                        "from DataOp"
```
jeromedockes: not important but we could still show the number of splits

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -64,7 +71,11 @@ def evaluate(
         (e.g. ``{"X": X_df, "other_table": df, ...}``).
 
     splitter : float, int, str, or cross-validation object, default=0.2
-        Determines how the data is split:
+        Determines how the data is split. When omitted, a skrub learner whose
```
jeromedockes, line 75: the 80/20 is really restricting a shufflesplit to its first split. i wonder if we could have an api to do that for all splitters, ie separate the "do only the first split" part from the choice of splitter

skore/src/skore/_sklearn/evaluate.py
```diff
@@ -203,6 +214,20 @@ def evaluate(
...
+    if splitter is _DEFAULT:
+        data_op = get_data_op(estimator)
+        if data_op is not None and data_op_has_explicit_cv(data_op):
+            return CrossValidationReport(
+                estimator,
+                X,
+                y,
+                data=data,
+                pos_label=pos_label,
+                splitter=_DEFAULT,
```
jeromedockes: I guess we could pass splitter=None here and not need to handle _DEFAULT in the crossvalidation report?

skore/src/skore/_sklearn/evaluate.py
jeromedockes: also I find the behavior of this default a bit hard to understand 🤔

jeromedockes (dismissed):
it's a great improvement thanks @glemaitre  . i had a few comments but I'm also happy to merge and open a follow-up pr myself

skore/src/skore/_sklearn/evaluate.py
glemaitre: Since we do a single 20% train test split by default, I think we might be in trouble for the others not?

skore/src/skore/_sklearn/_cross_validation/report.py
jeromedockes: something like

```python
X_y_cv = skrub.as_data_op(pred.skb.find_X_y()).skb.eval(env)
data = {"_skrub_X": X_y_cv["X"], "_skrub_y": X_y_cv["y"]}
if X_y_cv.get("cv") is not None:
    split_indices = list(X_y_cv["cv"].split(X_y_cv["X"], X_y_cv["y"], **X_y_cv["split_kwargs"]))
    splitter = X_y_cv["cv"]
```

skore/src/skore/_sklearn/evaluate.py
jeromedockes: i just have the impression there is some manipulation around None and default that could be simplified, here we pass default, which crossvalidationreport converts to default and passes to resolve_data_split_indices which converts back to None 🤔 i guess the only place we really need a mechanism to distinguish None vs "not provided" is here in evaluate because of the 0.2 default

skore/src/skore/_sklearn/evaluate.py
jeromedockes: we will discuss that in another issue

jeromedockes (dismissed):

jeromedockes (approved):

PR #3071, opened 2026-06-20
fix(skore-hub): Handle multioutput regression in the Skore Hub

### LINKED ISSUE #3021: error when put to hub with multioutput regression
### Describe the bug

scipy gaussian kde takes ndim x nsamples input not nsamples x ndim causing an error when doing put on a multi-output regression report

### Steps/Code to Reproduce

```python
import skore
import numpy as np
from sklearn.model_selection import KFold

rng = np.random.default_rng()
X = rng.normal(size=(100, 10))
y = rng.normal(size=(100, 2))

report = skore.evaluate(DummyRegressor(), X=X, y=y, splitter=KFold(3, shuffle=False))
skore.login()
project = skore.Project("jerome-workspace-1/project-2", mode="hub")
project.put("report-2d", report)
```

```
ValueError: Number of dimensions is greater than number of samples. This results in a singular data covariance matrix, which cannot be treated using the algorithms implemented in `gaussian_kde`. Note that `gaussian_kde` interprets each *column* of `dataset` to be a point; consider transposing the input to `dataset`.
```

### Expected Behavior

no error

### Actual Behavior

error

### Environment

```shell
any environment
```

glemaitre:
closes #3021 

Make it possible to send multioutput regression data to the hub for the target distribution

### DIFF
--- a/skore/src/skore/_plugins/hub/report/cross_validation_report.py
+++ b/skore/src/skore/_plugins/hub/report/cross_validation_report.py
@@ -55,6 +55,36 @@ SPLITTERS = {
 }
 
 
+def _regression_target_kdes(
+    y: np.ndarray,
+) -> tuple[list[list[float]], list[tuple[float, float]]]:
+    """Estimate per-target density curves for regression splitting strategy."""
+    sample_count = TARGET_DISTRIBUTION_REPR_SAMPLE_COUNT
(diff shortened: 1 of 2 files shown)


thomass-dev (approved):
LGTM!

Issue #3056, opened 2026-06-18
Improve tests for checks


GaetandeCast:
Here is a bunch of issues that should be resolved for the test suite for the checks:
- coverage is low, especially for the `CheckNotApplicable` cases
- checks could be covered more unitly by testing all the helpers separately from the checks
- reorganize into a dedicated folder. One file per report type, one for helper, one for not applicable cases, one for summarize, etc.
- create fixtures to speed up

auguste-probabl:
Also test error messages explicitly

PR #3074, opened 2026-06-22
feat(checks): Show reason for non applicability

### LINKED ISSUE #3055: Surface the reason a check was not applicable
Have each instance of `CheckNotApplicable` contain an explanation string that can be shown in the checks summary that explains why the check was not applicable.

GaetandeCast:
Adds custom messages to `CheckNotApplicable` instances that explain was the check was aborted. The reason is then stored in the explanation field of the check result and is shown in the display.

This PR also updates the types of `report.X_train` and `report.y_train` to not be able to `None` anymore since we forced them to exist. 

To check it out, this snippet has a lot of not applicable checks:

```python
from skore import evaluate
import pandas as pd
from sklearn.linear_model import RidgeCV

X, y = pd.DataFrame(X, columns=[f"f_{i}" for i in range(X.shape[1])]), pd.Series(y)

model.fit(X, y)
report_ridge = evaluate(model, X, y, splitter="prefit")
report_ridge.checks.summarize()
```

Closes #3055


skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -437,10 +454,12 @@ def check_function(self, report: _BaseReport) -> str | None:
         X = get_preprocessed_X(report, data_source="train")
 
         if X is None:
-            raise CheckNotApplicable()
+            raise CheckNotApplicable("Train data is unavailable.")
         X = nw.from_native(X).select(nw.selectors.numeric())
         if X.shape[1] < 2 or X.shape[1] > 1000:
-            raise CheckNotApplicable()
+            raise CheckNotApplicable(
+                "Train data has less than 2 or more than 1000 features."
```
auguste-probabl: Most of our error messages are of the form:
```suggestion
                f"Expected train data to have between 2 and 1000 features; got {X.shape[1]}."
```

Looking through the codebase we're actually not very consistent, but at least showing the offending value where it makes sense would be great.

skore/tests/unit/reports/estimator/test_checks.py
auguste-probabl, line 1: we need to check our error messages in the tests too. I know it may seem like overkill, but it ensures that the tests pass for the right reasons.

auguste-probabl (changes requested):

skore/tests/unit/reports/estimator/test_checks.py
GaetandeCast: I totally agree. I plan on doing it soon as part of a larger refactor of tests, see #3056.

auguste-probabl (approved):

PR #3076, opened 2026-06-22
fix: Ensure y_test is provided to estimator report

GaetandeCast:
Follow-up to #2992. 

I realized while working on #3074 that `y_test` is not mandatory for non skrub estimators, despite our philosophy of requiring test data to create a report.

For instance, this snippet works on main:
```python
from skore import evaluate
import pandas as pd
from sklearn.linear_model import RidgeCV

X, y = pd.DataFrame(X, columns=[f"f_{i}" for i in range(X.shape[1])]), pd.Series(y)

model.fit(X, y)
report_ridge = evaluate(model, X, splitter="prefit")
```
But any other action, such as displaying the metrics summary or the report's repr, fails.

### DIFF
--- a/skore/tests/unit/reports/comparison/estimator/test_report.py
+++ b/skore/tests/unit/reports/comparison/estimator/test_report.py
@@ -39,14 +39,6 @@ def test_different_test_data(
             ]
         )
 
-    # If there is an X_test but no y_test, it should not raise an error
-    ComparisonReport(
-        [
-            EstimatorReport(estimator, fit=False, X_test=X_test, y_test=None),
-            EstimatorReport(estimator, fit=False, X_test=X_test, y_test=None),
-        ]
-    )
-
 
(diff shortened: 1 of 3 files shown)


skore/tests/unit/reports/comparison/estimator/test_report.py
```diff
@@ -39,14 +39,6 @@ def test_different_test_data(
...
         )
 
-    # If there is an X_test but no y_test, it should not raise an error
-    ComparisonReport(
-        [
-            EstimatorReport(estimator, fit=False, X_test=X_test, y_test=None),
-            EstimatorReport(estimator, fit=False, X_test=X_test, y_test=None),
-        ]
-    )
-
```
GaetandeCast, line 45: We no longer support unsupervised learning and the `y_test=None` error is covered in the next test file.

auguste-probabl (approved):

PR #3079, opened 2026-06-23
fix(checks): Don't modify data before preprocessor

GaetandeCast:
On main, this crashes:
```python

report_ridge = evaluate(model, X, y, splitter=0.2)
report_ridge.checks.summarize()
```

`get_preprocessed_X` normalized numpy input to a DataFrame (columns: `Feature 0`, …) before calling `preprocessor.transform`, but `tabular_pipeline` had been fit on skrub’s auto-generated column names (`0`, `1`, …), causing a column mismatch. The fix is to re-transform the raw stored features from fit and normalize only the output.


glemaitre (dismissed):
LGTM

glemaitre (approved):

PR #3084, opened 2026-06-23
feat(skore): Improve the rendering of the HTML representation for checks

### LINKED ISSUE #3057: Show more checks information in report repr
Currently, in the "Check Summary" section of the report html repr, we only show the header, e.g. `1 issue(s), 1 tip(s), 3 passed, 7 not applicable, 0 ignored.`

Let's show the summary of checks with `fast_mode` instead. We can initially just inject the jinja template of the checks summary html into the report html. We might want to have a more refined view later such as folded lists. Reusing the model selector of the comparison report to only show check results of a single model would be nice too.

glemaitre:
closes #3057 

Improve the rendering of the HTML representation for the checks.

### DIFF
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 from datetime import UTC, datetime
 from importlib.metadata import version
-from typing import Generic, Literal, TypeVar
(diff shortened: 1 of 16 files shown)


GaetandeCast (dismissed):
Thanks a lot!

rouk1 (approved):

thomass-dev (approved):

PR #3086, opened 2026-06-23
chore: Improve support for DataOp and checks

### LINKED ISSUE #3058: Make checks compatible with skrub input
Checks were not implemented with skrub compatibility in mind. Investigate where they break with skrub input and fix them.

glemaitre:
closes #3058 

Make sure that we are supporting skore DataOp in the checks.


glemaitre:
OK I think it is quite ready for a review. @jeromedockes Would you have time to have a look at the skrub part and check that finding the nodes is making sense.

skore/src/skore/_utils/_skrub.py
```diff
@@ -45,6 +49,94 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
...
+    if apply_node is None:
+        raise NotFittedError("No supervised apply step found in the skrub learner.")
+    return apply_node
+
+
+def _supervised_fitted_estimator(learner: SkrubLearner) -> BaseEstimator:
+    impl = _supervised_apply_node(learner.data_op)._skrub_impl
+    if not isinstance(impl, Apply):
+        raise TypeError(
+            f"The supervised step does not represent an estimator application: "
```
auguste-probabl:
```suggestion
            "The supervised step does not represent an estimator application: "
```

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -162,14 +167,21 @@ def split_preprocessor_estimator(estimator):
     """Return ``(preprocessor, predictor)`` from a possibly wrapped estimator.
 
     Splits sklearn :class:`~sklearn.pipeline.Pipeline` into its preprocessing
-    steps and final predictor.
+    steps and final predictor. For :class:`~skrub.SkrubLearner`, returns the
+    inner :class:`~sklearn.pipeline.Pipeline` preprocessor when the supervised
+    apply wraps one (e.g. :func:`~skrub.tabular_pipeline`).
```
auguste-probabl: what is a "supervised apply"

skore/src/skore/_utils/_skrub.py
```diff
@@ -45,6 +49,94 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
     return data_op.skb.find_X_y().get("cv") is not None
 
 
+def _supervised_apply_node(data_op: DataOp) -> DataOp:
```
auguste-probabl:
```suggestion
def _supervised_apply_node(data_op: DataOp) -> Apply:
```

skore/src/skore/_utils/_skrub.py
```diff
@@ -45,6 +49,94 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
...
+    apply_node = find_first_apply(data_op)
+    if apply_node is None:
+        raise NotFittedError("No supervised apply step found in the skrub learner.")
+    return apply_node
+
+
+def _supervised_fitted_estimator(learner: SkrubLearner) -> BaseEstimator:
+    impl = _supervised_apply_node(learner.data_op)._skrub_impl
+    if not isinstance(impl, Apply):
```
auguste-probabl: how can this not be an `Apply`?

skore/src/skore/_utils/_skrub.py
```diff
@@ -45,6 +49,94 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
...
+    if isinstance(x_input, DataOp) and isinstance(x_input._skrub_impl, Apply):
+        return x_input
+    return None
+
+
+def collect_fitted_apply_estimators(data_op: DataOp) -> list[BaseEstimator]:
+    """Return fitted estimators along nested applies from preprocessing to predictor."""
+    apply_node = _supervised_apply_node(data_op)
+    chain: list[BaseEstimator] = []
+    node: DataOp | None = apply_node
```
auguste-probabl:
```suggestion
    node: Apply = apply_node
```

skore/src/skore/_utils/skrub.py
```diff
@@ -45,6 +49,94 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
...
+                "fitted sub-estimators."
+            )
+        chain.append(impl.estimator_)
+        x_input = impl.X
+        node = x_input if isinstance(x_input, DataOp) else None
+    chain.reverse()
+    return chain
+
+
+def iter_fitted_estimator_steps(
```
auguste-probabl, line 179: this function is used only once, and the result is immediately converted to a list

skore/tests/unit/reports/estimator/test_checks.py
```diff
@@ -325,6 +370,26 @@ def test_skd011_detects_golden_feature(estimator):
     assert "Feature 3" not in explanation
 
 
+def test_skd011_skrub_learner_golden_feature():
```
auguste-probabl: can this be moved into the `parametrize` of the previous test?

skore/tests/unit/reports/estimator/test_checks.py
```diff
@@ -234,6 +271,16 @@ def test_skd007_mdi_bias_with_high_cardinality(regression_data, estimator):
     )
 
 
+def test_skd007_skrub_learner_random_forest(regression_pandas_data):
```
auguste-probabl: can this be moved into the previous test's `parametrize`?

skore/tests/unit/utils/test_skrub.py
```diff
@@ -0,0 +1,82 @@
...
+    df, y = regression_xy
+    learner = (
+        skrub.X()
+        .skb.apply(StandardScaler())
+        .skb.apply(Ridge(), y=skrub.y())
+        .skb.make_learner()
+    )
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        report = evaluate(learner, data={"X": df, "y": y})
```
auguste-probabl: what warnings should be ignored? can you add `category` to the filter

skore/tests/unit/utils/test_skrub.py
```diff
@@ -0,0 +1,82 @@
...
+
+    predictor = resolve_fitted_predictor(report.estimator_)
+    assert type(predictor).__name__ == "Ridge"
+    assert not isinstance(predictor, Pipeline)
```
auguste-probabl:
```suggestion
    assert isinstance(predictor, Ridge)
```

skore/tests/unit/utils/test_skrub.py
```diff
@@ -0,0 +1,82 @@
...
+    df, y = regression_xy
+    learner = (
+        skrub.X().skb.apply(tabular_pipeline(Ridge()), y=skrub.y()).skb.make_learner()
+    )
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore")
+        report = evaluate(learner, data={"X": df, "y": y})
+
+    predictor = resolve_fitted_predictor(report.estimator_)
+    assert type(predictor).__name__ == "Ridge"
```
auguste-probabl:
```suggestion
    assert isinstance(predictor, Ridge)
```

auguste-probabl (dismissed):
some comments

skore/src/skore/_sklearn/_checks/_utils.py
glemaitre: `dataop.apply(learner, y=y)`: It is calling the `.apply` method where we are passing the target.

skore/src/skore/_utils/_skrub.py
glemaitre: indeed it cannot be.

skore/src/skore/_utils/skrub.py
glemaitre: Since it is super specific to skrub, I would not just put the code in the checks directly. However, we can avoid the conversion to a list.

skore/src/skore/_utils/_skrub.py
```diff
@@ -45,6 +49,89 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
...
+        if not isinstance(impl, Apply):
+            break
+        if not hasattr(impl, "estimator_"):
+            raise NotFittedError(
+                "The skrub learner has not been fitted. Call fit() before inspecting "
+                "fitted sub-estimators."
+            )
+        chain.append(impl.estimator_)
+        x_input = impl.X
+        node = x_input if isinstance(x_input, DataOp) else None
```
auguste-probabl: nit: using `None` to signal the end of the loop is a bit confusing

```suggestion
    node: DataOp = apply_node
    while True:
        impl = node._skrub_impl
        if not isinstance(impl, Apply):
            break
        if not hasattr(impl, "estimator_"):
            raise NotFittedError(
                "The skrub learner has not been fitted. Call fit() before inspecting "
                "fitted sub-estimators."
            )
        chain.append(impl.estimator_)
        if not isinstance(impl.X, DataOp):
            break
        node = impl.X
```

skore/src/skore/_utils/skrub.py
auguste-probabl: One compromise we found with @GaetandeCast for cases like this where the helper is only used in one check, is to put it in the check class itself as a static method -- i'll let you decide in the end.

glemaitre: It is too much related to skrub learner to be inside the specific class. Basically, it is a function to find the estimator within a skrub learner and I think that we can end up reusing it in several checks when it comes to hyperparameters. I would rather keep it here.

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: ok so it's a super specific thing. can you update the docstring to make that clearer?

skore/src/skore/_utils/_skrub.py
```diff
@@ -45,6 +49,89 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
...
+
+
+def collect_fitted_apply_estimators(data_op: DataOp) -> list[BaseEstimator]:
+    """Return fitted estimators along nested applies from preprocessing to predictor."""
+    chain: list[BaseEstimator] = []
+    node: Apply = _supervised_apply_node(data_op)
+    while True:
+        impl = node._skrub_impl
+        if not isinstance(impl, Apply):
+            break
```
jeromedockes: I think this will only work if the data op is a single unbroken chain of scikit-learn estimators so basically a scikit-learn pipeline. if we are using private APIs to find all the estimators in a data op you can do something like this

```python
import skrub
from skrub._data_ops._evaluation import _DataOpTraversal

data_op = skrub.X().skb.apply(skrub.TableVectorizer()).skb.apply(Ridge(), y=skrub.y())

X, y = make_regression()
learner = data_op.skb.make_learner().fit({'X': X, 'y': y})

# ==================================================================
# For inspecting estimators in the dataop before fitting, e.g. to check if all
# their parameters are the default ones:

class FindEstimators(_DataOpTraversal):
    def __init__(self, include_nested):
        self.include_nested = include_nested
        self.estimators = {}

    def handle_estimator(self, estimator):
        self.estimators[id(estimator)] = estimator
        if not self.include_nested:
            return estimator
        return (yield from super().handle_estimator(estimator))


def find_estimators(data_op, include_nested=False):
    finder = FindEstimators(include_nested=include_nested)
    finder.run(data_op)
    return list(finder.estimators.values())


print(find_estimators(data_op))
# [TableVectorizer(), Ridge()]
print(find_estimators(data_op, True))
# [TableVectorizer(), DatetimeEncoder(), StringEncoder(), OneHotEncoder(drop='if_binary', dtype='float32', handle_unknown='ignore', sparse_output=False), PassThrough(), Ridge()]


# ==================================================================
# For finding the fitted estimators in a learner:

def find_fitted_estimators(learner):
    fitted_estimators = {}
    def check_node(node):
        try:
            estimator = node._skrub_impl.estimator_
        except AttributeError:
            pass
        else:
            fitted_estimators[id(estimator)] = estimator
        return False
    learner.data_op.skb.find(check_node)
    return list(fitted_estimators.values())

print(find_fitted_estimators(learner))
# [Ridge(), TableVectorizer()]

# (actually this one should become doable with the public API we just need to
# support IDs in addition to names in find_fitted_estimator)

```

IIUC this is to look at each of them and see if its params are the scikit-learn default right?

later we can discuss if and how skrub should have a public API way of doing this kind of inspection

skore/src/skore/_utils/_skrub.py
```diff
@@ -45,6 +49,89 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
     return data_op.skb.find_X_y().get("cv") is not None
 
 
+def _supervised_apply_node(data_op: DataOp) -> Apply:
```
jeromedockes: there can be several predictors in a pipeline. for example have one histgradientboosting with different alpha and pinball loss for different quantiles

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -175,14 +180,21 @@ def split_preprocessor_estimator(estimator):
...
+    fitted = (
+        _supervised_fitted_estimator(estimator)
+        if is_skrub_learner(estimator)
+        else estimator
+    )
+
+    if isinstance(fitted, Pipeline):
+        if len(fitted.steps) > 1:
+            return fitted[:-1], fitted[-1]
+        return None, fitted.steps[0][1]
```
jeromedockes: but the actual preprocessing will not be done in a pipeline in the final node but rather in the rest of the graph no?

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -277,7 +289,30 @@ def get_preprocessed_X(
         else:
             data = report.X_test
 
-    preprocessor, _ = split_preprocessor_estimator(get_fitted_estimator(report))
+    estimator = get_fitted_estimator(report)
+
+    if report._initialized_with_data_op:
+        learner = estimator
+        if report._report_type == "cross-validation":
+            skrub_env = {"_skrub_X": data, "_skrub_y": get_report_y(report)}
```
jeromedockes: we are missing the other items than X and y from the env?

jeromedockes: also note the y that goes into an estimator is not necessarily the same as the y of the skore report (for example I could do some filtering of samples before fitting)

jeromedockes (commented):
thanks @glemaitre 

I put 2 ways of finding the scikit-learn estimators contained in a learner

for the other checks, I'm not sure the separation of "preprocessor" vs "predictor" really works or that it always makes sense, it may also be that what it means should be decided on a check-by-check basis. maybe we should discuss offline

for some checks I am not sure we should be looking at the "preprocessed" data anyway for example we can have a "golden column" but once it is vectorized none of the dimensions on its own is enough

also note ATM feature importances are not implemented for dataops so checks related to that will only be applicable when they are

skore/src/skore/_utils/_skrub.py
jeromedockes: for the second part (looking for fitted estimators) it could be done with public API only once

https://github.com/skrub-data/skrub/pull/2194

is merged

although in any case the private attribute it uses is stable so maybe this is simpler anyway

jeromedockes:
so IIUC this PR tackles 2 kinds of checks

- (i) those that need the "preprocessed X" + the predictor, to inspect them. eg check the variance of preprocessed X + whether predictor has a coef_ attribute

what you have is close but I think it can be simplified a bit and to evaluate the X node we must pass a full environment not just X and y. so

```python
import skrub
from skrub._data_ops._evaluation import find_first_apply


def get_predictor_and_compute_input(learner, env):
    # !! assumes there is a single predictor in the pipeline
    #
    # !! env must be an environment provided by the user ie
    #   CrossValidationReport._data or EstimatorReport._train_data or
    #   EstimatorReport._test_data . We cannot "concatenate" train and test env

    predictor_node = find_first_apply(learner.data_op)
    predictor = predictor_node._skrub_impl.estimator_
    predictor_input_node = predictor_node._skrub_impl.X
    input_value = learner.truncated_after(predictor_input_node.skb.id).transform(env)
    return input_value, predictor

# - we can look at input_value to check correlations, high cardinality etc.
# - we can inspect predictor to see if it has coef_ and feature_importance_
# - with what this simple function provides we don't have enough to compute
#   predictions or skores, none of the checks seem to require that ATM
```

- (ii) those that inspect the estimator parameters to check their values. for now just checking if they are different from the default, but actually it will be similar to check if they are being tuned, because in dataops it just means the value of that parameter in the unfitted estimator contains a skrub choice


```python
import skrub
from skrub._data_ops._evaluation import _DataOpTraversal


class FindEstimators(_DataOpTraversal):
    def __init__(self, include_nested):
        self.include_nested = include_nested
        self.estimators = {}

    def handle_estimator(self, estimator):
        self.estimators[id(estimator)] = estimator
        if not self.include_nested:
            return estimator
        return (yield from super().handle_estimator(estimator))


def find_estimators(data_op, include_nested=True):
    finder = FindEstimators(include_nested=include_nested)
    finder.run(data_op)
    return list(finder.estimators.values())


def is_tunable(obj):
    data_op = skrub.as_data_op(obj)
    found = data_op.skb.find(
        lambda o: not isinstance(o, skrub.DataOp)  # exclude dataops, keep choices
        #
        # or, a bit less strict, we consider user is doing something
        # special also if it depends on a dataop not only choices:
        # lambda o: o is not param_data_op
    )
    return found is not None
    # or simpler but less robust alternative:
    # return skrub.as_data_op(obj).skb.describe_param_grid() == "<empty parameter grid>\n"


def check_hyperparams(data_op):
    for e in find_estimators(data_op):

        # to check if different than default:
        check_different_than_default(e, _changed_params(e))

        # to check if being tuned:
        tunable_params = [
            param_name
            # filter to avoid calling is_tunable on params we don't care about
            # as it is not free
            for param_name in params_to_consider(type(e))
            if is_tunable(getattr(e, param_name))
        ]
        check_tunable(e, tunable_params)
```

finally, another check we could add is if the learner is not a search (randomized or grid or optuna) and `is_tunable(learner.data_op)`, mention in a tip that the user should pass a search object unless they are doing a test run on the default pipeline

thomass-dev:
/e2e (don't worry about that; i'm running a test)

skore/src/skore/_utils/_skrub.py
auguste-probabl: The suggestion has been applied. Even though the skrub PR has been merged, it doesn't simplify the code by a lot because `find_fitted_estimator` only returns the first matching node, not all of them. IMO this is sufficient, let's open an issue later if we find a better way to do it.

skore/src/skore/_utils/skrub.py
```diff
@@ -45,6 +49,151 @@ def data_op_has_explicit_cv(data_op: DataOp) -> bool:
...
+    _ensure_single_supervised_apply(data_op)
+    return apply_node
+
+
+def _ensure_single_supervised_apply(data_op: DataOp) -> None:
+    supervised_applies: list[DataOp] = []
+
+    def check_node(node: DataOp) -> bool:
+        impl = getattr(node, "_skrub_impl", None)
+        if isinstance(impl, Apply) and impl.y is not None:
```
jeromedockes, line 86: some preprocessors such as feature selectors or targetencoder also accept y

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: first point is solved, not sure yet about the second

skore/src/skore/_utils/skrub.py
auguste-probabl: is there any solution to this?

jeromedockes: we could check on the fitted learner the type of estimator (whether it has predict() or it is a transformer)

but basically the issue we are having here and in other places is that the assumption that an estimator must be a linear sequence of transformers followed by one final predictor is the sklearn.pipeline vision, which is too narrow for many data-science pipelines

however, as long as the only consequence is that one of the checks is skipped but this does not otherwise prevent the user from using skore (besides that check), we can do whatever is simple and figure out later if there are some cases where we can "rescue" this check

auguste-probabl: added an extra condition to differentiate transformers from predictors

skore/src/skore/_utils/_skrub.py
auguste-probabl: for now, we raise when this happens

auguste-probabl:
`tests/unit/reports/cross_validation/test_checks.py::test_skd010_detects_slower_than_baseline` is flaky; this is fixed in https://github.com/probabl-ai/skore/pull/3176

skore/src/skore/_checks/utils.py
auguste-probabl, line 1: > finally, another check we could add is if the learner is not a search (randomized or grid or optuna) and is_tunable(learner.data_op), mention in a tip that the user should pass a search object unless they are doing a test run on the default pipeline

Todo: write an issue

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -200,12 +205,19 @@ def split_preprocessor_estimator(estimator):
...
     """
+    if is_skrub_learner(estimator):
+        return None, resolve_fitted_predictor(estimator)
+
     if isinstance(estimator, Pipeline):
         if len(estimator.steps) > 1:
             return estimator[:-1], estimator[-1]
-        else:
-            return None, estimator[0]
+        return None, estimator.steps[0][1]
```
GaetandeCast: Out of curiosity, what does `estimator.steps[0][1]` change compared to `estimator[0]`?

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -299,18 +311,55 @@ def get_preprocessed_X(
...
+                    env = report.train_data
+
+                    data, _ = get_predictor_and_input(learner, env)
+                elif data_source == "test":
+                    if report.test_data is None:
+                        raise CheckNotApplicable("Test data is unavailable.")
+                    env = report.test_data
+
+                    data, _ = get_predictor_and_input(learner, env)
+                else:  # data_source == "test"
```
GaetandeCast:
```suggestion
                else:  # data_source == "both"
```

GaetandeCast (commented):
Only looked at the checks part which looks good to me. I expect us to merge #3176 first so I'll look at tests after it's merged.

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: sklearn overrides `__getitem__` so that `estimator[i]` does `estimator.steps[i][1]`. simplified

GaetandeCast (dismissed):
Looked at tests, they LGTM, I'll let @glemaitre have an other look then we can merge.

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: I didn't get this comment sorry

skore/src/skore/_checks/skd016_estimator_not_tuned.py
```diff
@@ -13,15 +13,43 @@
...
+            [step for _, step in unfitted.steps]
+            if isinstance(unfitted, Pipeline)
+            else [unfitted]
+        )
+        for est in estimators:
+            class_name = type(est).__name__
+            if class_name not in HYPERPARAMETERS_TO_TUNE:
+                continue
+            for param_name in HYPERPARAMETERS_TO_TUNE[class_name]:
+                if param_name in INFRASTRUCTURE_PARAMS:
```
jeromedockes, line 46: I don't think any infrastructure params are considered as hyperparameters to tune right?

```python
import itertools
from skore._checks.tunable_hyperparameters import (HYPERPARAMETERS_TO_TUNE, INFRASTRUCTURE_PARAMS)

print(set(itertools.chain(*HYPERPARAMETERS_TO_TUNE.values())).intersection(INFRASTRUCTURE_PARAMS))
```

gives empty set

skore/src/skore/_checks/skd016_estimator_not_tuned.py
```diff
@@ -47,24 +75,24 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+
+        skrub_tuned_classes = (
+            skrub_classes_with_tunable_recommended_params(estimator)
+            if is_skrub_learner(estimator)
+            else set()
+        )
 
         messages: list[str] = []
         for class_name, step in candidates:
+            if class_name in skrub_tuned_classes:
```
jeromedockes, line 94: I guess in theory we could fail to warn eg for a pipeline with 2 PCA where only one is tuned but it's probably not a big deal

jeromedockes (commented):
wip

skore/src/skore/_checks/skd016_estimator_not_tuned.py
jeromedockes: but it doesn't do any harm to check it anyway

jeromedockes (approved):
LGTM, thanks a lot @auguste-probabl and @glemaitre

PR #3087, opened 2026-06-23
feat: Remove multi-dataset X from evaluate

### LINKED ISSUE #3077: remove the possibility of passing several datasets to evaluate?
evaluate can be passed a list of estimators to compare several estimators as a shorthand for doing first their evaluation then a comparison

moreover, it can be passed a different dataset for each of those, so that they each use different "preprocessing"
I would suggest removing that option, as the processing must be part of the estimator. otherwise in the resulting reports there is no way to know what produced these different datasets or how the pipelines differ. moreover it complicates the interface and the code of that function

for users who really want to do this they can always map evaluate + compare which is not much harder (possibly easier) than doing it all at once by passing lists to evaluate

moujanrastgoo:
Closes #3077

#### Change description

This PR removes the option to pass a **list** or **dict** as `X` in `skore.evaluate()`.

Before, you could do something like `evaluate([model1, model2], [X1, X2], y)` to give each model its own dataset. I am removing that because preprocessing should live in the model/pipeline, and it’s clearer to evaluate separately and then compare.

**What still works:**
- `evaluate(model, X, y)` 
- `evaluate([m1, m2], X, y)`

**What to do instead if you had different X per model:**
```python
r1 = evaluate(model, X_raw, y, splitter=0.2)
r2 = evaluate(model, X_engineered, y, splitter=0.2)
compare({"raw": r1, "engineered": r2})
```

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:

I used AI to understand the code and find places where changes were necessary, and to check my code before testing.

### DIFF
--- a/examples/technical_details/plot_skore_api.py
+++ b/examples/technical_details/plot_skore_api.py
@@ -149,7 +149,7 @@ comparison_report = evaluate(
     y,
     splitter=0.2,
 )
-comparison_report.help()
+comparison_report
 
 # %%
 comparison_report.metrics.summarize().frame()
--- a/examples/technical_details/plot_skore_api.py
+++ b/examples/technical_details/plot_skore_api.py
@@ -152,7 +152,7 @@ comparison_report = evaluate(
 comparison_report.help()
 
 # %%
-_ = comparison_report.metrics.summarize().plot()
+comparison_report.metrics.summarize().frame()
 
 # %%
 # To compare evaluations on different feature matrices, call
@@ -168,7 +168,7 @@ comparison_report = compare([report_logistic, report_forest])
 comparison_report.help()
 
 # %%
-_ = comparison_report.metrics.summarize().plot()
+comparison_report.metrics.summarize().frame()
 
 # %%
 # Summary
--- a/examples/technical_details/plot_skore_api.py
+++ b/examples/technical_details/plot_skore_api.py
@@ -45,8 +45,9 @@ produce a visualization return a **Display** object with ``plot()``, ``frame()``
 # :class:`~skore.EstimatorReport`. The accessors and display API shown below
 # are the same for the other report types.
 from sklearn.datasets import load_breast_cancer
+from sklearn.ensemble import RandomForestClassifier
 from sklearn.linear_model import LogisticRegression
-from skore import evaluate
+from skore import compare, evaluate
 from skrub import tabular_pipeline
 
 X, y = load_breast_cancer(return_X_y=True, as_frame=True)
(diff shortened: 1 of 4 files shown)


examples/technical_details/plot_skore_api.py
```diff
@@ -131,10 +131,10 @@
...
 # =============================
 #
-# Passing a **list or dict of estimators** to :func:`~skore.evaluate` returns a
-# :class:`~skore.ComparisonReport`. It exposes the same ``metrics`` and
-# ``inspection`` accessors (no ``data`` accessor, since compared models can
-# use different datasets). The display API is unchanged.
+# ``inspection`` accessors (no ``data`` accessor on
+# :class:`~skore.ComparisonReport`). To compare evaluations on different
+# feature matrices, use :func:`~skore.evaluate` then :func:`~skore.compare`.
+# The display API is unchanged.
```
glemaitre: It is actually weird that we don't show the report here. I think that we should implement it.

Here, we need to improve a bit the explanation because one should be able to do:

```python
comparison_report = skore.evaluate([Ridge(), Lasso()], X, y)
```

It is just that the two models will be provided on the same input `X`.

However, if `X` is not the same, we will need:

```python
report_ridge = skore.evaluate(Ridge(), X_1, y)
report_lasso = skore.evaluate(Lasso(), X_2, y)
comparison_report = skore.compare(report_ridge, report_lasso)
```

So I think that it is worth to show this two pattern here how to get a comparison report.

glemaitre (commented):
It looks almost good. Could we make sure that we have a test for the case with multiple estimator and a single X to be sure that it returns a comparison report.

examples/technical_details/plot_skore_api.py
moujanrastgoo: I checked, I think there is  already a test that assures that  multiple estimators with a single X will return a comparison report. Here it is: 

```python 
def test_multiple_estimators(regression_data):
    """A list of estimators returns a ComparisonReport."""
    X, y = regression_data
    estimators = [LinearRegression(), LinearRegression()]
    report = evaluate(estimators, X, y, splitter=0.2)
    assert isinstance(report, ComparisonReport) 
```
Would you like to change it in any way? 
As for the example, I will get started on it right away!

glemaitre: Perfect. So no test needed.

skore/tests/unit/dispatch/test_evaluate.py
```diff
@@ -83,23 +83,18 @@ def test_multiple_estimators_cv(regression_data):
...
     X, y = regression_data
-    with pytest.raises(TypeError, match="cannot be a dict"):
-        evaluate(
-            [LinearRegression(), LinearRegression()],
-            {"a": X, "b": X},
-            y,
-            splitter=0.2,
-        )
+    with pytest.raises(TypeError, match="single array-like or None"):
+        evaluate(LinearRegression(), {"a": X}, y)
```
GaetandeCast:
```suggestion
def test_list_or_dict_X_raises(regression_data):
    """Passing list X raises TypeError."""
X, y = regression_data
    with pytest.raises(TypeError, match="single array-like or None"):
        evaluate(LinearRegression(), [X], y)
        evaluate(LinearRegression(), {"a": X}, y)
```

We can test both at the same time to avoid the (I admit, minimal) overhead of reloading the `regression_data` fixture.

GaetandeCast: Sorry, the github rendering of the suggestion is hard to read. Basically I suggest collapsing `test_list_X_raises` and `test_dict_X_raises`.

moujanrastgoo: ok sure i see!

examples/technical_details/plot_skore_api.py
```diff
@@ -131,10 +132,43 @@
...
+comparison_report = evaluate(
+    [
+        tabular_pipeline(LogisticRegression()),
+        tabular_pipeline(RandomForestClassifier()),
+    ],
+    X,
+    y,
+    splitter=0.2,
+)
+comparison_report.help()
```
glemaitre:
```suggestion
comparison_report
```

glemaitre (approved):

PR #3100, opened 2026-06-26
feat!: Stop deep-copying reports every time

### LINKED ISSUE #3075: EstimatorReport: Avoid deep-copy when possible
Extracted from https://github.com/probabl-ai/skore/pull/3046#discussion_r3441180492

Currently we deep-copy the estimator in `EstimatorReport.__init__` to avoid surprising the user if they keep using the estimator after passing it to skore. This may be a bit overkill if the estimator is already fitted:

> For most estimators making predictions should not modify their state, and if a user wants to be 100% sure it is not modified for whatever reason they can always deep-copy themselves. some estimators are very large objects so avoiding a deep copy could be beneficial

If we make this change for fitted estimators, I guess we would also stop cloning unfitted estimators.

auguste-probabl:
Closes https://github.com/probabl-ai/skore/issues/3075

Performance test:

```py
import statistics
from sklearn.ensemble import RandomForestRegressor
from time import perf_counter

from skore import evaluate
import sklearn

X, y = sklearn.datasets.make_classification(n_samples=10_000, return_X_y=True)

model = RandomForestRegressor()
model.fit(X, y)

n = 5
results = []
for _ in range(n):
    start = perf_counter()
    evaluate(model, X, y)
    end = perf_counter()
    results.append(end - start)

mean = statistics.mean(results)
stdev = statistics.stdev(results, xbar=mean)

print(f"{mean:.2f} +/- {stdev:.2f} ({n} runs)")
```

Results:

- without this PR: 10.41 +/- 0.07 seconds (5 runs)
- with this PR: 6.95 +/- 0.30 (5 runs)

Note: this test uses a RandomForest estimator which is the kind of case where deep-copying is costly (the estimator contains 100 decision trees, each containing ~1000 nodes/floats)

### DIFF
--- a/skore/src/skore/_sklearn/_estimator/report.py
+++ b/skore/src/skore/_sklearn/_estimator/report.py
@@ -1,9 +1,7 @@
 from __future__ import annotations
 
-import copy
 import html
 import uuid
-import warnings
 from dataclasses import asdict
 from functools import cached_property
 from typing import TYPE_CHECKING, Any, Literal, NotRequired, TypedDict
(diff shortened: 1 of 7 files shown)


thomass-dev:
/e2e (don't worry about that; i'm running a test)

glemaitre (approved):
Thanks @auguste-probabl. LGTM.

PR #3105, opened 2026-06-30
feat(checks): Compute fast checks at report init

### LINKED ISSUE #3059: Compute fast checks at init time
Compute checks in `fast_mode` at the init of the report, to populate the check cache with the fast checks and have them for free later.

GaetandeCast:
Closes #3059

We clear the checks cache when a custom metrics is added/removed since a lot of the checks should take custom metrics into account.


GaetandeCast:
Let's improve the way we define fast checks first (#3267)

PR #3106, opened 2026-06-30
fix(cache)!: Add scope to cache key

### LINKED ISSUE #3096: make cache key function names more specific
the second item of the cache key  is informally "the function name" but nothing is enforced. it would help if they were all at least prefixed with the accessor name (if any) to avoid name clashes and make it easier to inspect

auguste-probabl:
Closes #3096


GaetandeCast (approved):
LGTM, thanks !

PR #3108, opened 2026-07-01
chore(plugin/hub/uri): Add quick note about `SKORE_HUB_URI`

thomass-dev:


jeromedockes (approved):

jeromedockes:
thank you very much!

skore/src/skore/_plugins/hub/authentication/uri.py
```diff
@@ -1,4 +1,14 @@
-"""URI used for ``skore hub`` authentication."""
+"""Hub backend URI used for ``skore hub`` authentication.
```
auguste-probabl, line 1: I feel like this should be in the user docs too (i have been confused by it before)

thomass-dev: It is for internal use only. No lambda user or external contributor can start their own hub instance.

So i don't think it should be exposed in the user documentation.
The hub documentation should be a better place.

jeromedockes:
> It is for internal use only. No lambda user can start their own hub instance.
>
> So i don't think it should be exposed in the user documentation.
> The hub documentation should be a better place.

we can always revisit if one day some users have their own on-premise instance of the hub

PR #3109, opened 2026-07-01
test(skore): Enforce warnings-as-errors in tests

auguste-probabl:
Supersedes #2082 

#### AI usage disclosure

AI tools were involved for:

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -139,7 +139,24 @@ addopts = [
 xfail_strict = true
 log_cli_level = "info"
 filterwarnings = [
-  'ignore:codecs\.open\(\) is deprecated.*:DeprecationWarning:mlflow',
+  "error",
+  # mlflow-related
+  # NOTE: Only affects python 3.11
+  "ignore:SelectableGroups dict interface is deprecated. Use select.:DeprecationWarning",
+  # sklearn needs to update its scipy.optimize calls
+  'ignore:.*`disp` and `iprint` options of the L-BFGS-B solver are deprecated:DeprecationWarning',
+  # seaborn's boxplot still passes the deprecated `vert=` kwarg to matplotlib.
+  # Hit whenever any plot Display renders a boxplot, so keep at the top level.
+  'ignore:vert.*was deprecated in Matplotlib:matplotlib._api.deprecation.MatplotlibDeprecationWarning:seaborn',
+  # polars deprecates the dataframe interchange protocol; sklearn and plotly
+  # still call into it when handed a polars frame. Triggered across many
+  # container-parametrized tests spread over the suite.
+  'ignore:Support for the dataframe interchange protocol is deprecated:DeprecationWarning:(sklearn|plotly)',
+  # `TemporaryDirectory` finalizers may fire at arbitrary GC points during
+  # doctest teardown, blaming whichever test happens to run at that moment,
+  # so these have to stay global.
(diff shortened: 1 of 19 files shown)


skore/pyproject.toml
```diff
@@ -139,7 +139,25 @@ addopts = [
...
+  # call into it when handed a polars frame.
+  'ignore:Support for the dataframe interchange protocol is deprecated:DeprecationWarning',
+  # `TemporaryDirectory` finalizers may fire at arbitrary GC points during
+  # doctest teardown, blaming whichever test happens to run at that moment.
+  'ignore:Implicitly cleaning up:ResourceWarning',
+  'ignore::pytest.PytestUnraisableExceptionWarning',
+  # Doctests accumulate figures across the suite; the `figure.max_open_warning`
+  # rcParam is set to 0 in `tests/conftest.py` but that conftest is not loaded
+  # when running doctests inside `src/`.
+  'ignore:More than \d+ figures have been opened:RuntimeWarning',
```
thomass-dev: IMO, it is not a good idea to swallow warnings at the top level. It should be done case-by-case, in the tests.
Even if the messages look specific per use-case, we have no guarantee that several use-cases don't share the same message (and so we swallow the wrong warnings).
Also, the origin of the warning should be specified: `ignore:<message>:<category>:<package>`.

auguste-probabl: The last 3 warnings could happen at any point during the tests run so they have to stay; I'll see what I can do about the others. For example the seaborn one fires pretty much every time a plot Display is involved, so that would add *a lot* of cruft.

My hope here is that we can deal with each warning in followup PRs, so that warnings-as-errors is finally enabled.

> `ignore:<message>:<category>:<package>`

Thanks, will do!

skore/pyproject.toml
```diff
@@ -139,7 +139,24 @@ addopts = [
 xfail_strict = true
 log_cli_level = "info"
 filterwarnings = [
-  'ignore:codecs\.open\(\) is deprecated.*:DeprecationWarning:mlflow',
+  "error",
+  # mlflow-related
+  # NOTE: Only affects python 3.11
+  "ignore:SelectableGroups dict interface is deprecated. Use select.:DeprecationWarning",
```
auguste-probabl, line 145: occurs during collection so it has to stay here

skore/pyproject.toml
```diff
@@ -139,7 +139,24 @@ addopts = [
...
+  # sklearn needs to update its scipy.optimize calls
+  'ignore:.*`disp` and `iprint` options of the L-BFGS-B solver are deprecated:DeprecationWarning',
```
auguste-probabl, line 147: occurs during collection if i remember correctly

skore/pyproject.toml
auguste-probabl: ok, I think this will be it for this PR. Again, my goal is that we finally enable warnings-as-errors for all future dev right now, and later we will work on progressively removing the ignores.

thomass-dev (changes requested):
Please remove `test/unit/plugins/conftest.py`

```python
import pytest                                   
                                                
pytestmark = pytest.mark.filterwarnings("error")
```

warnings-as-errors was already enabled for plugins 😉 .

thomass-dev (dismissed):

thomass-dev (approved):

PR #3110, opened 2026-07-01
fix(skore): Exclude score from metrics for pipeline with default score

jeromedockes:


GaetandeCast (approved):
Thanks @jeromedockes !

PR #3120, opened 2026-07-03
fix(plugin/hub/cross_validation_report/target_range): Compute only on regression

### LINKED ISSUE #3117: put on hub error when y is a dataframe
now put() on the hub work for multioutput regression, but there is an error if y is stored in a dataframe

```python
import os

import dotenv
import skore
import numpy as np
import pandas as pd

dotenv.load_dotenv()

rng = np.random.default_rng()
X = rng.normal(size=(100, 10))
y = pd.DataFrame(rng.normal(size=(100, 2)))

report = skore.evaluate(DummyRegressor(), X=X, y=y, splitter=3)
skore.login()
project = skore.Project(
    workspace=os.environ["SKORE_HUB_WORKSPACE"], name="project-100", mode="hub"
)
project.put("report", report)
```

```
╭───────────────────────────────── Login to Skore Hub ─────────────────────────────────╮
│                                                                                      │
│                        Successfully logged in, using API key.                        │
│                                                                                      │
⠦ Putting report 0:00:00
  File "/home/jerome/workspace/scratch/2026/06/multioutput_report.py", line 18, in <module>
    project.put("report", report)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/jerome/workspace/repos/skore/skore/src/skore/_project/project.py", line 301, in put
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jerome/workspace/repos/skore/skore/src/skore/_plugins/hub/project/project.py", line 293, in put
    payload_dict = payload.model_dump()
  File "/home/jerome/miniforge3/envs/skore/lib/python3.13/site-packages/pydantic/main.py", line 475, in model_dump
    return self.__pydantic_serializer__.to_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        self,
        ^^^^^
    ...<13 lines>...
        polymorphic_serialization=polymorphic_serialization,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/home/jerome/miniforge3/envs/skore/lib/python3.13/functools.py", line 1025, in __get__
    val = self.func(instance)
  File "/home/jerome/workspace/repos/skore/skore/src/skore/_plugins/hub/report/cross_validation_report.py", line 344, in target_range
    return [float(target.min()), float(target.max())]
            ~~~~~^^^^^^^^^^^^^^
  File "/home/jerome/miniforge3/envs/skore/lib/python3.13/site-packages/pandas/core/series.py", line 251, in wrapper
    raise TypeError(f"cannot convert the series to {converter}")
TypeError: cannot convert the series to <class 'float'>
```

(the error is on the same line but slightly different for polars dataframes)

thomass-dev:
Closes https://github.com/probabl-ai/skore/issues/3117.

### DIFF
--- a/skore/src/skore/_plugins/hub/report/cross_validation_report.py
+++ b/skore/src/skore/_plugins/hub/report/cross_validation_report.py
@@ -132,6 +132,7 @@ class CrossValidationReportPayload(ReportPayload[CrossValidationReport]):
         self.__classes: list[str] | None
         self.__target_names: list[str] | None
         self.__target_ranges: list[list[float]] | None
+        self.__target_range: list[float] | None
 
         if "classification" in self.ml_task and (self.report.y is not None):
             class_to_class_indice: defaultdict[Any, int] = defaultdict(
@@ -148,24 +149,35 @@ class CrossValidationReportPayload(ReportPayload[CrossValidationReport]):
             self.__classes = [str(class_) for class_ in class_to_class_indice]
 
             assert max(self.__sample_to_class_index) == (len(self.__classes) - 1)
+
             self.__target_names = None
             self.__target_ranges = None
+            self.__target_range = None
         elif self.ml_task == "multioutput-regression" and (self.report.y is not None):
             from skore._utils._dataframe import _normalize_y_as_dataframe
 
             y_df = _normalize_y_as_dataframe(self.report.y)
+            y_arr = np.asarray(self.report.y)
+
             self.__sample_to_class_index = None
             self.__classes = None
             self.__target_names = [str(name) for name in y_df.columns]
+            self.__target_ranges = [[float(c.min()), float(c.max())] for c in y_arr.T]
+            self.__target_range = None
+        elif self.ml_task == "regression" and (self.report.y is not None):
             y_arr = np.asarray(self.report.y)
-            self.__target_ranges = [
-                [float(col.min()), float(col.max())] for col in y_arr.T
-            ]
+
+            self.__sample_to_class_index = None
+            self.__classes = None
+            self.__target_names = None
+            self.__target_ranges = None
+            self.__target_range = [float(y_arr.min()), float(y_arr.max())]
         else:
             self.__sample_to_class_index = None
             self.__classes = None
             self.__target_names = None
             self.__target_ranges = None
+            self.__target_range = None
 
     @computed_field  # type: ignore[prop-decorator]
     @cached_property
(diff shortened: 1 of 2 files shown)


skore/src/skore/_plugins/hub/report/cross_validation_report.py
```diff
@@ -148,24 +149,35 @@ def model_post_init(self, _: Any) -> None:  # noqa: D102
...
             assert max(self.__sample_to_class_index) == (len(self.__classes) - 1)
+
             self.__target_names = None
             self.__target_ranges = None
+            self.__target_range = None
         elif self.ml_task == "multioutput-regression" and (self.report.y is not None):
             from skore._utils._dataframe import _normalize_y_as_dataframe
 
             y_df = _normalize_y_as_dataframe(self.report.y)
+            y_arr = np.asarray(self.report.y)
```
jeromedockes: maybe we don't need to turn into an array

```python
import narwhals as nw

y_df_nw = nw.from_native(y_df)
self.__target_ranges = [[y_df_nw[c].min(), y_df_nw[c].max()] for c in y_df_nw.columns]
```

or something like that?

thomass-dev: I can do even better:

```python
self.__target_ranges = list(zip(df.min(), df.max()))
```

The question is : should i use directly nw over the wrapper `_normalize_y_as_dataframe` ?

jeromedockes: I think the list(df.min()) above would work on pandas dataframes, but on a polars dataframe min() returns a dataframe and on a narwhals dataframe it does not seem to exist

jeromedockes: an I would still use normalize_y_as_dataframe because it will handle the case where it is a numpy array, ensure pandas column names are strings, and raise an error for sparse matrices (btw in that case I think we shoud show the repr of the sparse matrix instead but not raise an error, but that is a niche use case so let's ignore it for now)

thomass-dev: Fixed in 129d825.

jeromedockes:
thanks @thomass-dev I checked it works with polars and pandas dataframes

jeromedockes (approved):

PR #3128, opened 2026-07-06
refactor: Auto-close figures after every test

auguste-probabl:
Add top-level `conftest.py` which also applies to doctests in `src`.

Remove redundant `pyplot` fixture.

### DIFF
--- a/skore/tests/conftest.py
+++ b/skore/tests/conftest.py
@@ -1,6 +1,5 @@
 from datetime import UTC, datetime
 
-import matplotlib
 import numpy as np
 import pytest
 from sklearn.base import BaseEstimator, ClassifierMixin, clone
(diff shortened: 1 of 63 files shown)


skore/tests/conftest.py
thomass-dev, line 1: If pytest only reads the conftest file from a parent, this means that this part does not work for doctest either:

```python
def pytest_configure(config):                                                
    """Set up global test configuration.                                     
                                                                             
    Some of these could be set in fixtures, but doctests do not run fixtures.
    """                                                                      
    matplotlib.use("agg")                                                    
                                                                             
    # Disable progress bars during tests to avoid rich interfering with      
    # doctest stdout capture.                                                
    configuration.show_progress = False                                      
```

I'm a bit surprised that was not detected before.

thomass-dev (changes requested):

skore/tests/conftest.py
auguste-probabl: Confirmed by putting a breakpoint in a doctest, without the changes in this PR:
```
>>> import skore; skore.configuration.show_progress
True
```

With these changes, it is correctly set to False.

thomass-dev: If so, could you please move this part in the root conftest?

auguste-probabl: Moved, thanks!

thomass-dev (approved):

PR #3133, opened 2026-07-08
chore(docs): Remove GTM

thomass-dev:
Following https://github.com/probabl-ai/skore-hub/pull/1643, addresses https://github.com/probabl-ai/skore-hub/issues/1648.


rouk1:
Do we really want to remove the only user insight we have ?

Issue #2622, opened 2026-03-19
Auto diagnosis: Create a curated list of examples


GaetandeCast:
### What would you like to say?

Create and curate a set of examples to refer the user to, detailing each pitfall in a practical scenario, showing why and how it can be problematic and how to fix/avoid it. It could then also act as a reference for an AI agent trained to fix and/or build proper ML pipelines.
This is a long term project and not the top priority for the automatic diagnosis initiative.

PR #3134, opened 2026-07-08
docs: Add pitfalls and solutions gallery with SKD001 and SKD002 example

moujanrastgoo:
#### Change description
Related to #2622

- Adds new sphinx-gallery subsection **Checking for modeling pitfalls and remediation**
- Adds SKD001 overfitting and SKD002 underfitting example and links from automated checks user guide

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:


skore/pyproject.toml
```diff
@@ -80,7 +80,7 @@ sphinx-base = [
   "xgboost ; platform_system == 'Darwin'",
   "xgboost-cpu ; platform_system != 'Darwin'",
 ]
-sphinx = ["mlflow", "polars", "skore[sphinx-base, hub, mlflow]"]
+sphinx = ["mlflow", "openml", "polars", "skore[sphinx-base, hub, mlflow]"]
```
thomass-dev: Please put `openml` in `sphinx-base`.

moujanrastgoo: Sure!

moujanrastgoo: Done, its now in sphinx-base.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+"""
+
+# %%
+# Load the electricity dataset
+# ============================
+#
+# Rows are time-ordered. ``shuffle=False`` keeps the hold-out as future data,
+# a realistic split for this table.
+
+import openml
```
GaetandeCast: We have had issues with openml in the past: it is often down which blocks the building of the docs which in turns blocks the merging of PRs. So I'd rather we use other sources of datasets.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+- improve feature engineering,
+- use better validation protocols or more data.
+"""
+
+# %%
+# Load the electricity dataset
+# ============================
+#
+# Rows are time-ordered. ``shuffle=False`` keeps the hold-out as future data,
+# a realistic split for this table.
```
GaetandeCast: General note: the style of writing is very concise here. While I like it, it would be better to align this example with the style of writing of the rest of the docs, which is heavily inspired by the scikit-learn docs. This means more details in an effort to be educational.

For example, I would start the example with an introductory phrase such as:

This example will go through the different ways to mitigate overfitting when it is detected by check `SDK001`. Here is a list of steps that can be taken to mitigate the risk.

- regularize more strongly,
- simplify the model,
- improve feature engineering,
- use better validation protocols or more data.

We will show how to apply these suggestions on the [...] dataset.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+
+import openml
+import pandas as pd
+from skore import TrainTestSplit, evaluate
+from skrub import TableReport, tabular_pipeline
+
+RANDOM_STATE = 42
+POS_LABEL = "UP"
```
GaetandeCast: I would personally not use such constants in a docs example and just reuse the same value everywhere.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+# ============================
+#
+# Rows are time-ordered. ``shuffle=False`` keeps the hold-out as future data,
+# a realistic split for this table.
+
+import openml
+import pandas as pd
+from skore import TrainTestSplit, evaluate
+from skrub import TableReport, tabular_pipeline
+
```
GaetandeCast: imports are a little over the place. Import what you need as late as possible: at the top of the first cell that uses an imported object.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+# %%
+# Trigger SKD001 — untuned gradient boosting
+# ==========================================
+#
+# ``tabular_pipeline("classifier")`` has enough capacity to overfit on a single
+# hold-out. Compare train vs test in the metrics table, then run checks.
+
+from sklearn.ensemble import HistGradientBoostingClassifier
+
+estimator = tabular_pipeline(HistGradientBoostingClassifier(random_state=RANDOM_STATE))
```
GaetandeCast: You mention `tabular_pipeline("classifier")` and then use `tabular_pipeline(HistGradientBoostingClassifier())`. You need to be consistent. Under the hood it is exactly the same but you can't expect the reader to know that.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+    y=y,
+    pos_label=POS_LABEL,
+    splitter=splitter,
+)
+
+report_regularized.metrics.summarize(data_source="both").frame(favorability=True)
+
+# %%
+report_regularized.checks.summarize()
+# SKD001 often clears once the train/test gap narrows.
```
GaetandeCast:
```suggestion
# SKD001 clears once the train/test gap narrows.
```

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+
+# %%
+# Simplify the model
+# ==================
+#
+# A linear model has less capacity to memorize training quirks.
+
+from sklearn.linear_model import LogisticRegression
```
GaetandeCast: Actually, I think that it is not good advice to recommend using a simpler model to reduce overfitting as it is very likely to result in a worse model. It can be good for other reasons (interpretability, computational efficiency, ..) but when we want to reduce overfitting it's to improve model performance not worsen it. Let's remove this section and remove the bullet point from the user guide.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+# %%
+report_simple.checks.summarize()
+# The gap often shrinks; SKD001 may still fire depending on the split.
+
+# %%
+# Use more training data
+# ======================
+#
+# A smaller hold-out leaves more rows for fitting — sometimes enough to reduce
+# overfitting without changing the estimator.
```
GaetandeCast: Here the argument is that collecting more data usually makes models generalize better. To show it, let's initially fix a test set, and train two models on two subsets of the training set of different size, and show that the model with more training data performs better. It is important to use the same test set to be able to compare the results.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+# own.
+
+# %%
+# Improve feature engineering
+# ===========================
+#
+# A per-row ``row_id`` is pure noise but lets the tree memorize rows (train
+# accuracy inflates). We add it first to worsen overfitting, then drop it along
+# with ``vicprice`` that has little to no signal. (Verify with the inspection
+# module # ``report.inspection.impurity_decrease().plot()``)
```
GaetandeCast: Would be cool to also create new features that allow the model to predict better, reducing train/test gap further. In this time series forecasting problem, creating lagged features like in [this example](https://scikit-learn.org/stable/auto_examples/applications/plot_time_series_lagged_features.html) may help.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,213 @@
...
+
+# %%
+report_fe.checks.summarize()
+# Dropping spurious columns narrows the gap vs ``report_spurious``.
+
+# %%
+# Compare mitigations
+# ===================
+#
+# ``compare()`` needs the same test set — ``report_more_data`` used a different
```
GaetandeCast: Would be fixed by one of my comments above 😉

GaetandeCast (changes requested):
Thanks @moujanrastgoo ! here is a first set of remarks.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: Ok sure i will look into that!

moujanrastgoo: i decided to go with California housing for now

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: I tried to add more comments and explanations. Tell me if the style has improved!

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: Done, I modified them.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: Done as well

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: Normally fixed as well.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: Done, i removed the bullet point as well.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: I tried to use the features in California housing to combine them and create new features, but the reduction in the gap between the test and train scores was not noticeable, so I decided to deliberately keep one feature out of all the models and add it here. What do you think?

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
moujanrastgoo: Done as well, tell me if it corresponds with what you were asking for!

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+
+housing = fetch_california_housing(as_frame=True)
+y_full = housing.target * 100
+
+X_full, _, y, _ = train_test_split(
+    housing.data,
+    y_full,
+    train_size=5_000,
+    random_state=42,
+)
```
GaetandeCast:
```suggestion
X, X_heldout, y, y_heldout = train_test_split(
    housing.data,
    y_full,
    train_size=0.5,
    random_state=42,
)
```
Let's use half the dataset (~10k) then we can use the other half to showcase "more data = less underfitting"

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+
+X_full = X_full.reset_index(drop=True)
+y = pd.Series(y, name="MedHouseVal_k$").reset_index(drop=True)
+
+house_age = X_full["HouseAge"]
+X = X_full.drop(columns=["HouseAge"])
+
```
GaetandeCast:
```suggestion

```
Let's keep `"HouseAge"` in, we'll create features that will improve the score.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+# :class:`~sklearn.ensemble.HistGradientBoostingRegressor` with default settings
+# has enough capacity to overfit the training data on this table. Compare train and
+# test rows in the metrics table, then run checks to see if SKD001 is triggered.
+
+from sklearn.ensemble import HistGradientBoostingRegressor
+from skore import evaluate
+
+
+def untuned_hgbr():
+    return HistGradientBoostingRegressor(random_state=42)
```
GaetandeCast: No need to define this, use the model directly in evaluate.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+
+from skore import TrainTestSplit
+
+splitter = TrainTestSplit(random_state=42)
+
+# %%
+# Trigger SKD001 — untuned gradient boosting
+# ===========================================
+#
+# :class:`~sklearn.ensemble.HistGradientBoostingRegressor` with default settings
```
GaetandeCast: I think the point would be more compelling if we used a `RandomForest` instead. They overfit more by default and can gain performance when regularized.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+        max_depth=3,
+        min_samples_leaf=50,
+        learning_rate=0.05,
+        max_iter=200,
+        random_state=42,
+    ),
+    X=X,
+    y=y,
+    splitter=splitter,
+)
```
GaetandeCast:
```suggestion
report_regularized = evaluate(
    RandomForestRegressor(
        min_samples_leaf=10,
        max_features=0.5,
        random_state=42,
    ),
    X=X,
    y=y,
    splitter=splitter,
)
```
Something like this keep the test score high and reduces the train score. While it does not eliminate the SKD001 issue, we can say that it reduces the gap and that combined with the following steps it will.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+
+report_more_data = EstimatorReport(
+    untuned_hgbr(),
+    X_train=X_train_pool,
+    y_train=y_train_pool,
+    X_test=X_test,
+    y_test=y_test,
+)
+
```
GaetandeCast:
```suggestion
from sklearn.ensemble import HistGradientBoostingRegressor

X_train, X_test, y_train, y_test = train_test_split(
    X,
    y,
    test_size=0.2,
    random_state=42,
)  # same 20% hold-out as ``splitter``

X_train_augmented = pd.concat([X_train, X_heldout])
y_train_augmented = pd.concat([y_train, y_heldout])

report_less_data = EstimatorReport(
    HistGradientBoostingRegressor(random_state=42),
    X_train=X_train,
    y_train=y_train,
    X_test=X_test,
    y_test=y_test,
)

report_more_data = EstimatorReport(
    HistGradientBoostingRegressor(random_state=42),
    X_train=X_train_augmented,
    y_train=y_train_augmented,
    X_test=X_test,
    y_test=y_test,
)

data_comparison = compare(
    {"less_training_data": report_less_data, "more_training_data": report_more_data}
)

```
Now switch to a more powerful model and show that adding the heldout data reduces overfit significantly.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+# %%
+# The gap between test and train metrics is typically reduced when the model sees more
+# training rows.
+
+report_more_data.metrics.summarize(data_source="both").frame(favorability=True)
```
GaetandeCast:
```suggestion
data_comparison.metrics.summarize(data_source="both").frame(favorability=True)
```

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+
+# %%
+report_more_data.checks.summarize()
+
+# Even though the gap is reduced, SKD001 is still triggered.
```
GaetandeCast: SKD001 does not appear here with the previous suggestion.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+
+# %%
+# Improve feature engineering
+# ===========================
+#
```
GaetandeCast: You can take inspiration from this section of our docs to create some good feature engineering. https://docs.skore.probabl.ai/stable/auto_examples/use_cases/plot_feature_importance.html#more-complex-model

I think also that it would work better to put this section before the previous one in terms of narrative. Adding these features alone also won't fix the issue completly but will reduce the gap by improving test score. But then adding more data to an engineered pipeline with a regularized model should do it, which would make the example flow nicely.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,275 @@
...
+# %%
+# Removing ``house_id`` and restoring ``HouseAge`` reduces the gap between test and train
+# metrics relative to the spurious table. SKD001 may still fire because the booster remains
+# untuned; feature quality and regularization should be done together in practice.
+
+report_fe.checks.summarize()
+
+# %%
+# Compare mitigations
+# ===================
```
GaetandeCast: I think this section is a bit messy right now. Either make a single comparison will all reports or drop it.

GaetandeCast (changes requested):
Here is an other pass, we're getting there !

sphinx/user_guide/automated_checks.rst
```diff
@@ -77,11 +77,12 @@ training data and may generalize poorly.
 How to reduce the risk
 ^^^^^^^^^^^^^^^^^^^^^^
 
-- simplify the model,
 - regularize more strongly,
 - improve feature engineering,
 - use better validation protocols or more data.
 
+Check out the :ref:`example for this check <sphx_glr_auto_examples_pitfalls_and_solutions_plot_skd001_overfitting.py>`.
```
auguste-probabl:
```suggestion
Check out the :ref:`example for this check <example_skd001_overfitting>`.
```

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
+"""
+.. _example_skd001_overfitting:
+
+SKD001 — Potential overfitting
+==============================
+
+When a model fits the training set much better than the hold-out test set, it may
+have memorized patterns of the training data instead of generalizing.
```
glemaitre:
```suggestion
have memorized patterns of that only exist into the training data instead of only focusing on pattern leading to generalization to the testing data as well.
```

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+:ref:`SKD001 <skd001-overfitting>` flags this situation by comparing train and test
+scores across the report's default metrics.
+
+This example walks through practical mitigations suggested in
+:ref:`automated_checks`:
+
+- regularize more strongly,
```
glemaitre:
```suggestion
- regularize more strongly to limit the expressiveness of the model,
```

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+y_full = housing.target * 100
+
+X, X_heldout, y, y_heldout = train_test_split(
+    housing.data,
+    y_full,
+    train_size=0.5,
+    random_state=42,
+)
+
+X = X.reset_index(drop=True)
```
glemaitre: Why do we need to reset the index here? Are we using those index later?

If we are not using, I would not mind to keep original index.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+# %%
+
+TableReport(y)
+
+# %%
+# A single random train-test split is used in the :func:`~skore.evaluate` cells below.
+
+from skore import TrainTestSplit
+
+splitter = TrainTestSplit(random_state=42)
```
glemaitre: Let's pass the amount for the `test_size` such that it is explicit.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+from skore import evaluate
+
+report = evaluate(
+    RandomForestRegressor(random_state=42),
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+
+report.metrics.summarize(data_source="both").frame(favorability=True)
```
glemaitre: Let's drop the favorability in this example

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+    splitter=splitter,
+)
+
+report_regularized.metrics.summarize(data_source="both").frame(favorability=True)
+
+# %%
+# Train scores move closer to test scores. The gap narrows even if SKD001 may
+# still fire until the other mitigations below are applied.
+
+report_regularized.checks.summarize()
```
glemaitre: Here, we set the hyper parameters by hands and typically it is weird the good practice will require to make hyperparameter tuning. So even if in the example we don't want to have it to not increase the computation time, we should have a narrative in this regards. Let me make a proposal.

glemaitre: Here is my proposal:

--

Here, we set the hyperparameters of the model by hand. In practice, it is rather hard to know in advance which combination of hyperparameters leads to generalization.

In this case, one should lean towards tuning hyperparameters. We would advocate for using randomized search, with successive halving when the amount of samples is large, to find this combination. Refer to the following scikit-learn documentation for more details: https://scikit-learn.org/stable/modules/grid_search.html#tuning-the-hyper-parameters-of-an-estimator. In this particular example, we are not implementing the search in order to keep the execution time of the example short.

--

We need to make sure that the URL is an hyperlink using the RST syntax.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+When a model fits the training set much better than the hold-out test set, it may
+have memorized patterns of the training data instead of generalizing.
+:ref:`SKD001 <skd001-overfitting>` flags this situation by comparing train and test
+scores across the report's default metrics.
+
+This example walks through practical mitigations suggested in
+:ref:`automated_checks`:
+
+- regularize more strongly,
+- improve feature engineering,
```
glemaitre: When reviewing, I see that we forget a point which does not fit well into the "regularize more strongly" section but that is highly relevant to overfitting: early-stopping.

To provide some context: https://scikit-learn.org/stable/auto_examples/ensemble/plot_gradient_boosting_early_stopping.html

Basically, so for iterative algorithm, we know that if we had new iteration, the model will end-up overfitting. So one way is to create a validation set where we measure a metric on this validation set and if we don't see any improvement then we stop the iterative learning process because whatever further iteration is assume to make the model overfit.

So I think that we should have a section in this notebook after the hyperparameter search showing that it is another model parameter but that does not require any search.

Also, we might need to change slightly the documentation in the user guide to reflect this new addition.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+report_regularized.metrics.summarize(data_source="both").frame(favorability=True)
+
+# %%
+# Train scores move closer to test scores. The gap narrows even if SKD001 may
+# still fire until the other mitigations below are applied.
+
+report_regularized.checks.summarize()
+
+# %%
+# Improve feature engineering
```
glemaitre: Reading this I'm not sure that it is the right recommendation.

Adding new features means that you add more flexibility into your model and get a more expressive model: it is typically what you try to solve when your model underfit.

So here, the metrics improve meaning that our model is more expressive and it allow to close the gap between train and test as a side-effect. But if you would continue increasing the degree of the polynomial features, then the gap will increase again because your model will be to expressive capturing pattern that does not allow you to generalize, and thus lead to overfitting.

So to me, this part is really how to avoid a model to underfit, at risk to lead to overfitting.

Maybe I missed something @GaetandeCast

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+    y_test=y_test,
+)
+
+report_more_data = EstimatorReport(
+    engineered_hgbr,
+    X_train=X_train_augmented,
+    y_train=y_train_augmented,
+    X_test=X_test,
+    y_test=y_test,
+)
```
glemaitre: I would like to still use `skore.evalute` here:

```python
model.fit(X_train, y_train)
skore.evaluate(model, X_test, y_test, cv="prefit")
```

It should lead to the same.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+
+# %%
+# Test error typically improves when the model sees the held-out rows during
+# training, even though train scores remain optimistic.
+
+report_more_data.metrics.summarize(data_source="both").frame(favorability=True)
```
glemaitre: Let's drop favorability

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+# income, occupancy, and location. The train/test gap also narrows compared with
+# the default forest on raw columns.
+
+report_engineered.checks.summarize()
+
+# %%
+# Use more training data
+# ======================
+#
+# We hold out one fixed test set, then fit the engineered pipeline on the
```
glemaitre: At the end, of the section, it could actually be cool to have something around the learning-curve: https://scikit-learn.org/stable/modules/learning_curve.html#learning-curve

We don't have it in skore so maybe it could only be a narrative paragraph that mention that one, at the cost of multiple fit, could actually see the effect of adding new observations and potentially extrapolate the cost/benefit of acquiring new observations in the future.

examples/pitfalls_and_solutions/plot_skd001_overfitting.py
```diff
@@ -0,0 +1,242 @@
...
+        "more_training_data": report_more_data,
+    }
+)
+comparison.metrics.summarize(data_source="both").frame(favorability=True)
+
+# %%
+# Conclusion
+# ==========
+#
+# SKD001 warns against models that excel on training data but underperform on
```
glemaitre: In the conclusion, I would also like that we mention that overfitting that the solution of the overfitting is not necessarily only a single point that have been mention but a combination of them.

examples/pitfalls_and_solutions/README.rst
```diff
@@ -0,0 +1,5 @@
+Pitfalls and solutions
+----------------------
+Examples for each automated check. Each script
+reproduces a check, and walks
+through fixes described in :ref:`automated_checks`.
```
glemaitre: Try to fit the 88 characters minimum on a line.

examples/pitfalls_and_solutions/README.rst
```diff
@@ -0,0 +1,5 @@
+Pitfalls and solutions
+----------------------
```
glemaitre:
```suggestion
Checking for modeling pitfalls and remediation
----------------------------------------------
```

sphinx/conf.py
```diff
@@ -102,6 +102,7 @@
         "../examples/getting_started",
         "../examples/use_cases",
         "../examples/model_evaluation",
+        "../examples/pitfalls_and_solutions",
```
glemaitre, line 105: note that we can keep this naming for the folder

glemaitre (commented):
Here, is a couple of comments

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
+"""
+.. _example_skd001_skd002_overfitting_underfitting:
+
+SKD001 & SKD002 — Overfitting and underfitting
```
GaetandeCast:
```suggestion
SKD001 & SKD002 - Overfitting and underfitting
```
Let's try to avoid the use of em dashes, they are used by LLMs a lot and may let the reader think that the example was fully generated without review.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+==============================================
+
+:ref:`SKD002 <skd002-underfitting>` and :ref:`SKD001 <skd001-overfitting>`
+describe opposite ends of the same problem: model *expressiveness*.
+
+- Too little expressiveness: train and test scores stay close to a weak
+  learner / dummy baseline (SKD002).
```
GaetandeCast: "Weak learner" is a specific term used in the gradient boosting litterature (see [here](https://en.wikipedia.org/wiki/Boosting_(machine_learning))) to describe a machine learning model that is only slightly better than random guessing.
Therefore I would avoid the use of this term here and use something like:
`train and test scores are close to those of a model that guesses randomly (often called a dummy model).`

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+- Too much expressiveness: train scores pull far ahead of test scores
+  (SKD001).
+
+The correct the model beats that baseline on both train and test, without
+memorizing the training set. This notebook walks that path while showing
+different mitigations
```
GaetandeCast:
```suggestion
A performant model learns from the training data without memorizing its specificities,
so that learned knowledge can be generalized to new data. This notebook walks the
underfitting to overfitting path while showing different mitigations techniques:
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+
+- capacity (model family / complexity)
+- features
+- regularization
+- early stopping
+- more data
```
GaetandeCast:
```suggestion
- tweaking model capacity (model family / complexity)
- adding and removing features
- tuning regularization
- using early stopping
- using more data
```
Let's use verbs :)

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+    y=y,
+    splitter=splitter,
+)
+report_hgbr.metrics.summarize(data_source="both").frame()
+
+# %%
+report_hgbr.checks.summarize(fast_mode=True)
+
+# %%
+# Feature engineering
```
GaetandeCast, line 117: I think we could extend that section to showcase how feature selection can reduce overfitting: when properly done, feature selection allows the model to focus on the features that contain signal and not learn patterns of useless features. I suggest doing a recursive feature elimination using [`RFECV`](https://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.RFECV.html) with permutation importance.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+# %%
+from skrub import TableReport
+
+TableReport(X)
+
+# %%
+TableReport(y)
+
+# %%
+# We use the same split for every :func:`~skore.evaluate` call.
```
GaetandeCast:
```suggestion
# We use the same split for every :func:`~skore.evaluate` call with a fixed seed,
# so that we are sure that we are comparing models on the same test data.
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+report_underfit = evaluate(
+    Ridge(alpha=1_000_000),
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+report_underfit.metrics.summarize(data_source="both").frame()
+
+# %%
+# SKD002 should be present.
```
GaetandeCast:
```suggestion
# For a model that learns so little, SKD002 fires.
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+report_underfit.checks.summarize(fast_mode=True)
+
+# %%
+# Increase model capacity
+# =======================
+#
+# Moving away from underfitting means giving the model enough expressiveness to
+# use the inputs. Dropping to a default :class:`~sklearn.linear_model.Ridge`
+# (mild regularization) already learns useful weights on each feature and is
+# usually enough to clear SKD002 on this table.
```
GaetandeCast:
```suggestion
# enough to clear SKD002 on this table.
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,304 @@
...
+report_rf_reg.checks.summarize(fast_mode=True)
+
+# %%
+# Early stopping
+# ==============
+#
+# For iterative learners, early stopping is another way to limit expressiveness,
+# but it does not require a hyperparameter search. Hold out a validation set,
+# monitor a metric, and stop when that metric stops improving — further
+# iterations are assumed to overfit. See scikit-learn's example on
```
GaetandeCast:
```suggestion
# For iterative learners, early stopping is another way to limit expressiveness
# that does not require a hyperparameter search. Hold out a validation set,
# monitor a metric, and stop when that metric stops improving. Further
# iterations are assumed to overfit. Either pass a validation set to `.fit` or use the
# `validation_fraction` parameter. See scikit-learn's example on
```

GaetandeCast (commented):
This is shaping up nicely! Here are a few more suggestions.

GaetandeCast:
Hmm I think some of my suggestions got broken, you may have to apply them by hand. They don't seem to be on the right lines.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
moujanrastgoo: Hello, I made a section with that. Tell me what you think about it and if I should change anything!

moujanrastgoo: Actually, I am not satisfied with it. I'm working on it again.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+# %%
+from skrub import TableReport
+
+TableReport(X)
+
+# %%
+TableReport(y)
+
+# %%
+# We use the same split for every :func:`~skore.evaluate` call with a fixed
```
glemaitre: Here, we are not using `evaluate` right away. We need to move both the import and narrative where we are going to use it for the first time. Only describe the `TrainTestSplit` object.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+)
+report_underfit.metrics.summarize(data_source="both").frame()
+
+# %%
+# For a model that learns so little, SKD002 fires.
+
+report_underfit.checks.summarize(fast_mode=True)
+
+# %%
+# Increase model capacity
```
glemaitre: Instead of capacity, I would "expressiveness"

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+# =======================
+#
+# Moving away from underfitting means giving the model enough expressiveness to
+# use the inputs. Dropping to a default :class:`~sklearn.linear_model.Ridge`
+# (mild regularization) already learns useful weights on each feature and is
+# enough to clear SKD002 on this table.
+
+report_ridge = evaluate(Ridge(), X=X, y=y, splitter=splitter)
```
glemaitre: set a specific alpha to be explicit to see that it is smaller than the previous one.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+# ==========================
+#
+# We start with a linear model that is overly regularized.
+# :class:`~sklearn.linear_model.Ridge` with a very large ``alpha`` shrinks
+# coefficients toward zero, so predictions stay close to a dummy baseline.
+
+from sklearn.linear_model import Ridge
+
+report_underfit = evaluate(
+    Ridge(alpha=1_000_000),
```
glemaitre:
```suggestion
    Ridge(alpha=1e6),
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+)
+report_rf_all_features.metrics.summarize(data_source="both").frame()
+
+# %%
+# A small wrapper fits the forest, then stores permutation importances so
+# RFECV can drop the least useful column at each step.
+
+
+class PermutationImportanceEstimator(RegressorMixin, BaseEstimator):
+    def __init__(self, n_estimators=20, n_repeats=1, random_state=42, n_jobs=4):
```
glemaitre:
```suggestion
    def __init__(self, n_estimators=20, n_repeats=1, random_state=42, n_jobs=None):
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+            random_state=self.random_state,
+            n_jobs=self.n_jobs,
+        ).importances_mean
+        return self
+
+    def predict(self, X):
+        return self.estimator_.predict(X)
+
+
+# Subsample for the selector so the gallery stays fast; apply the chosen
```
glemaitre:
```suggestion
# %%
# Subsample for the selector so the gallery stays fast; apply the chosen
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+    y=y,
+    splitter=splitter,
+)
+
+compare(
+    {
+        "rf_all_features": report_rf_all_features,
+        "rf_after_rfecv": report_rf_selected,
+    }
+).metrics.summarize(data_source="both").frame()
```
glemaitre: can we make that we call `skore.evaluate` and `skore.compare` by importing `skore`.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+# <https://scikit-learn.org/stable/modules/grid_search.html#tuning-the-hyper-parameters-of-an-estimator>`_.
+# In this particular example, we are not implementing the search in order to
+# keep the execution time of the example short.
+
+report_rf_reg = evaluate(
+    RandomForestRegressor(
+        min_samples_leaf=20,
+        max_features=0.5,
+        random_state=42,
+    ),
```
glemaitre: Let's define this model outside of `evaluate`.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+# For iterative learners, early stopping is another way to limit expressiveness
+# that does not require a hyperparameter search. Hold out a validation set,
+# monitor a metric, and stop when that metric stops improving. Further
+# iterations are assumed to overfit. Either pass a validation set to ``.fit``
+# or use the ``validation_fraction`` parameter. See scikit-learn's example on
+# `gradient boosting with early stopping
+# <https://scikit-learn.org/stable/auto_examples/ensemble/plot_gradient_boosting_early_stopping.html>`_.
+
+report_hgbr_es = evaluate(
+    HistGradientBoostingRegressor(
```
glemaitre: Same here.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,395 @@
...
+# %%
+# More training data
+# ==================
+#
+# Extra labeled rows help both regimes, but they are especially useful against
+# overfitting: the model has fewer opportunities to memorize a small sample.
+# Keep one fixed test fold, fit on the original train fold, then refit after
+# concatenating the held-out half of the dataset. Use
+# :func:`~skore.evaluate` with ``splitter="prefit"`` once the estimator is
+# fitted.
```
glemaitre: We need to mention later that the side-effect is that we will not have access to the `fit` time (it explains the `NaN` value in the reported metrics).

glemaitre (commented):
A couple of additional remarks. But it looks good, almost mergeable.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
moujanrastgoo: Should I do from skore import compare or evaluate, and then use them as just evaluate and compare, or do you want me to do skore.evaluate and skore.compare directly?

glemaitre: Something like:

```python
import skore

report = skore.compare(...)
report.metrics.summarize()
```

moujanrastgoo: OK I see Thanks!

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,400 @@
...
+# expansions (very high-degree interactions, huge one-hot spaces) can reopen an
+# overfit gap. Think of features as capacity you add to the *inputs*, not only
+# to the estimator.
+
+X_fe = X.assign(RoomsPerPerson=X["AveRooms"] / X["AveOccup"].clip(lower=0.1))
+
+report_ridge_fe = skore.evaluate(Ridge(alpha=1.0), X=X_fe, y=y, splitter=splitter)
+report_ridge_fe.metrics.summarize(data_source="both").frame()
+
+# %%
```
GaetandeCast, line 136:
```suggestion
# %%
# Our test score increased from `0.609` to `0.639` while keeping the train score close:
# we reduced underfitting without creating overfitting by adding a signal-rich feature to
# the model.
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,400 @@
...
+report_ridge_fe.checks.summarize(fast_mode=True)
+
+# %%
+# Feature selection is the other direction: drop columns that do not carry
```
GaetandeCast: I think it would flow more naturally to split the section off and add it in the overfitting mitigation steps. Then you can name the sections "Feature engineering pt.1" and "Feature engineering pt.2" or something.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,400 @@
...
+    splitter=splitter,
+)
+report_rf_all_features.metrics.summarize(data_source="both").frame()
+
+# %%
+# A small wrapper fits the forest, then stores permutation importances so
+# RFECV can drop the least useful column at each step.
+
+
+class PermutationImportanceEstimator(RegressorMixin, BaseEstimator):
```
GaetandeCast: You can pass scikit-learn's `permutation_importance` method to the `importance_getter` of `RFECV` instead of wrapping the RF.

GaetandeCast (commented):
A few more remarks

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,433 @@
...
+X_sel = X.sample(3_000, random_state=42)
+y_sel = y.loc[X_sel.index]
+X_fs, X_val, y_fs, y_val = train_test_split(
+    X_sel, y_sel, test_size=0.2, random_state=42
+)
+X_val_np = np.asarray(X_val)
+y_val_np = np.asarray(y_val)
+
+
+def permutation_importance_getter(model, X_val, y_val, random_state=42):
```
glemaitre: This seems far to complex: you should never have to call the Python frame :)

Let me check how to make this work with a simple callable.

glemaitre: @GaetandeCast I think that we would need https://github.com/scikit-learn/scikit-learn/pull/32340 to make it work easily?

moujanrastgoo: After some discussion, it was decided to remove this section for now.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,314 @@
...
+the main walkthrough; the other half is reserved to show the effect of adding
+training data.
+"""
+
+# %%
+# Load the California housing dataset
+# ===================================
+#
+# Each row is a census block group. The target ``MedHouseVal`` is in 100k USD
+# units; we multiply by 100 so errors read in kUSD.
```
GaetandeCast: We never really inspect the errors or anything that reads in the target unit so I would argue for dropping this conversion. IMO it adds overhead to the reader without justifying value.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,314 @@
...
+y = pd.Series(y, name="MedHouseVal_kUSD")
+y_heldout = pd.Series(y_heldout, name="MedHouseVal_kUSD")
+
+# %%
+from skrub import TableReport
+
+TableReport(X)
+
+# %%
+TableReport(y)
```
GaetandeCast, line 67: I'm questioning the value of these `TableReports`. Since we don't do any analysis of the features or the target distribution, I fear this impedes on the purpose of the example by dumping unexploited information.

I would either completely drop these or add some analysis so that the plots have value.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,314 @@
...
+report_ridge.metrics.summarize(data_source="both").frame()
+
+# %%
+report_ridge.checks.summarize(fast_mode=True)
+
+# %%
+# A :class:`~sklearn.ensemble.HistGradientBoostingRegressor` adds nonlinear
+# capacity and can lift test scores further. Watch the train/test gap as you do
+# this: the same method that cures underfitting is the one that creates
+# overfitting.
```
GaetandeCast: I can't help but feel like this section is a bit out of place. 
On the one hand I understand the idea of showing two ways to increase "model expressiveness": tuning regularization and using a model that makes less (unverified) assumptions on the data.
But at the same time, triggering SKD001 with this HGBR seems a bit premature: we did not introduce the check yet. Also it breaks the continuity with the next section where we go back to the linear model which can be missed by someone reading fast. 

Here is how I would fix this: swap this section with the feature engineering section (preserve continuity) and use a RF model instead, justified as an other way of capturing non linearity (other than a hand crafted non linear combination of features). This then transitions nicely with the part on SKD001.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,314 @@
...
+# Early stopping
+# ==============
+#
+# For iterative learners, early stopping is another way to limit expressiveness
+# that does not require a hyperparameter search. Hold out a validation set,
+# monitor a metric, and stop when that metric stops improving. Further
+# iterations are assumed to overfit. Either pass a validation set to ``.fit``
+# or use the ``validation_fraction`` parameter. See scikit-learn's example on
+# `gradient boosting with early stopping
+# <https://scikit-learn.org/stable/auto_examples/ensemble/plot_gradient_boosting_early_stopping.html>`_.
```
GaetandeCast, line 208: Here I would make a comparison report of an HGBR with and one without early stopping, so that we can see the effect clearly. Especially if you remove the earlier (not stopped) HBGR as I suggested.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,314 @@
...
+# (many refits on growing subsets) shows whether more labels are still worth
+# the cost. skore does not wrap that API yet; treat it as a scikit-learn-side
+# diagnostic next to these checks.
+
+# %%
+# Summary comparison
+# ==================
+#
+# Reading left to right: over-regularized Ridge → enough capacity → too much
+# capacity → controlled fit → more data. The same mitigations appear once; only
```
GaetandeCast: I would prefer a more natural phrase rather than the use of these unicode arrows :)

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,314 @@
...
+# the *direction* (add expressiveness vs limit it) changes.
```
GaetandeCast: And honestly I don't really understand the purpose of giving meaning to the order of the results beyond the order in which they were introduced in. I would just say something along the lines of "Here is a side-by-side comparison of the impact of the different techniques we applied, measured on the same test set."

GaetandeCast (commented):
A few more remarks, thanks for your patience :)

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
GaetandeCast: (I meant to comment on the code block just under)

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
moujanrastgoo: ok sure, I see that!

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
moujanrastgoo: I think because of the feature engineering blocks later and the fact that we mention specific columns it might be better to show the original table in each example because if not, the reader would not have any idea of what the columns are that we are talking about, but if you think it does not add anything i am happy to remove them!

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
moujanrastgoo: Sure, I see your point!

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
moujanrastgoo: sure i will fix that

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
moujanrastgoo: ok i see!

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
GaetandeCast: Good point. Any opinion @glemaitre ?

glemaitre: We can keep it but I think that I agree with @GaetandeCast that a small comment to point out what to pay attention is quite necessary.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,319 @@
...
+
+X, X_heldout, y, y_heldout = train_test_split(
+    housing.data,
+    housing.target,
+    train_size=0.5,
+    random_state=42,
+)
+
+y = pd.Series(y, name="MedHouseVal")
+y_heldout = pd.Series(y_heldout, name="MedHouseVal")
```
GaetandeCast:
```suggestion
```

This conversion is not needed anymore.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,319 @@
...
+
+# %%
+# Note ``AveRooms`` and ``AveOccup``: we combine them later into a
+# rooms-per-person feature.
```
GaetandeCast: Let's describe the dataset a little more. Let's mention how it is aggregate data around in a certain geographical region whose coordinates we have, and how we have aggregated information on characteristics of the houses (age, number of rooms, occupation etc.) and on the geographical region (median income, population). We can then foreshadow the feature engineering part saying that while the number of features is limited, they are rich and can be combined to create further informative markers such as e.g. creating a `AveRooms / AveOccup`.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,319 @@
...
+
+from skrub import TableReport
+
+TableReport(X)
+
+# %%
+TableReport(y)
+
```
GaetandeCast, line 68: Either drop this one or mention the ceiling effect: MedHouseVal is capped at $500k. You can suggest clicking the column to see the histogram.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,319 @@
...
+# %%
+# Overfitting: SKD001 fires
+# =========================
+#
+# Hand-crafted features are one way to capture nonlinearity for a linear model.
+# Another is to switch to a model family that learns nonlinear structure on its
+# own. A default :class:`~sklearn.ensemble.RandomForestRegressor` does that,
+# but unrestricted leaves can also memorize the training data: train metrics
+# look excellent, test metrics lag, and SKD001 flags the gap.
```
GaetandeCast:
```suggestion
# but unrestricted leaves can also memorize the training data: train metrics
# look excellent, test metrics improve but not as much as train ones, and
# SKD001 flags the gap.
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,319 @@
...
+# For iterative learners, early stopping is another way to limit expressiveness
+# that does not require a hyperparameter search. Hold out a validation set,
+# monitor a metric, and stop when that metric stops improving. Further
+# iterations are assumed to overfit. Either pass a validation set to ``.fit``
+# or use the ``validation_fraction`` parameter. See scikit-learn's example on
+# `gradient boosting with early stopping
+# <https://scikit-learn.org/stable/auto_examples/ensemble/plot_gradient_boosting_early_stopping.html>`_.
+#
+# Compare a :class:`~sklearn.ensemble.HistGradientBoostingRegressor` with and
+# without early stopping on the same split.
```
GaetandeCast:
```suggestion
# Let us compare a :class:`~sklearn.ensemble.HistGradientBoostingRegressor`
# with and without early stopping on the same split to see the effect.
```

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
```diff
@@ -0,0 +1,319 @@
...
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+
+model_hgbr_es = HistGradientBoostingRegressor(
+    early_stopping=True,
+    validation_fraction=0.1,
+    n_iter_no_change=10,
+    max_iter=40,
```
GaetandeCast:
```suggestion
    HistGradientBoostingRegressor(max_iter=1000, random_state=42),
    X=X,
    y=y,
    splitter=splitter,
)

model_hgbr_es = HistGradientBoostingRegressor(
    early_stopping=True,
    validation_fraction=0.1,
    n_iter_no_change=10,
    max_iter=1000,
```

Let's really bump the number of iterations to see the effect.
The little drop in test performance is probably due to training on only 90% of the train dataset. I think we should have similar test performance if we passed `X_heldout` and `y_heldout` instead of `validation_fraction`. However it is annoying to wire in fit parameters in skore right now (you have to use skrub) so I would just mention it in a comment.

@glemaitre do we have any plans to make it simple to pass e.g. `X_val` and `y_val` to HGBR ?

GaetandeCast (commented):
Final remarks, I think that then we can merge !

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
glemaitre: @GaetandeCast it is already the case: https://scikit-learn.org/stable/modules/generated/sklearn.ensemble.HistGradientBoostingRegressor.html#sklearn.ensemble.HistGradientBoostingRegressor.fit

glemaitre: For a short example you can check: https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_1_7_0.html#custom-validation-set-for-histogram-based-gradient-boosting-estimators

glemaitre: It however requires metadata routing in this case.

glemaitre:
I'm fine merging as well without adding new changes to the one propose by Gaetan.

examples/pitfalls_and_solutions/plot_skd001_skd002_overfitting_underfitting.py
GaetandeCast: Yes I meant wiring it with skore. We don't have an easy way of passing fit parameters, do we ?

GaetandeCast (approved):
LGTM, thanks a lot @moujanrastgoo !!

PR #3141, opened 2026-07-13
chore(sklearn/base): Remove deprecated `_hash` property

thomass-dev:


glemaitre (approved):
Yep let's break stuff.

PR #3142, opened 2026-07-13
feat(plugins): Add module to infer requirements

thomass-dev:
This approach is naive and basic: based on the fact that at put time, the predictions have been made, the report has been filled, so by transitivity, all the imports required by the model or the scorer have already be done.

This PR intends to inspect `sys.modules`, filter-out all builtins and disambiguate module versus distribution (`sklearn/scikit-learn`). It is python packages managers agnostic.

It raises a warning when editable/local packages are detected. In a near future, we should be able to automatically ship editable code needed for models and scorers, or do a "a-la-mlflow" declarative API.

### DIFF
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -32,6 +32,7 @@ dependencies = [
   "matplotlib",
   "narwhals>=2.0.1",
   "numpy",
+  "packaging",
   "pandas",
   "plotly",
   "rich>=14.2",
--- a/skore/pyproject.toml
+++ b/skore/pyproject.toml
@@ -32,7 +32,7 @@ dependencies = [
   "matplotlib",
   "narwhals>=2.0.1",
   "numpy",
-  "packaging",
+  "packaging>=24.2",
   "pandas",
   "plotly",
   "rich>=14.2",
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -4,7 +4,6 @@ import importlib.metadata
 import logging
 import pathlib
 import platform
-import site
 import sys
 import sysconfig
 import types
@@ -15,30 +14,23 @@ logger = logging.getLogger(__name__)
 
 def is_local_module(module: types.ModuleType) -> bool:
     """
-    Return whether ``module`` was loaded from outside site-packages.
+    Return whether ``module`` was loaded from outside the standard library.
 
-    True for editable installs, source trees, and other paths that are neither the
-    standard library nor an installed distribution in site-packages.
+    True for editable installs, source trees, and other non-stdlib paths.
+    False when there is no usable file origin (``None``, ``built-in``,
+    ``frozen``) or the origin is under the stdlib path.
     """
-    sitepackages = [pathlib.Path(path).resolve() for path in site.getsitepackages()]
-
-    if site.ENABLE_USER_SITE and site.getusersitepackages():
-        sitepackages.append(pathlib.Path(site.getusersitepackages()).resolve())
-
     origin = (module.__spec__ and module.__spec__.origin) or getattr(
         module, "__file__", None
     )
 
-    if (not origin) or (origin in {"built-in", "frozen"}):
+    if (origin is None) or (origin in {"built-in", "frozen"}):
         return False
 
     path = pathlib.Path(origin).resolve()
     stdlib = pathlib.Path(sysconfig.get_path("stdlib")).resolve()
 
-    if path.is_relative_to(stdlib):
-        return False
-
-    return not any(map(path.is_relative_to, sitepackages))
+    return not path.is_relative_to(stdlib)
 
 
 class Requirement(typing.TypedDict):
@@ -61,8 +53,8 @@ def infer() -> Requirements:
 
     Maps each imported top-level package to its distribution via
     :func:`importlib.metadata.packages_distributions`, then records the installed
-    version. Local or editable packages outside site-packages are skipped and a
-    warning is emitted once per package name.
+    version. Unmapped packages loaded from outside the standard library are
+    skipped and a warning is emitted once per package name.
     """
     module_to_requirement = importlib.metadata.packages_distributions()
     requirement_to_version = {}
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -22,7 +22,7 @@ import packaging.utils
 logger = logging.getLogger(__name__)
 
 version = functools.cache(importlib.metadata.version)
-MODULE_TO_REQUIREMENT = importlib.metadata.packages_distributions()
+MODULE_TO_DISTRIBUTIONS = importlib.metadata.packages_distributions()
 SITE_PACKAGES_DIRPATHS = tuple(
     {
         pathlib.Path(path).resolve()
@@ -67,6 +67,60 @@ class Requirement(typing.TypedDict):
     version: str
 
 
+@functools.cache
+def __distribution_files(name: str) -> frozenset[pathlib.Path]:
+    files = importlib.metadata.distribution(name).files or []
+    located = set()
+
+    for file in files:
+        try:
+            located.add(pathlib.Path(file.locate()).resolve())
+        except OSError:
+            continue
+
+    return frozenset(located)
+
+
+def __distribution(top_level_name: str, module: types.ModuleType) -> str | None:
+    """
+    Return the distribution name that owns ``module``.
+
+    Uses :func:`importlib.metadata.packages_distributions` for the top-level
+    name. When several distributions share that name (namespace packages), only
+    the distribution whose installed files include the module origin is returned.
+    A bare namespace root with no file origin owns nothing on its own.
+    """
+    candidates = MODULE_TO_DISTRIBUTIONS.get(top_level_name)
+
+    if not candidates:
+        return None
+
+    if len(candidates) == 1:
+        return candidates[0]
+
+    # When several distributions share that top-level name (namespace packages),
+    # we have to inspect each distribution and retrieve the one that contains the module
+    # file.
+
+    origin = (module.__spec__ and module.__spec__.origin) or getattr(
+        module, "__file__", None
+    )
+
+    if (origin is None) or (origin in {"built-in", "frozen"}):
+        return None
+
+    origin = pathlib.Path(origin).resolve()
+
+    return next(
+        (
+            candidate
+            for candidate in candidates
+            if origin in __distribution_files(candidate)
+        ),
+        None,
+    )
+
+
 def infer() -> list[Requirement]:
     """
     Infer distribution requirements from modules currently in ``sys.modules``.
@@ -75,11 +129,12 @@ def infer() -> list[Requirement]:
     moment of the call. Two successive calls can therefore differ: any import (or
     module removal) between them changes which distributions are discovered.
 
-    Maps each imported top-level package to its distribution via
-    :func:`importlib.metadata.packages_distributions`, then records the installed
-    version. Packages loaded from outside known install locations (editable or
-    local installs) are skipped and a warning is emitted once per package name,
-    even when a distribution mapping exists.
+    Maps each imported module to its distribution via
+    :func:`importlib.metadata.packages_distributions`. For namespace packages,
+    only the distribution that owns the imported file is recorded. Packages
+    loaded from outside known install locations (editable or local installs) are
+    skipped and a warning is emitted once per package name, even when a
+    distribution mapping exists.
     """
     requirement_to_version = {}
     warned = set()
@@ -111,13 +166,12 @@ def infer() -> list[Requirement]:
                 )
             continue
 
-        if requirements := MODULE_TO_REQUIREMENT.get(top_level_name):
-            for requirement in requirements:
-                if requirement not in requirement_to_version:
-                    requirement_to_version[requirement] = version(requirement)
+        if distribution := __distribution(top_level_name, module):
+            if distribution not in requirement_to_version:
+                requirement_to_version[distribution] = version(distribution)
         else:
             # No distribution mapping: Cython/C-extension aliases, runtime entrypoints
-            # (__main__), etc.
+            # (__main__), bare namespace roots, etc.
             logger.debug(module)
 
     return sorted(
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -9,6 +9,8 @@ import types
 import typing
 import warnings
 
+import packaging.utils
+
 logger = logging.getLogger(__name__)
 
 
@@ -90,6 +92,12 @@ def infer() -> list[Requirement]:
                 )
 
     return [
-        Requirement(name=name, version=version)
+        Requirement(
+            name=packaging.utils.canonicalize_name(name),
+            version=packaging.utils.canonicalize_version(
+                version,
+                strip_trailing_zero=False,
+            ),
+        )
         for name, version in sorted(requirement_to_version.items())
     ]
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -8,6 +8,7 @@ import sys
 import sysconfig
 import types
 import typing
+import warnings
 
 logger = logging.getLogger(__name__)
 
@@ -85,12 +86,15 @@ def infer() -> Requirements:
 
             if is_local_module(module) and (top_level_name not in warned):
                 warned.add(top_level_name)
-                logger.warning(
-                    "\033[38;5;208m"
-                    f"Package \033[1;3m{top_level_name}\033[22;23m seems to be an "
-                    "editable or local install (loaded from outside site-packages). "
-                    "It will not be recorded in the inferred requirements."
-                    "\033[0m"
+                warnings.warn(
+                    (
+                        "\033[38;5;208m"
+                        f"Package \033[1;3m{top_level_name}\033[22;23m seems to be an "
+                        "editable or local install (loaded from outside site-packages)."
+                        " It will not be recorded in the inferred requirements."
+                        "\033[0m"
+                    ),
+                    stacklevel=2,
                 )
 
     return Requirements(
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -3,6 +3,7 @@
 import functools
 import importlib.metadata
 import logging
+import operator
 import pathlib
 import sys
 import sysconfig
@@ -91,13 +92,16 @@ def infer() -> list[Requirement]:
                     stacklevel=2,
                 )
 
-    return [
-        Requirement(
-            name=packaging.utils.canonicalize_name(name),
-            version=packaging.utils.canonicalize_version(
-                version,
-                strip_trailing_zero=False,
-            ),
-        )
-        for name, version in sorted(requirement_to_version.items())
-    ]
+    return sorted(
+        (
+            Requirement(
+                name=packaging.utils.canonicalize_name(name),
+                version=packaging.utils.canonicalize_version(
+                    version,
+                    strip_trailing_zero=False,
+                ),
+            )
+            for name, version in requirement_to_version.items()
+        ),
+        key=operator.itemgetter("name"),
+    )
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -37,7 +37,7 @@ class Requirement(typing.TypedDict):
     """A distribution name and its installed version."""
 
     name: str
-    version: str | None
+    version: str
 
 
 def infer() -> list[Requirement]:
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -82,11 +82,9 @@ def infer() -> list[Requirement]:
                 warned.add(top_level_name)
                 warnings.warn(
                     (
-                        "\033[38;5;208m"
-                        f"Package \033[1;3m{top_level_name}\033[22;23m seems to be an "
-                        "editable or local install (loaded from outside site-packages)."
-                        " It will not be recorded in the inferred requirements."
-                        "\033[0m"
+                        f"Package {top_level_name} seems to be an editable or local "
+                        "install (loaded from outside site-packages). It will not be "
+                        "recorded in the inferred requirements."
                     ),
                     stacklevel=2,
                 )
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -76,7 +76,7 @@ def infer() -> list[Requirement]:
         else:
             # No distribution mapping: Cython/C-extension aliases, runtime entrypoints
             # (__main__), or local/dev imports outside site-packages.
-            logging.debug(module)
+            logger.debug(module)
 
             if is_local_module(module) and (top_level_name not in warned):
                 warned.add(top_level_name)
--- a/skore/src/skore/_plugins/requirements.py
+++ b/skore/src/skore/_plugins/requirements.py
@@ -3,7 +3,6 @@
 import importlib.metadata
 import logging
 import pathlib
-import platform
 import sys
 import sysconfig
 import types
@@ -41,14 +40,7 @@ class Requirement(typing.TypedDict):
     version: str | None
 
 
-class Environment(typing.TypedDict):
-    """Python version and inferred distribution requirements."""
-
-    python: str
-    requirements: list[Requirement]
-
-
-def infer() -> Environment:
+def infer() -> list[Requirement]:
     """
     Infer distribution requirements from modules currently in ``sys.modules``.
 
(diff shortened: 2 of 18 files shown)


skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,53 @@
...
+
+        if top_level_module in sys.stdlib_module_names:
+            continue
+
+        if requirements := module_to_requirement.get(top_level_module):
+            for requirement in requirements:
+                if requirement not in requirement_to_version:
+                    requirement_to_version[requirement] = version(requirement)
+        else:
+            ...
```
thomass-dev: To do: deal with non-packaged code.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+    requirement_to_version = {}
+    warned = set()
+
+    # Snapshot: importlib.metadata/console I/O can load modules and mutate sys.modules
+    # while we iterate.
+    for module in list(sys.modules.values()):
+        # sys.modules values are usually modules, but can be None (cleared/failed
+        # imports) or non-modules (legacy typing.io/re classes) that lack __spec__.
+        if not isinstance(module, types.ModuleType):
+            continue
```
glemaitre, line 146: Apparently, this is not compatible with Python 3.11 and 3.12:

```pytb
File "skore/src/skore/_plugins/requirements.py", line 62, in infer
    name = (module.__spec__ and module.__spec__.name) or module.__name__
AttributeError: type object 'typing.io' has no attribute '__spec__'
```

The tests does not catch it because we are using a mock.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+class Requirement(typing.TypedDict):
+    """A distribution name and its installed version."""
+
+    name: str
+    version: str
+
+
+def infer() -> list[Requirement]:
```
glemaitre, line 121: I might prefer `infer_requirements` to get what it does without looking at the import path.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+import sys
+import sysconfig
+import types
+import typing
+import warnings
+
+logger = logging.getLogger(__name__)
+
+
+def is_local_module(module: types.ModuleType) -> bool:
```
glemaitre: Apparently, a package installed with `setuptools` using `pip install -e .` will not work as expected. I get the following explanation:

`setuptools` writes `top_level.txt` into the `.dist-info`, so `packages_distributions()` maps `mypkg`, the if branch is taken, and `is_local_module` is never consulted — even though it returns `True`. The editable package is silently recorded as `mypkg==0.1.0`, which is exactly the unreproducible pin the warning was meant to prevent.

Conversely, skore's own hatchling/uv editable install writes no top_level.txt, so it does warn. So the behaviour is determined by the build backend's metadata rather than by whether the install is editable, which makes it unpredictable across users.

The two checks should be independent: test is_local_module (or read direct_url.json → dir_info.editable, which is the authoritative PEP 610 signal) before the mapping lookup, not only when the lookup fails.

glemaitre: My friend is complaining about the naming which is actually true:

is_local_module actually means "not in the stdlib directory"

```
is_local_module(numpy)   -> True
is_local_module(sklearn) -> True
```

Both are ordinary site-packages installs. In a virtualenv, `sysconfig.get_path("stdlib")` resolves against the base interpreter, so site-packages is never under it:

```
stdlib   /Users/.../uv/python/cpython-3.12.12-.../lib/python3.12
purelib  /Users/.../skore/.venv/lib/python3.12/site-packages
site-packages under stdlib? False
```

The function is correct against its docstring's first line, but the name and the way it's used as an editable-install proxy are misleading, and the docstring's "True for editable installs, source trees" reads as if those were the only cases. The false-positive rate only looks acceptable because the mapping filter in `infer()` removes most site-packages modules first. Comparing against purelib/platlib (i.e. "is this outside site-packages") would match the intent and the warning text.

Separately, platstdlib is not considered. It differs from stdlib in venvs and, on some distros, holds lib-dynload, so stdlib C extensions would be misclassified. That's latent rather than active, since `infer()` filters sys.stdlib_module_names first — but is_local_module is module-level, tested on its own, and will be used elsewhere.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+                if requirement not in requirement_to_version:
+                    requirement_to_version[requirement] = importlib.metadata.version(
+                        requirement
+                    )
+        else:
+            # No distribution mapping: Cython/C-extension aliases, runtime entrypoints
+            # (__main__), or local/dev imports outside site-packages.
+            logging.debug(module)
+
+            if is_local_module(module) and (top_level_name not in warned):
```
glemaitre: Apparently the following will raise a bit too much false positive:

```python
Package __main__ seems to be an editable or local install (loaded from outside site-packages)...
Package _virtualenv seems to be an editable or local install...
Package skore seems to be an editable or local install...
```

The two first warning will be problematic. The last one is normal in this case because I was in an editable case. Maybe we should stress test a bit depending on the use case that we will have with different configuration to avoid as much as possible false positive.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+            continue
+
+        if requirements := module_to_requirement.get(top_level_name):
+            for requirement in requirements:
+                if requirement not in requirement_to_version:
+                    requirement_to_version[requirement] = importlib.metadata.version(
+                        requirement
+                    )
+        else:
+            # No distribution mapping: Cython/C-extension aliases, runtime entrypoints
```
glemaitre: I think that you want it to use the `logger` declared above.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+            # (__main__), or local/dev imports outside site-packages.
+            logging.debug(module)
+
+            if is_local_module(module) and (top_level_name not in warned):
+                warned.add(top_level_name)
+                warnings.warn(
```
glemaitre, line 157: Let's use rich if we want to show specific characters.

glemaitre: Otherwise, we can strip it because if it ends up inside the logging, it will be annyoing.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+def infer() -> list[Requirement]:
+    """
+    Infer distribution requirements from modules currently in ``sys.modules``.
+
+    Maps each imported top-level package to its distribution via
+    :func:`importlib.metadata.packages_distributions`, then records the installed
+    version. Unmapped packages loaded from outside the standard library are
+    skipped and a warning is emitted once per package name.
```
glemaitre: I think that we can document that basically what is recorded depends on the current imported modules. It can explain within the same notebook we can end-up with a different solution.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+    stdlib = pathlib.Path(sysconfig.get_path("stdlib")).resolve()
+
+    return not path.is_relative_to(stdlib)
+
+
+class Requirement(typing.TypedDict):
+    """A distribution name and its installed version."""
+
+    name: str
+    version: str
```
glemaitre, line 67: It seems that `None` is not possible.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,97 @@
...
+                        "\033[38;5;208m"
+                        f"Package \033[1;3m{top_level_name}\033[22;23m seems to be an "
+                        "editable or local install (loaded from outside site-packages)."
+                        " It will not be recorded in the inferred requirements."
+                        "\033[0m"
+                    ),
+                    stacklevel=2,
+                )
+
+    return [
```
glemaitre: It seems that we can get something like:

```
Jinja2, MarkupSafe, Pygments, asttokens, certifi, charset-normalizer, cloudpickle, comm,
cycler, decorator, executing, idna, iniconfig, ipython, ipywidgets, jedi, joblib, kiwisolver,
matplotlib, narwhals, numpy, packaging, pandas, parso, pillow, pluggy, polars,
polars-runtime-32, prompt_toolkit, psutil, pure_eval, pyarrow, pyparsing, pytest,
python-dateutil, pytz, requests, rich, scikit-learn, scipy, seaborn, six, skrub, stack-data,
threadpoolctl, traitlets, typing_extensions, urllib3, wcwidth
```

Apparently it is not PEP503 compliant (e.g. `Jinja2, MarkupSafe, Pygments`) even if `pip` is tolerant. Apparently `packaging.utils.canonicalize_name` might help there

skore/tests/unit/plugins/test_requirements.py
```diff
@@ -0,0 +1,193 @@
...
+from types import ModuleType
+
+from pytest import warns
+
+from skore._plugins import requirements as requirements_module
+from skore._plugins.requirements import infer, is_local_module
+
+
+class Module(ModuleType):
+    def __init__(self, name: str, origin: str | None):
```
glemaitre, line 10: Apparently `origin` can never be `None` and we don't run `mypy` in test so it does not complain.

glemaitre (commented):
Here is a first round of review using my friend. It is quite interesting to get in the mood :).

skore/src/skore/_plugins/requirements.py
thomass-dev: Already fixed in https://github.com/probabl-ai/skore/pull/3142/commits/6c1fdd9507794fcb4378ca3d8a21f6c695602628.

skore/src/skore/_plugins/requirements.py
thomass-dev: Already fixed in f19ae82edc284b0c929a28a76479cc221e831dc7.

skore/src/skore/_plugins/requirements.py
thomass-dev: Fixed in 108b96ce9889d41f7ac622b62a80d486e108791b.

skore/tests/unit/plugins/test_requirements.py
thomass-dev: It is used in : 

```python
def test_no_origin(self, tmp_path):                                       
    assert not requirements.is_local_module(Module("_cython_3_2_4", None))
```

thomass-dev: And it's just a shortcut to simulate the real `cython` module:

```python
<module '_cython_3_2_4'>

repr(module.__spec__)
'None'

repr(module.__file__)
*** AttributeError: module '_cython_3_2_5' has no attribute '__file__'.
```

thomass-dev: Also fyi:

```python
class ModuleSpec:                           
    def __init__(                           
        self,                               
        name: str,                          
        loader: importlib.abc.Loader | None,
        *,                                  
        origin: str | None = None,          
        loader_state: Any = None,           
        is_package: bool | None = None,     
    ) -> None: ...                          
```

skore/src/skore/_plugins/requirements.py
thomass-dev: I want it to be a warning rather than a log because we should be able to silence it.
Formatting removed in c7baea4aa79f8c248f8513599b450fcc382274cf.

skore/src/skore/_plugins/requirements.py
thomass-dev: Oh oh good catch, typing mistake! Thanks.

thomass-dev: Fixed in dc05ab2dfb94694e38644c8bd59f462b7da92103.

skore/src/skore/_plugins/requirements.py
thomass-dev: I'm not convinced by the need: the package is installed in the user-environment, so at least one package manager installed it, and package manager are compliant.

But i added in cb5ceed4cf22adce75b9a04664ddaeb89d677236

```python
return [                                                       
    Requirement(                                               
        name=packaging.utils.canonicalize_name(name),          
        version=packaging.utils.canonicalize_version(          
            version,                                           
            strip_trailing_zero=False,                         
        ),                                                     
    )                                                          
    for name, version in sorted(requirement_to_version.items())
]                                                              
```

skore/src/skore/_plugins/requirements.py
thomass-dev: I can't reproduce on my side. `__main__` was filtered-out during my successive iterations by `origin is None`:

```python
origin = (module.__spec__ and module.__spec__.origin) or getattr(
    module, "__file__", None                                     
)                                                                
                                                                 
if (origin is None) or (origin in {"built-in", "frozen"}):       
    return False                                                 
```

Can you ask your agent for reproducer?

glemaitre: This would be the full reproducer:

```python
#!/usr/bin/env python3
"""Minimal reproducer for probabl-ai/skore#3142.

`infer()` silently records an editable install as a normal pinned requirement,
without emitting the "editable or local install" warning, whenever the build
backend writes `top_level.txt` into the `.dist-info` (setuptools does).

The two checks in `infer()` are mutually exclusive: `is_local_module()` is only
consulted in the `else` branch, i.e. only for packages that
`importlib.metadata.packages_distributions()` failed to map. So whether you get
warned depends on the build backend's metadata, not on whether the install is
editable.

Usage:
    python reproduce_3142.py [path/to/_plugins/requirements.py]

Defaults to skore/src/skore/_plugins/requirements.py relative to the cwd.
"""

import pathlib
import shutil
import subprocess
import sys
import tempfile
import venv

PYPROJECT = """\
[build-system]
requires = ["setuptools>=68"]
build-backend = "setuptools.build_meta"

[project]
name = "mypkg"
version = "0.1.0"
"""

PROBE = '''\
import importlib.metadata
import sys
import warnings

sys.path.insert(0, {tmp!r})

import requirements_under_test as R
import mypkg

mapping = importlib.metadata.packages_distributions().get("mypkg")

with warnings.catch_warnings(record=True) as caught:
    warnings.simplefilter("always")
    requirements = R.infer()

messages = [str(w.message) for w in caught]
about_mypkg = [m for m in messages if "mypkg" in m]
recorded = [r for r in requirements if r["name"] == "mypkg"]

print("  mypkg.__file__                       :", mypkg.__file__)
print("  is_local_module(mypkg)               :", R.is_local_module(mypkg))
print("  packages_distributions()['mypkg']    :", mapping)
print("  warnings mentioning mypkg            :", about_mypkg or "NONE")
print("  mypkg recorded in requirements       :", recorded or "NONE")
print()
print("  incidental warnings (unrelated bug)  :")
for m in messages:
    print("    -", m)
print()

editable = R.is_local_module(mypkg)
warned = bool(about_mypkg)
if editable and recorded and not warned:
    print("  RESULT: BUG REPRODUCED")
    print("  An editable install was pinned as mypkg==0.1.0 with no warning,")
    print("  even though is_local_module() correctly identifies it as local.")
    sys.exit(0)

print("  RESULT: not reproduced (behaviour changed)")
sys.exit(1)
'''


def main() -> int:
    if len(sys.argv) > 1:
        source = pathlib.Path(sys.argv[1])
    else:
        source = pathlib.Path("skore/src/skore/_plugins/requirements.py")

    if not source.is_file():
        print(f"error: cannot find {source}", file=sys.stderr)
        print("pass the path to the PR's requirements.py as an argument", file=sys.stderr)
        return 2

    tmp = pathlib.Path(tempfile.mkdtemp(prefix="skore-3142-"))
    try:
        package = tmp / "mypkg"
        (package / "src" / "mypkg").mkdir(parents=True)
        (package / "src" / "mypkg" / "__init__.py").write_text("")
        (package / "pyproject.toml").write_text(PYPROJECT)

        environment = tmp / ".venv"
        print(f"building venv in {environment} ...", flush=True)
        venv.EnvBuilder(with_pip=True).create(environment)
        python = environment / "bin" / "python"

        run = lambda *args, **kw: subprocess.run(  # noqa: E731
            [python, "-m", "pip", "install", "-q", *args], check=True, **kw
        )
        run("setuptools", "packaging")
        print("installing mypkg in editable mode (setuptools backend) ...", flush=True)
        run("-e", ".", cwd=package)

        shutil.copy(source, tmp / "requirements_under_test.py")

        print(f"\nrunning {source} ...\n", flush=True)
        probe = subprocess.run([python, "-c", PROBE.format(tmp=str(tmp))], cwd=tmp)
        return probe.returncode
    finally:
        shutil.rmtree(tmp, ignore_errors=True)


if __name__ == "__main__":
    raise SystemExit(main())
```

```
building venv in /tmp/skore-3142-y586ie8l/.venv ...
installing mypkg in editable mode (setuptools backend) ...

running /tmp/pr_requirements.py ...

  mypkg.__file__                       : /tmp/skore-3142-y586ie8l/mypkg/src/mypkg/__init__.py
  is_local_module(mypkg)               : True
  packages_distributions()['mypkg']    : ['mypkg', 'mypkg']
  warnings mentioning mypkg            : NONE
  mypkg recorded in requirements       : [{'name': 'mypkg', 'version': '0.1.0'}]

  incidental warnings (unrelated bug)  :
    - Package sitecustomize seems to be an editable or local install ...
    - Package requirements_under_test seems to be an editable or local install ...

  RESULT: BUG REPRODUCED
```

The three middle lines are the whole story: the helper says True, the mapping succeeded anyway, so the helper is never consulted and the editable install is pinned as mypkg==0.1.0 in silence. (The two incidental warnings are the separate false-positive problem — sitecustomize here, __main__ and _virtualenv on other setups.)

This would be a unit test variant:

<details>

```python
"""Unit-level regression test for probabl-ai/skore#3142."""

import importlib.metadata
import warnings
from importlib.machinery import ModuleSpec
from types import ModuleType

from skore._plugins import requirements


class Module(ModuleType):
    def __init__(self, name: str, origin: str):
        super().__init__(name)
        self.__file__ = origin
        self.__spec__ = ModuleSpec(name, loader=None, origin=origin)


def test_editable_install_is_recorded_without_warning(tmp_path, monkeypatch):
    origin = str(tmp_path / "mypkg" / "src" / "mypkg" / "__init__.py")
    module = Module("mypkg", origin)

    # setuptools writes top_level.txt for editable installs, so the package is
    # mapped exactly like a regular site-packages install.
    monkeypatch.setattr(
        importlib.metadata, "packages_distributions", lambda: {"mypkg": ["mypkg"]}
    )
    monkeypatch.setattr(importlib.metadata, "version", lambda name: "0.1.0")
    # `warnings` has to stay reachable: replacing sys.modules wholesale would
    # otherwise break catch_warnings. The PR's own tests do the same.
    monkeypatch.setattr(
        requirements.sys, "modules", {"mypkg": module, "warnings": warnings}
    )

    # The helper knows perfectly well that this is a local install...
    assert requirements.is_local_module(module) is True

    with warnings.catch_warnings(record=True) as caught:
        warnings.simplefilter("always")
        result = requirements.infer()

    # ...but infer() never asks, because is_local_module() is only reachable
    # from the `else` branch. The editable install is pinned silently.
    assert result == [{"name": "mypkg", "version": "0.1.0"}]
    assert caught == []
 ```
 
 </details>

File: skore/pyproject.toml [hunk 12]
```diff
@@ -32,6 +32,7 @@ dependencies = [
   "matplotlib",
   "narwhals>=2.0.1",
   "numpy",
+  "packaging",
```
glemaitre: Apparently you need the following lower bound: ≥ 22.0.

It is required for `strip_trailing_zero`

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,103 @@
...
+                warnings.warn(
+                    (
+                        f"Package {top_level_name} seems to be an editable or local "
+                        "install (loaded from outside site-packages). It will not be "
+                        "recorded in the inferred requirements."
+                    ),
+                    stacklevel=2,
+                )
+
+    return [
```
glemaitre: If we want to preserve the previous behaviour, we need to call `sort`.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,103 @@
...
+        name = (module.__spec__ and module.__spec__.name) or module.__name__
+        top_level_name = name.partition(".")[0]
+
+        if top_level_name in sys.stdlib_module_names:
+            continue
+
+        if requirements := module_to_requirement.get(top_level_name):
+            for requirement in requirements:
+                if requirement not in requirement_to_version:
+                    requirement_to_version[requirement] = importlib.metadata.version(
```
glemaitre: We might have a nice caching optimization to do:

`infer()` is expensive enough to notice. In your conda env (237 distributions, 2779 modules in `sys.modules` after building an `EstimatorReport`):

```
first  infer(): 1011 ms
second infer(): 552 ms
```

`packages_distributions()` alone is ~295 ms and it re-scans every `sys.path` entry on each call; `importlib.metadata.version()` × 237 adds ~56 ms. If this runs on every project.put() that's a second per put. Caching the mapping, and reading versions from the Distribution objects the mapping already visited instead of calling `version()` by name, would remove most of it.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,103 @@
...
+    # Snapshot: importlib.metadata/console I/O can load modules and mutate sys.modules
+    # while we iterate.
+    for module in list(sys.modules.values()):
+        # sys.modules values are usually modules, but can be None (cleared/failed
+        # imports) or non-modules (legacy typing.io/re classes) that lack __spec__.
+        if not isinstance(module, types.ModuleType):
+            continue
+
+        name = (module.__spec__ and module.__spec__.name) or module.__name__
+        top_level_name = name.partition(".")[0]
```
glemaitre, line 149: Apparently, here we could do better by just getting the library effectively imported:

4. Namespace packages pull in unrelated distributions. `packages_distributions()` maps one top-level name to every distribution that contributes to it. Importing only `protobuf` records `google-auth` too:

```
google  ->  ['google-auth', 'protobuf']
google-auth recorded?    True
protobuf recorded?       True
is google-auth imported? False
```

Same for `sphinxcontrib`, `ruamel`, `zope`, `backports`. The fix is to look up the specific submodule's distribution rather than the top-level name, or at minimum to record only the distribution that actually owns the imported file.

--

I'm not sure if it is the behaviour that we want actually.

skore/src/skore/_plugins/requirements.py
```diff
@@ -0,0 +1,103 @@
...
+class Requirement(typing.TypedDict):
+    """A distribution name and its installed version."""
+
+    name: str
+    version: str
+
+
+def infer() -> list[Requirement]:
+    """
+    Infer distribution requirements from modules currently in ``sys.modules``.
```
Aljutor, line 123: Does this assume that all scorer dependencies are imported before `put()`?
Since, `metrics.add()` only registers the scorer. The scorer can run later when we computes metrics.

thomass-dev: > Does this assume that all scorer dependencies are imported before put()?

Not exactly, it assumes that all dependencies are imported before the call of this function.
So as long as we compute the metrics first, everything will be fine.

skore/src/skore/_plugins/requirements.py
thomass-dev: Fixed in 5b352bbba5c9109a9f53854f3b4f73b9ce6c1f59.

skore/pyproject.toml
thomass-dev: Fixed in 2d4f955a2ccf7f30e5a533fa9c17d8bcaecb6edc.

skore/src/skore/_plugins/requirements.py
thomass-dev: Fixed in fb6084c835df69b13ea79d9c62e69864b8d3be92.

skore/src/skore/_plugins/requirements.py
thomass-dev: Fixed in 264bec5a8ffd8cb5bc6f07413ed05fe6eb359e5a.

skore/src/skore/_plugins/requirements.py
thomass-dev: Can't reproduce: `RESULT: not reproduced (behaviour changed)`.

skore/src/skore/_plugins/requirements.py
thomass-dev: Fixed in bd019b53.

skore/src/skore/_plugins/requirements.py
thomass-dev: My initial intention was to voluntary include all the namespace packages in this first iteration.
Inspecting files in distribution has a non-negligible cost...

I updated in c480065180865503edb258c48a6a53c222f65677.

glemaitre (approved):
LGTM for a first version. We will see if we find some caveats that we did not think of.

PR #3145, opened 2026-07-15
docs: Add SKD003 inconsistent performance pitfall example

moujanrastgoo:
#### Change description
Related to #2622 and #3134 
- Adds SKD003 check exmaple and links from automated checks user guide

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:


examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,214 @@
...
+
+This example demonstrates mitigations when check
+:ref:`SKD003 <skd003-inconsistent-performance>` fires on a
+:class:`~skore.CrossValidationReport`. The check compares per-fold test scores
+and flags splits whose metrics diverge sharply from the median — often a sign
+of a bad data batch, leakage, or an unlucky fold draw.
+
+Mitigations from the :ref:`automated_checks` user guide:
+
+- use stratified or grouped cross-validation,
```
glemaitre: Stratification is usually never a good solution. Usually, it is advised but it is to solve an engineering problem: your predictor might not be trained on a rare class and then it will fail during predict. This should be solve otherwise in scikit-learn.

Stratification might have an effect that reduce the variance observe during the cross-validation process. So I would avoid to mention it here.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,214 @@
...
+cross-validation scores once the root cause is addressed.
+"""
+
+# %%
+# Load Covertype (two classes) and corrupt labels
+# ===============================================
+#
+# The Covertype task predicts forest cover type from cartographic variables.
+# Classes 1 (Spruce/Fir) and 2 (Lodgepole Pine) make up about 85 % of the
+# table. We keep only those rows, draw a 2,000-row stratified subsample, then
```
glemaitre: I don't necessarily think that we need to force stratification. I think that here the changes that we could have due to the uniform sampling might representative from a new natural acquisition process with some variation.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,214 @@
...
+# ===================================================
+#
+# Five-fold cross-validation shuffles rows before splitting, so the corrupted
+# prefix is spread across folds — but fold #0 still sees enough bad labels to
+# drag its test scores away from the others. Inspect per-split metrics with
+# ``aggregate=None``, then summarize checks.
+
+from skore import evaluate
+
+report = evaluate(model, X=X, y=y_corrupted, splitter=5, n_jobs=-1)
```
glemaitre: Let's limit it.

```suggestion
report = evaluate(model, X=X, y=y_corrupted, splitter=5, n_jobs=4)
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,214 @@
...
+
+TableReport(y_clean)
+
+# %%
+y_corrupted = y_clean.copy()
+n_corrupt = len(y_corrupted) // 5
+y_corrupted.iloc[:n_corrupt] = np.random.RandomState(0).choice(
+    y_clean.unique(),
+    size=n_corrupt,
+)
```
glemaitre: You need to explain the process of corrupting the data here and what we intend to have with it. Also, we can reconnect on how it could happen on real use-case.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,214 @@
...
+
+# %%
+# What to check when the cause is unknown? SKD003 does not distinguish
+# label errors from leakage or temporal drift. If one fold still looks wrong
+# after auditing labels, review the feature pipeline for target leakage (e.g.
+# statistics fit on the full table before splitting, duplicate rows across
+# folds, features that encode future information) and, when rows are ordered in
+# time, whether cross-validation respects chronology — see
+# :ref:`SKD013 <skd013-train-test-time-overlap>`.
+# With clean labels, per-fold scores should cluster and SKD003 should be absent.
```
glemaitre: I think that we could further explain that one can get the corrupted `EstimatorReport` store and check the internal metric, plot, importances, etc. to debug. I think a small one liner to show how to access and show one plot could be nice.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,214 @@
...
+
+report_clean.checks.summarize(fast_mode=True, ignore=["SKD008"])
+
+# %%
+# Use stratified cross-validation
```
glemaitre: I would rather have an example with `groups` where we would get potential variance if the split are not done properly.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,249 @@
...
+
+from skrub import tabular_pipeline
+
+model = tabular_pipeline("classifier")
+
+# %%
+# Trigger SKD003 — cross-validate on corrupted labels
+# ===================================================
+#
+# Five-fold cross-validation shuffles rows before splitting, so the corrupted prefix is spread
```
GaetandeCast: This is false. In skore, `splitter=k` creates a `KFold` with `shuffle=False`, which is why only the first fold gets bad scores. The other are unaffected.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,249 @@
...
+report = evaluate(model, X=X, y=y_corrupted, splitter=5, n_jobs=4)
+
+report.metrics.summarize(data_source="test").frame(aggregate=None)
+
+# %%
+# SKD003 should list split #0 as an outlier relative to the other folds. We ignore SKD008 here: its
+# correlated-features check is unrelated to this walkthrough and can warn on Covertype's constant
+# one-hot soil columns.
+
+report.checks.summarize(fast_mode=True, ignore=["SKD008"])
```
GaetandeCast: I would not use `ingore` here and just mention that we focus only on `SDK003` for this example.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,249 @@
...
+
+report_clean = evaluate(model, X=X, y=y_clean, splitter=5, n_jobs=4)
+
+report_clean.metrics.summarize(data_source="test").frame(aggregate=None)
+
+# %%
+# What to check when the cause is unknown? SKD003 does not distinguish label errors from leakage or
+# temporal drift. If one fold still looks wrong after auditing labels, review the feature pipeline
+# for target leakage (e.g. statistics fit on the full table before splitting, duplicate rows across
+# folds, features that encode future information) and, when rows are ordered in time, whether
```
GaetandeCast: I don't really see how target leakage may induce unstable split performance. I think this paragraph does not bring much valuable information and can be removed.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,249 @@
+"""
+.. _example_skd003_inconsistent_performance:
+
+SKD003 — Inconsistent performance across splits
+===============================================
```
GaetandeCast: To me, the only realistic reason this check can fire is when either a fold is corrupted (which is good for demonstration puposes but not so realistic I think) or when there is structure in the data (e.g. temporal or groups).

Two concrete examples of that second case:
- groups: one group is way easier or way harder to predict than the others (either it's really similar to a group in the train set or it's really different to all groups in the train set). When that group is in the test set (with e.g. `GroupKFold`), the fold will have outlier performance.
- temporal structure: data points are time ordered events and the distribution of the target changes over time. For instance a class that was initially common and easy to predict become rarer. A model evaluated with a proper time series split will see it's performance on the earlier folds be better than on the later folds as it learns to predict the easy class which become gradually less present in the test sets.

Doing shuffled cv in these cases is a mistake: it hides the structure and over estimates the model's performance. So SKD003 should trigger when the splitting is done properly.

This leads me to thinking that SKD003 appearing is a 'good sign' as the splitting was done properly but is not really solvable as there is nothing to mitigate the structure of the dataset.

I think we need to reformulate the narrative around this check (user guide and example) to take this into account.

WDYT @glemaitre ?

GaetandeCast: I think the structure should then be: 

- artificial corruption
- bad group in test
- distribution shift in last time series fold
- advertise ignoring the check with the config once the problem is well understood. Maybe also suggest collecting more data either on similar groups as the "outlier" one or in the future after drift happened.

GaetandeCast (commented):
Here are my general thoughts on this check and a few other minor remarks

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
moujanrastgoo: oh i see thanks i will fix it!

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
moujanrastgoo: I put ignore here because apparently SKD008 was causing some warnings, so I wanted to not show those, but sure i can remove it !

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
moujanrastgoo: sure!

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
GaetandeCast: An other thing that may cause SKD003 to fire: the two classes of a binary classification problem are imbalanced and happen to not be homogeneously distributed across the dataset because of e.g. how the data was collected (iid data that is in collection order but a sensor broke so one part of the dataset a much more of the minority class). That's a case where there is no internal structure in the data but a default (unshuffled) `KFold` will create imbalanced folds. In this case it is better to shuffle (or stratify :)) to smooth out the irregularities, if we are sure that the distribution irregularities are accidental and won't appear in production.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+When structure is real, shuffled CV overestimates performance. SKD003 under a
+proper split is a good sign. The structure may not be fully fixable: once
+understood, mute with :func:`~skore.configuration` and consider collecting more
+data on the hard regime.
+
+This notebook walks four beats: artificial corruption, a bad group in test,
+distribution shift in the last time-series fold, then ignoring SKD003 once the
+problem is understood. We focus on SKD003 throughout and pass
+``ignore=["SKD008"]`` so Covertype's constant one-hot soil columns do not
+drown the summary in correlated-feature noise.
```
GaetandeCast: If we are not exploiting anything specific from this dataset (I think we don't but correct me if I'm wrong), let's just switch to a dataset that has no constant columns to avoid that concern altogether.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+# ==================================================
+#
+# ``reports_`` holds one :class:`~skore.EstimatorReport` per split; open the
+# flagged fold as usual.
+
+outlier_report = report.reports_[0]
+outlier_report.metrics.summarize(data_source="test").frame()
+
+# %%
+outlier_report.metrics.precision_recall().plot()
```
GaetandeCast: I think people are more used to roc plots so I'd rather use one here. It is easy to show that something is wrong on a roc plot since the plot will follow the diagonal "chance level" line.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+# metrics with ``aggregate=None``, then checks.
+
+import skore
+
+report = skore.evaluate(model, X=X, y=y_corrupted, splitter=5, n_jobs=4)
+
+report.metrics.summarize(data_source="test").frame(aggregate=None)
+
+# %%
+# ``SKD003`` should flag split #0.
```
GaetandeCast:
```suggestion
# ``SKD003`` correctly flags split #0.
```
Here and for other occurences.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+hard_start = n_time - (n_time // (n_splits + 1))
+
+X_arr, y_arr = make_classification(
+    n_samples=n_time,
+    n_features=6,
+    n_informative=2,
+    n_redundant=0,
+    weights=[0.5, 0.5],
+    random_state=0,
+)
```
GaetandeCast: I would rather we add a fake datetime column to the same dataset used in the previous sections for consistency, just like how we created fake groups, and then change the distribution in the last temporal group.

GaetandeCast (commented):
Some remarks

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
+"""
+.. _example_skd003_inconsistent_performance:
+
+SKD003 - Inconsistent performance across splits
+===============================================
+
+:ref:`SKD003 <skd003-inconsistent-performance>` flags folds whose test metrics
+diverge sharply from the median on a :class:`~skore.CrossValidationReport`.
```
glemaitre: We should focus on the stats words instead of the Python object.

```suggestion
diverge sharply from the median during a cross-validation evaluation.
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+(groups, time, or a corrupted batch) that shuffled CV would hide.
+
+Realistic triggers:
+
+- a contiguous bad batch of labels or features (mislabelled window, logging
+  bug, schema mix-up),
+- a much easier or harder group in one test fold under
+  :class:`~sklearn.model_selection.GroupKFold`,
+- temporal drift under :class:`~sklearn.model_selection.TimeSeriesSplit`
+  (e.g. an easy class that becomes rarer),
```
glemaitre: I would change the example:


```suggestion
  (e.g. fraudsters finding workarounds to circumvent controls),
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
+"""
+.. _example_skd003_inconsistent_performance:
+
+SKD003 - Inconsistent performance across splits
+===============================================
+
+:ref:`SKD003 <skd003-inconsistent-performance>` flags folds whose test metrics
+diverge sharply from the median on a :class:`~skore.CrossValidationReport`.
+With a proper splitter this is often a diagnostic check: the data have structure
+(groups, time, or a corrupted batch) that shuffled CV would hide.
```
glemaitre: Let's use whenever possible "cross-validation" instead of "CV"

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+drown the summary in correlated-feature noise.
+"""
+
+# %%
+# Load Covertype (two classes)
+# ============================
+#
+# Classes 1 and 2 dominate Covertype. We keep those rows and take a 2,000-row
+# subsample. SKD003 needs CV, so we use ``splitter=5`` or an explicit five-fold
+# splitter below.
```
glemaitre: To early to speak about the cross-validation. Let's only discuss it when we call `skore.evaluate`.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+TableReport(y_clean)
+
+# %%
+# Artificial corruption: a bad contiguous batch
+# =============================================
+#
+# Randomly reassign labels on the first fifth of rows (mislabelled batch,
+# logging bug, or merge mix-up). ``splitter=5`` is unshuffled
+# :class:`~sklearn.model_selection.KFold`, so that block stays in fold #0 and
+# SKD003 can flag it.
```
glemaitre:
```suggestion
# Below, we will use an unshuffled 5-fold cross-validation. Here, we permute
# labels on the first fold to break any association between `X` and `y`. We expect
# the score of this fold to be low due to this corruption.
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+
+y_corrupted = y_clean.copy()
```
glemaitre: So here we should use the new random state generator from numpy and I think that link to the change in the narrative above we can use `permutation`: https://numpy.org/doc/2.1/reference/random/generated/numpy.random.Generator.permutation.html#numpy.random.Generator.permutation

```python
y_corrupted, n_corrupt = y_clean.copy(), len(y_corrupted) // 5
rng = np.random.default_rng(seed=0)
y_corrupted.iloc[:n_corrupt] = rng.permutation(y_corrupted.iloc[:n_corrupt])
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+    y_clean.unique(),
+    size=n_corrupt,
+)
+
+# %%
+# Default :func:`~skrub.tabular_pipeline` classifier for preprocessing.
+
+from skrub import tabular_pipeline
+
+model = tabular_pipeline("classifier")
```
glemaitre:
```suggestion
model = tabular_pipeline("classifier")
model
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
moujanrastgoo: sure i will look into that!

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
moujanrastgoo: ok sure!

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
moujanrastgoo: sure!

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+
+# %%
+# Trigger SKD003: cross-validate on corrupted labels
+# ==================================================
+#
+# Corrupted rows land in split #0; later folds stay clean. Inspect per-split
+# metrics with ``aggregate=None``, then checks.
```
glemaitre: Can you review a bit the style of the narrative. We are using a lot of imperative tense here. I would prefer a less robotic narrative ;)

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+# ``SKD003`` should flag split #0.
+
+report.checks.summarize(fast_mode=True, ignore=["SKD008"])
+
+# %%
+# Debug the flagged fold through its EstimatorReport
+# ==================================================
+#
+# ``reports_`` holds one :class:`~skore.EstimatorReport` per split; open the
+# flagged fold as usual.
```
glemaitre:
```suggestion
# One can access to an underlying :class:`~skore.EstimatorReport` that contains
# all the information for the given split.
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
glemaitre: Indeed. Can you also catch the output:

```python
_ = outlier_report.metrics.roc().plot()
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+report_clean.metrics.summarize(data_source="test").frame(aggregate=None)
+
+# %%
+report_clean.checks.summarize(fast_mode=True, ignore=["SKD008"])
+
+# %%
+# Bad group in the test fold
+# ==========================
+#
+# With group ids (site, patient, batch), a grouped splitter keeps each group on
```
glemaitre: can you make a full descriptive sentence for the example of batch: "e.g. the medical center where patient data are collected"

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+    index=X.index,
+    name="batch_id",
+)
+
+y_batch = y_clean.copy()
+bad_batch = groups == 0
+y_batch.loc[bad_batch] = np.random.RandomState(1).choice(
+    y_clean.unique(),
+    size=int(bad_batch.sum()),
+)
```
glemaitre: use again an new `default_rng`

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+
+group_splits = list(GroupKFold(n_splits=5).split(X, y_batch, groups=groups))
```
glemaitre: We should not have to create a list of index here. But because it is an advance use case we need to use a skrub DataOp. We probably need to have some narrative to explain that to the user by:

1. generating the data
2. explaining the creation of the skrub data op and for which reason
3. make the evaluation and explain the problem at hand

The implementation of the skrub DataOp here will look something like:

```python
# %%
# Bad group in the test fold
# ==========================
#
# With group ids (site, patient, batch), a grouped splitter keeps each group on
# one side of every split. A much harder or easier group in one test fold
# triggers SKD003: expected, the splitter did its job. Shuffled CV would smear
# that group across folds, hide the gap, and overestimate performance.
#
# Below we invent batch ids and corrupt labels for batch ``0`` only. Instead of
# materializing a list of train/test indices, we keep features, target, and
# group ids in a single dataframe: the group ids then follow the rows through
# every split.

n_batches = 10
batch_id = np.minimum(np.arange(len(X)) // (len(X) // n_batches), n_batches - 1)

y_batch = y_clean.copy()
bad_batch = batch_id == 0
y_batch.iloc[bad_batch] = np.random.RandomState(1).choice(
    y_clean.unique(),
    size=int(bad_batch.sum()),
)

df_batch = X.assign(batch_id=batch_id, cover_type=y_batch)

# %%
# A skrub :class:`~skrub.DataOp` declares the split on the data itself:
# :meth:`~skrub.DataOp.skb.mark_as_X` takes the splitter through ``cv`` and the
# group ids through ``split_kwargs``. The resulting learner carries its own
# cross-validation scheme, so :func:`~skore.evaluate` needs no ``splitter``.

import skrub
from sklearn.model_selection import GroupKFold

data = skrub.var("data", df_batch)
groups = data["batch_id"]
X_op = data.drop(columns=["batch_id", "cover_type"]).skb.mark_as_X(
    cv=GroupKFold(n_splits=5),
    split_kwargs={"groups": groups},
)
y_op = data["cover_type"].skb.mark_as_y()
learner = X_op.skb.apply(model, y=y_op).skb.make_learner()

# %%
report_grouped = skore.evaluate(learner, data={"data": df_batch}, n_jobs=4)
report_grouped.metrics.summarize(data_source="test").frame(aggregate=None)
```

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+# ``SKD003`` should flag the fold that tests the corrupted batch.
+
+report_grouped.checks.summarize(fast_mode=True, ignore=["SKD008"])
+
+# %%
+# Distribution shift in the last time-series fold
+# ===============================================
+#
+# Under :class:`~sklearn.model_selection.TimeSeriesSplit`, later windows can
+# diverge from early training data (e.g. an easy class becomes rarer). Earlier
```
glemaitre: I prefer the frauds cases.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+# folds look strong, the last fold drops, and SKD003 fires.
+#
+# Build a balanced task with :func:`~sklearn.datasets.make_classification`,
+# treat row order as time, and replace labels only in the final test window
+# with a rare positive class so that one late fold stands out. No shuffle.
```
glemaitre: Same remarks about the imperative tense :)

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+    n_samples=n_time,
+    n_features=6,
+    n_informative=2,
+    n_redundant=0,
+    weights=[0.5, 0.5],
+    random_state=0,
+)
+X_time = pd.DataFrame(X_arr, columns=[f"f{i}" for i in range(6)])
+y_time = pd.Series(y_arr, name="label")
+y_time.iloc[hard_start:] = rng.choice([0, 1], size=n_time - hard_start, p=[0.96, 0.04])
```
glemaitre: Can you separate the data generative process from the evaluation.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
```diff
@@ -0,0 +1,256 @@
...
+
+time_splits = list(TimeSeriesSplit(n_splits=n_splits).split(X_time))
```
glemaitre: You should not have to materialize as a list. Pass directly the splitter:

```python
time_splitter = TimeSeriesSplit(n_splits=n_splits)
skore.evaluate(..., splitter=time_splitter, ...)
```

glemaitre: I'm thinking that since we introduce a skrub DataOp earlier, we could actually create a timestamp as index, and create our own `TimeSeriesSplit` that is based on the time. Then, in the data generative process, we could make that for instance the last fold that will be the year "2026", drift compared to previous year.

I find that showing the dataframe and the time and potentially making associated plot will be easier to read.

glemaitre (commented):
Here are a couple of comments.

examples/pitfalls_and_solutions/plot_skd003_inconsistent_performance.py
glemaitre: For the dataset, I think that the bike sharing (e.g. https://scikit-learn.org/stable/auto_examples/applications/plot_time_series_lagged_features.html) has this kind of drift: it is a 2 year collection data and the second year as more demand than the first one. Therefore, learning the trend on the first year will not help to compensate the trend of the second year.

GaetandeCast:
@glemaitre I've addressed our reviews. I have not added fraud detection as you suggested in order to keep the same dataset along the example. Instead for the time splitting part, I added a "more ill patients show up" narrative to give an example of distribution shift while keeping the same breast cancer dataset. LMK what you think.

glemaitre (approved):

Issue #3137, opened 2026-07-09
Store ID defined by skore-lib `report.id` in projects


thomass-dev:
Currently, when calling `project.summarize`, each report comes with an ID (referenced as `project_report_id` below) that is the UUID defined by the project, not the one defined in the skore-library. This ID has no meaning when changing from one mode to another.

To avoid copying a report multiple times, we need to identify if it's already put in the project.
To do that, `project_report_id` is not sufficient, because it is only relevant in the context of a specific mode : one project_report_id can have an existence in the hub, but not in local (and vice-versa).

So, the proposition is to add a new attribute in the `report` payload used during the put in all mode.
This attribute must refer to the `report.id` that is a UUID generated by the skore-lib.
This attribute must be served in `project.summarize` and must not conflict with the current ID. It must be an addition, not a replacement.

----

<ins>To do</ins>:

- [ ] hub
    - [ ] Send report.id to the hub
    - [ ] Store report.id in the hub
    - [ ] Get report.id from the hub in the current APIs
    - [ ] Expose report.id in project.summarize
- [ ] local
    - [ ] Send report.id to the local
    - [ ] Store report.id in the local
    - [ ] Get report.id from the local in the current APIs
    - [ ] Expose report.id in project.summarize
- [ ] mlflow
    - [ ] Send report.id to the local
    - [ ] Store report.id in the local
    - [ ] Get report.id from the local in the current APIs
    - [ ] Expose report.id in project.summarize

Aljutor:
@thomass-dev  
Should we update the skore hub side to store report_id? 
To avoid downloading the whole report just ot check it's ID

Current POC is fetching whole report to check 
https://github.com/probabl-ai/skore/pull/3073


```
def _canonical_id(project: Project, ref: _ReportRef) -> str:
    """Return the report's stable ``report.id``.

    For local projects the summarize id already is the canonical id; for remote
    projects the report must be fetched.
    """
    if project.mode == "local":
        return ref.id
    return str(project.get(ref.id).id)
```

thomass-dev:
> Should we update the skore hub side to store report_id?

Yes, definitely.

This is the task `Store report.id in the hub`.
You should a new metadata in the dedicated hub table.

Aljutor:
> Yes, definitely.
> 
> This is the task Store report.id in the hub.
> You should a new metadata in the dedicated hub table.

Thanks for clarification

thomass-dev:
[...]

Issue #3136, opened 2026-07-09
feat: Synchronize local/hub/mlflow projects


glemaitre:
Build upon this PoC: https://github.com/probabl-ai/skore/pull/3073

## Context

A user can store locally report into a project and would like to synchronize them at once into another `mode` (e.g. `hub` or `mlflow`). This experience could be extended to whatever direction between the different modes.

We don't need to handle conflict within the same report because we can consider at this time that a report is immutable. However, a user can try to synchronize the same report several time and we need to ensure that this operation does not copy the same reports  several time (no-op) by re-conciliation via the ID defined by skore-lib (i.e. `report.id`) between those reports.

Since that the synchronization is a `put` operation, the report date (i.e. timestamp of the report creation) will differ from the date of the synchronization of the report. We must not try to align those dates for now.

## API design

- add a `sync` method (the name might changed at the review) of a `skore.Project` with a parameter `other` that can be a `skore.Project` mode first (see the list of subtask to have minimal diff). If the project does not exist, we need to create the instance to create the project remotely on the hub.
- add a `direction` parameter with `"from"`, `"to"`, `"both"` options.
- add a `dry_run` parameter to see the execution plan without execution.

## Subtasks

- [x] https://github.com/probabl-ai/skore/issues/3137
- [x] create the skore lib API to sync between local <-> hub
- [x] create the skore lib API to sync between local <-> mlflow
- [x] create the skore lib API to sync between hub <-> mlflow
- [x] make that `other` can also be a `skore.Project` instance
- [x] add a new entry point for the skore CLI `skore sync`

thomass-dev:
Assignee : @Aljutor 
Reviewiers : @glemaitre @thomass-dev

Aljutor:
Should this PR https://github.com/probabl-ai/skore/pull/2905 be finalized in some capacity first?

glemaitre:
> Should this PR https://github.com/probabl-ai/skore/pull/2905 be finalized in some capacity first?

Nop it is orthogonal. The layout of the local storage is hidden behind the `.put` and we will just manipulate at the level of the abstraction. So we will only have limited conflicts if any.

glemaitre:
@Aljutor can you implement the two remaining features?

Aljutor:
> [@Aljutor](https://github.com/Aljutor) can you implement the two remaining features?

[...]

PR #3146, opened 2026-07-15
feat(project): Propagate canonical report IDs

Aljutor:
#### What

Implements the first library-side step for canonical cross-backend report IDs from #3137 and #3136.

A report now has a canonical `UUID` identity:

- Existing integer UUIDv4 IDs are converted losslessly when reports are read.
- New reports use UUIDv7. Python <3.14 uses the `uuid6` backport; Python >=3.14 uses the standard library.
- `_hash` remains available as `report.id.int`.
- IDs are serialized as strings at backend boundaries.

Project summaries expose this identity as `report_id`, while the existing `id` column remains the backend locator accepted by `project.get()`.

#### Backend adapters

- Hub sends and reads `report_id`, matching the nullable UUIDv4/UUIDv7 field from skore-hub#1665.
- MLflow stores the value in the `skore_report_id` tag and exposes it as `report_id`.
- Local storage keeps the physical `artifact_id` separate from canonical `report_id`. Legacy local metadata without `report_id` remains compatible.

Legacy Hub rows and MLflow runs may expose `report_id=None` until backfilled.

#### Compatibility

- Directly unpickled legacy reports are migrated through `__setstate__`.
- State deserialization migrates legacy integer IDs through `from_dict()`.
- Only RFC-compatible UUIDv4 and UUIDv7 IDs are accepted.

This PR does not introduce put idempotency, revisions, duplicate folding, or synchronization semantics.

#### Validation

- [x] 215 targeted unit tests pass
- [x] Ruff and formatting checks pass
- [ ] Full test suite
- [ ] Documentation changes

#### Contribution checklist

- [ ] Documentation was added or updated
- [x] The code passes style conventions
- [ ] The full test suite was run
- [ ] The documentation builds and renders properly
- [x] All commits in the PR are signed
- [x] The pull request title respects Conventional Commits

#### AI usage disclosure

AI tools were involved for code generation, test generation, and research.


thomass-dev:
It will be drafted until probabl-ai/skore-hub#1665 is merged.

skore/src/skore/_plugins/hub/report/report.py
```diff
@@ -58,6 +58,12 @@ class ReportPayload(BaseModel, ABC, Generic[Report]):
     report: Report = Field(repr=False, exclude=True)
     key: str
 
+    @computed_field  # type: ignore[prop-decorator]
+    @property
+    def report_id(self) -> str:
+        """The canonical skore report ID."""
+        return str(self.report.id)
```
thomass-dev: The id should be serialize in int: you reconstruct the UUID on the hub side using `UUID(int=int(value))`.

Aljutor: We can handle both formats
https://github.com/probabl-ai/skore-hub/blob/a6e6c746664d63aca1ab7a6e7d591bf1f2131908/hub/src/hub/projects/domain/typings.py#L35

And I would prefer sending UUID as a str because it's more readable and JSON-safe (128 int can be a problem)

skore/src/skore/_utils/_uuid.py
```diff
@@ -0,0 +1,30 @@
...
+
+if sys.version_info >= (3, 14):
+    from uuid import uuid7 as _generate_uuid7
+else:
+    from uuid6 import uuid7 as _generate_uuid7
+
+
+def uuid7() -> UUID:
+    """Generate a UUIDv7 as a standard-library UUID object."""
+    return UUID(str(_generate_uuid7()))
```
thomass-dev:
```suggestion
if sys.version_info < (3, 14):
    from uuid6 import uuid7
else:
    from uuid import uuid7
```

`uuid6.UUID` is already a subclass of `uuid.UUID`.

skore/src/skore/_plugins/local/project.py
```diff
@@ -130,6 +130,8 @@ def _write_metadata(
     name: str | None = None,
 ) -> None:
     metadata = report._metadata | {
+        "id": str(report.id),
```
thomass-dev: It's a bit unintuitive: `id` is already stored in the metadata, but since you want to store it as an `UUID`, you need here to cast it and overload the value from metadata. Why do we need to store `report.id` as an `UUID` and not as an `str`? It could avoid many casts in many places.

User doesn't need to manage `UUID` rather than `str` (and `UUID` is not natively JSON serializable).

Please store id directly as an `str` in the report metadata.

skore/src/skore/_project/types.py
```diff
@@ -8,6 +8,7 @@ class ReportMetadata(TypedDict):
     """Metadata and metrics for a single persisted report."""
 
     id: str
+    report_id: str | None
     key: str
```
thomass-dev, line 14: Please add a comment here something like: `id` and `report_id` can differ, the first can be the id from the backend, the second is always the report id. Also, mention that `report_id` is only needed for synchronisation, to avoid deserialising the report.

thomass-dev (changes requested):

skore/src/skore/_sklearn/_base.py
```diff
@@ -155,9 +162,26 @@ def __init__(self) -> None:
         }
 
     @property
-    def id(self) -> int:
+    def id(self) -> UUID:
         return self._metadata["id"]
 
+    @property
+    def _hash(self) -> int:
+        return self.id.int
```
thomass-dev: Dead code?

Aljutor: Yes, leaked with rebase

skore/src/skore/_plugins/local/project.py
Aljutor: Idea was to use more concrete types, but you are right.

Aljutor:
@thomass-dev do we have any blockers right now?

thomass-dev (approved):

PR #3147, opened 2026-07-16
docs: Add SKD005 underrepresented classes pitfall example

moujanrastgoo:
#### Change description
Related to #2622  and #3134  
- Adds SKD005 check exmaple and links from automated checks user guide

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:


moujanrastgoo:
i'm working on removing the warnings and also on improving the last part

examples/pitfalls_and_solutions/plot_skd005_underrepresented_classes.py
```diff
@@ -0,0 +1,294 @@
...
+:ref:`SKD005 <skd005-underrepresented-classes>` fires on a multiclass task.
+The check flags classes that each represent less than 10 % of rows — a pattern
+where overall accuracy can mask near-zero recall on rare labels.
+
+Mitigations from the :ref:`automated_checks` user guide:
+
+- use per-class metrics (precision, recall, F1 per class),
+- resample the dataset,
+- use class weights in the estimator,
+- collect more data for the underrepresented classes if possible.
```
glemaitre: Here, we should follow the what we advocate in the notebook here: https://probabl-ai.github.io/calibration-cost-sensitive-learning/content/notebooks/imbalanced_classification.html

From what is written:

- do not use F1: it is an average between precision and recall and usually it is never a good choice by default
- precision and recall refer to thresholded metrics and in the imbalanced case, you probably want to look at the estimate of the probability because rare classes will lead to rare probability so one should probably look at metric related to assess how good those estimate are
- do not use class weight. Instead post tune the classifier once you are sure it is calibrated to optimize the thresholded metric of your choice.
- collecting more data is good, but only from underrepresented might change your distribution (the prevalence will be different from the general population) and thus there is a need to correct for this bias to have a properly calibrated learner before to take decision.

glemaitre (commented):
So here, I don't think that I'll go into details in the section because what you got is what the literature is advertising but it is rather wrong. I think that the notebook that I forwarded can just correct and make the narrative towards what we should be advocating for.

examples/pitfalls_and_solutions/plot_skd005_underrepresented_classes.py
```diff
@@ -0,0 +1,210 @@
...
+SKD005 - Underrepresented classes
+=================================
+
+:ref:`SKD005 <skd005-underrepresented-classes>` flags a multiclass task when
+one or more classes each represent less than 10 % of rows. Overall accuracy can
+look acceptable while rare labels are barely learned. This notebook is about
+how to work with that rarity once the check fires: we do not try to make SKD005
+disappear by reshaping the class histogram.
+
+What to do instead (see also :ref:`automated_checks`):
```
GaetandeCast:
```suggestion
What to do instead:
```

examples/pitfalls_and_solutions/plot_skd005_underrepresented_classes.py
```diff
@@ -0,0 +1,210 @@
...
+multiclass model honestly, then see what extra rare-class rows can do.
+"""
+
+# %%
+# Load the Covertype dataset
+# ==========================
+#
+# The full Covertype task has seven forest types. A small stratified subsample
+# keeps frequent classes well represented while types 3-7 drop below 10 % each.
+# We keep the unused rows as a pool for the "more rare-class data" beat later.
```
GaetandeCast:
```suggestion
# We keep the unused rows as a pool for the "more rare-class data" section later.
```

examples/pitfalls_and_solutions/plot_skd005_underrepresented_classes.py
```diff
@@ -0,0 +1,210 @@
...
+
```
GaetandeCast, line 38:
```suggestion
# Let us first import the data, split it and inspect the distribution of classes,
# both in absolute count and relative frequencies.

```

examples/pitfalls_and_solutions/plot_skd005_underrepresented_classes.py
```diff
@@ -0,0 +1,210 @@
...
+# %%
+# Let us inspect the feature matrix with :class:`~skrub.TableReport`.
+
+from skrub import TableReport
+
+TableReport(X)
+
+# %%
+# The class histogram shows that the classes are not evenly distributed and
+# types 3-7 will trigger SKD005.
```
GaetandeCast:
```suggestion
# Clicking the column in the target's `TableReport` brings
# a class histogram that shows that the classes are not evenly distributed.
```

examples/pitfalls_and_solutions/plot_skd005_underrepresented_classes.py
```diff
@@ -0,0 +1,210 @@
...
+print("Rare labels added from the pool:", list(rare_labels))
+print("Extra rare-class rows added:", len(y_rare_extra))
+print("Train size (baseline):", len(y_train))
+print("Train size (more rare-class rows):", len(y_train_more))
+print("\nBaseline train counts:")
+print(y_train.value_counts().sort_index())
+print("\nEnriched train counts:")
+print(y_train_more.value_counts().sort_index())
+
+# %%
```
GaetandeCast, line 160:
```suggestion
# %%
# Let us now fit a model on the enriched train set and compare the results with the
# original model. We can observe that the model on the enriched train set has a better
# log-loss and accuracies on the common test set.

```

examples/pitfalls_and_solutions/plot_skd005_underrepresented_classes.py
```diff
@@ -0,0 +1,210 @@
...
+skore.compare(
+    {
+        "baseline_train": report_less,
+        "more_rare_class_rows": report_more,
+    }
+).metrics.summarize(
+    metric=["accuracy", "precision", "recall", "log_loss"],
+    data_source="test",
+).frame()
+
```
GaetandeCast, line 187:
```suggestion

# %%
# Enriching the train set with more rare-class rows also clears SKD005.

report_more.checks.summarize(fast_mode=True, ignore=["SKD008"])

```

glemaitre (approved):

PR #3148, opened 2026-07-16
docs: Add SKD006 coefficient interpretation pitfall example

moujanrastgoo:
#### Change description
Related to #2622   and #3134 
- Adds SKD006 check exmaple and links from automated checks user guide

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:


examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+.. _example_skd006_coefficient_interpretation:
+
+SKD006 — Coefficient interpretation
+===================================
+
+This example walks through mitigations when check
+:ref:`SKD006 <skd006-unscaled-coefficients>` tips on a linear model fitted on
+mixed-scale features. The check compares per-feature standard deviations and
+warns when raw coefficient magnitudes are not directly comparable — or when
+scaled coefficients are comparable but no longer in original units.
```
glemaitre: We should quickly reference: https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
+"""
+.. _example_skd006_coefficient_interpretation:
+
+SKD006 — Coefficient interpretation
+===================================
+
+This example walks through mitigations when check
```
glemaitre: So here, I would like that we change language to make sure that nothing is wrong with the code but rather than it is just tricky when interpreting and depending of what you want to do then you need to make changes or interpret in a specific manner.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+:ref:`SKD006 <skd006-unscaled-coefficients>` tips on a linear model fitted on
+mixed-scale features. The check compares per-feature standard deviations and
+warns when raw coefficient magnitudes are not directly comparable — or when
+scaled coefficients are comparable but no longer in original units.
+
+Mitigations from the :ref:`automated_checks` user guide:
```
glemaitre: So I would not call this mitigation but rather describe as "in this case, think of X"

Also can we make sure to cross-reference the scikit-learn section example for each part.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+# The target is continuous house value in \$100k units.
+
+TableReport(y)
+
+# %%
+# A single :class:`~skore.TrainTestSplit` feeds every evaluation below.
+
+from skore import TrainTestSplit
+
+splitter = TrainTestSplit(random_state=42)
```
glemaitre: Let's mention explicitly the size of the test.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+
+# %%
+report.checks.summarize()
+
+# %%
+report.inspection.coefficients().frame()
```
glemaitre: We need to describe to the user what to check in the report here.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+
+# %%
+# Scale coefficients by feature standard deviation
+# ================================================
+#
+# Multiplying each fitted coefficient by its feature standard deviation gives
```
glemaitre: Let's first recall the previous "pitfalls": if trying to compare feature between them then you cannot because the model was trained on feature with different  dynamic range.

Then we need to link the section in the scikit-learn example

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+# without refitting the model.
+
+coef = report.inspection.coefficients().frame()
+feature_std = X.std()
+
+comparable = coef[coef["feature"] != "Intercept"].copy()
+comparable["effect_per_std"] = comparable["coefficient"] * comparable["feature"].map(
+    feature_std
+)
+comparable.sort_values("effect_per_std", key=abs, ascending=False)
```
glemaitre: Looking at the code here, I think that we should open an issue where we should implement this behaviour via a parameter in the `frame` or `plot`.

```python
report.inspection.coefficients().frame(compare_between_feature=True)
```

maybe the name should be something different depending of the literature. Maybe it would be a way to return the effect plot instead: https://christophm.github.io/interpretable-ml-book/limo.html

```python
report.inspection.coefficients().frame(as_effect=True)
```

glemaitre: I would not address it here.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+
+# %%
+report_scaled.checks.summarize()
+
+# %%
+report_scaled.inspection.coefficients().frame()
```
glemaitre, line 193: We need more explanation

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+# Permutation importance (scale-invariant)
+# ========================================
+#
+# When features need to be interpreted in original units, or when the
+# model is nonlinear, permutation importance on held-out data offers a
+# scale-invariant ranking that does not depend on coefficient magnitudes.
+
+import matplotlib.pyplot as plt
+
+perm_display = report.inspection.permutation_importance(
```
glemaitre: call it `display`

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+    seed=42,
+    n_repeats=5,
+)
+perm_display.frame()
+
+# %%
+fig = perm_display.plot()
+plt.show()
```
glemaitre: you should not need this line

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+# model is nonlinear, permutation importance on held-out data offers a
+# scale-invariant ranking that does not depend on coefficient magnitudes.
+
+import matplotlib.pyplot as plt
+
+perm_display = report.inspection.permutation_importance(
+    seed=42,
+    n_repeats=5,
+)
+perm_display.frame()
```
glemaitre:
```suggestion

# %%
perm_display.frame()
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,150 @@
...
+
+# %%
+fig = perm_display.plot()
+plt.show()
+
```
glemaitre, line 222: We also need to go more in details in the analysis and explain what are the bias of using the permutation importance. @GaetandeCast should have some material that would fit here.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
GaetandeCast: That would be cool indeed, I can write the issue. Maybe add a parameter `type: Literal["coefficient", "effect"] = "coefficients"`. Problem is when people scaled their features before fitting, the parameter makes no sense anymore. But we can detect it and warn in this situation.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
+"""
+.. _example_skd006_coefficient_interpretation:
+
+SKD006 — Coefficient interpretation
```
GaetandeCast:
```suggestion
SKD006 - Coefficient interpretation
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+===================================
+
+:ref:`SKD006 <skd006-unscaled-coefficients>` is a tip about interpretation, not a sign that the model failed. This check
+only flags that on mixed-scale features, raw coefficient magnitudes are not directly comparable across columns, and
+after standardization they are comparable but no longer expressed in the original feature units. Which path you take
+depends on what you want to read from the coefficients. See also scikit-learn's example on `linear model coefficient interpretation <https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html>`_.
+
+In this case, think of:
```
GaetandeCast: Which case ? Let's be a little clearer here.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+
+# %%
+# A single :class:`~skore.TrainTestSplit` with a 20 % test set feeds every evaluation below.
+
+from skore import TrainTestSplit
+
+splitter = TrainTestSplit(test_size=0.2, random_state=42)
+
+# %%
+# Trigger SKD006 — Ridge on raw mixed-scale features
```
GaetandeCast:
```suggestion
# Trigger SKD006: Ridge on raw mixed-scale features
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+# ==================================================
+#
+# Ridge with a moderate ``alpha`` fits quickly on unscaled inputs. SKD006 is not a sign that the fit failed — other
+# issues may still exist — but it does flag that the features are not on the same scale, so raw coefficient magnitudes
+# should not be read as a ranking of importance.
```
GaetandeCast:
```suggestion
# Ridge with a moderate ``alpha`` fits quickly on unscaled inputs. SKD006 indicates that the features are not on the same scale, so raw coefficient magnitudes
# should not be read as a ranking of importance.
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+from skore import evaluate
+
+report = evaluate(
+    Ridge(alpha=1.0),
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+
+# %%
```
GaetandeCast, line 83:
```suggestion
# %%
# Find `SKD006` in the `Tips` tab below.
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+report.checks.summarize()
+
+# %%
+# Inspect coefficients with :meth:`~skore.EstimatorReport.inspection.coefficients`. Compare each coefficient's magnitude
+# to the feature's typical range (or standard deviation): a large coefficient on a small-scale column is not necessarily
+# more important than a small coefficient on a large-scale column such as population.
+
+report.inspection.coefficients().frame()
```
GaetandeCast: Here I would put the coefficients plot side by side with a feature variance plot and give a concrete example of a feature having a "misleading" coefficient value due to large or low variance in the feature.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+# discussion of `scale matters <https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#interpreting-coefficients-scale-matters>`_.
+
+coef = report.inspection.coefficients().frame()
+feature_std = X.std()
+
+comparable = coef[coef["feature"] != "Intercept"].copy()
+comparable["effect_per_std"] = comparable["coefficient"] * comparable["feature"].map(
+    feature_std
+)
+comparable.sort_values("effect_per_std", key=abs, ascending=False)
```
GaetandeCast: It would be nice to then comment on how scaling coefficients erased the potentially surprisingly low/high coefficient value marked in the previous section.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+
+# %%
+# Standardize inputs in a pipeline
+# ================================
+#
+# :class:`~sklearn.preprocessing.StandardScaler` inside a :class:`~sklearn.pipeline.Pipeline` puts every feature on a
+# common scale at fit time, so coefficients become directly comparable as "effect per one standard deviation." That is
+# useful for ranking features, but you lose statements in original units (for example, "one extra year of age"). SKD006
+# then tips that the coefficients are comparable but no longer in the original feature units — the other side of the
```
GaetandeCast:
```suggestion
# then tips that the coefficients are comparable but no longer in the original feature units: the other side of the
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+from sklearn.preprocessing import StandardScaler
+
+report_scaled = evaluate(
+    Pipeline([("scale", StandardScaler()), ("ridge", Ridge(alpha=1.0))]),
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+
+# %%
```
GaetandeCast, line 175:
```suggestion
# %%
# `SKD006` is still reported in the `Tips` tab but the warning message changed.
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+report_scaled.checks.summarize()
+
+# %%
+# The coefficient table is now on a shared scale. Read magnitudes as relative importance after standardization, not as
+# effects per original unit of income, rooms, or population.
+
+report_scaled.inspection.coefficients().frame()
```
GaetandeCast, line 193: Mention that we get exactly the same values as with the post normalization with standard deviation.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+
+# %%
+# Permutation importance (scale-invariant)
+# ========================================
+#
+# When you need a ranking that does not depend on coefficient units — for example because you want to keep features in
+# original units, or the model is nonlinear — permutation importance on held-out data is a scale-invariant alternative.
```
GaetandeCast: The real reason for this would be to do feature selection or model interpretation (answering questions such as "Which feature did my model use the most to make it's prediction).

GaetandeCast: Also this is just an other way of doing it that is an alternative to scaled coefficients and should be presented as such.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
GaetandeCast: I assume you mean to mention how feature correlation can distort permutation importance ? Since it also does it for coefficients importance, we should probably just link to the SKD008 example which should cover it extensively.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+
+# %%
+display.plot()
+
+# %%
+# Conclusion
+# ==========
+#
+# SKD006 is a reminder about scale and interpretation, not a sign that the fit failed (other problems may still be
+# present). When features are on different scales, linear coefficients mix feature importance with feature magnitude;
```
GaetandeCast:
```suggestion
# SKD006 is a reminder to be careful when interpreting coefficients of linear models. 
# When features are on different scales, linear coefficients mix feature importance with feature magnitude;
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,157 @@
...
+# when features are standardized, coefficients are comparable but no longer in original units. Choose standardization,
+# an effect-per-std rescaling, or a scale-invariant method such as permutation importance depending on the question you
+# want to answer.
```
GaetandeCast: You should mention that unscaled coefficients are still useful by themselves: they answer the question "If I change this feature by this much, how much does the prediction change in it's unit ?"

Actually also worth mentioning earlier too, like at the end of the first section.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,228 @@
...
+# deviations tell different stories. ``AveBedrms`` often carries one of the
+# largest absolute coefficients while having a small standard deviation, so its
+# "per unit bedroom" effect looks large. ``Population`` has a tiny coefficient
+# because a one-person change is negligible on a head-count scale, even though
+# the column varies a lot across districts.
+
+import matplotlib.pyplot as plt
+
+coef = coef_display.frame()
+coef_no_intercept = coef[coef["feature"] != "Intercept"].set_index("feature")
```
glemaitre: we have a parameter to exclude or not the intercept in `frame`.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,228 @@
...
+feature_std = report.X_train.std()
+
+_, axes = plt.subplots(1, 2, figsize=(10, 4), constrained_layout=True)
```
glemaitre: I think that we can replace using the `scale_features`: https://docs.skore.probabl.ai/dev/reference/api/skore.CoefficientsDisplay.html#skore.CoefficientsDisplay.frame

It will simplify the code.

glemaitre: Ah my bad, it is in the next section.

Here, you can directly show the standard deviation, and it should already be a dataframe so you can use the plotting api from pandas with the `.plot` method that should avoid quite some matplotlib code.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,228 @@
...
+#
+# Recall the pitfall above: when the model was trained on features with
+# different dynamic ranges, you cannot compare features by looking at raw
+# coefficients alone. Multiplying each fitted coefficient by its feature
+# standard deviation gives an "effect per one standard deviation" that is
+# comparable across columns without refitting. See scikit-learn's discussion of
+# `scale matters
+# <https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#interpreting-coefficients-scale-matters>`_.
+
+comparable = coef_no_intercept.copy()
```
glemaitre: see my previous comment

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
glemaitre: yes. We can cross-link example to say to look at SKD008

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,228 @@
...
+display = report.inspection.permutation_importance(
+    seed=42,
+    n_repeats=5,
+)
+
+# %%
+display.frame()
+
+# %%
+display.plot()
```
glemaitre, line 221:
```suggestion
_ = display.plot()
```

glemaitre (commented):
Some final comments

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+
+# %%
+# Load the California housing dataset
+# ===================================
+#
+# Median income is measured in 10k USD blocks, latitude in degrees, and
+# population in head counts. Fitting Ridge on these raw columns produces
+# coefficients whose magnitudes reflect units as much as predictive strength.
+# We load the table via skrub so Sphinx CI does not depend on scikit-learn's
+# ``cal_housing.tgz`` download cache.
```
GaetandeCast, line 38: Implem detail that does not need to be mentioned

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+
+# %%
+# Find ``SKD006`` in the Tips tab below.
+
+report.checks.summarize()
+
+# %%
+# Inspect coefficients with
+# :meth:`~skore.EstimatorReport.inspection.coefficients`.
+# Compare each coefficient's magnitude to the feature's typical range (or
```
GaetandeCast, line 91:
```suggestion
# Let us inspect coefficients with
# :meth:`~skore.EstimatorReport.inspection.coefficients`, and
# compare each coefficient's magnitude to the feature's typical range (or
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+#
+# Unscaled coefficients are still useful on their own: they answer "if I change
+# this feature by one unit in its original scale, how much does the prediction
+# change in target units?" They are misleading regarding feature importance.
+coef_display = report.inspection.coefficients()
+coef_display.frame()
+
+# %%
+# One figure under another, the raw coefficient magnitudes and the feature standard
+# deviations tell different stories. ``AveBedrms`` often carries one of the
```
GaetandeCast, line 104: I don't really understand what this first sentence tries to say.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+# <https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#interpreting-coefficients-scale-matters>`_.
+
+comparable = coef.copy()
+comparable["effect_per_std"] = comparable["coefficient"] * feature_std
+_ = comparable.plot.barh()
+
+# %%
+# That rescaling erases the surprises from the side-by-side bars above. The
+# large raw ``AveBedrms`` coefficient shrinks once multiplied by its small
+# standard deviation, so it usually drops in the ranking. The tiny raw
```
GaetandeCast, line 145:
```suggestion
# standard deviation, so it drops in the ranking. The tiny raw
```

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+# ``Population`` coefficient grows with that column's large std, and features
```
GaetandeCast, line 146: It does not grow that much, 0x1000=0 😆 
I'd not mention it and focus on the feature that do change in the rankings.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+# :class:`~sklearn.preprocessing.StandardScaler` inside a
+# :class:`~sklearn.pipeline.Pipeline` puts every feature on a common scale at
+# fit time, so coefficients become directly comparable as "effect per one
+# standard deviation." That is useful for ranking features, but you lose
+# statements in original units (for example, "one extra year of age"). SKD006
+# then tips that the coefficients are comparable but no longer in the original
+# feature units: the other side of the same interpretation trade-off. See again
+# `scale matters
+# <https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#interpreting-coefficients-scale-matters>`_.
+
```
GaetandeCast, line 163: We mentioned it twice already, I don't think we need to keep referring to this example.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+report_scaled = evaluate(
+    Pipeline([("scale", StandardScaler()), ("ridge", Ridge(alpha=1.0))]),
+    X=X,
+    y=y,
+    splitter=splitter,
+)
+report_scaled
+
+# %%
+# ``SKD006`` is still reported in the Tips tab but the warning message changed.
```
GaetandeCast, line 176: Uh the message has actually not changed. We need to investigate why

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+#
+# These values are the same quantity as the post-hoc
+# ``coefficient * feature_std`` column from the previous section: both are an
+# effect per one standard deviation. Fitting with ``StandardScaler`` does that
+# rescaling inside the pipeline; multiplying raw coefficients by the train-set
+# std does it after the fact. Small numerical differences can remain because the
+# scaler's internal mean/std and ``X_train.std()`` are not bit-identical, but the
+# ranking and interpretation should match.
+
+report_scaled.inspection.coefficients().frame()
```
GaetandeCast, line 193: We never showed the post-scaled numbers so these are useless on their own. Let's show both tables side by side.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
```diff
@@ -0,0 +1,236 @@
...
+# Scaled coefficients are one way to rank features. Another is permutation
+# importance on held-out data: shuffle one column at a time and measure how
+# much the score drops. That answers questions such as "which features did the
+# model rely on most?" and works for nonlinear models as well. Present it as an
+# alternative to scaled coefficients when you care about predictive reliance
+# rather than a linear effect size.
+#
+# Correlated features can distort both coefficient rankings and permutation
+# importance (credit is shared or shifted between partners). See
+# :ref:`SKD008 <skd008-correlated-features>` and scikit-learn's note on
```
GaetandeCast, line 208: Link to the example rather than the user guide I think.

examples/pitfalls_and_solutions/plot_skd006_coefficient_interpretation.py
GaetandeCast: I looked at it, the standard deviations for each feature after preprocessing are: `[0.99768173 0.99868593 1.03638858 1.09397345 0.99598968 0.89701992 0.99969662 0.99897233]`. We are using an absolute tolerance of `0.05` so a few of the deviations are outside the range here. We can bump the tolerance to `0.1` I think. Also we are checking the difference against the first std but we could do against the average.

I think we can bump it in this PR.

PR #3164, opened 2026-07-23
feat(plugin/hub/project/summarize): Use new `GET:/reports` endpoint

thomass-dev:


thomass-dev:
@retromechs it is ready for your review. I let the PR drafted because it needs https://github.com/probabl-ai/skore-hub/pull/1712/ to be merged+released.

retromechs:
> @retromechs it is ready for your review. I let the PR drafted because it needs [probabl-ai/skore-hub#1712](https://github.com/probabl-ai/skore-hub/pull/1712) to be merged+released.

We can schedule the hub's release for today, if you wish.

skore/src/skore/_plugins/hub/project/project.py
```diff
@@ -426,24 +389,35 @@ def dto(response: Any, /) -> Metadata:
...
+                        "sort_by": "created_at",
+                    },
+                )
+
+                page = response.json()
+
+                reports.extend(map(dto, page["items"]))
+
+                if (cursor := page["next_cursor"]) is None:
+                    break
```
retromechs, line 418: You should keep this previous code optimization:
```python
if len(page) < limit:
    # avoid one server call when we are sure that the current page is
    # the last one
    break
```
On the hub side, `next_cursor` is `None` only if `items` is empty.
I’m going to open a PR on the hub to make the `next_cursor` contract more intuitive, but IMO, this optimization should be kept.

skore/tests/unit/plugins/hub/project/test_project.py
```diff
@@ -506,76 +506,81 @@ def test_summarize(self, nowstr, respx_mock):
...
+                                "type": "cross-validation",
+                                "id": "<report_id_2>",
+                                "key": "<key>",
+                                "ml_task": "<ml_task>",
+                                "estimator_class_name": "<estimator_class_name>",
+                                "dataset_fingerprint": "<dataset_fingerprint>",
+                                "created_at": nowstr,
+                                "metrics": [
+                                    {
+                                        "name": "rmse_mean",
```
retromechs, line 566: The `*_mean` suffix was removed from the hub output, am I wrong?

retromechs (changes requested):
Apart from a small network optimization everything looks good to me!

skore/src/skore/_plugins/hub/project/project.py
thomass-dev: After thinking, the (side) case when `len(page) == limit` but it's the last page can't be detected by the client.
IMO, it could be cool that the server send `next_cursor=None` on the last page.
WDYT?

retromechs: Yes, this is exactly the PR I plan to open server side! :)

thomass-dev: Fixed by https://github.com/probabl-ai/skore-hub/pull/1716.

skore/tests/unit/plugins/hub/project/test_project.py
thomass-dev: Good catch, fixed by https://github.com/probabl-ai/skore-hub/pull/1717.

retromechs (approved):

PR #3094, opened 2026-06-25
feat(skore): Refactor MetricsSummaryDisplay to follow display API and handle non-scalar metrics

glemaitre:
closes #3093

This PR modifies the `MetricsSummaryDisplay` as follow:

- follow our API for display implemented a `_compute_for_display` factory and store the information inside a single dataframe `.summary`
- follow a refactoring to always return `summary` with only the column that are useful
- for allowing to get a close dataframe and the representation of the HTML repr of the display, add the possibility to whether or not using the verbose name of the metric and the multi index (which are not the default when calling `.frame`
- adapt the hub project such that we send all metrics, also with multiclass containing multi class labels and multi-output regression

PR #3169, opened 2026-07-28
fix(skore): Revert to compact metric HTML repr for cross-validation report

glemaitre:
Small regression introduced in #3094 when it comes to the cross-validation report. We are back to:

image


GaetandeCast (approved):
LGTM

thomass-dev (approved):

PR #3172, opened 2026-07-29
fix(skore): Allow to send averaged metric for binary classification

glemaitre:
This is a small fix when sending metrics on the hub. Currently we did not update the averaged metrics for the binary case because the built-in metrics blocking in this case.

But there is not reason that custom metric allowing it cannot be send to the hub.

So this PR is removing this lock.


thomass-dev (approved):

thomass-dev (approved):

thomass-dev (changes requested):
Tests need to be fixed.

thomass-dev (approved):

PR #3175, opened 2026-07-29
fix(skore): Metrics summary crash for cv report with report n_job != 1

GaetandeCast:
Checking a local bug against the CI.


glemaitre:
What is the bug that you get?

GaetandeCast:
The recursion error :wink:
I stumbled upon it while running the example in #3145 locally. It surprised me because the doc built properly. I want to check which of the versions we support break and which don't.

PR #3179, opened 2026-08-01
feat(skore): Add scale_features option to coefficients display

### LINKED ISSUE #3162: Add an option to normalize coefficients by the feature variance in coefficients display
So that we can quickly have comparable values for the effect of each feature.

>We should implement this behaviour via a parameter in the `frame` or `plot`.

```python
report.inspection.coefficients().frame(compare_between_feature=True)
```

>maybe the name should be something different depending of the literature. Maybe it would be a way to return the effect plot instead: https://christophm.github.io/interpretable-ml-book/limo.html

```python
report.inspection.coefficients().frame(as_effect=True)
```

_Originally posted by @glemaitre in https://github.com/probabl-ai/skore/pull/3148#discussion_r3602290610_

glemaitre:
## Summary

Add `scale_features` to coefficients `frame`/`plot`: multiply coeffs by training feature stds so magnitudes are comparable across features.

Internal `feature_std` column (unit std for intercept; NaN if no train data). Dropped from the user-facing frame.

Closes https://github.com/probabl-ai/skore/issues/3162

## Test plan

- [x] Unit tests for scaling, intercept, pipeline stds, prefit error, CV, plot xlabel
- [x] Existing coefficients display tests


skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -158,6 +165,13 @@ def frame(
             Can be used independently of `select_k`. Sorting is performed within the
             same groups as selection.
 
+        scale_features : bool, default=False
+            If `True`, multiply each non-intercept coefficient by the training-set
+            standard deviation of the corresponding feature (as seen by the
+            predictor). The resulting values are comparable across features as
+            the effect of a one-standard-deviation change. Requires
+            :attr:`feature_std` to have been computed at display construction.
```
GaetandeCast: I don't think we should mention such an implementation detail in a public facing docstring. I'd rather say that it requires `X_train`.

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -289,6 +320,11 @@ def plot(
             Can be used independently of `select_k`. Sorting is performed within the
             same groups as selection.
 
+        scale_features : bool, default=False
+            If `True`, multiply each non-intercept coefficient by the training-set
+            standard deviation of the corresponding feature (as seen by the
+            predictor). See :meth:`frame` for details.
```
GaetandeCast: I'd rather not refer to frame for more details and duplicate them instead.

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -708,11 +766,28 @@ def _compute_data_for_display(
         )
         n_features = len(feature_names)
 
+        if X is not None:
+            X_transformed: Any = (
+                preprocessor.transform(X)
+                if preprocessor is not None and len(preprocessor.steps) > 0
+                else X
```
GaetandeCast, line 775: Note that does not need a change:
I realize that here we are still pretty naive with the way we separate the predictor and preprocessor, which is probably broken for skrub inputs. We should transfer what we do in #3086 here. The other inspection displays probably also suffer from the same pb.

skore/tests/unit/displays/coefficients/test_cross_validation.py
```diff
@@ -71,3 +76,18 @@ def test_valid_subplot_by(fixture_name, subplot_by_tuples, request):
             assert isinstance(axes[0], mpl.axes.Axes)
         else:
             assert len(axes) == expected_len
+
+
+def test_scale_features():
+    """CV displays store feature_std and can scale coefficients."""
+    X, y = make_regression(n_samples=80, n_features=3, random_state=0)
+    report = evaluate(Ridge(), X, y, splitter=2)
```
GaetandeCast: Is there a particular reason for not using one of our data fixtures?

skore/tests/unit/displays/coefficients/test_cross_validation.py
```diff
@@ -71,3 +76,18 @@ def test_valid_subplot_by(fixture_name, subplot_by_tuples, request):
             assert isinstance(axes[0], mpl.axes.Axes)
         else:
             assert len(axes) == expected_len
+
+
+def test_scale_features():
```
GaetandeCast: Could you add a similar simple test for comparisons ?

skore/tests/unit/displays/coefficients/test_estimator.py
```diff
@@ -96,3 +99,115 @@ def test_include_intercept_multioutput_fit_intercept_false(request):
...
+    np.testing.assert_allclose(display.coefficients["feature_std"], 1.0, atol=0.05)
+    raw = display.frame(include_intercept=False)["coefficient"].to_numpy()
+    scaled = display.frame(include_intercept=False, scale_features=True)[
+        "coefficient"
+    ].to_numpy()
+    np.testing.assert_allclose(raw, scaled, rtol=0.05)
+
+
+def test_scale_features_sparse_preprocessor(regression_train_test_split):
+    """Feature stds are computed on sparse transformed data without densifying."""
```
GaetandeCast: I don't see how this tests that the data has not been densified, but it would be hard to test I think.

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -515,9 +563,12 @@ def _plot_single_estimator(
             barplot_kwargs=barplot_kwargs,
             boxplot_kwargs=boxplot_kwargs,
             stripplot_kwargs=stripplot_kwargs,
+            scale_features=scale_features,
         )
 
         title = f"Coefficients of {estimator_name}"
+        if scale_features:
+            title = f"Scaled coefficients of {estimator_name}"
```
GaetandeCast, line 573: This is not covered by tests

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
```diff
@@ -650,9 +702,10 @@ def _plot_comparison(
             barplot_kwargs={"sharey": has_same_features} | barplot_kwargs,
             boxplot_kwargs=boxplot_kwargs,
             stripplot_kwargs={"sharey": has_same_features} | stripplot_kwargs,
+            scale_features=scale_features,
         )
 
-        title = "Coefficients"
+        title = "Scaled coefficients" if scale_features else "Coefficients"
```
GaetandeCast, line 710: Neither is this

GaetandeCast (changes requested):
Some remarks. Also there is no special case for when features have unit variance. Do we silently let the parameter have no effect ?

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
glemaitre: Yes I agree.

skore/src/skore/_sklearn/_plot/inspection/coefficients.py
glemaitre: It is indeed broken as far as I remember in #3086. I'm not sure that it would be super straight-forward to solve. I think we could have an isolated PR/issue indeed.

skore/tests/unit/displays/coefficients/test_estimator.py
glemaitre: Yep the title is not the best. Here, it is more of a smoke tests that it does not break. While working on it, I made sure that we don't densify the intermediate output but it is hard to probe this in a test. I'll modify the docstring.

glemaitre:
> Also there is no special case for when features have unit variance. Do we silently let the parameter have no effect ?

You mean in the case that the feature are already scaled? I don't think that would be an issue. It would just be a no-op operation.

GaetandeCast (approved):
Thanks @glemaitre @cursoragent !

PR #3189, opened 2026-08-03
fix(skore): Do not create skore workspace if directory exists

### LINKED ISSUE #3187: fix(plugins/local): The default workspace must not corrupt the user's directory
```python
def init_workspace(workspace_dir: str | Path | None = None) -> Path:
    workspace_dir = (
        (Path(".") / "skore" if workspace_dir is None else Path(workspace_dir))
        .expanduser()
        .resolve()
    )
    new_workspace = not workspace_dir.exists()
    workspace_dir.mkdir(parents=True, exist_ok=True)
    if new_workspace:
        (workspace_dir / ".gitignore").write_text("*\n")
    (workspace_dir / ".SKORE_WORKSPACE").touch()
    (workspace_dir / "projects").mkdir(exist_ok=True)
    (workspace_dir / "datasets").mkdir(exist_ok=True)
    return workspace_dir
```

I have an issue with this behavior: as a user, and especially as a contributor, creating a project should not overlap with my `skore/` or `skore/skore/` sources.

If a `skore/` directory exists at cwd but `.SKORE_WORKSPACE` is missing, we should ask for an explicit validation (or something else). Otherwise, this could lead to conflicts or corruption of the user's directory. We don't manage the user's environment and what might be inside it compared to `.local/share` & co.

_Originally posted by @thomass-dev in https://github.com/probabl-ai/skore/pull/2905#discussion_r3702310401_

jeromedockes:
fixes #3187

### DIFF
--- a/skore/src/skore/_plugins/local/project.py
+++ b/skore/src/skore/_plugins/local/project.py
@@ -31,7 +31,7 @@ def init_workspace(workspace_dir: str | Path | None = None) -> Path:
         workspace_dir.is_file()
         or workspace_dir.is_dir()
         and not (workspace_dir / ".SKORE_WORKSPACE").exists()
-        and next(workspace_dir.iterdir(), None) is not None
+        and any(workspace_dir.iterdir())
     ):
         raise FileExistsError(
             f"Cannot create skore workspace: {workspace_dir} is not empty "
--- a/skore/src/skore/_plugins/local/project.py
+++ b/skore/src/skore/_plugins/local/project.py
@@ -27,11 +27,24 @@ def init_workspace(workspace_dir: str | Path | None = None) -> Path:
         .expanduser()
         .resolve()
     )
+    if (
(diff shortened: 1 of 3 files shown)


skore/src/skore/_plugins/local/project.py
```diff
@@ -27,11 +27,24 @@ def init_workspace(workspace_dir: str | Path | None = None) -> Path:
         .expanduser()
         .resolve()
     )
+    if (
+        workspace_dir.is_file()
+        or workspace_dir.is_dir()
+        and not (workspace_dir / ".SKORE_WORKSPACE").exists()
+        and next(workspace_dir.iterdir(), None) is not None
```
thomass-dev: Nitpick (more readable):

```suggestion
        and any(workspace_dir.iterdir())
```

thomass-dev (approved):
Thanks for the quick fix!

PR #3190, opened 2026-08-03
feat(plugins/hub/report): Add a new `environment` artifact

thomass-dev:
Ships an environment report artifact so the hub can recreate the python environment to deserialize reports.

- The environment artifact is a lockfile-style `name==version` file, uploaded as `text/plain`
- The environment artifact is always computed last: custom `@model_serializer` dumps everything except environment, then environment, so `sys.modules` includes imports from metrics/medias/pickle etc.
- The potential editable/local-install warnings are filtered in pytest (it's common to use editable packages for tests).

Related to https://github.com/probabl-ai/skore-hub/pull/1730.


skore/src/skore/_plugins/hub/artifact/environment.py
```diff
@@ -0,0 +1,52 @@
...
+        Returns
+        -------
+        str or None
+            A newline-separated list of ``name==version`` requirements from that
+            environment, or ``None`` if requirements cannot be inferred.
+        """
+        try:
+            return "\n".join(
+                f"{requirement['name']}=={requirement['version']}"
+                for requirement in infer()
```
glemaitre, line 48: Could we store directly this string into each report itself.

To me it belongs to it the same way we are storing a commit hash. When we will want to send it to the different backend, we can just get the string.

glemaitre: The only twist here is when to trigger the writing I think

thomass-dev: > The only twist here is when to trigger the writing I think

We can't know directly in the report when to infer the requirements without infering many times (one time at init, one time after each metrics summary etc), especially because the metric registry is dynamic and can contain custom metrics.

glemaitre (commented):
Everything looks clean. I'm just wondering to which structure the requirements are belonging. I think that they should be stored in the report as well but it comes with the question when and how to store this info because of the necessity to have imported most of the thing.

skore/src/skore/_plugins/hub/artifact/environment.py
glemaitre: OK so let's start like this.

glemaitre (approved):

PR #3191, opened 2026-08-04
perf(ci): Speed up pytest via xdist and faster MLflow tracking

glemaitre:
This PR should speed up the different tests using `xdist` and using the file system instead of sqlite for MLflow.

We go from 25 minutes on the window tests to 6-10 minutes.


glemaitre:
@thomass-dev I addressed your review here.

thomass-dev (approved):
Happy to see the xdist comeback 🤞🏼 .

PR #3194, opened 2026-08-04
fix(checks): Ensure every check is present in the summary

### LINKED ISSUE #3052: Make sure every check is surfaced in the summary
Make sure that in the different views of the `ChecksSummaryDisplay` (`.frame`, `__repr__` and `_repr_html_`), every check appears. Currently we only show checks that raised an issue or a tip or checks that finished running without finding anything.

Outcomes that are not yet surfaced:
- check raised `CheckNotApplicable` #3049
- check was skipped due to `fast_mode=True` #3103
- check was ignored #3103
- checks has incompatible `report_types`

auguste-probabl:
Closes https://github.com/probabl-ai/skore/issues/3052

### DIFF
--- a/skore/src/skore/_sklearn/_base.py
+++ b/skore/src/skore/_sklearn/_base.py
@@ -130,24 +130,37 @@ class _BaseReport(ReportHelpMixin):
         summary: dict[CheckCode, CheckResult] = {}
         for check in self._checks_registry:
             if self._report_type not in check.report_types:
-                continue
-            code = check.code
-            if code in ignored_codes:
-                summary[code] = {
+                summary[check.code] = {
+                    "title": check.title,
+                    "docs_url": check.docs_url,
+                    "explanation": f"Not applicable to {self._report_type} reports.",
+                    "section": "not_applicable",
+                }
+            elif check.code in ignored_codes:
+                summary[check.code] = {
                     "title": check.title,
                     "docs_url": check.docs_url,
                     "explanation": None,
                     "section": "ignored",
                 }
-            elif fast_mode and check.slow and code not in self._check_results_cache:
-                summary[code] = {
+            elif (
+                fast_mode
+                and check.slow
+                and (check.code not in self._check_results_cache)
+            ):
+                summary[check.code] = {
                     "title": check.title,
                     "docs_url": check.docs_url,
                     "explanation": None,
                     "section": "skipped",
                 }
-            elif code in self._check_results_cache:
-                summary[code] = self._check_results_cache[code]
+            elif check.code in self._check_results_cache:
+                summary[check.code] = self._check_results_cache[check.code]
+
+        # Every check appears exactly once in the summary
+        assert len(set(summary)) == len(summary)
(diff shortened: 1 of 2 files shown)


skore/src/skore/_sklearn/_base.py
```diff
@@ -130,24 +130,37 @@ def _get_checks_results(
...
                     "section": "skipped",
                 }
-            elif code in self._check_results_cache:
-                summary[code] = self._check_results_cache[code]
+            elif check.code in self._check_results_cache:
+                summary[check.code] = self._check_results_cache[check.code]
+
+        # Every check appears exactly once in the summary
+        assert len(set(summary)) == len(summary)
+        assert set(summary) == {check.code for check in self._checks_registry}
```
auguste-probabl: kind of ambivalent about this, I am under the influence of the SQLite talk :)

glemaitre: I'm not a fan of assertion when there is a possibility to test exactly the same within the test suite.

So I would move this in a proper test. Raising an `AssertionError` to the user without context and way to go around it is not an nice UX.

thomass-dev: IMHO, putting assertion in the source-code can be useful for DX only to drive the developpers.
When you are sure that the assertion will be triggered in the unitest, before to be triggered in the user runtime.

thomass-dev: Aside of the debate:

```suggestion
        # Every check appears exactly once in the summary
        assert sorted(summary) == sorted(check.code for check in self._checks_registry)
```

GaetandeCast: Not a big fan either but it's not a super strong opinion.

GaetandeCast (dismissed):
I'll let you decide on the `assert` question but otherwise LGTM

skore/src/skore/_sklearn/_base.py
auguste-probabl: TL;DR: Still no strong opinions, but the discussion is really thought-provoking, thanks :) I will remove the assert; I lay out my latest thoughts here.

> Raising an AssertionError to the user without context and way to go around it is not an nice UX.

An assertion is not a typical "We found something wrong, the user should know about it". It is an executable comment that states an invariant of the program. If this invariant is not upheld, some downstream logic might fail in a strange way.

With that in mind, if the user encounters a bug related to this, I would rather they report the `assert` failing, than an obscure error in the downstream code. The UX is basically the same for them (in both cases they see an obscure error) but catching the issue earlier makes it much easier to debug.

> test exactly the same within the test suite.

Here my view has also changed. A unit test tests up to a handful of scenarios (usually just one); even if the unit test technically covers the code, my *confidence* that the code is not broken is not super high. An assert is a test that runs *all the time, in a lot of strange situations*; it therefore makes me a lot more confident that the code is not broken.

GaetandeCast (approved):
Thanks @auguste-probabl !

PR #3195, opened 2026-08-04
chore(skore): Align default metrics with is_computable and discouraged

glemaitre:
## Summary

- Replace the task-branching `default_metrics` factory with a flat `DEFAULT_METRICS` catalog.
- Rename `Metric.available` → `Metric.is_computable` (technical: can we compute it?).
- Add `Metric.discouraged(report) -> str | None` (methodological: should we register it?). Defaults skip discouraged metrics silently; `metrics.add(..., force=False)` raises with the reason unless `force=True`.
- Move the `has_default_score` display filter onto `Score.discouraged`, so sklearn's default `score` is no longer registered for LogisticRegression / RandomForest / etc.
- Allow `metrics.add(Score(), force=True)` to put a discouraged score back into `summarize` / `get`; `metrics.score()` remains the dedicated escape hatch.

## Behavior change

Defaults only register metrics that are both computable and not discouraged. User `.add` never checks `is_computable`; only `discouraged` can block it (unless `force=True`).

For estimators that use scikit-learn's default `ClassifierMixin.score` / `RegressorMixin.score`, `"score"` is skipped from the registry for the same reason (use `report.metrics.score()`, or `add(Score(), force=True)` to put it back).

## Example

The point of `discouraged` is to refuse registering a metric that *can* be computed but should not be reported for the problem at hand. Override it on your metric class; `.add` raises with that reason unless the caller passes `force=True`:

```python
from sklearn.metrics import accuracy_score
from skore import evaluate
from skore._sklearn.metrics import FunctionKind, Metric

class AccuracyOnImbalanced(Metric):
    """Accuracy that refuses to be registered on binary classification reports."""

    name = "accuracy_on_imbalanced"
    verbose_name = "Accuracy (imbalanced)"
    function = staticmethod(accuracy_score)
    response_method = "predict"
    greater_is_better = True
    function_kind = FunctionKind.METRIC

    @staticmethod
    def discouraged(report):
        if report._ml_task == "binary-classification":
            return (
                "accuracy is often misleading for binary classification; "
                "prefer precision, recall, or ROC AUC"
            )
        return None

report = evaluate(LogisticRegression(max_iter=10_000), X, y, splitter=0.2)

# Refused with the reason from discouraged().
report.metrics.add(AccuracyOnImbalanced())
# ValueError: Cannot add 'accuracy_on_imbalanced': accuracy is often misleading
# for binary classification; prefer precision, recall, or ROC AUC. Pass
# force=True to register it anyway.

# Opt in explicitly when you still want it.
report.metrics.add(AccuracyOnImbalanced(), force=True)
report.metrics.accuracy_on_imbalanced()
# 0.94...
```

Custom metrics without a `discouraged` override keep today's behavior: `.add` registers them unconditionally (still not filtered by `is_computable`).

## Out of scope (flagged, not fixed here)

- The `neg_` alias claim in `summarize` / `get` docstrings that lookup does not implement
- `Metric.new` silently dropping some arguments
- `MetricRow` / `MetricsSummaryRow` duplication
- Stale `'score'` listing in accepted design record `docs/design/0007-show-neg-when-used.md` (left as history)

## Test plan

- [x] Registry / discouraged / force-add unit tests
- [x] CV and comparison `metrics.score()` still work when score is not registered
- [x] Doctests on metrics accessors
- [x] No circular `registry ↔ report` reference (state size test)
- [x] Merged with `main` (`#3196` registry-key row names kept via `resolved.name`)

### DIFF
--- a/skore/src/skore/_sklearn/metrics.py
+++ b/skore/src/skore/_sklearn/metrics.py
@@ -875,7 +875,7 @@ def default_metrics(report: EstimatorReport) -> list[Metric]:
         The metric instances to register, in default display order.
     """
     ml_task = report._ml_task
-    metrics: list[Metric] = []
+    metrics: list[Metric] = [Score()]
 
     if ml_task == "binary-classification":
         metrics += [Accuracy(), Precision(), Recall(), RocAuc(), LogLoss(), Brier()]
@@ -894,7 +894,6 @@ def default_metrics(report: EstimatorReport) -> list[Metric]:
         metrics += [R2(), Rmse(), Mae(), Mape()]
 
     metrics += [FitTime(), PredictTime()]
-    metrics.insert(0, Score())
 
     return metrics
 
--- a/skore/src/skore/_sklearn/metrics.py
+++ b/skore/src/skore/_sklearn/metrics.py
@@ -857,24 +857,46 @@ class Score(Metric):
         )
 
 
-# Order matters for default display
-BUILTIN_METRICS: list[Metric] = [
-    Accuracy(),
-    Precision(),
-    PrecisionMacro(),
-    Recall(),
-    RecallMacro(),
-    RocAuc(),
-    RocAucMacro(),
-    LogLoss(),
-    Brier(),
-    R2(),
-    Rmse(),
-    Mae(),
-    Mape(),
-    FitTime(),
-    PredictTime(),
-]
+def default_metrics(report: EstimatorReport) -> list[Metric]:
+    """Build the metrics registered by default for ``report``, in display order.
+
+    Selection is driven by the ML task. This is the single place to evolve the
+    default set. Metrics that the estimator cannot support are filtered out by
+    :meth:`Metric.available`.
+
+    Parameters
+    ----------
+    report : EstimatorReport
+        The report to build the default metrics for.
+
+    Returns
+    -------
+    list of :class:`Metric`
+        The metric instances to register, in default display order.
+    """
+    ml_task = report._ml_task
+    metrics: list[Metric] = []
+
+    if ml_task == "binary-classification":
+        metrics += [Accuracy(), Precision(), Recall(), RocAuc(), LogLoss(), Brier()]
+    elif ml_task == "multiclass-classification":
+        metrics += [
+            Accuracy(),
+            Precision(),
+            PrecisionMacro(),
+            Recall(),
+            RecallMacro(),
+            RocAuc(),
+            RocAucMacro(),
+            LogLoss(),
+        ]
+    elif ml_task in ("regression", "multioutput-regression"):
+        metrics += [R2(), Rmse(), Mae(), Mape()]
+
+    metrics += [FitTime(), PredictTime()]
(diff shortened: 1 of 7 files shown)


thomass-dev (dismissed):
Honestly i have no strong opinion on that.

Making the availability tests in a factory is the same as delegating to each `metric.available`.
The only benefit is one compute in the factory vs re-doing many computes in each `metric.available`.

LGTM, wait for the @auguste-probabl review.

skore/src/skore/_sklearn/metrics.py
```diff
@@ -857,24 +857,45 @@ def _raw(
...
+    ----------
+    report : EstimatorReport
+        The report to build the default metrics for.
+
+    Returns
+    -------
+    list of :class:`Metric`
+        The metric instances to register, in default display order.
+    """
+    ml_task = report._ml_task
```
auguste-probabl: inline

auguste-probabl (changes requested):
As I understand it, the point of `default_metrics` is to allow more flexibility in defining which metrics should be shown: `available` tells you if the metric is technically computable, whereas `default_metrics` would tell you what we maintainers recommend showing. The thing is:
- I can't think of a motivating example for this change. What would you like to do with this that you can't do with `available`?
- If we accept that we need this distinction, I don't see why that information couldn't be defined within each metric class, e.g. with a new `recommended` method.

glemaitre:
`available` for me is related to what we had with the `available_if` in scikit-learn. It answers to the question "is this metric can be computed on the given predictions".

The factory with the default choices is different semantic to me. It answers to the following question "From the available metric for the problem, is this particular metric advisable to be used given the problem, data, and model at hand".

So in short, `available` is for a technical/engineering purpose while the factory is for a methodological purpose.

glemaitre:
I need to think a bit more for this problem. Basically, what I don't like here is more the language actually: built-in and available are the two terms that I'm concerned with.

However, the fact that the metric itself, given the report, can tell you what to do (that could be extended for which reason), is actually nice.

I'll think about more about the metric API to see if we can make this distinction between the `available` and for instance something like `is_recommended`. Registering a metric will be a combination of the two and `is_recommended` could return a sort of reason why it is not recommended.

thomass-dev:
@glemaitre if a user adds a metric, do you want `available` to filter it out or you want it in the summary whether it's computable or not?

glemaitre:
So I'm putting in draft mode because it seems that it is not as urgent as what I thought.

glemaitre:
> @glemaitre if a user adds a metric, do you want available to filter it out or you want it in the summary whether it's computable or not?

My last thought where that the `.add` could have a parameter that would bypass or not a `discouraged` function that could return the reason for which a metric should not be added.

If the user is bypassing this, the metric is added in the registry and will be computed. If not bypassing the check, then we raise an error with the reason. In the factory, we could add any available method and make sure it does not raise any error then.

glemaitre:
I open https://github.com/probabl-ai/skore/pull/3202 to isolate the part that should not require discussion for the moment.

PR #3196, opened 2026-08-04
refactor(skore): Use the registry key as the summarize row name

glemaitre:
`Metric.summary_name` existed only so `PrecisionMacro`, `RecallMacro` and `RocAucMacro` reported themselves as `precision`, `recall` and `roc_auc` in `summarize().summary`. Metrics registered through `metrics.add` already report their registry key, so built-ins and custom metrics followed different rules.

The idea is to treat those metrics as any other metrics.


thomass-dev (approved):
No impact on the hub, as the hub is compliant with unknown metric types.

LGTM.

PR #3197, opened 2026-08-05
feat(checks): Improve baseline checks

### LINKED ISSUE #3064: Improve baseline checks (SKD009 & SKD010)
Here are some improvements to make on the two baseline checks:
- Show the performance of the baselines we compare against. Easy when the check finds an issue: put it in the explanation. Harder but valuable when the check passes: there is no explanation currently but it would be nice to know how much better we are to a decent baseline.
- Use `LogisticRegressionCV` instead of `LogisticRegression` for the fast baseline in classification, if the computational cost increase is not too dramatic. Needs quick investigation
- Compare both fit and predict time in SKD010, and use the maximum time ratio.
- Increase the floor of fit time in SKD010, 0.05 seconds is too little and raises false positives on small datasets.
- Problem: when using the same model as the baseline, e.g. `tabular_pipeline(HistGradientBoostingClassifier())`, SKD009 can raise an issue, despite using the same model.
  -> allow some tolerance, especially if the fit and predict times are close to each other.

GaetandeCast:
Closes #3064. Improves the two baseline checks (SKD009, SKD010) on all five points raised in the issue.

1. Show the baseline's performance, always
SKD009 now always reports the `HistGradientBoosting` baseline's per-metric test scores — previously this only showed up when the check flagged an issue, and the check emitted nothing at all when the model passed. Since SKD009 now always has something informative to say, it's been reclassified from issue to tip: a warning when the model is significantly worse than the baseline, or "on par with or better" with the baseline's scores for reference otherwise.

2. `LogisticRegressionCV` for the SKD010 fast baseline
Benchmarked it against `LogisticRegression` and found it's on the order of 20x slower to fit across dataset sizes — switching would defeat the purpose of a fast baseline, so `LogisticRegression` stays.

3. Compare both fit and predict time in SKD010
The slowness gate now computes both the fit-time and predict-time ratios (report vs. baseline) and uses whichever is larger, instead of only looking at fit time. A model that's fine to fit but slow to predict is now caught too.

4. Raise the SKD010 timing floor
The absolute-gap floor moved from 0.05s to 1s — differences below a second are negligible in practice regardless of the ratio, and the old floor was prone to false positives from timing noise on small/fast fits.

5. Tolerate SKD009 firing on a model that is the baseline
Reworked the significance check to require the baseline to be significantly better than the model (rather than the model failing to significantly beat the baseline) before calling it a regression. This directly fixes the reported case — a model built from `tabular_pipeline(HistGradientBoostingClassifier())` compared against SKD009's own identical baseline — since a tie or marginal loss from fit-to-fit noise no longer crosses the threshold; only a genuine, meaningful gap does.


skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -268,6 +268,14 @@ def get_fit_time(report: EstimatorReport | CrossValidationReport) -> float:
     return report._fit_time
 
 
+def get_predict_time(report: EstimatorReport | CrossValidationReport) -> float:
+    if report._report_type == "cross-validation":
+        return float(
+            report.metrics.timings(aggregate="mean").loc["Predict time test (s)"]
+        )
```
auguste-probabl, line 275: should be able to use `report.metrics.predict_time`

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -268,6 +268,14 @@ def get_fit_time(report: EstimatorReport | CrossValidationReport) -> float:
...
+    return cast(float, report.metrics.predict_time(data_source="test"))
```
auguste-probabl, line 276:
```suggestion
    return float(report.metrics.predict_time(data_source="test"))
```

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -525,37 +529,55 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+        )
 
-        votes = [
-            not check_score_gap_to_baseline(
-                score=report_test[key]["score"],
-                baseline=baseline_test[key]["score"],
+        worse_votes = [
+            check_score_gap_to_baseline(
+                score=baseline_test[key]["score"],
+                baseline=report_test[key]["score"],
```
auguste-probabl, line 542: confused at to why `baseline != baseline_test...`

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -525,37 +529,55 @@ def check_function(self, report: _BaseReport) -> str | None:
...
         ]
-        majority, n_positive, total = majority_vote(votes)
+        majority, n_worse, total = majority_vote(worse_votes)
+
+        baseline_performance = ", ".join(
+            f"{key[0]}"
+            f"{f' ({key[1]})' if key[1] is not None else ''}"
+            f"={baseline_test[key]['score']:.3g}"
+            for key in common_keys
+        )
```
auguste-probabl: Confusing, can you make this less terse

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -525,37 +529,55 @@ def check_function(self, report: _BaseReport) -> str | None:
...
+
         if majority:
             return (
-                "Test scores are not significantly better than a "
-                "HistGradientBoosting baseline for "
-                f"{n_positive}/{total} default predictive metrics."
+                "Test scores are significantly worse than a HistGradientBoosting "
+                f"baseline for {n_worse}/{total} default predictive metrics. "
+                f"Baseline performance on the test set: {baseline_performance}."
```
auguste-probabl:
```suggestion
                "Test scores are significantly worse than a HistGradientBoosting"
                f" baseline for {n_worse}/{total} default predictive metrics."
                f" Baseline performance on the test set: {baseline_performance}."
```
I find this way I forget spaces less often

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -525,37 +529,55 @@ def check_function(self, report: _BaseReport) -> str | None:
...
             )
-        return None
+        return (
+            "Your model is on par with or better than a HistGradientBoosting "
+            "baseline. Baseline performance on the test set, for reference: "
+            f"{baseline_performance}."
```
auguste-probabl:
```suggestion
            "Your model is on par with or better than a HistGradientBoosting"
            " baseline. Baseline performance on the test set, for reference:"
            f" {baseline_performance}."
```

skore/src/skore/_sklearn/_checks/model_checks.py
```diff
@@ -592,8 +627,8 @@ def check_function(self, report: _BaseReport) -> str | None:
         majority, n_positive, total = majority_vote(votes)
         if majority:
             return (
-                f"Fit is ~{slowness_ratio:.1f}x slower than a fast linear baseline "
-                f"without significantly better test scores "
+                f"{dimension} is ~{slowness_ratio:.1f}x slower than a fast linear "
+                "baseline without significantly better test scores "
                 f"({n_positive}/{total} default predictive metrics)."
```
auguste-probabl:
```suggestion
                f"{dimension} is ~{slowness_ratio:.1f}x slower than a fast linear"
                " baseline without significantly better test scores"
                f" ({n_positive}/{total} default predictive metrics)."
```

skore/tests/unit/checks/test_skd010_slower_than_baseline.py
```diff
@@ -33,6 +33,34 @@ def test_detects_slower_than_baseline(report_type, regression_data):
...
 
+@pytest.mark.parametrize("report_type", ["estimator", "cross-validation"])
+@pytest.mark.filterwarnings(
+    "ignore:Only pandas and polars DataFrames are supported:UserWarning:skrub"
+)
+def test_not_detected_when_gap_below_floor(monkeypatch, report_type, regression_data):
+    """SKD010 does not fire when the ratio is high but the absolute gap is tiny."""
+    X, y = regression_data
+    report = evaluate(
+        RidgeCV(), X, y, splitter=0.2 if report_type == "estimator" else 3
```
auguste-probabl: Maybe the estimator could be dumber since we patch?

auguste-probabl (changes requested):
AFAIU the check needs to become a "tip" so that it does not show up in issues when it is in fact fine. But then the check fails, will it also be in "tip"? Won't that make it super easy to gloss over and forget?

GaetandeCast:
> But then the check fails, will it also be in "tip"? Won't that make it super easy to gloss over and forget?

Yep it's a tradeoff. I would argue that it is fine because as opposed to the "slower than linear model" check, there are cases where being worse than an HGB is not an issue e.g. when you deliberately choose a simpler model than a gradient boosting model for computational/interpretability reasons. In this case knowing the perf of HGB is a good tip but not an issue.

But yes this change makes the information about the performance baseline less loud, which is a loss for people actively trying to improve their score.

skore/src/skore/_sklearn/_checks/_utils.py
GaetandeCast: mypy is not happy with that. The return type of `predict_time` is `float | None` which is not compatible with what `float` expects, according to mypy.

auguste-probabl: ok

skore/src/skore/_sklearn/_checks/_utils.py
GaetandeCast: AFAIK there is no `metric.predict_time` for cv reports.

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: This is on purpose. To make the check less sensitive, we now vote when the HBR is significantly better than the given report, not when the report is not significantly better than the HGB.

GaetandeCast: Adding an inline comment to clarify.

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: There is now that metrics are defined dynamically

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: Done, also added potential info about the output and average that were skipped.

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: done, good catch

skore/src/skore/_sklearn/_checks/model_checks.py
auguste-probabl: maybe a better solution could be to rename `check_score_gap_to_baseline`? Maybe `check_score_better_than_baseline`?

skore/src/skore/_sklearn/_checks/_utils.py
GaetandeCast: cool, using it now.

skore/src/skore/_sklearn/_checks/_utils.py
```diff
@@ -268,6 +268,11 @@ def get_fit_time(report: EstimatorReport | CrossValidationReport) -> float:
     return report._fit_time
 
 
+def get_predict_time(report: EstimatorReport | CrossValidationReport) -> float:
+    time = report.metrics.predict_time(data_source="test")
+    return cast(float, time)
```
GaetandeCast: I think it's safe to not catch the `None` case since it cannot happen naturally as we compute test predictions at init. Covering a `raise CheckNotApplicable` would need an artificial monkeypatch.

skore/src/skore/_sklearn/_checks/model_checks.py
GaetandeCast: done

skore/src/skore/_sklearn/_checks/_utils.py
auguste-probabl: oh, i was hoping you could avoid the dataframe operations by using it. oh well

auguste-probabl (approved):

PR #3201, opened 2026-08-05
feat(project): Synchronize reports across storage modes

Aljutor:
## Summary
- add one-way and bidirectional project synchronization using canonical report IDs
- preserve report keys and support dry-run transfer plans
- validate project compatibility and make retries skip completed transfers

```python
local = Project("demo", mode="local", workspace="./data")
hub = Project("demo", mode="hub", workspace="team")

local.sync(hub)  # local → Hub
local.sync(hub, bidirectional=True)  # both directions
```

### DIFF
--- a/skore/src/skore/__init__.py
+++ b/skore/src/skore/__init__.py
@@ -16,7 +16,6 @@ from rich.theme import Theme
 from skore._config import configuration
 from skore._externals._sklearn_compat import parse_version
 from skore._project._summary import Summary
-from skore._project._sync import SyncOperation, SyncResult
 from skore._project.login import login
 from skore._project.project import Project
 from skore._sklearn import (
@@ -83,8 +82,6 @@ __all__ = [
     "RocCurveDisplay",
     "CalibrationDisplay",
     "Summary",
-    "SyncOperation",
-    "SyncResult",
     "TableReportDisplay",
     "TrainTestSplit",
     "compare",
--- a/skore/src/skore/__init__.py
+++ b/skore/src/skore/__init__.py
@@ -16,6 +16,7 @@ from rich.theme import Theme
 from skore._config import configuration
 from skore._externals._sklearn_compat import parse_version
 from skore._project._summary import Summary
+from skore._project._sync import SyncOperation, SyncResult
 from skore._project.login import login
 from skore._project.project import Project
 from skore._sklearn import (
@@ -82,6 +83,8 @@ __all__ = [
     "RocCurveDisplay",
     "CalibrationDisplay",
     "Summary",
+    "SyncOperation",
+    "SyncResult",
     "TableReportDisplay",
     "TrainTestSplit",
     "compare",
--- a/skore/src/skore/_project/_sync.py
+++ b/skore/src/skore/_project/_sync.py
@@ -7,25 +7,14 @@ from typing import TYPE_CHECKING, Literal, cast
 from pandas import NA, DataFrame, Index, concat
 
 if TYPE_CHECKING:
-    from typing import Any, Protocol
-
-    class _Summary(Protocol):
-        def frame(self) -> DataFrame: ...
-
-    class _Project(Protocol):
-        def summarize(self) -> _Summary: ...
-
-        def get(self, id: str) -> Any: ...
-
-        def put(self, key: str, report: Any) -> Any: ...
+    from skore._project.project import Project
 
 
-_Direction = Literal["outbound", "inbound"]
-_Status = Literal["planned", "transferred", "skipped"]
-_RESULT_COLUMNS = ("key", "direction", "status")
+Direction = Literal["outbound", "inbound"]
+Status = Literal["planned", "transferred", "skipped"]
 
 
-def _snapshot(project: _Project) -> DataFrame:
+def _snapshot(project: Project) -> DataFrame:
     frame = project.summarize().frame()
     if frame.empty or "report_id" not in frame:
         # Treat legacy summaries from before `report_id` was stored as empty snapshots.
@@ -45,34 +34,36 @@ def _snapshot(project: _Project) -> DataFrame:
 
 def _transfer(
     reports: DataFrame,
-    source: _Project,
-    destination: _Project,
-    direction: _Direction,
+    source: Project,
+    destination: Project,
 ) -> None:
     for report_id, row in reports.iterrows():
         try:
             report = source.get(cast(str, row["backend_id"]))
             destination.put(cast(str, row["key"]), report)
         except Exception as exc:
-            exc.add_note(f"Failed to synchronize report {report_id!r} {direction}.")
+            exc.add_note(
+                f"Failed to synchronize report {report_id!r} "
+                f"from {source} to {destination}."
+            )
             raise
 
 
-def _result(
+def _build_result_frame(
     reports: DataFrame,
     *,
-    direction: _Direction | None,
-    status: _Status,
+    direction: Direction | None,
+    status: Status,
 ) -> DataFrame:
-    result = reports.loc[:, ["key"]].copy()
+    result = reports.loc[:, ["key"]]
     result["direction"] = NA if direction is None else direction
     result["status"] = status
     return result
 
 
 def synchronize(
-    source: _Project,
-    destination: _Project,
+    source: Project,
+    destination: Project,
     *,
     bidirectional: bool,
     dry_run: bool,
@@ -81,28 +72,41 @@ def synchronize(
     source_reports = _snapshot(source)
     destination_reports = _snapshot(destination)
 
-    outbound = source_reports.loc[
+    outbound_reports = source_reports.loc[
         source_reports.index.difference(destination_reports.index, sort=False)
     ]
-    inbound = destination_reports.loc[
+    inbound_reports = destination_reports.loc[
         destination_reports.index.difference(source_reports.index, sort=False)
     ]
-    skipped = source_reports.loc[
+    skipped_reports = source_reports.loc[
         source_reports.index.intersection(destination_reports.index, sort=False)
     ]
 
     if not dry_run:
-        _transfer(outbound, source, destination, "outbound")
+        _transfer(outbound_reports, source, destination)
         if bidirectional:
-            _transfer(inbound, destination, source, "inbound")
+            _transfer(inbound_reports, destination, source)
 
-    transfer_status: _Status = "planned" if dry_run else "transferred"
+    transfer_status: Status = "planned" if dry_run else "transferred"
     frames = [
-        _result(outbound, direction="outbound", status=transfer_status),
+        _build_result_frame(
+            outbound_reports,
+            direction="outbound",
+            status=transfer_status,
+        ),
     ]
     if bidirectional:
-        frames.append(_result(inbound, direction="inbound", status=transfer_status))
-    frames.append(_result(skipped, direction=None, status="skipped"))
+        frames.append(
+            _build_result_frame(
+                inbound_reports,
+                direction="inbound",
+                status=transfer_status,
+            )
+        )
+    frames.append(
+        _build_result_frame(skipped_reports, direction=None, status="skipped")
+    )
 
-    result = concat(frames).reindex(columns=_RESULT_COLUMNS).rename_axis("report_id")
+    result = concat(frames).reindex(columns=["key", "direction", "status"])
+    result = result.rename_axis("report_id")
     return cast(DataFrame, result.astype("string"))
--- a/skore/src/skore/_project/_sync.py
+++ b/skore/src/skore/_project/_sync.py
@@ -1,131 +1,52 @@
-"""Cross-mode project synchronization."""
+"""Project synchronization."""
 
 from __future__ import annotations
 
-from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
-from pandas import isna
-
-from skore._project.types import ProjectMode
+from pandas import NA, DataFrame, concat
 
 if TYPE_CHECKING:
     from skore._project.project import Project
 
 
-@dataclass(frozen=True)
-class SyncOperation:
-    """A report transfer in a synchronization result.
-
-    Parameters
-    ----------
-    report_id : str
-        Canonical Skore report ID.
-    key : str
-        Source key used when storing the report at the destination.
-    source_mode : {"hub", "local", "mlflow"}
-        Storage mode from which the report is loaded.
-    destination_mode : {"hub", "local", "mlflow"}
-        Storage mode in which the report is stored.
-    """
-
-    report_id: str
-    key: str
-    source_mode: ProjectMode
-    destination_mode: ProjectMode
+_RESULT_COLUMNS = ["key", "direction", "status"]
 
 
-@dataclass(frozen=True)
-class SyncResult:
-    """The planned or completed operations of a project synchronization.
-
-    Parameters
-    ----------
-    operations : tuple of SyncOperation
-        Planned operations for a dry-run, otherwise completed operations.
-    skipped : tuple of str
-        Canonical report IDs already present in both projects.
-    dry_run : bool
-        Whether the operations were only planned.
-    """
-
-    operations: tuple[SyncOperation, ...]
-    skipped: tuple[str, ...]
-    dry_run: bool
-
-
-@dataclass(frozen=True)
-class _ReportRef:
-    report_id: str
-    backend_id: str
-    key: str
-
-
-@dataclass(frozen=True)
-class _Transfer:
-    operation: SyncOperation
-    source: Project
-    destination: Project
-    backend_id: str
-
-
-def _snapshot(project: Project) -> dict[str, _ReportRef]:
+def _snapshot(project: Project) -> DataFrame:
     frame = project.summarize().frame()
     if frame.empty:
-        return {}
+        return DataFrame(columns=["backend_id", "key"]).rename_axis("report_id")
 
-    reports: dict[str, _ReportRef] = {}
-    missing_backend_ids: list[str] = []
-
-    for backend_id, row in zip(
-        frame.index.get_level_values("id"), frame.to_dict("records"), strict=True
-    ):
-        report_id = row.get("report_id")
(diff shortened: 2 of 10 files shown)


thomass-dev:
@Aljutor can you please sign your commits? That way we are able to use the merge-queue. Thanks!

Aljutor:
> @Aljutor can you please sign your commits? That way we are able to use the merge-queue. Thanks!

Thanks, done, ssh agent broke again and I forgot to re-enable ssh sign

skore/src/skore/_project/project.py
```diff
@@ -373,6 +375,56 @@ def summarize(self) -> Summary:
...
+        self,
+        other: Project,
+        *,
+        bidirectional: bool = False,
+        dry_run: bool = False,
+    ) -> SyncResult:
+        """Copy missing reports to another project.
+
+        Reports are matched by canonical ID. Set ``bidirectional=True`` to copy missing
+        reports in both directions.
```
glemaitre: I would not use the term "canonical ID". I would rather point to ID and mention how this ID could be found.

skore/src/skore/_project/project.py
```diff
@@ -373,6 +375,56 @@ def summarize(self) -> Summary:
...
+
+        Parameters
+        ----------
+        other : Project
+            Destination project. It must have the same name as this project and use a
+            different storage mode.
+        bidirectional : bool, default=False
+            If ``False``, transfer reports from this project to ``other``. If ``True``,
+            also transfer reports missing from this project. Both differences are
+            computed before any transfer.
```
glemaitre: I think that you can drop the last sentence.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+        Planned operations for a dry-run, otherwise completed operations.
+    skipped : tuple of str
+        Canonical report IDs already present in both projects.
+    dry_run : bool
+        Whether the operations were only planned.
+    """
+
+    operations: tuple[SyncOperation, ...]
+    skipped: tuple[str, ...]
+    dry_run: bool
```
glemaitre: I'm wondering if I would not just return a plain dataframe to quickly see the status. It would have the benefit that both HTML and text repr will be readable.

skore/src/skore/_project/project.py
```diff
@@ -373,6 +375,56 @@ def summarize(self) -> Summary:
...
+            Planned operations when ``dry_run=True``, otherwise completed operations,
+            together with canonical IDs already present in both projects.
+        """
+        if not isinstance(other, Project):
+            raise TypeError(f"`other` must be a Project (found {type(other)!r}).")
+        if other.name != self.name:
+            raise ValueError(
+                "Synchronization requires projects with the same name "
+                f"(found {self.name!r} and {other.name!r})."
+            )
```
glemaitre: I'm not sure about this limitation. Basically, `other={"hub", "local", ...}` was a shortcut when the name are matching.

If the name are not matching then it is handy to provide a `skore.Project` for this purpose.

skore/src/skore/_project/project.py
```diff
@@ -373,6 +375,56 @@ def summarize(self) -> Summary:
...
+        if other.mode == self.mode:
+            raise ValueError(
+                "Synchronization requires two different project modes "
+                f"(both were {self.mode!r})."
+            )
```
glemaitre: Actually do we need to force that? It could be an handy way to do a quick copy paste in a local machine.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+    backend_id: str
+
+
+def _snapshot(project: Project) -> dict[str, _ReportRef]:
+    frame = project.summarize().frame()
+    if frame.empty:
+        return {}
+
+    reports: dict[str, _ReportRef] = {}
+    missing_backend_ids: list[str] = []
```
glemaitre: I'm wondering if we should take care about this part. Basically, since we are raising an error and nobody can do anything about it, I would just drop the rows with missing IDs and just ignore.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+    ):
+        report_id = row.get("report_id")
+        if report_id is None or isna(report_id):
+            missing_backend_ids.append(str(backend_id))
+            continue
+
+        report_id = str(report_id)
+        # Summaries are ordered by date. Reinsert duplicate IDs so the latest stored
+        # copy also determines transfer order and the key used at the destination.
+        reports.pop(report_id, None)
```
glemaitre: I'm confused here. Why we would need to pop if we just change the value associated with the key right after

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+
+    for backend_id, row in zip(
+        frame.index.get_level_values("id"), frame.to_dict("records"), strict=True
+    ):
+        report_id = row.get("report_id")
+        if report_id is None or isna(report_id):
+            missing_backend_ids.append(str(backend_id))
+            continue
+
+        report_id = str(report_id)
```
glemaitre: Can we avoid casting and just use whatever comes from the dataframe?

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+def _snapshot(project: Project) -> dict[str, _ReportRef]:
+    frame = project.summarize().frame()
+    if frame.empty:
+        return {}
+
+    reports: dict[str, _ReportRef] = {}
+    missing_backend_ids: list[str] = []
+
+    for backend_id, row in zip(
+        frame.index.get_level_values("id"), frame.to_dict("records"), strict=True
```
glemaitre: Actually, I'm not sure why we need to iterate other the dataframe here. At this stage we should just return the frame as is. Once, we have the two frame from both mode, it should comeback to a set operation on the index?

What are the corner cases that does not allow to do that.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+
+    if (
+        source.ml_task is not None
+        and destination.ml_task is not None
+        and source.ml_task != destination.ml_task
+    ):
+        raise ValueError(
+            "Cannot synchronize projects with different ML tasks: "
+            f"{source.ml_task!r} and {destination.ml_task!r}."
+        )
```
glemaitre: I'm almost sure that it will fail later when calling `put`. I would delegate.

skore/src/skore/_project/_sync.py
glemaitre: At least we should consider it, when send back to the user.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+    if not dry_run:
+        for transfer in transfers:
+            try:
+                report = transfer.source.get(transfer.backend_id)
+                if str(report.id) != transfer.operation.report_id:
+                    raise RuntimeError(
+                        "The loaded report ID does not match its project summary: "
+                        f"expected {transfer.operation.report_id!r}, got "
+                        f"{str(report.id)!r}."
+                    )
```
glemaitre: What is the actual reason for this case to happen?

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,192 @@
...
+                source,
+                destination_reports,
+                source_reports,
+            )
+        )
+
+    operations = tuple(transfer.operation for transfer in transfers)
+    skipped = tuple(
+        report_id for report_id in source_reports if report_id in destination_reports
+    )
```
glemaitre: Do we really care for those to be immutable?

skore/src/skore/__init__.py
```diff
@@ -82,6 +83,8 @@
     "RocCurveDisplay",
     "CalibrationDisplay",
     "Summary",
+    "SyncOperation",
+    "SyncResult",
```
glemaitre: We should not expose those to the public API.

skore/tests/unit/plugins/mlflow/test_project.py
```diff
@@ -144,6 +144,18 @@ def test_init_reuses_existing_storage_experiment(self, mlflow_tracking_uri):
             == second._Project__storage_experiment_id
         )
 
+    def test_init_existing_project_with_sql_tracking_uri(self, tmp_path):
+        tracking_uri = f"sqlite:///{tmp_path}/existing-mlflow.db"
```
glemaitre: To speed up in windows, we are using the file system with MLFlow instead of SQLite. It is much faster.

skore/tests/unit/project/test_sync.py
```diff
@@ -0,0 +1,367 @@
...
+import pytest
+from pandas import DataFrame, Index, MultiIndex, RangeIndex
+from sklearn.datasets import make_regression
+from sklearn.linear_model import LinearRegression, Ridge
+
+from skore import Project, SyncOperation, evaluate
+from skore._project._sync import synchronize
+
+
+class FakeProject:
```
glemaitre, line 25: Could we put in common some mock project such as `FakeHubProject` and `FakeLocalProject` in `testing.mock` module. It could avoid to redefine those in a each new module.

sphinx/reference/project.rst
```diff
@@ -29,8 +29,16 @@ These functions and classes are meant for managing a `Project` and its reports.
    Project.put
    Project.get
    Project.summarize
+   Project.sync
    Project.delete
 
+.. autosummary::
```
glemaitre: Let's remove this.

sphinx/user_guide/project.rst
```diff
@@ -56,3 +56,29 @@ returned by :meth:`Project.summarize`. This method returns a list of
...
+   result = project_local.sync(project_hub)
+
+The caller is the source; reverse the call for the opposite direction. With
+``bidirectional=True``, both differences are computed before transfers start.
+
+Reports are matched by canonical ID and copied with their keys. Existing IDs are skipped;
+contents and metadata are not compared. ``dry_run=True`` returns the transfer plan
+without calling ``get`` or ``put``.
+
+:class:`SyncResult` contains transfer operations and IDs already present in both
```
glemaitre: We would need to update that. I think a dataframe will be better for the user.

glemaitre (commented):
A first round of review.

skore/src/skore/_project/_sync.py
Aljutor: I don't know why, but using DF here feels wrong for me here, but maybe it's a taste thing and I just prefer structs.  We can do it anyways.

skore/src/skore/_project/project.py
Aljutor: It's to avoid push report to the wrong project, as initial idea for sync was closer to "push all".  

We can lift this requirement.

skore/tests/unit/plugins/mlflow/test_project.py
Aljutor: Oh, windows

skore/src/skore/_project/_sync.py
glemaitre: Be aware that I was thinking the dataframe when reporting to the user not necessarily in the intermediate representation.

skore/tests/unit/project/test_sync.py
Aljutor: Agreed, let's do this latter in dedicated PR.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,108 @@
...
+
+    class _Summary(Protocol):
+        def frame(self) -> DataFrame: ...
+
+    class _Project(Protocol):
+        def summarize(self) -> _Summary: ...
+
+        def get(self, id: str) -> Any: ...
+
+        def put(self, key: str, report: Any) -> Any: ...
```
glemaitre: This is weird to have those protocol declared here.

I would advocate to declare them without `_` in the module where we declare the project and summary.
Then we can import it here.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+_Direction = Literal["outbound", "inbound"]
+_Status = Literal["planned", "transferred", "skipped"]
+_RESULT_COLUMNS = ("key", "direction", "status")
```
glemaitre: I would not make them with an `_` because you are already in a private module. So whatever live in here is scope to this module.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,108 @@
...
+            raise
+
+
+def _result(
+    reports: DataFrame,
+    *,
+    direction: _Direction | None,
+    status: _Status,
+) -> DataFrame:
+    result = reports.loc[:, ["key"]].copy()
```
glemaitre: I'm almost sure that we don't need the copy here.

glemaitre (commented):
Just some small nitpicks but it looks good globally.  I'll look more in details the tests.

skore/tests/unit/project/test_sync.py
```diff
@@ -0,0 +1,369 @@
+from itertools import permutations
+from types import SimpleNamespace
+from typing import cast
+from unittest.mock import Mock
+
+import mlflow
+import pytest
+from pandas import NA, DataFrame, Index, MultiIndex, RangeIndex, isna
```
glemaitre: In general we tend to do `import pandas as pd` and just use those.

skore/tests/unit/project/test_sync.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def report_ids(project):
+    return {row["report_id"] for row in project.records}
+
+
+@pytest.mark.parametrize(
+    ("source_mode", "destination_mode"), permutations(("local", "hub", "mlflow"), 2)
+)
+def test_synchronize_supports_all_mode_pairs(source_mode, destination_mode):
```
glemaitre, line 81: Can you add a small docstring of one line for each test to have a narrative on what we are testing.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,108 @@
...
+    source: _Project,
+    destination: _Project,
+    direction: _Direction,
+) -> None:
+    for report_id, row in reports.iterrows():
+        try:
+            report = source.get(cast(str, row["backend_id"]))
+            destination.put(cast(str, row["key"]), report)
+        except Exception as exc:
+            exc.add_note(f"Failed to synchronize report {report_id!r} {direction}.")
```
auguste-probabl: nit
```suggestion
            exc.add_note(f"Failed to synchronize report {report_id!r} from {source} to {destination}.")
```
or a more readable approximation (`direction` should not be a user-facing concern)

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,108 @@
...
+            raise
+
+
+def _result(
```
auguste-probabl: Can you find a more explicit name for this?

skore/src/skore/_project/_sync.py
auguste-probabl: inline

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def synchronize(
+    source: _Project,
+    destination: _Project,
+    *,
+    bidirectional: bool,
+    dry_run: bool,
+) -> DataFrame:
+    """Synchronize two projects using report IDs."""
```
auguste-probabl: nit: a bit confusing that IDs are never mentioned in the body of the function. I guess `outbound` and such are IDs? Maybe renaming them to `outbound_ids` would help? Not sure

sphinx/user_guide/project.rst
```diff
@@ -56,3 +56,33 @@ returned by :meth:`Project.summarize`. This method returns a list of
...
+
+The caller is the source; reverse the call for the opposite direction. Set
+``bidirectional=True`` to transfer missing reports in both directions.
+
+Reports are matched using the ``report_id`` column returned by
+``Project.summarize().frame()`` and copied with their keys. Existing IDs are skipped;
+contents and metadata are not compared. Reports without a ``report_id`` are ignored.
+Set ``dry_run=True`` to return the transfer plan without loading or storing reports.
+Two MLflow projects must use the same tracking URI because MLflow uses process-global
+tracking state. Synchronization does not provide concurrency control.
```
auguste-probabl: This could be a note added at the end

auguste-probabl (commented):
some nitpicks; looks good!

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,112 @@
+"""Project synchronization."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Literal, cast
+
+from pandas import NA, DataFrame, Index, concat
```
glemaitre: I missed this one `import pandas as pd` as well

Aljutor: Noooo

Aljutor: Done in `d2818843`. `_sync.py` now uses `import pandas as pd` consistently for `DataFrame`, `Index`, `NA`, and `concat`.

glemaitre (dismissed):
LGTM on my side. @auguste-probabl do you want to have a look regarding your reviews.

skore/src/skore/_project/_sync.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def synchronize(
+    source: Project,
+    destination: Project,
+    *,
+    bidirectional: bool,
+    dry_run: bool,
+) -> pd.DataFrame:
+    """Synchronize two projects using report IDs."""
```
auguste-probabl: nit
```suggestion
    """Synchronize two projects."""
```

auguste-probabl (dismissed):
Just a final nit but good to go anyway!

glemaitre (approved):

glemaitre:
LGTM. Merging. Thanks @Aljutor

PR #3207, opened 2026-08-06
docs(examples): Stop using skrub open payments data

moujanrastgoo:
Here is an updated version of the ``EstimatorReport`` example that does not use open payments. The reason for this change is that maybe in the future the open-payments dataset might not be available in Skrub. Feel free to make suggestions. 

#### Contribution checklist

#### AI usage disclosure

AI tools were involved for:

### DIFF
--- a/examples/model_evaluation/plot_estimator_report.py
+++ b/examples/model_evaluation/plot_estimator_report.py
@@ -205,11 +205,9 @@ def operational_decision_gain(y_true, y_pred, *, amount):
 
 # %%
 #
-# In our example use case, each classification decision changes monetary gain.
+# In our example use case, each classification decision has a different monetary gain.
 # The function above models this by translating the confusion matrix into a gain
-# (payoff) matrix; this also depends on an extra parameter named ``amount``, to
-# illustrate that skore can handle custom metrics with non-standard arguments.
-# Here ``amount`` is the real basket cash value from the feature matrix.
+# (payoff) matrix that depends on the basket cash value.
 # Let's test adding this metric to our report.
 from sklearn.metrics import make_scorer
 
--- a/examples/model_evaluation/plot_estimator_report.py
+++ b/examples/model_evaluation/plot_estimator_report.py
@@ -170,13 +170,6 @@ report.metrics.log_loss(data_source="train")
 # scikit-learn scorers, i.e. functions taking `estimator`, `X` and `y` (and optional
 # keyword arguments). Let's take a look at an example.
 
-
-def operational_decision_gain(y_true, y_pred, *, amount):
-    mask_true_positive = (y_true == pos_label) & (y_pred == pos_label)
-    mask_true_negative = (y_true == neg_label) & (y_pred == neg_label)
-    mask_false_positive = (y_true == neg_label) & (y_pred == pos_label)
-    mask_false_negative = (y_true == pos_label) & (y_pred == neg_label)
-    # TP: recover basket amount, pay review cost (20)
 def operational_decision_gain(y_true, y_pred, *, amount):
     """The monetary gain we obtain depending on our predictions.
 
@@ -201,27 +194,18 @@ def operational_decision_gain(y_true, y_pred, *, amount):
     # Legitimate basket wrongly flagged: we pay the review costs, but we also annoy
     # the customer and risk losing them, so it is penalized compared to a correct
     # refusal
-    legitimate_refuse = mask_false_positive.sum() * (REVIEW_COST+REPUTATION_COST)
+    legitimate_refuse = mask_false_positive.sum() * (REVIEW_COST + REPUTATION_COST)
 
     # Legitimate basket correctly accepted: we earn a margin on the sale
     legitimate_accept = (amount[mask_true_negative] * MARGIN).sum()
 
     return fraudulent_refuse + fraudulent_accept + legitimate_refuse + legitimate_accept
-    # FN: miss fraud, lose full basket amount
-    fraudulent_accept = -amount[mask_false_negative].sum()
-    # FP: wrongly flag legitimate basket, pay review cost (20)
-    legitimate_refuse = mask_false_positive.sum() * -20
-    # TN: correctly accept, small margin on basket amount
-    legitimate_accept = (amount[mask_true_negative] * 0.02).sum()
-    return fraudulent_refuse + fraudulent_accept + legitimate_refuse + legitimate_accept
-
-
 # %%
 #
-# In our example use case, each classification decision we make has a cost.
-# The function above models this by translating the confusion matrix into a cost
-# matrix; this cost also depends on an extra parameter named ``amount``, to illustrate
-# that skore can handle custom metrics with non-standard arguments.
+# In our example use case, each classification decision changes monetary gain.
+# The function above models this by translating the confusion matrix into a gain
+# (payoff) matrix; this also depends on an extra parameter named ``amount``, to
+# illustrate that skore can handle custom metrics with non-standard arguments.
 # Here ``amount`` is the real basket cash value from the feature matrix.
 # Let's test adding this metric to our report.
 from sklearn.metrics import make_scorer
--- a/examples/model_evaluation/plot_estimator_report.py
+++ b/examples/model_evaluation/plot_estimator_report.py
@@ -13,16 +13,57 @@ quickly get insights from any scikit-learn estimator.
 # Loading our dataset and defining our estimator
 # ==============================================
 #
-# First, we load a dataset from skrub. Our goal is to predict if a healthcare
-# manufacturing company paid medical doctors or hospitals, in order to detect
-# potential conflicts of interest.
+# First, we load a dataset from skrub. Our goal is to predict whether an online
+# shopping basket is fraudulent, so that the payment can be reviewed before
+# money leaves the account.
+#
+# Baskets and products live in two tables, so we aggregate product-level
+# information (including the cash amount of the basket) into one feature matrix
+# with pandas. Using a skrub DataOp to keep those joins inside the estimator
+# (and replay them on unseen data) is shown in
+# :ref:`example_tracking_all_the_data_processing`.
 
 # %%
-from skrub.datasets import fetch_open_payments
+from skrub.datasets import fetch_credit_fraud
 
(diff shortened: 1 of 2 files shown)


examples/model_evaluation/plot_estimator_report.py
```diff
@@ -13,16 +13,57 @@
...
-# potential conflicts of interest.
+# First, we load a dataset from skrub. Our goal is to predict whether an online
+# shopping basket is fraudulent, so that the payment can be reviewed before
+# money leaves the account.
+#
+# Baskets and products live in two tables, so we aggregate product-level
+# information (including the cash amount of the basket) into one feature matrix
+# with pandas. Using a skrub DataOp to keep those joins inside the estimator
+# (and replay them on unseen data) is shown in
+# :ref:`example_tracking_all_the_data_processing`.
```
auguste-probabl: I can't find that example, maybe it's from the skrub docs?

moujanrastgoo: It's in  [Tracking all the data processing](https://docs.skore.probabl.ai/stable/auto_examples/use_cases/plot_tracking_all_the_data_processing.html). We do the processing there with DataOpps.

auguste-probabl: Right, I missed it because it doesn't have a stable ID like the other examples. Could you add one in this PR? Something like:

```
"""
.. _example_data_processing:
```

at the top of `plot_tracking_all_the_data_processing.py`

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -13,16 +13,57 @@
...
+# 8,000-row stratified subsample that preserves the natural fraud rate.
 
 # %%
-from skrub.datasets import fetch_open_payments
+from sklearn.model_selection import train_test_split
 
-dataset = fetch_open_payments()
-df = dataset.X
-y = dataset.y
+df, _, y, _ = train_test_split(df, y, train_size=8_000, stratify=y, random_state=0)
```
auguste-probabl: Does it really make the example much slower to use the full dataset? If not we can remove this block

examples/model_evaluation/plot_estimator_report.py
moujanrastgoo: Sure i will do that!

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -144,9 +185,9 @@ def operational_decision_cost(y_true, y_pred, *, amount):
     mask_true_negative = (y_true == neg_label) & (y_pred == neg_label)
     mask_false_positive = (y_true == neg_label) & (y_pred == pos_label)
     mask_false_negative = (y_true == pos_label) & (y_pred == neg_label)
-    fraudulent_refuse = mask_true_positive.sum() * 50
+    fraudulent_refuse = (amount[mask_true_positive] - 20).sum()
     fraudulent_accept = -amount[mask_false_negative].sum()
-    legitimate_refuse = mask_false_positive.sum() * -5
+    legitimate_refuse = mask_false_positive.sum() * -70
     legitimate_accept = (amount[mask_true_negative] * 0.02).sum()
```
auguste-probabl: why was this changed?

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -157,12 +198,11 @@ def operational_decision_cost(y_true, y_pred, *, amount):
...
 # matrix; this cost also depends on an extra parameter named ``amount``, to illustrate
 # that skore can handle custom metrics with non-standard arguments.
+# Here ``amount`` is the real basket cash value from the feature matrix.
 # Let's test adding this metric to our report.
-import numpy as np
 from sklearn.metrics import make_scorer
 
-rng = np.random.default_rng(42)
-amount = rng.integers(low=100, high=1000, size=len(report.y_test))
+amount = report.X_test["basket_amount"].to_numpy()
```
auguste-probabl: Maybe
```suggestion
amount = report.X_test["basket_amount"]
```
is sufficient

auguste-probabl (changes requested):

examples/model_evaluation/plot_estimator_report.py
moujanrastgoo: I can check,  the full number of rows of the final table after preprocessing is around 93k. it might still be okay to use the full thing or more of the rows. If the runtime turns out to be less than 5 mins, I will keep the whole dataset then.

examples/model_evaluation/plot_estimator_report.py
moujanrastgoo: I thought it would be nice to penalize more expensive missed fraudulent baskets more, but if you think not, I can revert it!

examples/model_evaluation/plot_estimator_report.py
auguste-probabl: > 5 mins

I think even 30 seconds is long enough

examples/model_evaluation/plot_estimator_report.py
auguste-probabl: what i'm confused about is that the cost for `fraudulent_refuse` was a `*` before, but now it's a `-`

moujanrastgoo: Because, before in the open-payment case, the amount was a random number, so our reward for a true positive was 50 times the random number, here, our reward for a true positive can be the amount of the money of the basket that was fraudulent, minus a review cost (adding the review cost was suggested by the agent).

moujanrastgoo: but i can go back to the original version or a version without review or customer dissatisfaction cost

auguste-probabl: Either way, could you add comments explaining the calculation?

moujanrastgoo: sure !

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -185,9 +176,13 @@ def operational_decision_cost(y_true, y_pred, *, amount):
     mask_true_negative = (y_true == neg_label) & (y_pred == neg_label)
     mask_false_positive = (y_true == neg_label) & (y_pred == pos_label)
     mask_false_negative = (y_true == pos_label) & (y_pred == neg_label)
+    # TP: recover basket amount, pay review cost (20)
     fraudulent_refuse = (amount[mask_true_positive] - 20).sum()
```
auguste-probabl: Ok so this should be `* -20` right?

moujanrastgoo: I just realized where the problem might be, I forgot to change the name of the custom metric from operational_cost(the name in the original example) to operational_gain, because here I made a metric that we actually have to maximize rather than minimize. Here, 20 is a flat review cost. When we catch a fraud case successfully, the money that we gain compared to not noticing it is the cost of the basket minus the money that we spent on reviewing it. If it had been the amount of the basket * -20, we would actually lose 20 times the amount of the fraudulent basket that we caught, so we would punish catching a fraud.

I'm sorry, I will change that.

auguste-probabl: but if we change `operational_decision_cost` is converted to a gain, shouldn't we also be changing `fraudulent_accept` and `legitimate_accept`?

moujanrastgoo: Sure, we can change them, what should we change them to?

auguste-probabl: ok, sorry for the back-and-forths, i had to dig in and get a better
understanding. Here is what I ended up with:

```suggestion
def operational_decision_gain(y_true, y_pred, *, amount):
    """The monetary gain we obtain depending on our predictions.

    May be negative, in which case our predictions actually *cost* us money.
    """
    mask_true_positive = (y_true == pos_label) & (y_pred == pos_label)
    mask_true_negative = (y_true == neg_label) & (y_pred == neg_label)
    mask_false_positive = (y_true == neg_label) & (y_pred == pos_label)
    mask_false_negative = (y_true == pos_label) & (y_pred == neg_label)

    REVIEW_COST = -20
    REPUTATION_COST = -30
    MARGIN = 0.02

    # Fraud correctly flagged: we pay the review costs, but do not incur
    # more cost
    fraudulent_refuse = mask_true_positive.sum() * REVIEW_COST

    # Fraud missed: the payment goes through and we lose the full basket amount
    fraudulent_accept = -amount[mask_false_negative].sum()

    # Legitimate basket wrongly flagged: we pay the review costs, but we also annoy
    # the customer and risk losing them, so it is penalized compared to a correct
    # refusal
    legitimate_refuse = mask_false_positive.sum() * (REVIEW_COST+REPUTATION_COST)

    # Legitimate basket correctly accepted: we earn a margin on the sale
    legitimate_accept = (amount[mask_true_negative] * MARGIN).sum()

    return fraudulent_refuse + fraudulent_accept + legitimate_refuse + legitimate_accept
```

Indeed `fraudulent_accept` and `legitimate_accept` don't need to change. The change to `operational_decision_gain` is good because that is actually what we calculate (and what we already did on `main` -- this was actually an error).

examples/model_evaluation/plot_estimator_report.py
```diff
@@ -139,34 +171,53 @@
...
 #
-# In our example use case, each classification decision we make has a cost.
-# The function above models this by translating the confusion matrix into a cost
-# matrix; this cost also depends on an extra parameter named ``amount``, to illustrate
-# that skore can handle custom metrics with non-standard arguments.
+# In our example use case, each classification decision changes monetary gain.
+# The function above models this by translating the confusion matrix into a gain
+# (payoff) matrix; this also depends on an extra parameter named ``amount``, to
+# illustrate that skore can handle custom metrics with non-standard arguments.
+# Here ``amount`` is the real basket cash value from the feature matrix.
```
auguste-probabl: We no longer need to justify the artificial `amount` parameter, so this can be simplified:
```suggestion
# In our example use case, each classification decision has a different monetary gain.
# The function above models this by translating the confusion matrix into a gain
# (payoff) matrix that depends on the basket cash value.
```

auguste-probabl (approved):
Thank you!!

PR #3208, opened 2026-08-08
feat(hub): Send CheckSummaryDisplay info

glemaitre:
This PR is sending the dataframe containing all the information to be displayed on the Skore Hub side.

### DIFF
--- a/skore/src/skore/_plugins/hub/artifact/media/media.py
+++ b/skore/src/skore/_plugins/hub/artifact/media/media.py
@@ -35,9 +35,11 @@ class Media(Artifact, ABC, Generic[Report]):
         The name of the media.
     data_source : str | None
         The source of the data used to generate the media.
+    parameters : Parameters
+        The parameters used to generate the media.
     """
 
     report: Report = Field(repr=False, exclude=True)
     name: str = Field(init=False)
     data_source: str | None = Field(init=False)
-    parameters: Parameters | None = Field(init=False, default=None)
+    parameters: Parameters = Field(init=False, default=None)
--- a/skore/src/skore/_plugins/hub/artifact/media/media.py
+++ b/skore/src/skore/_plugins/hub/artifact/media/media.py
@@ -14,6 +14,7 @@ Parameters = (
     | dict[Literal["with_roc_auc"], Literal[True]]
     | dict[Literal["threshold_value"], Literal["all"]]
     | dict[Literal["threshold_value"], None]
+    | dict[Literal["fast_mode"], Literal[True]]
     | None
 )
 
--- a/skore/src/skore/_plugins/hub/artifact/pickle/pickle.py
+++ b/skore/src/skore/_plugins/hub/artifact/pickle/pickle.py
@@ -31,7 +31,8 @@ class Pickle(Artifact):
     -----
     It uploads the pickled report to the artifacts storage in a lazy way.
 
-    The report is uploaded without its cache, to avoid salting the checksum.
+    The report is uploaded without its prediction/display cache and without its
+    checks-results cache, to avoid salting the checksum.
     The report is primarily pickled on disk to reduce RAM footprint.
     """
 
@@ -45,13 +46,21 @@ class Pickle(Artifact):
 
         Notes
         -----
-        The report is pickled without its cache, to avoid salting the checksum.
+        The report is pickled without its prediction/display cache and without its
+        checks-results cache, to avoid salting the checksum.
         """
         reports = [self.report] + getattr(self.report, "reports_", [])
         reports_with_cache = [
             (report, report._cache) for report in reports if hasattr(report, "_cache")
         ]
+        reports_with_check_results_cache = [
+            (report, report._check_results_cache)
+            for report in reports
+            if hasattr(report, "_check_results_cache")
+        ]
         self.report._clear_cache()
+        for report, _ in reports_with_check_results_cache:
+            del report._check_results_cache
 
         try:
             with BytesIO() as stream:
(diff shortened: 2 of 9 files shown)


skore/src/skore/_plugins/hub/artifact/media/media.py
```diff
@@ -14,6 +14,7 @@
     | dict[Literal["with_roc_auc"], Literal[True]]
     | dict[Literal["threshold_value"], Literal["all"]]
     | dict[Literal["threshold_value"], None]
+    | dict[Literal["fast_mode"], Literal[True]]
     | None
```
auguste-probabl, line 18: nit: redundant (the `| None` specifically)

skore/src/skore/_plugins/hub/artifact/pickle/pickle.py
```diff
@@ -45,13 +46,21 @@ def content_to_upload(self) -> Generator[bytes, None, None]:
 
         Notes
         -----
-        The report is pickled without its cache, to avoid salting the checksum.
+        The report is pickled without its prediction/display cache and without its
```
auguste-probabl, line 49: nit: redundant with the note in the class docstring

skore/src/skore/_plugins/hub/artifact/pickle/pickle.py
auguste-probabl, line 43: maybe we could take the opportunity to use `report.to_dict()` here?

skore/tests/unit/plugins/hub/artifact/media/test_checks.py
```diff
@@ -0,0 +1,47 @@
+from pydantic import ValidationError
+from pytest import mark, param, raises
```
auguste-probabl: we tend to `import pytest` in this project

skore/tests/unit/plugins/hub/report/test_cross_validation_report.py
auguste-probabl, line 53: nit: i know it doesn't originate from your PR, but it's a smell to me that we need to copy-paste the logic from the code in the tests.

skore/tests/unit/plugins/hub/report/test_estimator_report.py
auguste-probabl, line 31: see above

auguste-probabl (commented):
some nitpicks, but ok with me!

skore/tests/unit/plugins/hub/artifact/media/test_checks.py
thomass-dev: There is no convention (and also no PEP or whatever). If you want one for the whole project, open a discussion.

skore/src/skore/_plugins/hub/artifact/media/media.py
thomass-dev: Please explain. The parameters can be null or one of the options.

skore/src/skore/_plugins/hub/artifact/pickle/pickle.py
thomass-dev: Not in this PR/this issue, thanks. It's already planned and sync with the compute feature.

skore/src/skore/_plugins/hub/artifact/pickle/pickle.py
thomass-dev: It two different points of docstr, that are not displayed at the same point.
Repeating docstr for user is not a bad thing.

skore/src/skore/_plugins/hub/artifact/media/media.py
auguste-probabl: it's redundant with line 42, where we have `Parameters | None`. One of the two `| None` is unnecessary.

skore/tests/unit/plugins/hub/artifact/media/test_checks.py
auguste-probabl: By "we tend to" I mean "There is a convention, at least in my mind". `grep`ping for both shows we use `import pytest` about 5x more

glemaitre: For those imports, it is true that I'm usually importing `pytest`. I'll be consistent with myself here.

skore/tests/unit/plugins/hub/report/test_cross_validation_report.py
glemaitre: I'm not going to touch that here as well. If we refactor, it will be outside of this PR but maybe it is worth. I don't have the design in my mind.

skore/src/skore/_plugins/hub/artifact/media/media.py
thomass-dev: `| None` line 42 should be deleted.

skore/tests/unit/plugins/hub/artifact/media/test_checks.py
thomass-dev: IMO, we should not put convention on use "direct import" or "from import" if both are allowed by PEP.
It's only a matter of taste.

thomass-dev (approved):

thomass-dev (changes requested):
Sphinx fail because of public project: it tries to insert checks in the hub, while the hub is not compatible yet.

> Platform exception: Unknown MediaType ('checks_summary', 'application/vnd.dataframe', {'fast_mode': True})

thomass-dev:
I draft the PR while the hub is not compatible.

thomass-dev (approved):

PR #3214, opened 2026-08-12
fix(skore): Derive the MLflow storage experiment name on Databricks

### LINKED ISSUE #3213: Connect Skore to MLFlow backend on Databricks

`skore.Project(mode="mlflow")` cannot be used against Databricks managed MLflow
(`tracking_uri="databricks"`), regardless of the experiment name passed by the user.

Databricks requires MLflow experiment names to be absolute workspace paths
(e.g. `/Users/<user>/my-experiment`). `Project.__init__` in the MLflow plugin
unconditionally creates an internal storage experiment with a hardcoded name:

https://github.com/probabl-ai/skore/blob/main/skore/src/skore/_plugins/mlflow/project.py
(`skore/_plugins/mlflow/project.py`, line 121 in 0.23.1)

```python
self.__storage_experiment_id = mlflow.create_experiment("__skore-storage__")
```

`"__skore-storage__"` is not a valid Databricks experiment name, so `create_experiment`
raises `RestException(INVALID_PARAMETER_VALUE)`. The fallback
`mlflow.get_experiment_by_name("__skore-storage__")` returns `None` (the experiment
can never exist), so the exception is re-raised and the `Project` constructor fails —
even when the user-supplied `name` is a perfectly valid workspace path.

## Steps/Code to Reproduce

```python
# DATABRICKS_HOST and DATABRICKS_TOKEN set in the environment
import skore

project = skore.Project(
    "/Users/<user>@<domain>/skore-test",  # valid Databricks experiment path
    mode="mlflow",
    tracking_uri="databricks",
)
```

## Expected Behavior

The project is created and reports can be stored in and retrieved from
Databricks managed MLflow.

## Actual Behavior

```
mlflow.exceptions.RestException: INVALID_PARAMETER_VALUE: Got an invalid experiment
name '__skore-storage__'. An experiment name must be an absolute path within the
Databricks workspace, e.g. '/Users/<some-username>/my-experiment'.
```

(A bare, non-path `name` fails the same way once the storage experiment issue is
fixed — that part is expected Databricks behavior, but may be worth a note in the
MLflow integration docs.)

## Workaround / evidence the rest of the flow works

Remapping only the storage experiment name to a valid workspace path makes the full
`put()` → `summarize()` → `get()` roundtrip work against a real Databricks
workspace (verified with skore 0.23.1 / mlflow 3.15.1: metrics, model artifact,
and `EstimatorReport` reconstruction via `project.get(run_id)` all behave as on
open-source MLflow):

```python
import mlflow

STORAGE_PATH = "/Users/<user>@<domain>/__skore-storage__"
_create, _get = mlflow.create_experiment, mlflow.get_experiment_by_name
mlflow.create_experiment = lambda n, *a, **k: _create(
    STORAGE_PATH if n == "__skore-storage__" else n, *a, **k)
mlflow.get_experiment_by_name = lambda n: _get(
    STORAGE_PATH if n == "__skore-storage__" else n)
```

Any of:

1. Make the storage experiment name configurable — a `Project` kwarg and/or a
   `SKORE_MLFLOW_STORAGE_EXPERIMENT` environment variable.
2. Auto-detect a Databricks tracking URI (`mlflow.get_tracking_uri()` starting with
   `"databricks"`) and derive the storage experiment as
   `/Users/<current-user>/__skore-storage__`, resolving the user from the workspace
   identity (`mlflow.utils.databricks_utils` or the SCIM `Me` endpoint).
3. Derive the storage experiment name from the user-supplied project `name`
   (e.g. a sibling path), which sidesteps the identity lookup entirely.

## Environment where i tested it

- skore 0.23.1
- mlflow 3.15.1
- scikit-learn 1.9.0
- Python 3.14, macOS
- Databricks on AWS, managed MLflow, PAT auth

yanndebray:
Closes #3213.

Implements option 2 from the issue, as agreed by @cakedev0 and @thomass-dev.

## Problem

`Project(mode="mlflow")` could never be constructed against Databricks managed MLflow. The plugin created its internal storage experiment under the hardcoded name `__skore-storage__`, which Databricks rejects because an experiment name must be an absolute workspace path. The `get_experiment_by_name` fallback could not recover either, since an experiment under that name can never exist there — so the constructor always raised `INVALID_PARAMETER_VALUE`, however valid the `name` passed by the user was.

## Change

On a Databricks tracking URI, derive the storage experiment as `/Users/<current-user>/__skore-storage__`. The identity comes from the SCIM `Me` endpoint via `get_databricks_host_creds(tracking_uri)` — the same credential path MLflow's own Databricks tracking store uses, so profiles, PATs, env vars and in-runtime auth all work without extra configuration.

Every other tracking URI keeps the bare `__skore-storage__` name, so the storage experiment stays shared between projects and report objects stay deduplicated. `__init__` passes the derived name to both `create_experiment` and the existing fallback lookup, leaving the "reuse the storage experiment across projects" behaviour untouched.

A failed identity lookup is re-raised with a message naming what skore was attempting, since otherwise a bare SCIM `403` surfaces from inside a `Project()` call with no context.

The user guide gains a note about the other half of the issue: on Databricks the project `name` must itself be an absolute workspace path. That is Databricks behaviour rather than something the plugin should hide.

## Tests

Five new unit tests, no workspace required: the bare name is preserved off Databricks; `databricks` and `databricks://profile` both derive the home path and call the expected endpoint; the lookup-failure message; and a `Project.__init__` test asserting the name the created experiment actually carries.

Verified end to end against a real Databricks workspace (AWS, managed MLflow, PAT auth, mlflow 3.15.1): the storage experiment lands at `/Users/<user>/__skore-storage__`, and `put()` → `summarize()` → `get()` roundtrips with the `EstimatorReport` reconstructed intact.

Not covered, and worth a reviewer's eye: I only exercised PAT-via-env-vars auth. A `~/.databrickscfg` profile, a service principal token (whose `userName` is the application ID), and running inside a Databricks notebook all take the same code path but are untested.

## Note

`test_get` fails on my machine, on `main` as well as on this branch — numpy 2.5 deprecation raised through joblib unpickling under `filterwarnings = error`. Unrelated to this change.


skore/src/skore/_plugins/mlflow/project.py
```diff
@@ -98,6 +101,43 @@ def format_date(start_time: int | None) -> str:
     return datetime.fromtimestamp(start_time / 1_000, tz=UTC).isoformat()
 
 
+STORAGE_EXPERIMENT_NAME = "__skore-storage__"
+
+
+def _databricks_user_name(tracking_uri: str) -> str:
```
glemaitre, line 107: Let's inline this function, it is only called once.

skore/src/skore/_plugins/mlflow/project.py
```diff
@@ -98,6 +101,43 @@ def format_date(start_time: int | None) -> str:
...
+    response = http_request_safe(  # type: ignore[no-untyped-call]
+        host_creds=creds,
+        endpoint="/api/2.0/preview/scim/v2/Me",
+        method="GET",
+    )
+
+    return cast(str, response.json()["userName"])
+
+
+def _storage_experiment_name(tracking_uri: str) -> str:
```
glemaitre, line 127: The same with this function.

skore/src/skore/_plugins/mlflow/project.py
```diff
@@ -98,6 +101,43 @@ def format_date(start_time: int | None) -> str:
...
+        return STORAGE_EXPERIMENT_NAME
+
+    try:
+        user_name = _databricks_user_name(tracking_uri)
+    except MlflowException as exc:
+        raise MlflowException(
+            "Failed to resolve the Databricks workspace user needed to create the "
+            f"{STORAGE_EXPERIMENT_NAME!r} experiment. Make sure the credentials used "
+            f"for '{tracking_uri}' can read the current user."
+        ) from exc
```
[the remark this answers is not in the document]
glemaitre: I don't agree with this comment and it seems to much some defensive code here. If the code fails, the original failure would be OK.

glemaitre (commented):
A couple of comments.

skore/src/skore/_plugins/mlflow/project.py
auguste-probabl: i think this one is alright actually

skore/src/skore/_plugins/mlflow/project.py
```diff
@@ -98,6 +101,43 @@ def format_date(start_time: int | None) -> str:
...
+def _databricks_user_name(tracking_uri: str) -> str:
+    """Identity of the workspace user behind a Databricks ``tracking_uri``."""
+    creds = get_databricks_host_creds(tracking_uri)  # type: ignore[no-untyped-call]
+    response = http_request_safe(  # type: ignore[no-untyped-call]
+        host_creds=creds,
+        endpoint="/api/2.0/preview/scim/v2/Me",
+        method="GET",
+    )
+
+    return cast(str, response.json()["userName"])
```
cakedev0: `response.json()["userName"]` can raise a bare `KeyError` here, which escapes the `except MlflowException` in `_storage_experiment_name` below and propagates unwrapped out of `Project()`.

This happens for a Databricks service-principal identity (OAuth M2M via `DATABRICKS_CLIENT_ID`/`DATABRICKS_CLIENT_SECRET`, common in CI): per Databricks' SCIM schema, a service principal has no `userName` field, so the 200 response from `/api/2.0/preview/scim/v2/Me` omits it.

Worth guarding this so it surfaces as the friendly `MlflowException` this PR is adding, rather than a raw `KeyError`:

```suggestion
    body = response.json()
    try:
        return cast(str, body["userName"])
    except KeyError:
        raise MlflowException(
            "Databricks SCIM 'Me' response has no 'userName' field; this can happen "
            "when authenticating as a service principal, which is identified by "
            "'applicationId' rather than 'userName'."
        ) from None
```

cakedev0 (approved):
LGTM

PR #3233, opened 2026-08-25
perf: Lazy-load display and plotting backends

thomass-dev:
Based on https://peps.python.org/pep-0562/.


thomass-dev:
Superseded by https://github.com/probabl-ai/skore/pull/3236.

PR #3242, opened 2026-09-01
ci: Move uv installation to its own step

auguste-probabl:


.github/workflows/pytest.yml
```diff
@@ -89,6 +89,11 @@ jobs:
         with:
           fetch-depth: 2
 
+      - name: Install uv
+        uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d # v10.0.1
+        with:
+          version: "0.6.16"
```
rouk1, line 95: `uv`'s current version is [0.12.8](https://github.com/astral-sh/uv/releases#release-0.12.8), should we take this opportunity to upgrade ?

rouk1 (changes requested):
We should bump uv's version but otherwise LGTM.

.github/workflows/pytest.yml
auguste-probabl: i did it in a followup PR https://github.com/probabl-ai/skore/pull/3243, either way is fine for me

rouk1: oki !

rouk1 (approved):

Issue #3063, opened 2026-06-18
Refine SKD011: golden feature


GaetandeCast:
Here is a list of improvements and fixes for SKD011:
- When a model underfits, SKD011 will trigger a lot of false positives: if a model has poor performance, it will perform as poorly with all individual features isolated, marking many features as gold. Solution: filter low performance model or make it silent if SKD002 (underfitting) has raised an issue.
- Make a special case when the score is close to perfect: it almost necessarily means that the target is in the features.
- The check is only performed on pre-processed features. It would be nice to also try it on the raw input features although tricky as they may have incompatible type, missing values etc.
- a golden feature can happen in two scenarios: the target is in the features or a feature happens to carry almost all the information on the target, i.e. it is a good proxy. It would be nice to distinguish the two cases to have custom messages. Point 2 goes into that direction.

GaetandeCast:
#3252 addressed everything except point 3. Closing this in favor of a dedicated issue for point 3 that will also touch on SKD012

PR #3252, opened 2026-09-03
feat(checks): Improve the golden feature check (SKD011)

GaetandeCast:
Towards #3063:
- SKD011 skips underfit models. If SKD002 has already flagged underfitting on the same report, golden-feature detection is marked not applicable so a uniformly bad model is not treated as having many golden features.

- SKD011 splits the tip using a target-as-feature oracle. After golden features are found, the predictor is refit on the target as the only column. Features that match that oracle are described as likely copies of the target; the others keep the existing “matches the full model” wording. When both apply, the two sentences are joined with a newline.

### DIFF
--- a/skore/src/skore/_checks/skd011_golden_feature.py
+++ b/skore/src/skore/_checks/skd011_golden_feature.py
@@ -17,11 +17,71 @@ from skore._checks.utils import (
     split_preprocessor_estimator,
 )
 from skore._sklearn.feature_names import _get_feature_names
+from skore._utils.dataframe import _normalize_y_as_dataframe
 
 if TYPE_CHECKING:
+    from skore._checks.utils import MetricKey
+    from skore._displays.metrics.metrics_summary_display import MetricsSummaryRow
     from skore._reports.base import _BaseReport
     from skore._reports.cross_validation.report import CrossValidationReport
     from skore._reports.estimator.report import EstimatorReport
+    from skore._sklearn.types import EstimatorLike
+
+
+def _on_par(
+    candidate: dict[MetricKey, MetricsSummaryRow],
+    reference: dict[MetricKey, MetricsSummaryRow],
+) -> bool:
+    return majority_vote(
+        [
+            not check_score_better_than_baseline(
+                score=reference[key]["score"],
+                baseline=candidate[key]["score"],
+                greater_is_better=reference[key]["greater_is_better"],
+                floor=0.03,
+                fraction=0.10,
+            )
+            for key in reference.keys() & candidate.keys()
+        ]
+    )[0]
+
+
+def _predictor_test_scores(
+    report: EstimatorReport | CrossValidationReport,
+    predictor: EstimatorLike,
+    metric_registry,
+    X,
+    y,
+    X_test=None,
+    y_test=None,
+) -> dict[MetricKey, MetricsSummaryRow]:
+    if report._report_type == "cross-validation":
+        from skore._reports.cross_validation.report import CrossValidationReport
+
+        report = CrossValidationReport(
+            clone(predictor),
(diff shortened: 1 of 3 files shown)


skore/src/skore/_checks/skd011_golden_feature.py
```diff
@@ -44,6 +106,10 @@ class CheckGoldenFeature(Check):
     slow = True
 
     def check_function(self, report: _BaseReport) -> str | None:
+        skd002_result = getattr(report, "_check_results_cache", {}).get("SKD002")
+        if skd002_result is not None and skd002_result["section"] == "issue":
+            raise CheckNotApplicable("Skipped because SKD002 detected underfitting.")
```
auguste-probabl, line 114: so we're not going to run SKD002, just check whether it's been run before and rely on the fact that checks tend to be run in increasing number order? This seems a bit brittle. I'm okay with this code right now, but do we expect many future cases of dependencies between checks like this? If yes, we might have to do a bit of design work to be able to do "get the result of SKD002 -- run the check if you need to"

skore/src/skore/_checks/skd011_golden_feature.py
```diff
@@ -17,19 +17,81 @@
...
 
 class CheckGoldenFeature(Check):
     """Check for a golden feature (SKD011).
 
     Detects a single feature that, used alone to refit the estimator, reaches
     scores close to the full model on the report's default predictive metrics.
-    This often signals data leakage or excessive reliance on one feature.
+    Features whose scores also match a model trained on the target itself are
+    reported as likely target leakage. Skipped when SKD002 has already flagged
+    underfitting.
```
auguste-probabl: Can you separate the "skipped" comment? otherwise it looks like it is linked to the previous text. Also explain a bit why we skip, like what was done in the PR description

auguste-probabl (changes requested):
Looks okay, some remarks

skore/src/skore/_checks/skd011_golden_feature.py
GaetandeCast: Yes and it is intentional. We had discussed it when the issue was raise (offline I think) and agreed that for this first case of dependent checks it is fine. But I agree that it may be too brittle in the future if such occurrences multiply. Then we can fix the mess/debt :smile:

skore/src/skore/_checks/skd011_golden_feature.py
GaetandeCast: done

skore/src/skore/_checks/skd011_golden_feature.py
auguste-probabl: can you open an issue

skore/src/skore/_checks/skd011_golden_feature.py
```diff
@@ -17,19 +17,84 @@
...
 class CheckGoldenFeature(Check):
     """Check for a golden feature (SKD011).
 
     Detects a single feature that, used alone to refit the estimator, reaches
     scores close to the full model on the report's default predictive metrics.
-    This often signals data leakage or excessive reliance on one feature.
+    Features whose scores also match a model trained on the target itself are
+    reported as likely target leakage.
+
+    If SKD002 has already flagged underfitting on the same report, This check is marked
```
auguste-probabl:
```suggestion
    If SKD002 (underfitting) was already reported on the same report, this check is marked
```

skore/src/skore/_checks/skd011_golden_feature.py
GaetandeCast: #3258

auguste-probabl (approved):

Issue #3240, opened 2026-08-27
refactor(plugins/hub): Rethink the API key management / authentication


thomass-dev:
Description WIP

Prioritities `environ['SKORE_HUB_API_KEY']` > `~/.skore.hub/.credentials.json` > interactive login

in .credentials.json, we store the tuple (URL, workspace, API key)
based on the SKORE_HUB_URI, and the `Project(workspace)`, the authentication module is able to manage.

---

After https://github.com/probabl-ai/skore-cli/pull/29

Inspired by git and cursor

https://www.reddit.com/r/ClaudeCode/comments/1ogzcxa/claude_code_credentials_or_api_key_when_using/
https://github.com/anthropics/claude-code/issues/441#issuecomment-3207226042
https://cursor.com/docs/cli/reference/authentication

<img width="791" height="527" alt="Image" src="https://github.com/user-attachments/assets/e5f11cd0-1453-4df7-ab9f-fba580d0e089" />

PR #3255, opened 2026-09-04
feat(plugins/hub)!: Refactor authentication to be API key oriented

### LINKED ISSUE #3261: [discussion] login has a mode - which is surprising.
When we had trouble to make the agent find how to log on Tuesday, with @rouk1 we noticed that in login signature, you could give a mode (cf [docs](https://docs.skore.probabl.ai/stable/reference/api/skore.login.html)). Which is surprising, because we don't need login, unless we want to be connected to the hub. Therefore, why is there a mode?  
- option 1: we want to log, we call login
- option 2: we don't need/want to, we don't call the function. 

It's a bit weird to have a function which does things... sometimes. And in some other configuration nothing.  

Should we simplify login?

thomass-dev:
Part of https://github.com/probabl-ai/skore/issues/3240, closes #3261.

---

Authentication no longer goes through `skore.login` or JWT/OAuth.
Users are now forced to authenticate with an API key stored in a local credentials registry:

```bash
$ skore hub api-key generate --workspace=<workspace>
```
or 
```bash
$ skore hub api-key add <api-key> --workspace=<workspace>
```

The public login API, login plugins, and token flow are removed; docs, examples, and tests follow that model.
API keys are stored in ``~/.skore.hub/credentials.json`` as a JSON object:

    {
        "type": "plaintext",
        "keys": [
            {
                "host": "<host>",
                "workspace": "<workspace>",
                "key": "<key>"
            }
        ]
    }

    {
        "type": "secret",
        "keys": [
            {
                "host": "<host>",
                "workspace": "<workspace>"
            }
        ]
    }

The storage mode is chosen automatically when the registry file is created: ``secret`` if a recommended [keyring]( https://github.com/jaraco/keyring) backend is available, otherwise ``plaintext``. In ``secret`` mode the API key is stored in the system keyring rather than in the JSON file.


skore/pyproject.toml
```diff
@@ -85,17 +85,17 @@ sphinx = ["mlflow", "polars", "skore[sphinx-base, hub, mlflow]"]
...
 dev = ["ipykernel", "nbformat"]
 hub = [
+  "filelock",
   "httpx",
+  "keyring>=25.7",
   "orjson",
   "pydantic",
   "rich[jupyter]>=14.2",
   "scipy",
+  # "skore-cli>=0.5",
```
thomass-dev, line 95: To uncomment when available.

skore/src/skore/_plugins/hub/authentication/registry.py
```diff
@@ -0,0 +1,237 @@
...
+def keys() -> Generator[tuple[str, str]]:
+    """Yield ``(host, workspace)`` keys stored in the registry."""
+    filepath = setup()
+
+    with filepath.open() as file:
+        for credential in load(file)["keys"]:
+            yield (normalize(credential["host"]), credential["workspace"])
+
+
+@lock
```
rouk1, line 133: 🤩

rouk1 (dismissed):
Tested on OSX and it works fine.

glemaitre:
The two last commits are introduced to address the changes specified in https://github.com/probabl-ai/skore-cli/pull/47#issuecomment-5822711068

PR #3264, opened 2026-09-11
fix: compute coefficient feature std on a plain array so polars inputs work

### LINKED ISSUE #3263: pydanticserializationerror when putting report when y is a polars series or dataframe
### Describe the bug

```python
import polars as pl

import skore

X = pl.DataFrame(
    {
        "a": [0.0, 1.0, 0.5, 1.5, 2.0, 0.2, 0.9, 1.1, 0.3, 1.8],
        "b": [1.0, 0.0, 1.0, 0.0, 1.0, 1.0, 0.0, 0.5, 0.8, 0.1],
    }
)
# both those options fail with different errors:
# y = pl.DataFrame({"class": ["A", "B", "A", "B", "B", "A", "B", "A", "B", "A"]})
y = pl.Series("class", ["A", "B", "A", "B", "B", "A", "B", "A", "B", "A"])

report = CrossValidationReport(LogisticRegression(max_iter=1000), X, y, splitter=3)

skore.login()
project = skore.Project(workspace="jerome-workspace-1", name="project-2", mode="hub")
project.put("report", report)
```

```
⠙ Putting report 0:00:00
  File "/home/jerome/workspace/scratch/2026/09/put_report.py", line 20, in <module>
    project.put("report", report)
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^
  File "/home/jerome/workspace/repos/skore/skore/src/skore/_project/project.py", line 309, in put
           ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jerome/workspace/repos/skore/skore/src/skore/_plugins/hub/project/project.py", line 292, in put
    payload_dict = payload.model_dump()
  File "/home/jerome/miniforge3/envs/skore/lib/python3.13/site-packages/pydantic/main.py", line 475, in model_dump
    return self.__pydantic_serializer__.to_python(
           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        self,
        ^^^^^
    ...<13 lines>...
        polymorphic_serialization=polymorphic_serialization,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
pydantic_core._pydantic_core.PydanticSerializationError: Error calling function `serialize_model`: TypeError: DataFrame.std() got an unexpected keyword argument 'axis'
```

### Steps/Code to Reproduce

.

### Expected Behavior

.

### Actual Behavior

.

### Environment

```shell
main branch
```

L4XB:
#### Change description

`CoefficientsDisplay._compute_data_for_display` computed the training feature standard deviation with `np.std(X_transformed, axis=0)`. `np.std` defers to the container's own `std` method, and `polars.DataFrame.std()` does not accept `axis`, so with a polars `X` (and no preprocessor that turns it into an array) the call raised `TypeError: DataFrame.std() got an unexpected keyword argument 'axis'`. In the hub plugin that surfaced as the `PydanticSerializationError` from the report, because the `medias` computed field builds the coefficients artifact; the same error is reachable locally through `report.inspection.coefficients()`.

The reduction now runs on `np.asarray(X_transformed)`. For pandas and numpy inputs this is the same population standard deviation as before (`ddof=0`); the sparse branch is untouched.

Closes #3263

The polars `y` from the report is not involved: with the fix, the payload in the issue dumps successfully (verified with the repro script and the hub test fixtures).

#### Contribution checklist

- [x] Unit tests were added or updated (if necessary): `test_feature_std_with_polars_train_data` in `tests/unit/displays/coefficients/test_estimator.py`, fails on `main` with the error above.
- [ ] Documentation was added or updated (if necessary): not needed.
- [x] The code passes our style conventions (`pre-commit run --files …`: all hooks pass, including ruff and mypy)
- [x] All the tests pass (`tests/unit/displays/coefficients/test_estimator.py`: 16 passed)
- [ ] The documentation builds and renders properly: no documentation change.
- [x] The pull request title respects the Conventional Commits convention

### DIFF
--- a/skore/src/skore/_displays/inspection/coefficients.py
+++ b/skore/src/skore/_displays/inspection/coefficients.py
@@ -782,7 +782,9 @@ class CoefficientsDisplay(DisplayMixin):
                 _, variance = mean_variance_axis(X_transformed, axis=0)
                 std = np.sqrt(variance)
             else:
-                std = np.std(X_transformed, axis=0)
+                # ``np.std`` defers to the container's own ``std`` method, and a
+                # polars frame does not take ``axis``; reduce a plain array.
+                std = np.std(np.asarray(X_transformed), axis=0)
             # the intercept is given a unit standard deviation to leave it unscaled
             feature_std = np.concatenate([[1.0], std])
         else:
--- a/skore/tests/unit/displays/coefficients/test_estimator.py
+++ b/skore/tests/unit/displays/coefficients/test_estimator.py
@@ -212,3 +212,25 @@ def test_scale_features_plot_labels(regression_train_test_split):
     fig = report.inspection.coefficients().plot(scale_features=True)
     assert fig.axes[0].get_xlabel() == "Magnitude of scaled coefficient"
     assert fig.get_suptitle() == "Scaled coefficients of Ridge"
+
+
+def test_feature_std_with_polars_train_data():
+    """Feature standard deviations are computed on a polars frame too (#3263).
+
+    ``np.std`` defers to the container's ``std`` method, and a polars frame does
+    not accept ``axis``; the display has to reduce a plain array.
+    """
+    polars = pytest.importorskip("polars")
+    X, y = make_regression(n_samples=40, n_features=3, random_state=0)
+    X = polars.DataFrame(X, schema=["a", "b", "c"])
+    report = EstimatorReport(
+        LinearRegression(), X_train=X, y_train=y, X_test=X, y_test=y
+    )
+
+    display = report.inspection.coefficients()
+
+    feature_std = display.coefficients.set_index("feature")["feature_std"]
+    np.testing.assert_allclose(feature_std["Intercept"], 1.0)
+    np.testing.assert_allclose(
+        feature_std[["a", "b", "c"]].to_numpy(), np.std(X.to_numpy(), axis=0)
+    )


L4XB:
#3273 fixed #3263 on `main` while this was open, so I've narrowed this PR to the one thing left: narwhals' `std()` defaults to `ddof=1`, while the ndarray and sparse branches use the population std. The same training data therefore gives a different `feature_std` as a pandas or polars frame than as an array, and pandas matched the array before #3273. The fix is `std(ddof=0)`, with a test that fails on `main`. Measurements are in the description.

GaetandeCast:
@L4XB it's an interesting catch. A few thoughts:
- I don't think the motivation for this change should be "use the same ddof as numpy" but rather make sure we use the same ddof that scalers use across the libraries we support. We use this value to make coefficients comparable for feature importance, based on the fact that multiplying coefficients by the feature std has the same effect on the coefficient as dividing each feature by its std before fitting. (see [here](https://scikit-learn.org/stable/auto_examples/inspection/plot_linear_model_coefficient_interpretation.html#interpreting-coefficients-scale-matters)) So we should align with what value scalers use to scale feature.
- scikit-learn's [`StandardScaler`](https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.StandardScaler.html) divides by std with `ddof=0`
- interestingly the above scikit-learn example uses pandas' default `ddof` value which is one, meaning the equivalence is actually not respected in the code
- the ratio between the two scaled numbers is $\sqrt{n}/\sqrt{n-1}$ which very quickly goes to one so for non tiny datasets, the difference is negligible (n=500 leads to a ratio of 1.001).
- The ratio is the same for all coefficients so it does not make the comparison invalid

So all in all, I think we should scale coefficients by the same value no matter the library used for storing the data and I think that using `ddof=0` is the right call here, but it's not a super urgent matter.

Any thoughts @glemaitre @jeromedockes ?

L4XB:
@GaetandeCast thanks, that is the better motivation, and I've reframed the PR around it. The test now asserts that a frame's `feature_std` equals `StandardScaler().fit(X_train).scale_` rather than `np.std`, and the code comment names the scaler; the change itself is still `std(ddof=0)`. The ndarray and sparse branches already match `StandardScaler`, so after this all three containers do.

Agreed on the size: the factor is √(n/(n−1)) for every feature, so rankings are unaffected and it vanishes on real data. What made it worth a PR to me is only that the same data gives different absolute scaled values in a frame and in an array.

PR #3266, opened 2026-09-13
Making SKD011's and SKD002's dependency robust

### LINKED ISSUE #3258: Add a way to make checks dependent on each other
PR #3252 makes SKD011 dependent on the result of SKD002 (golden feature check is marked as not applicable when model undefits so a uniformly bad model is not treated as having many features). This is fine for this occurrence as checks are run in increasing order of code, but if more dependencies are added, we will need a more robust way of handling them

"so we're not going to run SKD002, just check whether it's been run before and rely on the fact that checks tend to be run in increasing number order? This seems a bit brittle. I'm okay with this code right now, but do we expect many future cases of dependencies between checks like this? If yes, we might have to do a bit of design work to be able to do 'get the result of SKD002 -- run the check if you need to'"

_Originally posted by @auguste-probabl in https://github.com/probabl-ai/skore/pull/3252#discussion_r3948133071_

m4nn2609-dot:
Previously, CheckGoldenFeature (SKD011) checked whether SKD002 (underfitting) had already flagged an issue by reading directly from report._check_results_cache.

This only worked because checks happen to run in increasing numeric order by default, so SKD002's result was already cached by the time SKD011 ran. If the registry were ever reordered (via .remove()/.add()), or a future dependency pointed from a lower-numbered check to a higher-numbered one, this would silently do nothing — a missing cache entry looks identical to "SKD002 ran and found no issue."

closes #3258

AI tools were involved for:

### DIFF
--- a/skore/src/skore/_checks/skd011_golden_feature.py
+++ b/skore/src/skore/_checks/skd011_golden_feature.py
@@ -109,9 +109,6 @@ class CheckGoldenFeature(Check):
     slow = True
 
     def check_function(self, report: _BaseReport) -> str | None:
-        skd002_result = getattr(report, "_check_results_cache", {}).get("SKD002")
-        if skd002_result is not None and skd002_result["section"] == "issue":
-            raise CheckNotApplicable("Skipped because SKD002 detected underfitting.")
 
         if report._report_type == "cross-validation":
             report = cast("CrossValidationReport", report)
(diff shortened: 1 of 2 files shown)


GaetandeCast (commented):
This PR is puzzling me. I don't see any functional changes with these changes:
- moving the block down in `skd011_golden_feature.py` just means we move the early exit condition to after more work.
- the `_run_check` abstraction doesn't change anything: the only thing it adds is to exit if the check is in cache but we already exclude cached checks in the for loop that runs this function.
- the introduced `_get_check_result` is never called.

I'm closing the PR for this reason and also because the issue was opened to remind us of a potential problem when adding other dependencies between checks in the future, not as something actionable for right now.

PR #3274, opened 2026-09-16
feat(plugins/hub/client): Fallback to API keys registry

thomass-dev:


skore/tests/unit/plugins/hub/client/test_client.py
thomass-dev, line 1: Add a test with registry.

